PR #2, opened 2018-01-28
Implemented min and mag filters.

Rinnegatamante:
I'm quite sure the other listed values in vita-headers are fake values ( https://github.com/vitasdk/vita-headers/blob/master/include/psp2/gxm.h#L813-L814 ).


Rinnegatamante:
Can't say why but using this causes a GL_INVALID_ENUM being raised when running OpenSyobon.

src/emulator/modules/SceGxm/src/gxm.h
```diff
@@ -28,9 +28,11 @@ namespace emu {
         uint16_t height;
         Ptr<const void> data;
         Ptr<void> palette;
+        GLenum min_filter;
+        GLenum mag_filter;
     };
 
-    static_assert(sizeof(SceGxmTexture) == 16, "Incorrect size.");
+    static_assert(sizeof(SceGxmTexture) == 24, "Incorrect size.");
```
petmac: SceGxmTexture is 16 bytes in size, so whatever we pack in there can't make it bigger. I suspect this causes the OpenGL error you see as the game is probably overwriting that memory.

PR #4, opened 2018-01-28
Added a proper semaphore implementation with SDL_Semaphore

Rinnegatamante:
This also adds support for semaphores with initVal != 0 (They're used for example in lpp-vita and OpenSyobon).

### DIFF
--- a/src/emulator/modules/SceLibKernel/src/SceLibKernel.cpp
+++ b/src/emulator/modules/SceLibKernel/src/SceLibKernel.cpp
@@ -27,6 +27,10 @@
 #include <psp2/kernel/threadmgr.h>
 
 struct Semaphore {
+    std::shared_ptr<SDL_semaphore> sema = nullptr;
+    Semaphore(int initVal){
+        sema = std::shared_ptr<SDL_semaphore>(SDL_CreateSemaphore(initVal), SDL_DestroySemaphore);
+    }
 };
 
 struct ThreadParams {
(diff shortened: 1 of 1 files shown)


petmac:
When I started work on threading primitives I also reached for the SDL versions. Unfortunately, they can't be used directly because threads need to be able to get woken up and quit at any time.

For example:
The main thread M creates a worker thread.
Worker thread blocks on signal from main thread.
The user closes the emulator window, so the main thread exits.
The worker thread is never signalled, so the emulator is keeps running in the background.

A better solution might be to add a set of waiting thread IDs to each semaphore. Then when the semaphore is signalled, attempt to wake up the thread by setting its `to_do` field to `ThreadToDo::run` and signalling `something_to_do`.

src/emulator/modules/SceLibKernel/src/SceLibKernel.cpp
```diff
@@ -744,11 +748,11 @@ EXPORT(int, sceKernelCreateRWLock) {
 EXPORT(SceUID, sceKernelCreateSema, const char *name, SceUInt attr, int initVal, int maxVal, SceKernelSemaOptParam *option) {
     assert(name != nullptr);
     assert(attr == 0);
-    assert(initVal == 0);
```
petmac: Thanks. Is the minimum bound still 0? If so, might be worth adding a >= check.

Rinnegatamante: Can't really say if the minimum bound is 0 but i think so.

PR #5, opened 2018-01-28
Fix for app0 and ux0 paths without '/'

Rinnegatamante:
PSVITA supports partition referencing even without the / (eg. app0:eboot.bin).
This will properly handle such cases.


petmac:
Thanks!

PR #6, opened 2018-01-28
Fix building on Unix

TingPing:
This could go much further of course such as using `pkg-config` for all dependencies but this is a start.


petmac:
I've no means to test this but approving anyway.

PR #12, opened 2018-01-29
Added several sceGxm implementations.

Rinnegatamante:
This commit adds:

- sceGxmSetCullMode implementation
- sceGxmTextureSetData implementation.
- sceGxmTextureSetWidth implementation.
- sceGxmTextureSetHeight implementation.
- Support for multitexturing (TEXUNIT1 and biggers support for fragment shaders)
- Support for almost all SceGxm primitives (except SCE_GXM_PRIMITIVE_TRIANGLE_EDGES)

### DIFF
--- a/src/emulator/gxm/include/gxm/state.h
+++ b/src/emulator/gxm/include/gxm/state.h
@@ -16,4 +16,5 @@ namespace emu {
 
 struct GxmState {
     emu::SceGxmInitializeParams params;
+    bool isInScene = false;
 };
--- a/src/emulator/modules/SceGxm/src/SceGxm.cpp
+++ b/src/emulator/modules/SceGxm/src/SceGxm.cpp
@@ -313,6 +313,21 @@ EXPORT(int, sceGxmBeginScene, SceGxmContext *context, unsigned int flags, const 
     
     glBindFramebuffer(GL_FRAMEBUFFER, renderTarget->framebuffer[0]);
 
+    // Re-load GL machine settings for multiple contexts support
+    switch (context->cull_mode){
+    case SCE_GXM_CULL_CCW:
+        glEnable(GL_CULL_FACE);
+        glCullFace(GL_FRONT);
+        break;
+    case SCE_GXM_CULL_CW:
+        glEnable(GL_CULL_FACE);
+        glCullFace(GL_BACK);
+        break;
+    case SCE_GXM_CULL_NONE:
+        glDisable(GL_CULL_FACE);
+        break;
+    }
+    
     // TODO This is just for debugging.
     glClear(GL_COLOR_BUFFER_BIT);
 
--- a/src/emulator/modules/SceGxm/src/SceGxm.cpp
+++ b/src/emulator/modules/SceGxm/src/SceGxm.cpp
@@ -273,10 +273,9 @@ static GLenum translate_primitive(SceGxmPrimitiveType primType){
     case SCE_GXM_PRIMITIVE_POINTS:
         return GL_POINTS;
     case SCE_GXM_PRIMITIVE_TRIANGLE_EDGES: // Todo: Implement this
-    default:
         return GL_TRIANGLES;
-        break;
     }
+    return GL_TRIANGLES;
 }
 
 EXPORT(int, sceGxmAddRazorGpuCaptureBuffer) {
@@ -297,7 +296,7 @@ EXPORT(int, sceGxmBeginScene, SceGxmContext *context, unsigned int flags, const 
     assert(colorSurface != nullptr);
     assert(depthStencil != nullptr);
 
-    if (context->isInScene){
+    if (host.gxm.isInScene){
         return SCE_GXM_ERROR_WITHIN_SCENE;
     }
     if (depthStencil == nullptr && colorSurface == nullptr){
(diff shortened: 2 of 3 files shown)


src/emulator/modules/SceGxm/src/SceGxm.cpp
```diff
@@ -605,6 +624,8 @@ EXPORT(int, sceGxmEndScene, SceGxmContext *context, const emu::SceGxmNotificatio
     glReadPixels(0, 0, width, height, GL_RGBA, GL_UNSIGNED_BYTE, pixels);
     flip_vertically(pixels, width, height, stride_in_pixels);
 
+    context->isInScene = false;
```
petmac: Is this required? If not, can we delete it?

Rinnegatamante: There are some functions that can be called only while in scene or out scene. I added it for this purpose even if at the moment it's not used anywhere.

src/emulator/modules/SceGxm/src/SceGxm.cpp
```diff
@@ -278,13 +297,37 @@ EXPORT(int, sceGxmBeginScene, SceGxmContext *context, unsigned int flags, const
...
     context->vertex_ring_buffer_used = 0;
     context->color_surface = *colorSurface;
-
+    
+    context->isInScene = true;
+    
     glBindFramebuffer(GL_FRAMEBUFFER, renderTarget->framebuffer[0]);
 
+    // Re-load GL machine settings for multiple contexts support
+    switch (context->cull_mode){
```
petmac: Is this needed by anything?

src/emulator/modules/SceGxm/src/gxm.h
```diff
@@ -65,6 +65,8 @@ struct SceGxmContext {
     emu::SceGxmColorSurface color_surface;
     const SceGxmVertexProgram *vertex_program = nullptr;
     GLObjectArray<1> texture;
+    bool isInScene = false;
```
petmac: Is it possible we can infer this from the color_surface field? It would be nice to not have extra fields that need kept in sync. There could be a helper function `is_in_scene` that derives it from the Context state.

src/emulator/modules/SceGxm/src/SceGxm.cpp
```diff
@@ -278,13 +297,37 @@ EXPORT(int, sceGxmBeginScene, SceGxmContext *context, unsigned int flags, const
     assert(colorSurface != nullptr);
     assert(depthStencil != nullptr);
 
+    if (context->isInScene){
+        return SCE_GXM_ERROR_WITHIN_SCENE;
+    }
+    if (depthStencil == nullptr && colorSurface == nullptr){
```
petmac: Should this be an `||`? We assert both are present.

src/emulator/modules/SceGxm/src/SceGxm.cpp
```diff
@@ -260,6 +260,25 @@ static GLenum translate_format(SceGxmTextureFormat src) {
...
+    case SCE_GXM_PRIMITIVE_TRIANGLE_STRIP:
+        return GL_TRIANGLE_STRIP;
+    case SCE_GXM_PRIMITIVE_TRIANGLE_FAN:
+        return GL_TRIANGLE_FAN;
+    case SCE_GXM_PRIMITIVE_LINES:
+        return GL_LINES;
+    case SCE_GXM_PRIMITIVE_POINTS:
+        return GL_POINTS;
+    case SCE_GXM_PRIMITIVE_TRIANGLE_EDGES: // Todo: Implement this
+    default:
```
petmac: Please remove the default case and make the function return the default GL_TRIANGLES. That way, Clang warns if/when not all switch cases are handled.

src/emulator/modules/SceGxm/src/SceGxm.cpp
```diff
@@ -566,12 +609,14 @@ EXPORT(int, sceGxmDisplayQueueFinish) {
...
-    assert(indexType == SCE_GXM_INDEX_FORMAT_U16);
     assert(indexData != nullptr);
     assert(indexCount > 0);
 
-    const GLenum mode = primType == SCE_GXM_PRIMITIVE_TRIANGLES ? GL_TRIANGLES : GL_TRIANGLE_STRIP;
+    if (!context->isInScene){
+        return SCE_GXM_ERROR_NOT_WITHIN_SCENE;
+    }
+    
+    GLenum mode = translate_primitive(primType);
```
petmac: Add const please.

src/emulator/modules/SceGxm/src/SceGxm.cpp
Rinnegatamante: We need this to handle multiple contexts properly since several commercial games use multiple direct contexts (eg. Duke Nukem 3D).

src/emulator/gxm/include/gxm/state.h
```diff
@@ -16,4 +16,5 @@ namespace emu {
 
 struct GxmState {
     emu::SceGxmInitializeParams params;
+    bool isInScene = false;
```
petmac: Can games draw to more than one context at the same time? Maybe this variable should be in the context after all.

Rinnegatamante: No, sceGxmBeginScene can't be called if a context is already in scene. It's not possible to have multiple scenes active at the same moment.

PR #17, opened 2018-01-30
Add SceKernelGetProcessTimeLow

frangarcj:


src/emulator/modules/SceLibKernel/src/SceLibKernel.cpp
```diff
@@ -886,6 +882,14 @@ EXPORT(SceUInt64, sceKernelGetProcessTimeWide) {
     return clocks * scale;
 }
 
+EXPORT(int, sceKernelGetProcessTimeLow) {
```
petmac: Thanks! I think the return type is SceUInt32.

PR #18, opened 2018-01-30
Add some TTY support

frangarcj:


src/emulator/io/src/io.cpp
```diff
@@ -184,6 +194,15 @@ int read_file(void *data, IOState &io, SceUID fd, SceSize size) {
         return fread(data, 1, size, file->second.get());
     }
 
+    const TtyFiles::const_iterator tty_file = io.tty_files.find(fd);
+    if (tty_file != io.tty_files.end()) {
+        if(tty_file->second == TTY_IN){
+            std::cin.read(reinterpret_cast<char *>(data), size);
+            return size;
```
petmac: This can probably be a static cast, no need to reinterpret.

PR #20, opened 2018-01-30
Initial implementation for some scePower functions.

Rinnegatamante:


VelocityRa:
@petmac Maybe this process should be documented in the Wiki?

petmac:
Agreed, it should. I mentioned "How module export stubs are generated from db.yml" in #51 which I should expand to cover this too.

src/emulator/modules/ScePower/src/ScePower.cpp
```diff
@@ -140,31 +140,49 @@ EXPORT(int, scePowerRequestSuspend) {
...
+    if (freq < 0){
+        return SCE_POWER_ERROR_INVALID_VALUE;
+    }
+    return 0;
 }
 
-EXPORT(int, scePowerSetConfigurationMode) {
-    return unimplemented("scePowerSetConfigurationMode");
+EXPORT(int, scePowerSetConfigurationMode, int mode) {
+    if (mode != 0x80 && mode != 0x800 && mode != 0x10880){
```
petmac: Do we have named constants for these values?

Rinnegatamante: vita-headers doesn't have them. I grabbed those values with IDA Pro.

PR #25, opened 2018-01-31
Added sceIoMkdir implementation.

Rinnegatamante:


src/emulator/io/src/io.cpp
```diff
@@ -269,3 +269,39 @@ void close_file(IOState &io, SceUID fd) {
...
 }
+
+int create_dir(const char *dir, int mode, const char *pref_path){
+    // TODO Hacky magic numbers.
+    assert((strncmp(dir, "ux0:", 4) == 0) || (strncmp(dir, "uma0:", 5) == 0));
+    if (strncmp(dir, "ux0:", 4) == 0) {
+        std::string dir_path = pref_path;
+        dir_path += "ux0/";
+		
+        int i = 4;
```
petmac: I think we're getting to the stage where we have too much duplication of this path conversion code, and we need to formalise it into a few functions.

PR #36, opened 2018-02-04
Better Logging (to file, console, VS Output)

### LINKED ISSUE #34: Improved logging
Log to file so messages can be seen after emulator is closed.

VelocityRa:
## Description

This improves logging using [spdlog](https://github.com/gabime/spdlog) and other than console output, it adds logging to a file (`vita3k.log` in the path of the executable) and Visual Studio's Output tab.

Existing logging code is ported to the new system.

**Copy from main commit:**
Defines `LOG_*` macros for different log levels, where * is either TRACE, DEBUG, INFO, WARN, ERROR or CRITICAL.

Additionally, it defines `LOG_*_IF` (same log levels), for conditional logging, with the first argument being the condition.

Resolves #34.

## Screenshot

Console:
image

Similar output for the VS Output tab.
File output is the same (without color obviously).

As you can see I log the **time**, **log level** (also denoted by the color) and the **calling function name**.

---
Non-windows code path _should_ work, but is untested.

Used my best judgmement to convert existing logging calls to the appropriate level. They can be tweaked as we see fit.

---

## Possible future work
* Make TTY output get saved to a different log channel and file.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -22,3 +22,6 @@
 [submodule "src/external/sdl2-cmake-scripts"]
 	path = src/external/sdl2-cmake-scripts
 	url = https://github.com/tcbrindle/sdl2-cmake-scripts.git
+[submodule "src/external/spdlog"]
+	path = src/external/spdlog
+	url = https://github.com/gabime/spdlog.git
--- a/src/emulator/audio/include/audio/functions.h
+++ b/src/emulator/audio/include/audio/functions.h
@@ -23,6 +23,6 @@
 
 struct AudioState;
 
-typedef std::function<void(SceUID)> ResumeThread;
+typedef std::function<void(SceUID)> ResumeAudioThread;
 
-bool init(AudioState &state, ResumeThread resume_thread);
+bool init(AudioState &state, ResumeAudioThread resume_thread);
--- a/src/emulator/cpu/src/cpu.cpp
+++ b/src/emulator/cpu/src/cpu.cpp
@@ -20,6 +20,7 @@
 #include <disasm/functions.h>
 #include <disasm/state.h>
 #include <mem/ptr.h>
+#include <util/log.h>
 
 #include <unicorn/unicorn.h>
 
@@ -59,12 +60,12 @@ static void code_hook(uc_engine *uc, uint64_t address, uint32_t size, void *user
     const size_t buffer_size = GB(4) - address;
     const bool thumb = is_thumb_mode(uc);
     const std::string disassembly = disassemble(state.disasm, code, buffer_size, address, thumb);
-    std::cout << std::hex << std::setw(8) << address << std::dec << " " << disassembly << std::endl;
+	LOG_TRACE("{:#08x} {}", address, disassembly);
 }
 
 static void log_memory_access(const char *type, Address address, int size, int64_t value, const MemState &mem) {
     const char *const name = mem_name(address, mem);
-    std::cout << type << " " << size << " bytes, address 0x" << std::hex << address << " (" << name << "), value 0x" << value << std::dec << std::endl;
+	LOG_TRACE("{} {} bytes, address {:#08x} ( {} ), value {:#x}", type, size, address, name, value);
 }
 
 static void read_hook(uc_engine *uc, uc_mem_type type, uint64_t address, int size, int64_t value, void *user_data) {
--- a/src/emulator/host/src/host.cpp
+++ b/src/emulator/host/src/host.cpp
@@ -27,6 +27,7 @@
 #include <kernel/thread_state.h>
 #include <nids/functions.h>
 #include <util/lock_and_find.h>
+#include <util/log.h>
 
 #include <SDL_events.h>
 #include <SDL_filesystem.h>
@@ -58,7 +59,7 @@ bool init(HostState &state) {
     const std::unique_ptr<char, void (&)(void *)> base_path(SDL_GetBasePath(), SDL_free);
     const std::unique_ptr<char, void (&)(void *)> pref_path(SDL_GetPrefPath(org_name, app_name), SDL_free);
 
-    const ResumeThread resume_thread = [&state](SceUID thread_id) {
+    const ResumeAudioThread resume_thread = [&state](SceUID thread_id) {
         const ThreadStatePtr thread = lock_and_find(thread_id, state.kernel.threads, state.kernel.mutex);
         const std::unique_lock<std::mutex> lock(thread->mutex);
         assert(thread->to_do == ThreadToDo::wait);
@@ -91,8 +92,7 @@ bool handle_events(HostState &host) {
 void call_import(HostState &host, uint32_t nid, SceUID thread_id) {
     if (LOG_IMPORT_CALLS) {
         const char *const name = import_name(nid);
-        const char prev_fill = std::cout.fill();
-        std::cout << "NID " << std::hex << std::setw(8) << std::setfill('0') << nid << std::setfill(prev_fill) << std::dec << " (" << name << ") called." << std::endl;
+		LOG_TRACE("NID {:#08x} ({})) called", nid, name);
     }
 
     ImportFn *const fn = resolve_import(nid);
--- a/src/emulator/io/CMakeLists.txt
+++ b/src/emulator/io/CMakeLists.txt
@@ -8,3 +8,4 @@ add_library(
 
 target_include_directories(io PUBLIC include)
 target_link_libraries(io PUBLIC miniz vita-headers)
+target_link_libraries(io PRIVATE util)
--- a/src/emulator/io/src/io.cpp
+++ b/src/emulator/io/src/io.cpp
@@ -18,6 +18,7 @@
 #include <io/functions.h>
 
 #include <io/state.h>
+#include <util/log.h>
 
 #include <psp2/io/fcntl.h>
 
@@ -141,7 +142,7 @@ SceUID open_file(IOState &io, const char *path, int flags, const char *pref_path
         int i = 5;
         if (path[5] == '/') i++;
         const ZipFilePtr file = open_zip(*io.vpk, &path[i]);
-		
+
         if (!file) {
             return -1;
         }
@@ -165,7 +166,6 @@ SceUID open_file(IOState &io, const char *path, int flags, const char *pref_path
         return fd;
     } else if (strncmp(path, "uma0:", 5) == 0) {
         std::string file_path = translate_path("uma0", path, pref_path);
-
         const char *const open_mode = translate_open_mode(flags);
         const FilePtr file(fopen(file_path.c_str(), open_mode), delete_file);
         if (!file) {
@@ -222,7 +222,7 @@ int write_file(SceUID fd, const void *data, SceSize size, const IOState &io) {
     if (tty_file != io.tty_files.end()) {
         if(tty_file->second == TTY_OUT){
             std::string s(reinterpret_cast<char const*>(data),size);
-            std::cout << s;
+            LOG_INFO("*** TTY: {}", s);
             return size;
         }
 
--- a/src/emulator/load_self.cpp
+++ b/src/emulator/load_self.cpp
@@ -20,6 +20,7 @@
 #include "relocation.h"
 
 #include <nids/functions.h>
+#include <util/log.h>
 
 #include <elfio/elf_types.hpp>
 #define SCE_ELF_DEFS_TARGET
@@ -45,8 +46,7 @@ static bool load_func_imports(const uint32_t *nids, const Ptr<uint32_t> *entries
 
         if (LOG_IMPORTS) {
             const char *const name = import_name(nid);
-            const char prev_fill = std::cout.fill();
-            std::cout << "\tNID " << std::hex << std::setw(8) << std::setfill('0') << nid << std::setfill(prev_fill) << " (" << name << ") at 0x" << entry.address() << std::dec << std::endl;
+			LOG_DEBUG( "\tNID {:#08x} ({}) at {:#x}", nid, name, entry.address());
         }
 
         uint32_t *const stub = entry.get(mem);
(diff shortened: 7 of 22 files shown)


VelocityRa:
Travis built successfully, this PR is now ready for review.

src/emulator/io/CMakeLists.txt
```diff
@@ -7,4 +7,4 @@ add_library(
 )
 
 target_include_directories(io PUBLIC include)
-target_link_libraries(io PUBLIC miniz vita-headers)
+target_link_libraries(io PUBLIC miniz spdlog util vita-headers)
```
petmac: `PRIVATE util` should work here too.

src/emulator/mem/CMakeLists.txt
```diff
@@ -7,3 +7,4 @@ add_library(
 )
 
 target_include_directories(mem PUBLIC include)
+target_link_libraries(mem PUBLIC util spdlog)
```
petmac: This can probably be removed until `mem` logs something.

src/emulator/module/CMakeLists.txt
```diff
@@ -10,5 +10,5 @@ add_library(
 )
 
 target_include_directories(module PUBLIC include)
-target_link_libraries(module PUBLIC host microprofile)
+target_link_libraries(module PUBLIC host microprofile util spdlog)
```
petmac: This can probably be reverted if `module` doesn't need logging stuff in a public header.

src/emulator/module/src/module.cpp
```diff
@@ -20,6 +20,7 @@
 #include <iostream>
 #include <mutex>
 #include <set>
+#include "util/log.h"
```
petmac: Could this include be moved up with the other project includes and changed to use angle brackets?

src/emulator/modules/CMakeLists.txt
```diff
@@ -259,3 +259,5 @@ SceVideodec
 SceVoice
 SceVoiceQoS
 SceVshBridge)
+
+target_link_libraries(modules INTERFACE util spdlog)
```
petmac: This can be reverted.

`modules` is just an `INTERFACE` library that groups the existing module libraries into one for easier dependency management, it doesn't need any dependencies itself.

This file is also automatically generated, so we shouldn't edit it without editing `gen-modules` too.

src/emulator/modules/SceGxm/src/SceGxm.cpp
```diff
@@ -18,6 +18,7 @@
 #include <SceGxm/exports.h>
 
 #include "gxm.h"
+#include "util/log.h"
```
petmac: Angle brackets here too please. Sorry!

src/emulator/modules/SceGxm/src/SceGxm.cpp
```diff
@@ -57,7 +58,7 @@ static void before_callback(const FunctionCall &fn) {
 static void after_callback(const FunctionCall &fn) {
     MICROPROFILE_LEAVE();
     for (GLenum error = glGetError(); error != GL_NO_ERROR; error = glGetError()) {
-        std::cerr << "OpenGL: " << fn.function->name() << " set error " << error << "." << std::endl;
+        LOG_ERROR("OpenGL: {} set error {}.", fn.function->name(), static_cast<unsigned int>(error));
```
petmac: `glbinding` does some kind of magic to convert OpenGL errors to strings here. Is that preserved? The cast to `unsigned int` suggests it's not.

src/emulator/modules/SceGxm/src/SceGxm.cpp
petmac: Come to think of it, rather than using `LOG_` macros directly, there should be a module wrapper around logging, which inserts the function name without the `export_` prefix.

There's some prerequisite work I need to do here in bridging, because I had some difficulty getting the export name to pass through as a template parameter with Visual Studio and just nuked it when I ran out of time during a train journey.

src/emulator/relocation.cpp
```diff
@@ -16,6 +16,7 @@
 // 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
 #include "relocation.h"
+#include "util/log.h"
```
petmac: Angle brackets too here please, sorry.

src/emulator/relocation.cpp
```diff
@@ -192,7 +193,7 @@ static bool relocate(void *data, Code code, uint32_t s, uint32_t a, uint32_t p)
         return true;
     }
 
-    std::cerr << "Unhandled relocation code " << code << "." << std::endl;
+    LOG_ERROR("Unhandled relocation code {}.", code);
```
petmac: This should actually be a warning. Some homebrew has unhandled codes and they seem to be OK to ignore.

src/emulator/util/CMakeLists.txt
```diff
@@ -1,6 +1,12 @@
 add_library(
 	util
-	INTERFACE
+	STATIC
```
petmac: I think the initialisation code to set up the logger actually belongs in `emulator` because nothing else needs to call it, and I think we should be passing the logger pointer around to wherever needs it rather than having it as a global.

src/emulator/util/include/util/log.h
```diff
@@ -0,0 +1,49 @@
...
+#include <memory>
+#include <string>
+
+// needed for cross-platform getcwd
+#ifdef _WIN32
+#include <direct.h>
+#define getcwd _getcwd // stupid MSFT "deprecation" warning
+#else
+#include <unistd.h>
+#endif
```
petmac: The directories to put things in are available from SDL. We want the "pref path."

Currently `host`'s `init` function puts it in HostState, but we could get it earlier if we need to.

src/external/CMakeLists.txt
```diff
@@ -27,6 +27,9 @@ set(capstone_INCLUDE_DIRS "${CMAKE_CURRENT_SOURCE_DIR}/capstone/include" PARENT_
 add_library(elfio INTERFACE)
 target_include_directories(elfio INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}/elfio")
 
+add_library(spdlog INTERFACE)
+target_include_directories(spdlog INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}/spdlog/include")
```
petmac: Any defines we need can be added here.

.gitmodules
```diff
@@ -22,3 +22,6 @@
 [submodule "src/external/sdl2-cmake-scripts"]
 	path = src/external/sdl2-cmake-scripts
 	url = https://github.com/tcbrindle/sdl2-cmake-scripts.git
+[submodule "src/external/spdlog"]
+	path = src/external/spdlog
+	url = ../../gabime/spdlog.git
```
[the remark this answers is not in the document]
VelocityRa: I'm not sure why git did that... will fix.

src/emulator/audio/include/audio/functions.h
```diff
@@ -23,6 +23,6 @@
 
 struct AudioState;
 
-typedef std::function<void(SceUID)> ResumeThread;
+typedef std::function<void(SceUID)> ResumeAudioThread;
```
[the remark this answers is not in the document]
VelocityRa: Nope, it was some other header in the Windows SDK, I don't remember which, I can check if you want.

src/emulator/cpu/src/cpu.cpp
```diff
@@ -37,7 +38,7 @@ struct CPUState {
     UnicornPtr uc;
 };
 
-static const bool LOG_CODE = true;
+static const bool LOG_CODE = false;
```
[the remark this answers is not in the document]
VelocityRa: This shouldn't be here, must have slipped in at my latest push, sorry...

src/emulator/glutil/CMakeLists.txt
```diff
@@ -9,3 +9,4 @@ add_library(
 
 target_include_directories(glutil PUBLIC include)
 target_link_libraries(glutil PUBLIC glbinding)
+target_link_libraries(glutil PUBLIC util spdlog)
```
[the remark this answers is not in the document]
VelocityRa: Futureproofing? I just added it everywhere so people don't have to think about it if/when they need to log stuff from anywhere. Same goes for `disasm`.

I'll delete it and it can be added if you think so too.

src/emulator/host/src/host.cpp
```diff
@@ -36,7 +37,7 @@
 #include <iomanip>
 #include <iostream>
 
-static const bool LOG_IMPORT_CALLS = false;
+static const bool LOG_IMPORT_CALLS = true;
```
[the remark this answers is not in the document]
VelocityRa: yea slipped in sorry

src/emulator/gxm/CMakeLists.txt
```diff
@@ -4,4 +4,5 @@ add_library(
 )
 
 target_include_directories(gxm INTERFACE include)
-target_link_libraries(gxm INTERFACE mem)
+target_link_libraries(gxm INTERFACE mem util spdlog)
+target_link_libraries(gxm INTERFACE util spdlog)
```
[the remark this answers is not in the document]
VelocityRa: *facepalm*

src/emulator/io/src/io.cpp
```diff
@@ -129,7 +130,7 @@ SceUID open_file(IOState &io, const char *path, int flags, const char *pref_path
         int i = 5;
         if (path[5] == '/') i++;
         const ZipFilePtr file = open_zip(*io.vpk, &path[i]);
-		
+
```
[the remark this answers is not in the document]
VelocityRa: I'm on Windows, will do.

src/emulator/load_self.cpp
```diff
@@ -36,7 +37,7 @@
 
 using namespace ELFIO;
 
-static const bool LOG_IMPORTS = false;
+static const bool LOG_IMPORTS = true;
```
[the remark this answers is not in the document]
VelocityRa: The calls are `LOG_DEBUG` so I suppose. Could demote one of them to `LOG_TRACE` if they're too noisy anyway.

src/emulator/mem/CMakeLists.txt
VelocityRa: CMake and particular structure of this project isn't necessarily apparent to everyone working on this project... so I thought I'd do this for futureproofing, it doesn't hurt anything, only saves time for future development. Same for a couple other modules I've mentioned.

src/emulator/module/src/module.cpp
VelocityRa: why yes it can

src/emulator/modules/CMakeLists.txt
VelocityRa: thank god it's autogenerated

src/emulator/modules/SceGxm/src/SceGxm.cpp
VelocityRa: yessir

src/emulator/modules/SceGxm/src/SceGxm.cpp
VelocityRa: Ohh I didn't know that. Guess I'll have to make a stringstream and feed it that way then, cool.

src/emulator/util/CMakeLists.txt
VelocityRa: Honestly I've tried doing that before and it's... not great. I think if we should have one global, logging has to be it. It just makes everything much easier and more readable.
I think using a global is justified in that case and every project I've come across does it this way.

src/emulator/util/include/util/log.h
VelocityRa: Wouldn't this put stuff in AppData on Windows? As a Windows user, that's terrible
It's convenient at all, in fact I'd expect the uma0 and ux0 folders to be in the same directory as the emulator too, or at least close to it, but that's another discussion.

These are all files that someone using this emulator needs to have access to, and this directory is hard to reach.

src/emulator/util/include/util/log.h
```diff
@@ -0,0 +1,49 @@
...
+
+
+namespace spd = spdlog;
```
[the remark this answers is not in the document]
VelocityRa: I'll remove it but well, the official example code did it...

src/emulator/util/include/util/log.h
```diff
@@ -0,0 +1,49 @@
...
+extern std::shared_ptr<spd::logger> g_logger;
```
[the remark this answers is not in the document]
VelocityRa: Honestly I've tried doing that before and it's... not great. I think if we should have one global, logging has to be it. There can be many places where you want to log but you won't have access to it. It just makes everything much easier and more readable.
I think using a global is justified in that case and almost every project I've come across does it this way.

src/emulator/util/include/util/log.h
```diff
@@ -0,0 +1,49 @@
...
+
+static const
+#ifdef _WIN32
+std::wstring& LOG_FILE_NAME = L"\\vita3k.log";
+#else
+std::string& LOG_FILE_NAME = "\\vita3k.log";
+#endif
```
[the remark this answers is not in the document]
VelocityRa: Hm probably not, I'll change it to a forward slash.
Windows, always doing things their own way.

src/emulator/util/include/util/log.h
```diff
@@ -0,0 +1,49 @@
...
+
+// Initialize logging
+extern void init_logging();
```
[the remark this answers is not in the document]
VelocityRa: Let's chat about the globals thing and depending on what we decide I may do that.

src/emulator/util/src/util.cpp
```diff
@@ -0,0 +1,70 @@
+#include "util/log.h"
+
+std::shared_ptr<spd::logger> g_logger;
+
+void init_logging()
+{
+	//! Whether we should log to Visual Studio's Output tab
```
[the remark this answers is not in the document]
VelocityRa: Yeah, most of this file is copied from an old project of mine...

src/emulator/util/src/util.cpp
```diff
@@ -0,0 +1,70 @@
...
+	static constexpr bool LOG_MSVC_OUTPUT = true;
+
+	std::vector<spdlog::sink_ptr> sinks;
+
+	// console window
+#ifdef _WIN32
```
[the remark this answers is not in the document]
VelocityRa: Seems like `WIN32` is defined by the SDK and `_WIN32` by the compiler, I'll just change it since you're using `WIN32` everywhere else.

VelocityRa:
Pushed as one commit for now, when everything's all good I'll separate everything to commits properly.

VelocityRa:
Any update on this? The sooner this particular PR is merged, the better.

PR #38, opened 2018-02-04
Add AppVeyor Configuration File

Nicba1010:
This will make it much easier for the general emulator userbase to get and test the emulator, because they won't have to build it themselves.


suhoparoff2017:
Thank you

PR #39, opened 2018-02-04
Added networking support.

Rinnegatamante:


Rinnegatamante:
@petmac need your help to solve current build error (look at travis or appveyor output).

src/emulator/net/src/net.cpp
```diff
@@ -0,0 +1,160 @@
...
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+#include <net/functions.h>
+
+#include <net/state.h>
+
+#include <cassert>
```
petmac: There aren't any asserts so this line can be deleted.

src/emulator/net/src/net.cpp
```diff
@@ -0,0 +1,160 @@
...
+#include <cstring>
+
+static void convertSceSockaddrToPosix(struct SceNetSockaddr *src, struct sockaddr *dst){
```
petmac: Can src be const? We could skip some casts later if so.

src/emulator/net/src/net.cpp
```diff
@@ -0,0 +1,160 @@
...
+    dst->sa_family = src->sa_family;
+    memcpy(&dst->sa_data, &src->sa_data, 14);
+}
+
+static void convertPosixSockaddrToSce(struct sockaddr *src, struct SceNetSockaddr *dst){
+    dst->sa_family = src->sa_family;
```
petmac: Can src be const? We could skip some casts later if so.

src/emulator/net/src/net.cpp
```diff
@@ -0,0 +1,160 @@
...
+    }
+    return -1;
+}
+
+int accept_socket(NetState &net, int id, SceNetSockaddr *name, unsigned int *addrlen){
+    const sockets::const_iterator socket = net.socks.find(id);
+    if (socket != net.socks.end()) {
+        struct sockaddr addr;
+        abs_socket new_socket = accept(socket->second, &addr, (socklen_t*)addrlen);
+#ifdef _MSC_VER
```
petmac: WIN32

src/emulator/net/include/net/state.h
```diff
@@ -0,0 +1,55 @@
...
+#include <vector>
+
+#include <net/functions.h>
+
+#include <psp2/types.h>
+
+#ifdef _MSC_VER
+# define NOMINMAX
+# define WIN32_LEAN_AND_MEAN
+# define _WINSOCK_DEPRECATED_NO_WARNINGS
```
petmac: These defines can be put in the CMakeLists.txt that sets up the Winsock interface library. I do something similar for microprofile, to make sure the settings are consistent whenever it's used.

src/external/CMakeLists.txt
```diff
@@ -52,6 +52,15 @@ add_library(sdl2 INTERFACE)
 target_include_directories(sdl2 INTERFACE "${SDL2_INCLUDE_DIR}")
 target_link_libraries(sdl2 INTERFACE "${SDL2_LIBRARY}")
 
+if(WIN32)
+	add_library(wsock32 INTERFACE)
+	find_library(WSOCK32 wsock32)
+	target_link_libraries(wsock32 INTERFACE WSOCK32)
+	add_library(ws2_32 INTERFACE)
+	find_library(WS2_32 ws2_32)
+	target_link_libraries(ws2_32 INTERFACE WS2_32)
```
petmac: I'd probably combine both of these into the same library.

Do we need both anyway? Can we just use Winsock 1 or 2 but not both?

petmac:
Regarding the errors related to bridging `sceNetInetNtop`. These are because there's no code to bridge from a C++ pointer (which is memory in macOS or Windows land) to the Vita, because that pointer is not necessarily in the memory that's mapped to the Vita.

You need to allocate memory that the Vita can see (check mem.h), and return that pointer as type `Ptr<const char *>` instead. So you'll need to change the return type of your `sceNetInetNtop` export.

Rinnegatamante:
The error is not only related to sceNetInetNtop it seems, i get another one for an unsigned short return it seems.

petmac:
You can add another case in https://github.com/Vita3K/Vita3K/blob/master/src/emulator/module/src/bridge_return.cpp

I just added them as I needed more return types, and I didn't need many.

src/emulator/net/src/net.cpp
Rinnegatamante: Making this const will just add even more casts since that function is called only for converting a temp non-const (since it came from a previous call of convertSceSockaddrToPosix) struct to a Sce struct as result.

src/emulator/host/CMakeLists.txt
```diff
@@ -12,5 +12,10 @@ version.cpp
 configure_file(src/version.cpp.in version.cpp)
 
 target_include_directories(host PUBLIC include)
-target_link_libraries(host PUBLIC audio cpu ctrl gxm io kernel mem)
-target_link_libraries(host PRIVATE nids sdl2)
+target_link_libraries(host PUBLIC audio cpu ctrl gxm io kernel mem net)
+if (WIN32)
```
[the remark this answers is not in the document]
Rinnegatamante: Without net here, host fails to find net/state.h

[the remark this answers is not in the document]
petmac: You need net (for state, as you say), but you don't need winsock.

Rinnegatamante:
Regarding the allocation of a Ptr<const char *>. Wouldn't this cause memleak since the function is supposed to return a temp ptr that will be erased at next call?

src/emulator/net/include/net/state.h
```diff
@@ -0,0 +1,44 @@
...
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+#pragma once
+
+#include <map>
+#include <memory>
+#include <vector>
```
petmac: memory and vector aren't used.

src/emulator/modules/SceNet/src/SceNet.cpp
```diff
@@ -105,60 +209,102 @@ EXPORT(int, sceNetGetpeername) {
...
-    return unimplemented("sceNetInetNtop");
+EXPORT(Ptr<const char>, sceNetInetNtop, int af, const void *src, Ptr<char> dst, unsigned int size) {
+    char *dst_ptr = dst.get(host.mem);
+#ifdef WIN32
+    const char *res = InetNtop(af, src, dst_ptr, size);
+#else
+    const char *res = inet_ntop(af, src, dst_ptr, size);
+#endif
+    if (res == nullptr){
+        return error("sceNetInetNtop", translate_errorcode());
```
petmac: This needs to return a `Ptr<char>()` (null pointer in Ptr language) rather than translating the error code.

src/emulator/net/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+STATIC
+include/net/functions.h
+include/net/state.h
+src/net.cpp
+)
+
+target_include_directories(net PUBLIC include)
+target_link_libraries(net PUBLIC sdl2 vita-headers)
+if (WIN32)
+    target_link_libraries(net PRIVATE winsock)
```
petmac: Oh, this should be `PUBLIC` because net/state.h includes winsock headers. I wonder would that fix the define issue.

Then again, if we can get rid of the headers out of the net/state.h header, we could keep this PRIVATE.

src/emulator/net/include/net/state.h
```diff
@@ -0,0 +1,44 @@
...
+# include <Ws2tcpip.h>
+typedef SOCKET abs_socket;
+typedef int socklen_t;
+#else
+# include <unistd.h>
+# include <sys/socket.h>
+# include <netinet/in.h>
+# include <netdb.h>
+# include <arpa/inet.h>
+typedef uint32_t abs_socket;
```
petmac: Hardly any of these includes need to be in this public header. int is a built-in type and uint32_t comes from `<cstdint>`. I'm a bit wary of accidentally including Windows headers into the host state header, because that will cause the whole emulator to take a lot longer to build.

Maybe we can just use an int instead of abs_socket?

Rinnegatamante: Windows uses SOCKET that's a struct as result for socket(). Don't get your point here, sorry.

petmac: The point was just to remove as many includes as possible. Looks like we don't need any here on macOS, but we need to include something on Windows.

Sorry for being annoying about this.

Rinnegatamante: Without those includes, we don't have required definitions on unix for sockets prototypes.

src/emulator/modules/SceNet/src/SceNet.cpp
```diff
@@ -241,18 +244,19 @@ EXPORT(Ptr<const char>, sceNetInetNtop, int af, const void *src, Ptr<char> dst,
 #else
     const char *res = inet_ntop(af, src, dst_ptr, size);
 #endif
-    if (res == nullptr){
```
petmac: I think you still want to check this, and return the equivalent null Ptr:

`Ptr<char>()`

src/emulator/modules/SceAppUtil/src/SceAppUtil.cpp
```diff
@@ -185,12 +185,14 @@ EXPORT(int, sceAppUtilSystemParamGetInt) {
 }
 
 EXPORT(int, sceAppUtilSystemParamGetString, unsigned int paramId, SceChar8 *buf, SceSize bufSize) {
+    char devname[80];
```
petmac: Should we reserve a buffer as big as `bufSize`?

Zangetsu38:
big conflict, need rebase and squach

Zangetsu38:
PR is wrong, log is already on master, need refactoreing commit and rmove WIP commit

PR #40, opened 2018-02-04
Added SCE_CTRL_PSBUTTON key and controller bindings

0nza1101:


petmac:
Thanks for this. Does it work OK?

I wonder why I missed that button.

0nza1101:
Yes it's working, tested with a simple homebrew.

PR #43, opened 2018-02-05
Improve git revision and add app version

Zangetsu38:
Improve git version and set v0.1 for version emulator, and fix also name "Development version" in Windows build if git missing.
Preview :
capture d ecran 2018-02-05 20 25 00


src/emulator/CMakeLists.txt
```diff
@@ -1,10 +1,28 @@
 set(VITA3K_ORG_NAME "Vita3K")
 set(VITA3K_APP_NAME "Vita3K")
 
-set(VITA3K_VERSION $ENV{TRAVIS_BRANCH})
+if(APPLE)
```
petmac: I don't think we need to switch on platform here.

We should be able to use similar logic for all platforms, and just get the things that change per CI provider up front.

src/emulator/CMakeLists.txt
```diff
@@ -1,10 +1,28 @@
...
+			OUTPUT_STRIP_TRAILING_WHITESPACE)
+		execute_process(
+			COMMAND git rev-parse --short HEAD
+			WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+			OUTPUT_VARIABLE GIT_HASH
+			OUTPUT_STRIP_TRAILING_WHITESPACE)
+			if(GIT_HASH STREQUAL "")
+				set(VITA3K_VERSION "Development version")
+			else()
+				set(VITA3K_VERSION "v0.1 ${GIT_COUNT}-${GIT_HASH}")
```
petmac: I think I'd prefer the version numbers to come from CI rather than manual git inspection.

https://www.appveyor.com/docs/environment-variables/
`APPVEYOR` exists and is true when building via AppVeyor.
`APPVEYOR_REPO_TAG_NAME` says the tag name. If that's undefined then `APPVEYOR_REPO_BRANCH` says the branch.
`APPVEYOR_BUILD_NUMBER` says the build number, but I'm not sure why we care.
`APPVEYOR_REPO_COMMIT` says the commit hash, but again, I don't think we need to display it.

Similarly for Travis:
https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables
`TRAVIS` is set to true
`TRAVIS_BRANCH` is the branch name, or the tag name when the build is a tag (release.)
`TRAVIS_BUILD_NUMBER` is the build number, but I don't think we need to display it.
I think `TRAVIS_COMMIT` is the commit hash, but I don't think we need to display it.

So I think I'd probably do something like this:

`set(VITA3K_VERSION "development version")`
`if AppVeyor is set`
`set VITA3K_VERSION from tag if it's set, branch if it's not`
`else if travis is set`
`set VITA3K_VERSION from TRAVIS_BRANCH`

src/emulator/CMakeLists.txt
Zangetsu38: No is needed for separed travis mac and Windows/appveyor

src/emulator/CMakeLists.txt
Zangetsu38: Appveyor code is no needed, like i have say, appveyor build works like this and fix regression for git version, and show exactly same think me with compile myself Visual Studio.
For travis mac, no idea, if you can test, but normally is same think travis

Zangetsu38:
I have added change for bin folder on travis, and for appveyor is better like that, it is universal works for Windows compile and Appveyor, so is good

src/emulator/CMakeLists.txt
Zangetsu38: because i don't kinow if "set(VITA3K_VERSION "")" works for travis, it is not me have set "set(VITA3K_VERSION $ENV{TRAVIS_BRANCH})", so i have juste thinked is needed for travis

src/emulator/CMakeLists.txt
petmac: OK, thanks for the explanation. I appreciate the preference for having one solution to the problem that works on both AppVeyor and Travis.

Would it be possible to get the version number from the current git tag? Something like `git describe --abbrev=0 --tags` should work.

I'm not sure what that returns if the current commit doesn't have a tag.

We should be able to get rid of the `if(APPLE)` and `if(VITA3K_VERSION STREQUAL "")` checks then, and just call execute_process from the outermost level.

Thanks again for this.

petmac:
I wanted to ask, what use is the build number?

src/emulator/CMakeLists.txt
Zangetsu38: no, i have tested "git describe --abbrev=0 --tags" on my git and don't works like you see so is bad
capture d ecran 2018-02-05 19 38 48

And for other, i don't know, if you can test my code for see if works, and after teste without If WIN32 and if APPLE for see if works same for you.
but for "if(VITA3K_VERSION STREQUAL "")" io think is obligatory, for user not have git or user gitwindows only like that return Devlopement version" and no empty

Zangetsu38:
you speed what exactly ? v0.1 or got git vesion ?

Zangetsu38:
Small refactoring for more clear for separate app version en git revision

src/emulator/host/include/host/version.h
```diff
@@ -2,4 +2,5 @@
 
 extern const char org_name[];
 extern const char app_name[];
+extern const char app_version[];
```
petmac: This isn't used anywhere. Should it be?

src/emulator/CMakeLists.txt
```diff
@@ -1,10 +1,22 @@
 set(VITA3K_ORG_NAME "Vita3K")
 set(VITA3K_APP_NAME "Vita3K")
+set(VITA3K_APP_VERSION "v0.1")
```
petmac: We should build this from the current git tag. That way, the tag is the source of truth and we don't have to manually keep this file updated with the tag string.

Zangetsu38:
Rebased

src/emulator/CMakeLists.txt
Zangetsu38: Hmm yes i think is good idea, but caused problem also for built without git, and i don't really know who set this with git tag hmm

PR #46, opened 2018-02-06
Preventing column-major matrices to be double transposed.

Rinnegatamante:
This also allows us to not double transpose the world-view-projection matrix when a shader with that matrix is set as column-major.


src/emulator/shaders/6959385632856496271.glsl
```diff
@@ -7,8 +7,16 @@ attribute vec2 aTexcoord;
...
+  return mat4(m[0][0], m[1][0], m[2][0], m[3][0],
+              m[0][1], m[1][1], m[2][1], m[3][1],
+              m[0][2], m[1][2], m[2][2], m[3][2],
+              m[0][3], m[1][3], m[2][3], m[3][3]);
+}
+
 void main()
 {
-    gl_Position = vec4(aPosition, 1) * wvp;
+    gl_Position = vec4(aPosition, 1) * transpose(wvp);
```
petmac: What I would try here instead is reordering the multiplication to be `wvp * vec4(aPosition, 1)`

src/emulator/modules/SceGxm/src/SceGxm.cpp
```diff
@@ -1205,7 +1191,7 @@ EXPORT(int, sceGxmSetUniformDataF, void *uniformBuffer, const SceGxmProgramParam
         break;
 
     case 16:
-        glUniformMatrix4fv(location, 1, GL_TRUE, sourceData);
+        glUniformMatrix4fv(location, 1, GL_FALSE, sourceData);
```
petmac: I think this transpose is a separate issue, and we should make a separate PR for it.

src/emulator/shaders/8854786452444500389.glsl
```diff
@@ -7,8 +7,16 @@ attribute vec4 aColor;
...
+  return mat4(m[0][0], m[1][0], m[2][0], m[3][0],
+              m[0][1], m[1][1], m[2][1], m[3][1],
+              m[0][2], m[1][2], m[2][2], m[3][2],
+              m[0][3], m[1][3], m[2][3], m[3][3]);
+}
+
 void main()
 {
-    gl_Position = vec4(aPosition, 1) * wvp;
+    gl_Position = vec4(aPosition, 1) * transpose(wvp);
```
petmac: Same as above. Should probably be `wvp * vec4(aPosition, 1)`

PR #48, opened 2018-02-06
Support Unicode paths on Windows

### LINKED ISSUE #45: Trouble with accented paths
If the rom path contains any accents the rom fails to load. Tested only on Windows.

unknown

Thanks to Zangetsu for the screenshot.

VelocityRa:
Resolves #45.

### DIFF
--- a/src/emulator/util/include/util/string_convert.h
+++ b/src/emulator/util/include/util/string_convert.h
@@ -1,0 +1,12 @@
+#pragma once
+
+#include <vector>
+#include <string>
+
+using ProgramArgsWide = std::vector<std::wstring>;
+
+std::wstring utf_to_wide(const std::string& str);
+
+std::string wide_to_utf(const std::wstring& str);
+
+ProgramArgsWide process_args(int argc, char* argv[]);
--- a/src/emulator/util/src/util.cpp
+++ b/src/emulator/util/src/util.cpp
@@ -1,0 +1,55 @@
+#include <util/string_convert.h>
+
+#ifdef _WIN32
+#include <windows.h>
+#include <processenv.h>
+#include <shellapi.h>
+#endif
+
+#include <codecvt>   // std::codecvt_utf8
+#include <locale>    // std::wstring_convert
+#include <iostream>
+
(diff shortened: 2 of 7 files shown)


src/emulator/util/include/util/string_convert.h
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+#include <vector>
+#include <codecvt>   // std::codecvt_utf8
+
+using ProgramArgsWide = std::vector<std::wstring>;
+
+// convert UTF-8 string to wstring
+inline std::wstring utf_to_wide(const std::string& str)
```
petmac: I'd probably move these into the .cpp file. They're not templates and I wouldn't think they're called enough to warrant inlining.

That'll mean we can move the `<codecvt>` include to the .cpp file too.

I think the macOS build may have failed because `<locale>` wasn't included.

src/emulator/util/src/util.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include <windows.h>
+#include <processenv.h>
+#include <shellapi.h>   // CommandLineToArgvW
+#else
+#include <unistd.h>
+#endif
+
+#include <iostream>
+#include <memory>
+#include <string>
```
petmac: I think these includes aren't needed, and `<string>` should be included in the string_covert header.

src/emulator/util/include/util/string_convert.h
VelocityRa: >I'd probably move these into the .cpp file.
To be honest I did, but I was getting unresolved symbol errors that I wasn't sure how how to fix so I moved them back haha. I'll try again and ask you on Discord if I can't figure it out.

>I think the macOS build may have failed because <locale> wasn't included.
Yeah I fixed that and another bug on the unix code path.

src/emulator/util/src/util.cpp
VelocityRa: `iostream` is needed for the `std::cerr`, won't be needed once the logging stuff's in.

PR #50, opened 2018-02-07
Fix UNIX Build

mrcmunir:
Restore cstring is needed for memset reference.


VelocityRa:
Weird that Travis didn't catch that. Unless it's not Unix, but Linux specific.

petmac:
Thanks!

PR #54, opened 2018-02-07
Better logging for errors.

Rinnegatamante:


src/emulator/module/src/module.cpp
```diff
@@ -27,31 +27,25 @@ static std::mutex mutex;
 static NameSet logged;
 
 int unimplemented(const char *name) {
+#ifndef VERBOSE
```
VelocityRa: I think this should be an `#if VERBOSE` and put a `#define VERBOSE 0` someplace so it can be easily disabled/enabled, especially since `VERBOSE` isn't a standard macro afaik.
Or better yet, don't use the preprocessor at all and just make a `constexpr` flag.

Rinnegatamante: VERBOSE afaik it's a common macro in big projects to enable extended logging.
Btw, "if VERBOSE" with VERBOSE set to 0 or "ifndef VERBOSE" have the same exact result.

VelocityRa: `#ifdef` checks whether it's **defined**, it doesn't care about the value, whereas `#if` actually checks if the **value** is 0 or not.
In your code you'd need to either define it, or comment out the define or undef it.

I didn't know `VERBOSE` was common, and even if so, you can't assume that everyone coming in to this project will assume it's there so still nice to have it nice somewhere, like the `LOG_*` flags.

petmac: I'd like to know the rationale behind logging this more than once. I'm not against if if there's a reason, just that I'm not sure at this point what the reason is.

src/emulator/module/src/module.cpp
```diff
@@ -27,31 +27,25 @@ static std::mutex mutex;
...
-
-    {
-        const std::lock_guard<std::mutex> lock(mutex);
-        inserted = logged.insert(name).second;
-    }
-
-    if (inserted) {
-        std::cerr << ">>> " << name << " <<< returned 0x" << std::hex << std::uppercase << error << std::endl;
-    }
+    std::cerr << ">>> " << name << " <<< returned 0x" << std::hex << std::uppercase << error << std::endl;
```
petmac: Thanks for this. You're right that we should always log these errors.

src/emulator/module/src/module.cpp
Rinnegatamante: As i explained on Discord, an application can have unwanted behaviours due to several calls to an unimplemented nid but you'll never know about how many times it got called due to it being logged once.

petmac:
I merged #36 so this logging should probably be reimplemented in terms of LOG_ macros rather than std::cerr. Sorry!

Rinnegatamante:
Bump.

Rinnegatamante:
VERBOSE is supposed to be a non default macro that can be used to have VERBOSE logging as the macro itself says.

petmac:
I understand.

CMake has a mechanism for specifying options like this:
https://cmake.org/cmake/help/latest/command/option.html

My preference I think would be to not introduce configuration, and to instead use logging levels. That way a change in logging verbosity doesn't cause a recompile.

VelocityRa:
Yeah that's what the trace logging level is for.

Currently logging level is set [here](https://github.com/Vita3K/Vita3K/blob/master/src/emulator/util/src/util.cpp#L87), but when get a config system we can use it to set the level at runtime from a config file.

VelocityRa:
Oh this is about logging the same unimplemented function multiple time vs once right?
Then it's irrelevant to logging levels @petmac, unless you mean that it should be removed entirely and always log everythig, now that we have logging levels.

petmac:
Yes, I think that's what it should do. Log every time, but at a low enough severity that it can be turned off, or is off by default.

I think the first time should be logged as a warning rather than disappear though, because an unimplemented import means all bets are off, the emulator is incorrect, and funky things may happen.

VelocityRa:
@Rinnegatamante You wanna do this? Remove everything you changed here and change existing code to log as a warning the first time, and trace all times after that. If not close the issue and someone else will do it.

Rinnegatamante:
No idea how to deal with this tbh (i should look into spdlog and how it works), so i suppose it would be better if someone else picks up this.

PR #61, opened 2018-02-11
[libRtc/libKernel] Fix clock implementation

### LINKED ISSUE #56: Clock is wrong on Windows
Affected functions are sceKernelGetProcessTime* functions and sceRtcGetTick function.
This causes ticks based homebrews to stall a lot of time for nothing (that's why Network PR has issues on Windows for example).

VelocityRa:
vitaQuake multiplayer works with this + #39 
Resolves #56

### DIFF
--- a/src/emulator/host/include/host/rtc.h
+++ b/src/emulator/host/include/host/rtc.h
@@ -1,0 +1,53 @@
+// Vita3K emulator project
+// Copyright (C) 2018 Vita3K team
+//
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation; either version 2 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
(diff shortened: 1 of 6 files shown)


src/emulator/modules/SceRtc/src/SceRtcUser.cpp
```diff
@@ -68,8 +68,13 @@ EXPORT(int, sceRtcGetCurrentNetworkTick) {
 EXPORT(int, sceRtcGetCurrentTick, SceRtcTick *tick) {
     assert(tick != nullptr);
```
Rinnegatamante: This should be replaced with error();

src/emulator/host/include/host/rtc.h
```diff
@@ -0,0 +1,53 @@
...
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+#pragma once
+
+#include <host/state.h>
+
+#include <chrono>
+#include <cstdint>
+
+#define VITA_CLOCKS_PER_SEC 1000000
```
petmac: I'd move as much of this as possible into `rtc.cpp`, as it doesn't need to be in a header or inlined.

VelocityRa: Only reason I didn't is because with the current code structure - as far as I understand it - it would be in `host.cpp` instead of an `rtc.cpp`.

PR #64, opened 2018-02-13
facepalm Fix transparency on icon

Zangetsu38:
Fix transparency fail on icon
Rename VIta3K on Vita3K, thx @Ghabry for catch that


Zangetsu38:
@Ghabry thx, i have not see that before, i have set good name here, but have using all time old name ^^

PR #65, opened 2018-02-14
[mem] Fix VirtualProtect call

VelocityRa:
The last argument being `nullptr` was causing the function to fail.

### DIFF
--- a/src/emulator/mem/src/mem.cpp
+++ b/src/emulator/mem/src/mem.cpp
@@ -21,6 +21,7 @@
 #include <cassert>
 #include <cstring>
 #include <cmath>
+#include "util/log.h"
 
 #ifdef WIN32
 #define WIN32_LEAN_AND_MEAN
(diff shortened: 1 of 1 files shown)


src/emulator/mem/src/mem.cpp
```diff
@@ -81,14 +82,18 @@ bool init(MemState &state) {
...
+        LOG_CRITICAL("VirtualAlloc failed");
         return false;
     }
 
     state.allocated_pages.resize(length / state.page_size);
     const Address null_address = alloc(state, 1, "NULL");
     assert(null_address == 0);
 #ifdef WIN32
-    const BOOL res = VirtualProtect(state.memory.get(), state.page_size, PAGE_NOACCESS, nullptr);
+    DWORD lpflOldProtect;
```
petmac: No Hungarian notation required. This prefix is wrong by the way, as lp = long pointer, which is what the & operator does. fl is flags I guess.

old_protect would match the existing code style.

PR #67, opened 2018-02-15
Added implementations of sceMt19937 functions

bentokun:
Implement:
- sceMt19937Init
- sceMt19937UInt


Rinnegatamante:
Sadly this module is useless to implement since that library is statically linked in the executable.

After that PR get merged, we should update vita-headers submodule in order to get rid of all those modules: https://github.com/vitasdk/vita-headers/pull/295

VelocityRa:
Thanks for the contribution though @bentokun, I recommend hanging around/asking devs on Discord, before implementing something like this (or most things really, if you're new to the project).

Zangetsu38:
You can Squash your PR for have only one commit, more propr 👍

petmac:
Sorry @bentokun it looks like we don't need this. It's probably worth noting too that C++11 (we require C++14) has Mersenne Twister in the standard library:

http://en.cppreference.com/w/cpp/numeric/random/mersenne_twister_engine

PR #71, opened 2018-02-17
Added vitaQuake/vitaQuakeII shaders.

Rinnegatamante:
Added shaders used from vitaQuake/vitaQuakeII with vitaGL.

vitaGL shaders however are still missing.


VelocityRa:
Most of the files lack a newline at the end - GitHub warns about this.

Zangetsu38:
Not change error on Vitaquake 2
>[19:39:31.873] [error] [ compile_shader ]:    Couldn't open 'D:\Emulateurs\Vita3K\shaders/8579410444553319554.glsl' for reading.

petmac:
I also have missing shaders on vitaQuake 2.8. I'll revert this, maybe the hashes are wrong.

Rinnegatamante:
That's cause vitaGL shaders are missing (vglInit still sets up vitaGL default shaders even if they're not used after that).
All shaders i pushed in the PR are fine.

shadowbrony33:
@Rinnegatamante This is somewhat off-topic but, later on how would you obtain / generate shaders for retail games?

petmac:
There's a shader binary reverse engineering effort underway that will hopefully pave the way for automatic translation to GLSL. Failing that, tedious, error-prone manual translation.

petmac:
On the plus side, we could use the github API to automatically log missing shaders and such, with their parameters and types, easing the work somewhat.

VelocityRa:
@shadowbrony33 Come back in a year or so.
This requires a proper shader recompiler. We're planning to recompile SGX543+ shaders to SPIR-V and from there we can transpile to GLSL or use it straight from Vulkan (in the even more far-off future when there will be a Vulkan renderer).

When we started, we knew almost nothing about .gxm and the instruction set of the Vita's GPU though. So the first step is to reverse that. motoharu is working on that, posting his findings here: https://wiki.henkaku.xyz/vita/SGX543

Rinnegatamante:
@shadowbrony33 You can actually dump shaders of commercial games using this: https://github.com/Rinnegatamante/GxmDumper
However you still get proprieary compiled GXP shaders that need to be RE and then converted to GLSL shaders.

PR #74, opened 2018-02-18
Revert "Added vitaQuake/vitaQuakeII shaders."

petmac:
Reverts Vita3K/Vita3K#71


shadowbrony33:
why

petmac:
The hashes were wrong so the emulator wouldn't load the files.

PR #76, opened 2018-02-19
Add ability to seek app0 files

bentokun:
I added the ability to seek app0 files by extracting and storing data in a InMemoryFile (a lightweight class).
This should fix the crash for game like Bernard Chronicles, and still keep other games functional.


src/emulator/io/include/io/file.h
```diff
@@ -0,0 +1,45 @@
...
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+#pragma once
+
+#include <psp2/types.h>
+#include <psp2/io/fcntl.h>
+#include <iostream>
```
VelocityRa: Space before this line

src/emulator/io/include/io/file.h
```diff
@@ -0,0 +1,45 @@
...
+#include <cassert>
+#include <cstring>
+#include <vector>
+
+class ReadOnlyInMemFile {
+	std::vector<char> buf;
+	size_t            currentPos = 0;
```
VelocityRa: I think it's better to avoid manual alighment like that. Maybe run the format script (that runs clang-format) to see what are the stardard formatting rules.

src/emulator/io/include/io/state.h
```diff
@@ -30,9 +30,12 @@ struct _WDIR;
 #include <dirent.h>
 #endif
 
+#include <io/file.h>
```
VelocityRa: Should be above, below `psp2/types.h`.

src/emulator/io/src/io.cpp
```diff
@@ -40,18 +40,25 @@
...
-    if (!zip_file) {
-        return ZipFilePtr();
-    }
+	const auto zip_file = mz_zip_reader_extract_iter_new(&zip, index, 0);
+	
+	if (!zip_file) {
+		return ReadOnlyInMemFile();	
+	}   
+
+	ReadOnlyInMemFile res; char* data = res.alloc_data(zip_file->file_stat.m_uncomp_size);
```
VelocityRa: Separate these lines please, 1 statement per line.

src/emulator/io/src/file.cpp
```diff
@@ -0,0 +1,66 @@
...
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+#include <io/file.h>
+
+ReadOnlyInMemFile::ReadOnlyInMemFile() = default; 
+
+ReadOnlyInMemFile::ReadOnlyInMemFile(const char* data, size_t size) 
+	: buf(data, data+size) {
```
VelocityRa: This whole file uses tabs. Use spaces as the rest of the project please! I recommend getting clang-format and running the script in the root of the project.

src/emulator/io/src/io.cpp
```diff
@@ -75,8 +82,9 @@ const char *translate_open_mode(int flags) {
                 } else {
                     return "wb+";
                 }
+
             } else {
-                return "rb+";
+            	return "rb+";
```
VelocityRa: tab -> spaces

src/emulator/io/include/io/state.h
```diff
@@ -46,14 +49,14 @@ enum TtyType {
 
 typedef std::map<SceUID, TtyType> TtyFiles;
 typedef std::map<SceUID, FilePtr> StdFiles;
-typedef std::map<SceUID, ZipFilePtr> ZipFiles;
+typedef std::map<SceUID, ReadOnlyInMemFile> AppFiles;
```
frangarcj: Why no just AppFilePtr?

src/emulator/io/src/io.cpp
```diff
@@ -275,12 +284,23 @@ int seek_file(SceUID fd, int offset, int whence, const IOState &io) {
         break;
     }
 
-    const int ret = fseek(file->second.get(), offset, base);
+    int ret = 0; long pos = 0;
```
VelocityRa: 1 statement per line

src/emulator/io/include/io/state.h
bentokun: I dont think the name AppFilePtr suits for it, since the class is just storing data using std::vector, it doesnt point to anything

src/emulator/io/src/file.cpp
bentokun: ok, sorry :P

src/emulator/io/include/io/state.h
VelocityRa: `AppFileBuffer`?

bentokun: better

petmac:
Would it be worth changing over our file IO code to use something like SDL's file abstraction layer?
https://wiki.libsdl.org/CategoryIO

petmac:
I'm still not 100% sure why we have to decompress the whole file just to seek. The file could be hundreds of MB, for example.

Seeking forwards should be trivial -- read some bytes and discard them until the file pointer is correct.

Seeking backwards is annoying -- either reopen the file and seek forwards, or cache bytes previously read?

VelocityRa:
@petmac Hm that's true. But this PR would still be faster right?
Feels like for something as slow as IO operations we should opt in for the faster-at-runtime approach than the memory conserving one. 

The games (homebrew) we're running atm aren't even that big.
By the time commercial games run, fs code should have been refactored to be more robust anyway.

PR #77, opened 2018-02-19
Hash shader binaries using SHA-256 rather than FNV-1a.

petmac:
Also dump binaries of missing shaders for reversing.


src/emulator/modules/SceGxm/SceGxm.cpp
```diff
@@ -32,20 +35,26 @@ using namespace glbinding;
 
 #define GXM_PROFILE(name) MICROPROFILE_SCOPEI("GXM", name, MP_BLUE)
 
-// https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function#FNV-1a_hash
-static uint64_t fnv1a(const void *data, size_t size) {
+static void sha256(char (&dst)[65], const void *data, size_t size) {
```
VelocityRa: Not important yet, but I'd abstract crypto stuff to a separate module.

src/emulator/modules/SceGxm/SceGxm.cpp
```diff
@@ -94,13 +103,24 @@ static bool compile_shader(GLuint shader, const GLchar *source) {
 static bool compile_shader(GLuint shader, const SceGxmProgram *program, const char *base_path) {
     GXM_PROFILE(__FUNCTION__);
 
-    const uint64_t hash = fnv1a(program, program->size);
+    char hash[65] = {};
```
VelocityRa: Could wrap this (and the `sha256` arg) to a struct to avoid the magic number here.

src/emulator/modules/SceGxm/SceGxm.cpp
petmac: Looks like I may have to anyway, to work around the Windows typedef clash.

PR #78, opened 2018-02-20
Add short type aliases and port existing code

VelocityRa:
Sorry for the possible conflicts, but we may as well do this as early as possible.

**Note to developers:** Please include `util/types.h` in new code and **use short-form types** from now on:
`u8`, `u16`, `u32`, `u64` and the `s` variants for signed types. There's also `f32` and `f64` for float/double. I didn't port those.

Should probably be added to the Wiki if this is merged.

The `types.h` code is from RPCS3, but it has no GPL2 header. Should I get one from a file that does and include it?
Btw, some macros in there will make more sense once I import more util files.

### DIFF
--- a/src/emulator/audio/include/audio/state.h
+++ b/src/emulator/audio/include/audio/state.h
@@ -25,6 +25,7 @@
 #include <vector>
 
 #include <psp2/types.h>
+#include <util/types.h>
 
 #include <SDL_audio.h>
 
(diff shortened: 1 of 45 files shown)


Zangetsu38:
rebase needed

VelocityRa:
Discussed in Discord, devs don't want this because it's non-standard.

Issue #75, opened 2018-02-19
SceNet implementation broken linux build


mrcmunir:
Returned in 
```
SceNet.cpp:35:30: error: duplicate case value
 # define ERROR_CASE(errname) case(errname): return SCE_NET_ERROR_##errname;
```

PR #79, opened 2018-02-20
Fix compile error on Non-Windows platform caused by ENOTSUP and EOPNOTSUPP

bentokun:
Fix for #75 . ENOTSUP and EOPNOTSUPP have the same value (3440), but ENOTSUP embraces the meaning of EOPNOTSUPP, so we should only use ENOTSUP.


Rinnegatamante:
That's not completely correct.
macOS compiles fine and has different values for those values.
This seems to be a Linux-related issue more than non-Windows one.

mrcmunir:
Not sure if this fix it

```
#ifndef ENOTSUP
#define ENOTSUP EOPNOTSUPP
ERROR_CASE(ENOTSUP)
ERROR_CASE(EOPNOTSUPP)
#endif
```
I can pass this but returned some error in the log but can connect
```
[23:49:00.180] [error] [ error ]:    >>> sceNetShowNetstat <<< returned -0X7FBEFE38
[23:50:24.987] [error] [ error ]:    >>> sceNetSendto <<< returned -0X7FBEFEF3
[23:50:24.987] [error] [ error ]:    >>> sceNetRecvfrom <<< returned -0X7FBEFEDD
```

bentokun:
yeah, that's why i close this :P

mrcmunir:
@bentokun if move these error cases into ifndef ENOTSUP define can pass it

bentokun:
proper fix would be remove the EOPNOTSUPP if its unix, this morning i was rushed >-<. but still, good for you :))

PR #80, opened 2018-02-22
[READY TO MERGE] Threads refactoring

frangarcj:
- [x] Move common code to thread_functions
- [x] Move main thread to a background thread
- [x] Add display code to main.cpp
- [x] Sync background over main loop 
- [x] Gxm display thread and queue
- [x] Some games show garbage (breakveetout)

Please show your opinion

Moved to other issues
- Gxm sync primitives
- Some games are pretty slow

### DIFF
--- a/src/emulator/gxm/include/gxm/state.h
+++ b/src/emulator/gxm/include/gxm/state.h
@@ -86,6 +86,8 @@ private:
 struct DisplayCallback {
     Address pc;
     Address data;
+    Address old_buffer;
+    Address new_buffer;
 };
 
 struct GxmState {
--- a/src/emulator/gxm/include/gxm/state.h
+++ b/src/emulator/gxm/include/gxm/state.h
@@ -1,9 +1,7 @@
 #pragma once
 
 #include <mem/ptr.h>
-#include <queue>
-#include <mutex>
-#include <condition_variable>
+#include <threads/queue.h>
 
 namespace emu {
     typedef void SceGxmDisplayQueueCallback(Ptr<const void> callbackData);
@@ -16,72 +14,6 @@ namespace emu {
         uint32_t parameterBufferSize = 0;
     };
 }
-
-template <typename T>
-class Queue
-{
-public:
-
-    uint32_t displayQueueMaxPendingCount_;
-
-    
-    T pop()
-    {
-        std::unique_lock<std::mutex> mlock(mutex_);
-        while(queue_.empty())
-        {
-            condempty_.wait(mlock);
-        }
-        auto item = queue_.front();
-        queue_.pop();
-        mlock.unlock();
-        cond_.notify_one();
-        return item;
-    }
-    
-    void pop(T& item)
-    {
-        std::unique_lock<std::mutex> mlock(mutex_);
-        while(queue_.empty())
-        {
-            condempty_.wait(mlock);
-        }
-        item = queue_.front();
-        queue_.pop();
-        mlock.unlock();
-        cond_.notify_one();
-    }
-    
-    void push(const T& item)
-    {
-        std::unique_lock<std::mutex> mlock(mutex_);
-        while (queue_.size()==displayQueueMaxPendingCount_)
-        {
-            cond_.wait(mlock);
-        }
-        queue_.push(item);
-        mlock.unlock();
-        condempty_.notify_one();
-    }
-    
-    void push(T&& item)
-    {
-        std::unique_lock<std::mutex> mlock(mutex_);
-        while (queue_.size()==displayQueueMaxPendingCount_)
-        {
-            cond_.wait(mlock);
-        }
-        queue_.push(std::move(item));
-        mlock.unlock();
-        condempty_.notify_one();
-    }
-    
-private:
-    std::queue<T> queue_;
-    std::mutex mutex_;
-    std::condition_variable cond_;
-    std::condition_variable condempty_;
-};
 
 struct DisplayCallback {
     Address pc;
--- a/src/emulator/gxm/include/gxm/state.h
+++ b/src/emulator/gxm/include/gxm/state.h
@@ -1,6 +1,9 @@
 #pragma once
 
 #include <mem/ptr.h>
+#include <queue>
+#include <mutex>
+#include <condition_variable>
 
 namespace emu {
     typedef void SceGxmDisplayQueueCallback(Ptr<const void> callbackData);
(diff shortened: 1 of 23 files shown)


src/emulator/main.cpp
```diff
@@ -110,29 +111,24 @@ int main(int argc, char *argv[]) {
...
+    const CallImport call_import = [&host](uint32_t nid, SceUID main_thread_id) {
         ::call_import(host, nid, main_thread_id);
     };
 
     const size_t stack_size = MB(1); // TODO Get main thread stack size from somewhere?
-    const bool log_code = false;
-    const ThreadStatePtr main_thread = init_thread(entry_point, stack_size, log_code, host.mem, call_import);
-    if (!main_thread) {
+   
+    const SceUID main_thread_id = create_thread(entry_point, host.kernel, host.mem, "main_thread",stack_size, call_import);
```
VelocityRa: Should be "Main" probably, the "thread" part is obvious.

src/emulator/host/src/host.cpp
```diff
@@ -63,8 +63,9 @@ bool init(HostState &state) {
     const ResumeAudioThread resume_thread = [&state](SceUID thread_id) {
         const ThreadStatePtr thread = lock_and_find(thread_id, state.kernel.threads, state.kernel.mutex);
         const std::unique_lock<std::mutex> lock(thread->mutex);
-        assert(thread->to_do == ThreadToDo::wait);
-        thread->to_do = ThreadToDo::run;
+        if(thread->to_do == ThreadToDo::wait) {
+            thread->to_do = ThreadToDo::run;
+        }
```
VelocityRa: Previous code errored if `thread->to_do != ThreadToDo::wait`, is this case handled here?

src/emulator/main.cpp
VelocityRa: Thread name should be "Main" probably, the "thread" part is obvious.

src/emulator/host/src/host.cpp
VelocityRa: Previous code errored if `thread->to_do != ThreadToDo::wait`, is this case handled here?

src/emulator/kernel/src/thread.cpp
```diff
@@ -16,16 +16,45 @@
...
 #include <cassert>
 #include <cstring>
 
-ThreadStatePtr init_thread(Ptr<const void> entry_point, size_t stack_size, bool log_code, MemState &mem, CallImport call_import) {
+struct ThreadParams {
+    KernelState *kernel = nullptr;
+    SceUID thid = SCE_KERNEL_ERROR_ILLEGAL_THREAD_ID;
+    SceSize arglen = 0;
+    Ptr<void> argp;
+    std::shared_ptr<SDL_semaphore> host_may_destroy_params = std::shared_ptr<SDL_semaphore>(SDL_CreateSemaphore(0), SDL_DestroySemaphore);
```
VelocityRa: You seem to only use `.get` so this can probably be a `unique_ptr`. Please always prefer that, or `weak_ptr` over `shared_ptr`.

petmac: My fault, I overuse shared_ptr because you don't need to specify the deleter type for incomplete types, as you do for unique_ptr (I think.)

VelocityRa: Yeah `unique_ptr` uses static deleters so the deleter type needs to be complete.

src/emulator/kernel/src/thread.cpp
```diff
@@ -16,16 +16,45 @@
...
+    write_reg(*thread->cpu, 0, params.arglen);
+    write_reg(*thread->cpu, 0, params.argp.address());
+    const bool succeeded = run_thread(*thread);
+    assert(succeeded);
+    const uint32_t r0 = read_reg(*thread->cpu, 0);
+    return r0;
+}
+
+typedef std::function<void(uint32_t)> CallImportForThread;
+ThreadStatePtr init_thread(Ptr<const void> entry_point, size_t stack_size, bool log_code, MemState &mem, CallImportForThread call_import) {
```
petmac: Is `init_thread` called anywhere other than `create_thread`?

If not, I would probably combine them into one function. That may help reduce the need for an extra `CallImportForThread ` type too.

src/emulator/host/src/host.cpp
frangarcj: thread->to_do can be ThreadToDo::exit so assert must be removed

src/emulator/kernel/src/thread.cpp
frangarcj: Someone could explain me what are all those ptr 😢

src/emulator/kernel/src/thread.cpp
frangarcj: It's not called anymore so I'll combine them

src/emulator/kernel/src/thread.cpp
frangarcj: If I change it to a unique_ptr then thread_function will complain of call to implicitly-deleted copy constructor

VelocityRa: I see PetMac already answered :p

Zangetsu38:
rebase needed for fix conflict

src/emulator/modules/SceGxm/SceGxm.cpp
```diff
@@ -719,11 +736,60 @@ EXPORT(int, sceGxmGetRenderTargetMemSizes) {
...
+struct GxmThreadParams {
+    KernelState *kernel = nullptr;
+    MemState *mem = nullptr;
+    SceUID thid = SCE_KERNEL_ERROR_ILLEGAL_THREAD_ID;
+    GxmState *gxm = nullptr;
+};
+
+static int SDLCALL thread_function(void *data) {
+    const GxmThreadParams params = *static_cast<const GxmThreadParams *>(data);
+    while(true){
```
petmac: There's a chance here that the parameters passed to the thread function are destroyed before the thread properly starts. When creating this thread, we need to wait for the thread to indicate that it's made a copy of the arguments before the original goes out of scope.

Search for `host_may_destroy_params` to see how it's done.

src/emulator/gxm/include/gxm/state.h
```diff
@@ -14,7 +17,81 @@ namespace emu {
...
+
+    uint32_t displayQueueMaxPendingCount_;
+
+    
+    T pop()
+    {
+        std::unique_lock<std::mutex> mlock(mutex_);
+        while(queue_.empty())
+        {
+            condempty_.wait(mlock);
```
petmac: When I tried to quit the emulator (Cmd-Q) it deadlocked here. It's very inconvenient, but threads also need to wake up and exit when the user wants to quit the emulator.

We need to find a good way around this I think, as it makes multithreading a lot more complex.

src/emulator/host/CMakeLists.txt
```diff
@@ -13,5 +13,5 @@ version.cpp
 configure_file(src/version.cpp.in version.cpp)
 
 target_include_directories(host PUBLIC include)
-target_link_libraries(host PUBLIC audio cpu ctrl gxm io kernel mem net util)
+target_link_libraries(host PUBLIC audio cpu ctrl glutil gxm io kernel mem net util)
```
petmac: I think this can probably be a private dependency.

src/emulator/kernel/CMakeLists.txt
```diff
@@ -10,4 +10,4 @@ add_library(
 )
 
 target_include_directories(kernel PUBLIC include)
-target_link_libraries(kernel PUBLIC cpu vita-headers util)
+target_link_libraries(kernel PUBLIC cpu sdl2 vita-headers util)
```
petmac: I think SDL can be a private dependency.

src/emulator/module/include/module/bridge_args.h
```diff
@@ -104,8 +104,9 @@ struct ArgLayout<R0, long long, StackHead, StackTail...> {
         static typename std::enable_if < index<2, T>::type read(CPUState &cpu, const MemState &mem) {
         if (sizeof(T) == sizeof(uint64_t)) {
             uint64_t value = 0;
+            value |= read_reg(cpu, index * 2 + 1);
+            value = value >> 32;
```
petmac: I think this should be a shift left.

src/emulator/kernel/src/thread.cpp
frangarcj: I have problems when using unique_ptr. Like there's no copy constructor. any idea?

src/emulator/threads/include/threads/queue.h
```diff
@@ -0,0 +1,77 @@
+//
+//  queue.h
+//  Vita3K
+//
+//  Created by Francisco José García García on 11/3/18.
+//
```
petmac: Probably should use GPL header comment.

src/emulator/modules/SceKernelThreadMgr/SceThreadmgr.cpp
```diff
@@ -104,18 +104,11 @@ EXPORT(int, sceKernelCreateThreadForUser) {
...
-        } else {
-            elapsed = delay_ms;
-        }
-    } while (elapsed < delay_ms);
-
+#ifdef _WIN32
+    Sleep(delay / 1000);
+#else
+    usleep(delay);
+#endif
```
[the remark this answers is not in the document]
frangarcj: We will address that on #100

PR #84, opened 2018-03-01
Work around for sceIoLseek using 64bit argument

frangarcj:


src/emulator/module/include/module/bridge_args.h
```diff
@@ -96,6 +96,37 @@ struct ArgLayout {
     }
 };
 
+template <typename R0, typename StackHead, typename... StackTail>
+struct ArgLayout<R0, long long, StackHead, StackTail...> {
```
petmac: It might be worth changing this long long to a specific 64-bit type. Maybe it's not an issue though if it's 64-bit everywhere.

src/emulator/module/include/module/bridge_args.h
```diff
@@ -96,6 +96,37 @@ struct ArgLayout {
...
+    using StackLayout2 = StackLayout<StackType<StackHead>, StackType<StackTail>...>;
+    
+    template <typename T, size_t index>
+    static typename std::enable_if < index<2, T>::type read(CPUState &cpu, const MemState &mem) {
+        if(sizeof(T)==sizeof(uint64_t)){
+            uint64_t value = 0;
+            value |= read_reg(cpu,index*2);
```
petmac: This looks quite fragile but we discussed in Discord that it was a temporary workaround until we make bridging arguments more understandable.

src/emulator/module/include/module/bridge_args.h
frangarcj: It didn't work with uint64_t so I used long long as was indicated in debugger

PR #87, opened 2018-03-05
Added macOS nightlies support.

Rinnegatamante:
This requires a new repo (Vita3K/autobuilds) where macOS nightlies will be deployed.


Rinnegatamante:
Now works as expected, REPO_TOKEN environment variable must be set with a security token of an user having push access to Vita3k/Vita3k-builds.

PR #90, opened 2018-03-07
Change license to GPL-3.0

VelocityRa:
We'll need this to be compatible with the Apache-2.0 license that SPIRV-Cross and SPIRV-Tools use.

Hopefully we don't need to change the header in every file ('2' to '3'), I don't think we do since it reads "2 or (at your option) any later version".

### DIFF
--- a/COPYING.txt
+++ b/COPYING.txt
@@ -1,281 +1,622 @@
                     GNU GENERAL PUBLIC LICENSE
-                       Version 2, June 1991
+                       Version 3, 29 June 2007
 
- Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
- 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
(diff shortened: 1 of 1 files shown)


Ghabry:
Btw when you change to GPLv3 you can't use unicorn anymore because GPLv2 is incompatible to GPLv3 and unicorn has no "and later" clause.

VelocityRa:
Well. That's a bummer.

Ghabry:
dynarmic is GPLv2 or later, thats fine.

Basicly your current license (GPLv2 or later) is sufficient for linking against SPIRV as long as the distribution at the end is GPLv3.

With Unicorn your current license is also fine as long as the distribution at the end is GPLv2.

Or you keep your code at "v2 or later" and just link in both and hope that nobody cares until you find a better solution ;)

VelocityRa:
I see, thanks for clearing that up @Ghabry.
I'm closing this for now since it'll be a while before we need SPIR-V anyway.

PR #91, opened 2018-03-16
[sceKernel] Implement sceKernelLibcGettimeofday

VelocityRa:
Allows **Vita Doom** to go in-game (still not playable because of other issues).

The `tzAddr` argument is unused, not even PPSSPP implements it apparently.

### DIFF
--- a/src/emulator/modules/SceProcessmgr/SceProcessmgr.cpp
+++ b/src/emulator/modules/SceProcessmgr/SceProcessmgr.cpp
@@ -19,6 +19,13 @@
 
 #include <psp2/kernel/error.h>
 #include <psp2/kernel/processmgr.h>
+#include <host/rtc.h>
+
+struct VitaTimeval
+{
+    uint32_t tv_sec;
+    uint32_t tv_usec;
+};
 
 EXPORT(int, sceKernelGetStderr) {
     return unimplemented("sceKernelGetStderr");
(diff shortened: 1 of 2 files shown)


src/emulator/modules/SceProcessmgr/SceProcessmgr.cpp
```diff
@@ -44,8 +45,18 @@ EXPORT(int, sceKernelLibcClock) {
...
-EXPORT(int, sceKernelLibcGettimeofday) {
-    return unimplemented("sceKernelLibcGettimeofday");
+EXPORT(int, sceKernelLibcGettimeofday, Ptr<VitaTimeval> timeAddr, Ptr<Address> tzAddr) {
+    if (timeAddr)
+    {
+        auto* tv = timeAddr.get(host.mem);
+
+        const auto ticks = rtc_get_ticks(host);
+
+        tv->tv_sec = ticks / 1'000'000;
```
petmac: I think we may have a define for 1000000 somewhere. `VITA_CLOCK_PER_SEC` or something like it?

src/emulator/modules/SceProcessmgr/SceProcessmgr.h
```diff
@@ -19,6 +19,12 @@
 
 #include <module/module.h>
 
+struct VitaTimeval
```
petmac: This file is auto-generated so the struct declaration will get erased when regenerating module files.

It might be better to put it in `SceProcessmgr.cpp` as it's only used in that one place at the moment. `SceProcessmgr.cpp` is also auto-generated, but at least everything will be in one place.

src/emulator/modules/SceProcessmgr/SceProcessmgr.cpp
VelocityRa: You're right, forgot about it.

src/emulator/modules/SceProcessmgr/SceProcessmgr.h
VelocityRa: >`SceProcessmgr.cpp` is also auto-generated

How does that work, how does code stay there then?

petmac: Good point. I added documentation in https://github.com/Vita3K/Vita3K/wiki/Module-generation, copied from #20.

Sorry, I forgot to do that sooner.

PR #105, opened 2018-03-22
Support all 16 paletted texture formats (fixes VitaDoom)

VelocityRa:
Fixes VitaDoom (**doom.wad**, **doom2.wad**, and some other fan .wads work).
Runs at 35 fps for me, pretty playable. Only other issue at the moment is garbled audio.

From commit message:

- Extend current support of `SCE_GXM_TEXTURE_FORMAT_P8_ABGR` to the 15 more palleted formats.

- Only tested ones are `SCE_GXM_TEXTURE_FORMAT_P8_ABGR` and `SCE_GXM_TEXTURE_FORMAT_P8_1BGR`.

- A warning message is generated if one of the untested ones is used, just to be sure.
  Can be removed in the future when more of them are tested.

---
#### To reviewers:
I included a refactoring commit (VS does this automatically on Save for me and it was too anoying trying to separate the lines I wanted) so go [here](https://github.com/Vita3K/Vita3K/pull/105/commits/d754d9ce1a25327ab613640e0e8cd61a8d706608) to review main commit only, for convenience.

### DIFF
--- a/src/emulator/modules/SceGxm/SceGxm.cpp
+++ b/src/emulator/modules/SceGxm/SceGxm.cpp
@@ -386,10 +386,10 @@ EXPORT(int, sceGxmDraw, SceGxmContext *context, SceGxmPrimitiveType primType, Sc
     // TODO Use some kind of caching to avoid setting every draw call?
     const SharedGLObject program = get_program(*context, host.mem);
     glUseProgram(program->get());
-    
+
     // TODO Use some kind of caching to avoid setting every draw call?
     set_uniforms(program->get(), *context, host.mem);
-    
+
     const GLenum mode = translate_primitive(primType);
     const GLenum type = indexType == SCE_GXM_INDEX_FORMAT_U16 ? GL_UNSIGNED_SHORT : GL_UNSIGNED_INT;
     glDrawElements(mode, indexCount, type, indexData);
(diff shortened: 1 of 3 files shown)


src/emulator/modules/SceGxm/SceGxm.cpp
```diff
@@ -1010,27 +1010,64 @@ EXPORT(int, sceGxmSetFragmentTexture, SceGxmContext *context, unsigned int textu
...
+        {
+            LOG_ERROR("Invalid swizzle for paletted texture foramt.");
+        }
+        }
+
+        glPixelMapfv(GL_PIXEL_MAP_I_TO_R, palette_indexes, map[R]);
+        glPixelMapfv(GL_PIXEL_MAP_I_TO_G, palette_indexes, map[G]);
+        glPixelMapfv(GL_PIXEL_MAP_I_TO_B, palette_indexes, map[B]);
+        if (A == A_max) {
+            constexpr float max_alpha[1] = { 255.0 };
```
petmac: Is this array big enough? Should it be 256 elements?

VelocityRa: Ah indeed it should, gonna fix that later.

VelocityRa: Pushed a fix.
Turns out constexpr float array initialization is impossible without really hacky stuff, so I just used `std::fill`.

PR #107, opened 2018-03-23
Implement sceRtcGetTime*_t

VelocityRa:
Used by:
* Starfield-Vita

### DIFF
--- a/src/emulator/host/CMakeLists.txt
+++ b/src/emulator/host/CMakeLists.txt
@@ -7,6 +7,7 @@ include/host/rtc.h
 include/host/state.h
 include/host/version.h
 src/host.cpp
+src/rtc.cpp
 version.cpp
 )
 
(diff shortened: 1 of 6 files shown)


VelocityRa:
If you remember from last time, I had said I didn't do this because with the current project structure they would go in `util.cpp`, which is weird.
Do you still think it's ok? Another solution would be to make rtc stuff its own project.

petmac:
That is probably true.

At the moment this is in the host module, so it shouldn't cause any change of dependencies if rtc implementations stay there but move to a .cpp file.

I worry a little about making rtc its own project as I don't want to end up in the same place as before with hundreds of libraries that take ages to load and build in Visual Studio.

VelocityRa:
Yeah it's in host not utils, my bad. Also you're right about rtc being its own module, it's not large enough to warrant it.

I put them in a _new_ .cpp file instead of host.cpp, which I guess violates the projects structure? Even if so, I think it's better this way, thoughts?

VelocityRa:
Bump @petmac.

PR #108, opened 2018-03-23
Better Crypto Interface and Dependency 

pent0:
**When the reviewing is done, i will add the binary and include headers**

Using LibTomCrypt as dependency and add new crypto operations interface.
Expect a fail build, i will add binary dependency soon, just put it here for anyone to review the code to fit with Vita3k


src/external/CMakeLists.txt
```diff
@@ -78,6 +76,9 @@ if(APPLE)
 elseif(WIN32)
 	target_include_directories(unicorn INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}/unicorn/windows/include")
 	target_link_libraries(unicorn INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}/unicorn/windows/unicorn.lib")
+elseif(UNIX)
+        target_include_directories(unicorn INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}/unicorn/unix/include")
```
VelocityRa: spacing

src/emulator/modules/ScePower/ScePower.cpp
```diff
@@ -63,7 +63,7 @@ EXPORT(int, scePowerGetBatteryLifeTime) {
     int res;
     SDL_GetPowerInfo(&res, NULL);
     if (res == -1){
-        return INT_MAX;
+        return 0xFFFFFFFF;
```
VelocityRa: Reason? Is this what `scePower` does (did you reverse it)?

VelocityRa: Reason? Is this what scePower does (did you reverse it)? Because it returns an int, not unsigned int.

src/external/CMakeLists.txt
VelocityRa: indendation

src/emulator/crypto/src/crypto_operations_factory.cpp
```diff
@@ -0,0 +1,17 @@
+#include <stdexcept>
+
+#include <crypto/crypto_operations_factory.h>
+#include <crypto/libtomcrypt_crypto_operations.h>
+
+#include <memory>
+
+std::shared_ptr<ICryptoOperations> CryptoOperationsFactory::create(CryptoOperationsTypes type)
```
VelocityRa: why not `std::unique_ptr`?

src/emulator/modules/ScePower/ScePower.cpp
bentokun: It's a quick fix, sorry :P

src/emulator/crypto/src/crypto_operations_factory.cpp
bentokun: Ask motoharu, I do not write this

src/emulator/modules/ScePower/ScePower.cpp
VelocityRa: If it's right, it should probably be `UINT_MAX`.

src/emulator/crypto/src/crypto_operations_factory.cpp
VelocityRa: @motoharu-gosuto

src/emulator/modules/ScePower/ScePower.cpp
bentokun: I just noticed, isn't this function return an int? Why UINT_MAX?

VelocityRa: Because I assumed you changed this because you reversed the function and saw that it was returning this value.

Rinnegatamante: Said function should return UNKNOWN state error when this value is not known but this could potentially break everything running on Desktop PC, that's why i think it's better to stick with INT_MAX.

src/emulator/crypto/src/crypto_operations_factory.cpp
motoharu-gosuto: @VelocityRa 
Because crypto interface is passed everywhere. Copy constructor of unique_ptr is not accessable by design. You will either pass it as a reference (which is meaningless). Or only store it in host state and never pass it as argument.

VelocityRa: > (which is meaningless)

Can you elaborate? You pass it as a pointer everywhere, copying it everywhere for no reason (as far as I can tell). Why not pass as a referece and have only one defined place of ownership.

motoharu-gosuto: @VelocityRa 
Because this is how you are supposed to use smart pointers?
They are designed to be an reference counted objects.
If you pass any smart pointer by reference - you are bypassing the core mechanism of these classes. Additionally if you are passing unique_ptr by reference - you are bypassing relationship control mechanism.

VelocityRa: >Because this is how you are supposed to use smart pointers?
I think you've misunderstood the goal of `shared_ptr` then.

In my opinion (and definitely not only mine) `shared_ptr` should be used when absolutely necessary.
If you say it's definitely neccesary here, fine by me, but passing smart pointers by reference is not as unheard of as you make it out to be.

motoharu-gosuto: @VelocityRa 
I dont know man. As I said using references for smart pointers sounds completely weird to me.
And I am definitely not a fan of move semantics. Especially with things like unique_ptr.
If you want - you may try to adjust the code to use unique_ptr. I am not against it.

VelocityRa: You're probably right, I haven't thought this through very much. Maybe I'll try to mess around with it after it's merged, not very important.

petmac: I'm sure it's fine, but I think it's strange to introduce a runtime cost (virtual functions) and the associated pointer management headaches when the implementation doesn't actually change at runtime.

I would probably have just declared the crypto functions and expect them to be provided.

motoharu-gosuto: Depends on if we would need to switch to another crypto library or not. It gives you just enough flexibility. Not sure what you mean by :
'I would probably have just declared the crypto functions and expect them to be provided.'
Just declarations without a class?

petmac: Yeah, and just let the linker sort it out.

Virtual functions are for dynamic dispatch -- implementations changing at runtime. We don't need that.

motoharu-gosuto: I dont know. This sounds so much C. 
Also this was supposed to be integrated with pfs stuff in the future. 
Not sure how we will handle it without a crypto class.

VelocityRa: >This sounds so much C.

have you looked at the rest of the project? :p

pent0:
No intention of working on this anymore

PR #118, opened 2018-04-02
Fixes

VelocityRa:
Sorry for the descriptive title and big PR but I couldn't bother opening like 6 separate ones.

The commit messages are ok though, I recommend going through each one...
Well, there are some wrong changes in some commits but they're generally fine, I didn't want to spend even more time fixing those.

Please DO NOT SQUASH/merge. Just review, thanks.


src/emulator/gxm/src/gxm.cpp
```diff
@@ -42,31 +42,47 @@ static const char *parameter_name(const SceGxmProgramParameter &parameter) {
     return reinterpret_cast<const char *>(bytes + parameter.name_offset);
 }
 
-static const char *scalar_type(SceGxmParameterType type) {
+static SceGxmParameterType parameter_type(const SceGxmProgramParameter &parameter) {
+    return static_cast<SceGxmParameterType>(
+        static_cast<uint16_t>(parameter.type));
+}
+
+static const char *scalar_type(SceGxmParameterType const& type) {
```
Rinnegatamante: Missing SCE_GXM_PARAMETER_TYPE_AGGREGATE, maybe add a warn for it?

VelocityRa: yea I had it initially but I removed it because I added `default` case, so it warns already.

src/emulator/main.cpp
```diff
@@ -75,6 +75,8 @@ int main(int argc, char *argv[]) {
         return SDLInitFailed;
     }
 
+    SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
+    SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 1);
```
petmac: Unfortunately for us, it actually isn't as easy as this for macOS. Above OpenGL 2.1, macOS only supports "Core" contexts. You also need:

`SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_CORE);`

Using a Core context removes all the deprecated OpenGL functionality, like vertex attrib pointers that point into RAM rather than index into a buffer object. So to upgrade OpenGL we will need to start managing vertex buffers ourselves.

src/emulator/gxm/include/gxm/functions.h
```diff
@@ -32,15 +32,16 @@ GLenum attribute_format_to_gl_type(SceGxmAttributeFormat format);
 bool attribute_format_normalised(SceGxmAttributeFormat format);
 void set_uniforms(GLuint program, const SceGxmContext &context, const MemState &mem);
 void flip_vertically(uint32_t *pixels, size_t width, size_t height, size_t stride_in_pixels);
-GLenum translate_blend_func(SceGxmBlendFunc src);
-GLenum translate_blend_factor(SceGxmBlendFactor src);
+GLenum translate_blend_func(const SceGxmBlendFunc& src);
```
petmac: Isn't `SceGxmBlendFunc` an integer? It's OK to pass by value.

Unless the goal is to forward declare them and remove `#include <psp2/gxm.h>` from this file?

VelocityRa: Yeah all of those args I changed are enums. Isn't passing enums by reference ok? I read that the compiler will do the same thing anyway (will actually pass by value).
Anyway, I'll change all of them back to by value.

src/emulator/main.cpp
petmac: Confirmed. Using the above code, `GL_VERSION` is still `2.1` on macOS.

src/emulator/modules/SceGxm/SceGxm.cpp
```diff
@@ -705,11 +705,10 @@ EXPORT(int, sceGxmPrecomputedVertexStateSetUniformBuffer) {
 EXPORT(int, sceGxmProgramCheck, const SceGxmProgram *program) {
     assert(program != nullptr);
 
-    assert(memcmp(program->magic, "GXP", 4) == 0);
-    assert(program->maybe_version[0] == 1);
```
VelocityRa: no idea how MSVC was ok with compiling that...

src/emulator/gxm/src/gxm.cpp
```diff
@@ -145,13 +167,17 @@ static void output_glsl_parameters(std::ostream &glsl, const SceGxmProgram &prog
 static std::string generate_fragment_glsl(const SceGxmProgram &program) {
     GXM_PROFILE(__FUNCTION__);
 
+    LOG_DEBUG("Generating fragment shader");
+
     std::ostringstream glsl;
     glsl << "// Fragment shader.\n";
-    glsl << "#version 120\n";
+    glsl << "#version 140\n";
```
petmac: macOS won't be able to load this yet.

src/emulator/gxm/src/gxm.cpp
```diff
@@ -512,7 +546,7 @@ void flip_vertically(uint32_t *pixels, size_t width, size_t height, size_t strid
     }
 }
 
-GLenum translate_blend_func(SceGxmBlendFunc src) {
+GLenum translate_blend_func(const SceGxmBlendFunc src) {
```
petmac: Passed by value, doesn't need to be const.

src/emulator/gxm/src/gxm.cpp
```diff
@@ -607,20 +642,23 @@ GLenum translate_internal_format(SceGxmTextureFormat src) {
...
     switch (src) {
         case SCE_GXM_TEXTURE_FORMAT_U8U8U8U8_ABGR:
+        case SCE_GXM_TEXTURE_FORMAT_U8U8U8_BGR:
             return GL_RGBA8;
+        case SCE_GXM_TEXTURE_FORMAT_U4U4U4U4_ABGR:
+            return GL_RGBA4;
         case SCE_GXM_TEXTURE_FORMAT_U8_R111:
+        case SCE_GXM_TEXTURE_FORMAT_U8_111R:
+        case SCE_GXM_TEXTURE_FORMAT_U8_1RRR:
+            // TODO: this is inaccurate
```
petmac: I think OpenGL uses intensity as the colour palette index, so I think it's OK?

src/emulator/gxm/src/gxm.cpp
```diff
@@ -684,12 +729,28 @@ GLenum translate_minmag_filter(SceGxmTextureFilter src){
...
+    const auto base_format = texture::get_base_format(format);
+    switch (base_format) {
+        case SCE_GXM_TEXTURE_BASE_FORMAT_U4U4U4U4:
+            return GL_UNSIGNED_SHORT_4_4_4_4;
+        case SCE_GXM_TEXTURE_BASE_FORMAT_U8:
+        case SCE_GXM_TEXTURE_BASE_FORMAT_U8U8U8U8:
+            return GL_UNSIGNED_BYTE;
+        default:
+        {
+            return GL_UNSIGNED_BYTE;
```
petmac: Should we warn about unhandled base formats?

src/emulator/modules/SceGxm/SceGxm.cpp
```diff
@@ -45,14 +45,7 @@ EXPORT(int, sceGxmBeginCommandList) {
...
 EXPORT(int, sceGxmBeginScene, SceGxmContext *context, unsigned int flags, const SceGxmRenderTarget *renderTarget, const SceGxmValidRegion *validRegion, SceGxmSyncObject *vertexSyncObject, SceGxmSyncObject *fragmentSyncObject, const emu::SceGxmColorSurface *colorSurface, const emu::SceGxmDepthStencilSurface *depthStencil) {
-    assert(context != nullptr);
-    assert(flags == 0);
-    assert(renderTarget != nullptr);
-    assert(validRegion == nullptr);
-    assert(vertexSyncObject == nullptr);
-    assert(fragmentSyncObject != nullptr);
-    assert(colorSurface != nullptr);
-    assert(depthStencil != nullptr);
+    verify(HERE), context, flags == 0, renderTarget, !validRegion, !vertexSyncObject, fragmentSyncObject, colorSurface, depthStencil;
```
petmac: I think we should avoid adding yet another way to trap/log errors.

I suspect most of the asserts can be replaced with:
- Proper error handling (which gets logged and returns a SCE error code),
- LOG_WARN_IF or similar
Or could just be deleted.

petmac: Additionally, the comma operator just looks weird and non-idiomatic. People need to do a double-take to understand what's going on.

src/emulator/util/CMakeLists.txt
```diff
@@ -1,11 +1,17 @@
 add_library(
 	util
 	STATIC
+	include/util/BitField.h
+	include/util/types.h
```
petmac: I'm really against adding a `types.h` header as it's way too nebulous and guaranteed to grow into a junkyard.

petmac: I'm also inclined to move source code appropriated from RPCS3 to src/external, as then we don't have to try and make it match in style or formatting, etc.

src/emulator/util/include/util/BitField.h
```diff
@@ -0,0 +1,257 @@
+#pragma once
+
+#include "util/types.h"
+#include "log.h"
+
+#define strcat_(x, y) x ## y
```
petmac: These macros and includes look unnecessary.

VelocityRa: Oh yea I added them when I was debugging `bf_t` `constexpr` stuff and forgot to take them out.

src/emulator/util/CMakeLists.txt
VelocityRa: All other utility files use it though... that's the main reason I added it.
I can try stripping it off to absolutely necessary stuff, or idk, splitting it into new files with appropriate names?

src/emulator/modules/SceGxm/SceGxm.cpp
VelocityRa: Ok.

src/emulator/gxm/src/gxm.cpp
VelocityRa: I removed these because we already warn _once_ in `sceGxmTextureInitLinear`, with the log here here it would spam the log every frame.

src/emulator/gxm/src/gxm.cpp
VelocityRa: Colour palette index? Paletted formats are handled above, with `GL_RGBA`.
`GL_INTENSITY8` is wrong for  these formats because it doesn't return the max value for the channel "1", it just returns this one value ("R") everywhere.

src/emulator/gxm/src/gxm.cpp
VelocityRa: Oh, true.

src/emulator/gxm/src/gxm.cpp
VelocityRa: Ah yes. The rest of the code runs ok for me with 120, so let's try that on Mac too.

src/emulator/gxm/src/gxm.cpp
petmac: Ah, that makes sense. Thanks!

src/emulator/modules/SceGxm/SceGxm.cpp
```diff
@@ -380,7 +380,7 @@ EXPORT(int, sceGxmDraw, SceGxmContext *context, SceGxmPrimitiveType primType, Sc
     assert(indexData != nullptr);
 
     if (!host.gxm.isInScene) {
-        return error("sceGxmDraw", SCE_GXM_ERROR_NOT_WITHIN_SCENE);
+        (__func__, SCE_GXM_ERROR_NOT_WITHIN_SCENE);
```
petmac: Missing the return statement.

src/emulator/modules/SceRtc/SceRtcUser.cpp
```diff
@@ -18,58 +18,59 @@
 #include "SceRtcUser.h"
 
 #include <host/rtc.h>
+#include <util/types.h>
```
petmac: I think we should delete `util/types.h`

src/emulator/gxm/include/gxm/functions.h
```diff
@@ -32,19 +32,20 @@ GLenum attribute_format_to_gl_type(SceGxmAttributeFormat format);
...
-    GLenum translate_wrap_mode(SceGxmTextureAddrMode src);
-    GLenum translate_minmag_filter(SceGxmTextureFilter src);
+    SceGxmTextureBaseFormat get_base_format(const SceGxmTextureFormat src);
+    std::uint32_t get_swizzle(const SceGxmTextureFormat src);
+    bool is_paletted_format(const SceGxmTextureFormat src);
+    GLenum translate_internal_format(const SceGxmTextureFormat src);
+    GLenum translate_format(const SceGxmTextureFormat src);
+    GLenum translate_wrap_mode(const SceGxmTextureAddrMode src);
+    GLenum translate_type(const SceGxmTextureFormat format);
+    GLenum translate_minmag_filter(const SceGxmTextureFilter src);
```
petmac: Some unnecessary consts here too.

src/emulator/gxm/src/gxm.cpp
petmac: Ahh you're right, sorry.

src/emulator/gxm/include/gxm/types.h
```diff
@@ -4,12 +4,13 @@
 #include <condition_variable>
 #include <glutil/object.h>
 #include <glutil/object_array.h>
+#include <rpcs3/BitField.h>
```
petmac: We probably need a public dependency added to gxm/CMakeLists.txt.

src/emulator/modules/SceGxm/SceGxm.cpp
VelocityRa: 🤦‍♂️

src/emulator/gxm/include/gxm/types.h
VelocityRa: I've already done that haven't I?
`target_link_libraries(gxm PUBLIC crypto glutil mem sdl2 rpcs3 threads vita-headers)`

src/emulator/gxm/src/gxm.cpp
```diff
@@ -42,31 +42,46 @@ static const char *parameter_name(const SceGxmProgramParameter &parameter) {
     return reinterpret_cast<const char *>(bytes + parameter.name_offset);
 }
 
+static SceGxmParameterType parameter_type(const SceGxmProgramParameter parameter) {
```
petmac: This should probably be a reference because `SceGxmProgramParameter` is a struct.

PR #121, opened 2018-04-04
Setting min/mag filters to linear filtering on init.

Rinnegatamante:
This fixes graphics in some homebrews (Happy Land for example) and stops homebrews from spamming on console for unimplemented mag/min filters translations.


src/emulator/modules/SceGxm/SceGxm.cpp
```diff
@@ -1667,6 +1667,7 @@ EXPORT(int, sceGxmTextureInitLinear, SceGxmTexture *texture, Ptr<const void> dat
     texture->data_addr = data.address() >> 2;
     texture->swizzle_format = (texFormat & 0x7000) >> 12;
     texture->normalize_mode = 1;
+    texture->min_filter = texture->mag_filter = SCE_GXM_TEXTURE_FILTER_LINEAR;
```
VelocityRa: should be 2 separate lines imo

petmac: Apart from that it's all good.

PR #124, opened 2018-04-04
[gxm] Using error() whenever an error is returned.

Rinnegatamante:


VelocityRa:
You could use `__FUNCTION__` everywhere as the string to pass `error`.

VelocityRa:
Or we could just use the logging system (`LOG_ERROR_IF`) to avoid all that altogether.

VelocityRa:
I found this
```cpp
 #if __STDC_VERSION__ < 199901L
 # if __GNUC__ >= 2
 #  define __func__ __FUNCTION__
 # else
 #  define __func__ "<unknown>"
 # endif
 #endif
```
that guarantees it existing. I'll add it in `types.h` to my Fixes PR.

PR #132, opened 2018-04-07
IO refactoring

VelocityRa:
- `VitaPartition` enum class
- Translation function instead of string compares
- Joined ux0/uma0 code paths
- Run clang-format

~~Just io.cpp changed.~~


VelocityRa:
Hm I thought about that so that I don't need to define everything twice, but sounds a bit too much work for what it's worth. I'll try it and see how it goes.

VelocityRa:
yea it was easy

PR #144, opened 2018-04-10
Log asserts

VelocityRa:
Adds custom assert: `v3k_assert` which:
- Works on Release mode
- Reports assertion fails to logging system
- Optionally aborts like normal asserts, or not
- Above is decided at runtime. **The argument `-s` or `--strict` makes it exit on failed `assert`s**

Existing codebase is ported to use it.

So:
1) **It functions the same as it did by default, except now it logs failed asserts.**
2) If you want to enable exiting on failed assertions, pass the above arg before the `.vpk` path. Probably more useful for developers.

### DIFF
--- a/src/emulator/audio/src/audio.cpp
+++ b/src/emulator/audio/src/audio.cpp
@@ -19,15 +19,15 @@
 
 #include <audio/state.h>
 #include <util/log.h>
+#include <util/v3k_assert.h>
 
-#include <cassert>
 #include <cstring>
 
 static const int stream_put_granularity = 512;
 
 static void mix_out_port(uint8_t *stream, uint8_t *temp_buffer, int len, AudioOutPort &port, const ResumeAudioThread &resume_thread) {
     int available_to_get = SDL_AudioStreamAvailable(port.callback.stream.get());
-    assert(available_to_get >= 0);
+    v3k_assert(available_to_get >= 0);
 
     while (available_to_get < len) {
         std::unique_lock<std::mutex> lock(port.shared.mutex);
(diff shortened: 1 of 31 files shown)


VelocityRa:
Everyone else was ok with it...

>The way I see it, assert is to trap an invalid program state. Ideally, it halts the program, because the program is essentially guaranteed to crash immediately or soon after an assertion fails anyway.

Well good because it's not `assert`, it's our custom assert. We control it, we can do whatever we want with it. `assert` is indeed really bad and useless for anything more than a simple console application.

>Additionally, assert should never be compiled into Release builds, because an assert should never fail for end users.

We're working on an emulator. Things break, unexpectedly. I'm quite adamant that it should **absolutely** be compiled to Release buids so that we know when something is wrong. Like I said, this is an emulator. We won't be able to test every game ourselves, logs will often be the only thing we have to work with.

>I feel doubling down on assert is probably the wrong way to go. Any cases where assert is used to indicate something that's unimplemented should be replaced with LOG_WARN_IF or similar. Any case where assert is used instead of proper error checking, should be replaced with proper error checking.

1) we can't control everything can control with v3k_assert, it just logs and that's it. Also it **needs** a third argument, I couldn't figure out how to do it with either 2, or vararg of arguments. On the other hand I got `v3k_assert` to work with either just printing the condition and if fmt/varargs are specified, it prints those too, but it's OPTIONAL.
2) >proper error checking

Which is? Exceptions? I feel this would complicate things for no reason.

>I take responsibility for the overuse of assert in the codebase. I've been sloppy in the past, using asserts to take a note of functionality I partially implemented, or just to postpone doing proper error checking to another time.

There's so much stuff to work on that this is often needed and that's ok.

VelocityRa:
I don't understand what the problem is honestly, the current way it worked was horrible. Release buids crashed and we had no idea why because asserts were just invisible. This is the right way to go about this.

VelocityRa:
Unless you're willing to properly handle every `assert` in the codebase at once.

PR #145, opened 2018-04-12
Fix warnings

petmac:


src/emulator/io/src/io.cpp
```diff
@@ -139,7 +139,7 @@ bool init(IOState &io, const char *pref_path) {
 
 static std::pair<VitaPartition, std::string>
 translate_partition(const std::string &full_path) {
-    std::uint32_t partition_end_idx = full_path.find(":");
+    std::string::size_type partition_end_idx = full_path.find(":");
```
VelocityRa: Maybe just do `const auto` here.

src/emulator/util/src/util.cpp
```diff
@@ -32,7 +32,9 @@ std::string& LOG_FILE_NAME = "/vita3k.log";
 
 void init_logging()
 {
+#ifdef _MSC_VER
```
VelocityRa: You may as well also change the #ifdef a few lines below on 41, to `_MSC_VER`.

src/emulator/io/src/io.cpp
VelocityRa: Maybe just do `const auto` here.

PR #153, opened 2018-04-16
 [gxm] Attrib location and set_uniform fixes. 

Rinnegatamante:
With this Flood It! works correctly.
What this PR does:

- Fixes some bad behaviours in set_uniform adding support for "float" uniforms.
- Fixes attrib location leak. Before we were using only glVertexAttribPointer with index 0,4,8,12,...; now we use 0,1,2,3,4... allowing us to use up to 16 vertex attributes like it should be.

This allows Flood It! to correctly work:
immagine
immagine


VelocityRa:
I had that `/ 4` hack locally as well, but I'm not sure if it's correct. IIRC you or PetMac had told me that it's wrong.

Rinnegatamante:
Welp the current code too is wrong, with it we won't be able to use more than 4 vertexattribs resulting in pratically every single commercial game not being playable probably.

PR #167, opened 2018-04-19
Implement sceKernelGetRandomNumber

bentokun:


src/emulator/modules/SceLibKernel/SceLibRng.cpp
```diff
@@ -15,10 +15,26 @@
...
+EXPORT(int, sceKernelGetRandomNumber, void *output, unsigned size) {
+    srand(time(nullptr));
+
+    if (size > 64) {
+        return -1;
+    }
+
+    int* int_output = reinterpret_cast<int*>(output);
+
+    for (unsigned int i = 0; i < size; i ++) {
```
mailwl: buffer overflow? size - maximum 64 bytes, but you use `size*sizeof(int)`

bentokun: I think it should be casted to char, working on it.

mailwl: you may drop for as well, just use std::generate, for example:
```
char* out = reinterpret_cast<chat*>(output);
std::generate(out , out  + size, std::rand);
```

VelocityRa: `RAND_MAX` can be 32767 though, and `std::rand` returns an int. Not sure if this would leave empty bytes? As I recommended on Discord, I'd do this the C++ way instead of using `rand()`

bentokun:
Reworking on this on a new branch using C++ random library.

mailwl:
heh, i tryed some time ago - it can't generate random chars (feature/bug in stl maybe?)

PR #176, opened 2018-04-25
Implemented sceGxmSetFrontDepthWriteEnable.

Rinnegatamante:


src/emulator/modules/SceGxm/SceGxm.cpp
```diff
@@ -1141,8 +1141,14 @@ EXPORT(int, sceGxmSetFrontDepthFunc) {
     return unimplemented("sceGxmSetFrontDepthFunc");
 }
 
-EXPORT(int, sceGxmSetFrontDepthWriteEnable) {
-    return unimplemented("sceGxmSetFrontDepthWriteEnable");
+EXPORT(void, sceGxmSetFrontDepthWriteEnable, SceGxmContext *context, SceGxmDepthWriteMode enable) {
+    if (!context->two_sided) {
+        glDepthMask(enable == SCE_GXM_DEPTH_WRITE_ENABLED ? GL_TRUE : GL_FALSE);
```
petmac: I would move this line out of the `if` since it's the same both times.

Then you can just warn if `context->two_sided` is set.

PR #179, opened 2018-04-26
Raising waiting threads when a thread returns.

Rinnegatamante:


VelocityRa:
@Rinnegatamante Merge?

Rinnegatamante:
This didn't solve the issue referred in the PR (before these commit, nxengine used to softlock at startup, with this, it softlocks as soon as you try to load Cave Story data).

Rinnegatamante:
Ready to review. It now fully fixes CaveStory regression.

src/emulator/kernel/src/thread/thread.cpp
```diff
@@ -105,6 +105,16 @@ SceUID create_thread(Ptr<const void> entry_point, KernelState &kernel, MemState
     return thid;
 }
 
+void raise_waiting_threads(ThreadState *thread) {
```
VelocityRa: Can you make this function available in `thread_functions.h`?

PR #192, opened 2018-04-28
Use export name rather than __func__, string, etc.

petmac:


Rinnegatamante:
Don't forget to run clang-format (i've spotted some minor things that should be fixed when running it)

VelocityRa:
Waiting for this to be merged to rebase #231

PR #196, opened 2018-04-29
Replaced unique_lock with lock_guard when possible.

Rinnegatamante:


VelocityRa:
To provide a reason, since I mentioned this on Discord:
`std::lock_guard` is a lighter version of `std::unique_lock`, it provides optional deferred locking and explicit unlocking and you can also transfer a lock to another.

We rarely need any of that, so `std::lock_guard` should be preferred.

PR #200, opened 2018-04-29
[io] Redirect ux0:app/<title_id> to app0:

d3m3vilurr:
retail vita system allow both `ux0:/...` and `ux0:...`

### DIFF
--- a/src/emulator/io/src/io.cpp
+++ b/src/emulator/io/src/io.cpp
@@ -41,6 +41,7 @@
 #include <iostream>
 #include <string>
 #include <tuple>
+#include <array>
 
 static ReadOnlyInMemFile open_zip(mz_zip_archive &zip, const char *entry_path) {
     const int index = mz_zip_reader_locate_file(&zip, entry_path, nullptr, 0);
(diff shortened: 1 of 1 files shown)


src/emulator/io/src/io.cpp
```diff
@@ -170,20 +170,25 @@ SceUID open_file(IOState &io, const std::string &path_, int flags, const char *p
     std::string device_name;
     std::tie(device, device_name) = translate_device(path);
 
-    const std::string ux_app_path{ "ux0:/app/" };
+    const std::vector<std::string> ux_app_paths = { "ux0:/app/", "ux0:app/" };
```
VelocityRa: use `std::array`

d3m3vilurr: done

VelocityRa:
Yeah, ideally we'd 'normalize' the string before doing any redirecting or processing.

Can you do it that way instead, perhaps? Instead of touching redirecting, or any other code, make a `normalize_path` function that removes this `/` beforehand (and in potentially in the future other things too).

VelocityRa:
Discussed in Discord, I'll do this in a different way at some point.

PR #207, opened 2018-05-03
[io] Support tty1:

VelocityRa:
Probably helps some commercial games log more.


TehPsychedelic:
When will this one be merged. I just noticed a bunch of games show that TTY1 message.

VelocityRa:
Good point.

PR #211, opened 2018-05-04
Fix status on loading modules

frangarcj:
Some games use status value as a way to call functions from the module

### DIFF
--- a/src/emulator/kernel/src/thread.cpp
+++ b/src/emulator/kernel/src/thread.cpp
@@ -190,12 +190,13 @@ bool run_callback(ThreadState &thread, Address &pc, Address &data) {
     return run_thread(thread, true);
 }
 
-bool run_on_current(ThreadState &thread, const Ptr<const void> entry_point, SceSize arglen, Ptr<void> &argp) {
+uint32_t run_on_current(ThreadState &thread, const Ptr<const void> entry_point, SceSize arglen, Ptr<void> &argp) {
     std::unique_lock<std::mutex> lock(thread.mutex);
     stop(*thread.cpu);
     write_reg(*thread.cpu, 0, arglen);
     write_reg(*thread.cpu, 1, argp.address());
     write_pc(*thread.cpu, entry_point.address());
     lock.unlock();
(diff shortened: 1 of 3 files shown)


src/emulator/kernel/src/thread.cpp
```diff
@@ -190,12 +190,13 @@ bool run_callback(ThreadState &thread, Address &pc, Address &data) {
...
+uint32_t run_on_current(ThreadState &thread, const Ptr<const void> entry_point, SceSize arglen, Ptr<void> &argp) {
     std::unique_lock<std::mutex> lock(thread.mutex);
     stop(*thread.cpu);
     write_reg(*thread.cpu, 0, arglen);
     write_reg(*thread.cpu, 1, argp.address());
     write_pc(*thread.cpu, entry_point.address());
     lock.unlock();
-    return run_thread(thread, false);
+    run_thread(thread, false);
+    return read_reg(*thread.cpu, 0);
```
VelocityRa: Does this basically return the return value of `module_start`?

frangarcj: yes

src/emulator/modules/SceLibKernel/SceLibKernel.cpp
```diff
@@ -1277,7 +1277,19 @@ EXPORT(int, sceKernelLoadStartModule, char *path, SceSize args, Ptr<void> argp,
...
-    run_on_current(*thread, entry_point, args, argp);
+    uint32_t result = run_on_current(*thread, entry_point, args, argp);
+    char *module_name = module->second.get()->module_name;
+
+    LOG_INFO("{} returned {:#08x}", module_name, result);
+
+    if (status)
+        *status = result;
+
+    thread->to_do = ThreadToDo::exit;
```
VelocityRa: I think we do these 4 lines pretty often, maybe it should be a function?

frangarcj: I think but it will be the purpose of another PR

PR #212, opened 2018-05-05
Add support for Linux

bentokun:
- Add a Unicorn binary (compiled by myself)
- Add Unix Unicorn library to CMakeLists
- Add Linux nightly


Sunderland93:
Why not add Unicorn as submodule?

Sunderland93:
> It doesn't have a CMakeLists.

Yuzu are using CMake, but Unicorn distributed as submodule https://github.com/yuzu-emu/yuzu/blob/master/CMakeLists.txt

d3m3vilurr:
pr is pretty good. and IMO, adding header and built binary into external space is also acceptable(because other platform also use that approach), but propably all platform have same headers, we should combine one in later :)

PR #215, opened 2018-05-06
Float args support

frangarcj:


src/emulator/module/include/module/lay_out_args.h
```diff
@@ -47,23 +48,39 @@ constexpr std::tuple<ArgLayout, LayoutArgsState> add_to_gpr_or_stack(const Layou
...
-    
+    const LayoutArgsState next_state = { next_gpr_used, state.stack_used, state.float_used };
+
+    return { layout, next_state };
+}
+
+template <typename Arg>
+constexpr std::tuple<ArgLayout, LayoutArgsState> add_to_float(const LayoutArgsState &state) {
+    const size_t float_index = state.float_used;
+    const size_t next_float_used = state.float_used + 1;
```
petmac: No time to properly review but floats probably run out of registers and need to spill onto stack like integers do.

frangarcj: Doc says that they will use fp registers and I see a lot of them. S0-S31

frangarcj: I've also checked that on assembly code for sceGxmSetViewport and it is all registers

frangarcj:
I need someone on Windows to fix this pr

src/emulator/module/include/module/read_arg.h
```diff
@@ -38,18 +38,27 @@ std::enable_if_t<sizeof(T) == 8, T> read_from_gpr(CPUState &cpu, const ArgLayout
...
+    case ArgLocation::gpr:
+        return read_from_gpr<T>(cpu, arg);
+    case ArgLocation::stack: {
+        const Address sp = read_sp(cpu);
+        const Address address_on_stack = static_cast<Address>(sp + arg.offset);
+        return *Ptr<T>(address_on_stack).get(mem);
+    }
+    case ArgLocation::fp:
+        if constexpr (std::is_same_v<T, float>)
+            return read_from_fp<T>(cpu, arg);
```
petmac: This `if` should be redundant, as `arg.location` should always be `ArgLocation::fp` for `float`s. If that's not true, we're going to be returning default values for a bunch of non-float args that somehow get assigned to a floating point register. (`ArgLocation::fp`)

petmac: Never mind, explained in Discord.

PR #218, opened 2018-05-06
Wrong signature on sceKernelGetProcessTime

frangarcj:


src/emulator/modules/SceLibKernel/SceLibKernel.cpp
```diff
@@ -1149,8 +1149,10 @@ EXPORT(int, sceKernelGetPMUSERENR) {
     return unimplemented("sceKernelGetPMUSERENR");
 }
 
-EXPORT(SceUInt64, sceKernelGetProcessTime) {
-    return rtc_get_ticks(host.kernel.base_tick.tick);
+EXPORT(SceUInt32, sceKernelGetProcessTime, SceUInt64 *time) {
```
VelocityRa: Yea probably returns int like every other func (that returns error code). Doesn't matter much since you return 0, but still let's have it correct.

PR #222, opened 2018-05-06
SceGxmParameter semantic field

frangarcj:


src/emulator/gxm/include/gxm/types.h
```diff
@@ -186,7 +186,7 @@ struct SceGxmProgramParameter {
         bf_t<uint16_t, 8, 4> component_count; // applicable for constants, not applicable for samplers (select size like float2, float3, float3 ...)
         bf_t<uint16_t, 12, 4> container_index; // applicable for constants, not applicable for samplers (buffer, default, texture)
     };
-    uint16_t unknown; // Maybe relevant to SCE_GXM_PARAMETER_CATEGORY_AUXILIARY_SURFACE or SCE_GXM_PARAMETER_CATEGORY_UNIFORM_BUFFER.
+    uint16_t semantic; // Maybe relevant to SCE_GXM_PARAMETER_CATEGORY_AUXILIARY_SURFACE or SCE_GXM_PARAMETER_CATEGORY_UNIFORM_BUFFER.
```
VelocityRa: Semantics are conventionally present in _vertex attributes_ (see [DX/HLSL semantics](https://msdn.microsoft.com/en-us/library/windows/desktop/bb509647(v=vs.85).aspx)). And in HLSL [they're only for code clarity](https://gamedev.stackexchange.com/questions/70821/do-the-predefined-hlsl-semantics-carry-any-meaning), the compiler ignores them.
So I think this is what this field is actually, not related to auxiliary surfaces or uniform buffers.

Edit: Actually yeah, [take a look at this](https://github.com/vitasdk/vita-headers/blob/master/include/psp2/gxm.h#L1179-L1195).

frangarcj: What should I do? remove comment?

PR #230, opened 2018-05-08
Several fixes done for getting Downwell to show things

frangarcj:


src/emulator/gxm/src/gxm.cpp
```diff
@@ -522,7 +522,12 @@ AttributeLocations attribute_locations(const SceGxmProgram &vertex_program) {
     for (uint32_t i = 0; i < vertex_program.parameter_count; ++i) {
         const SceGxmProgramParameter &parameter = parameters[i];
         if (parameter.category == SCE_GXM_PARAMETER_CATEGORY_ATTRIBUTE) {
-            locations.emplace(parameter.resource_index, parameter_name_raw(parameter));
+            std::string name = parameter_name_raw(parameter);
```
VelocityRa: Commit name is wrong here https://github.com/Vita3K/Vita3K/pull/230/commits/c681c6824a0cddcad6df361ee9e1fd05b64d44ca
Also should probably squash https://github.com/Vita3K/Vita3K/pull/230/commits/9a2c5021d171de001e27d4a799eefcf84a4d0b4b with this commit.
And add a `TODO:` here that it shoud be changed (to being an actual GLSL struct) when we're on GLSL >120.

PR #233, opened 2018-05-09
Stubbed some functions and implemented sceNpTrophySetupDialog.

Rinnegatamante:


src/emulator/modules/SceNgs/SceNgs.cpp
```diff
@@ -41,8 +43,10 @@ EXPORT(int, sceNgsPatchRemoveRouting) {
     return UNIMPLEMENTED();
 }
 
-EXPORT(int, sceNgsRackGetRequiredMemorySize) {
-    return UNIMPLEMENTED();
+EXPORT(int, sceNgsRackGetRequiredMemorySize, uint32_t hSysHandle, void *unk, uint32_t *size) {
```
VelocityRa: `hSysHandle` isn't in our style guide, change to `sys_handle` or something

src/emulator/modules/SceNgs/SceNgs.cpp
```diff
@@ -41,8 +43,10 @@ EXPORT(int, sceNgsPatchRemoveRouting) {
...
+    *size = 1;
```
VelocityRa: Isn't 1 too small? I'd do 1KB just to be safe.

src/emulator/modules/SceNgs/SceNgs.cpp
```diff
@@ -41,8 +43,10 @@ EXPORT(int, sceNgsPatchRemoveRouting) {
...
+    LOG_WARN("Stubbed sceNgsRackGetRequiredMemorySize called (size = 1)");
```
VelocityRa: Don't use yet another way to do stubs, either make `stubbed()` or do `unimplemented_impl("sceNgsRackGetRequiredMemorySize  => size=1");`

src/emulator/modules/SceNgs/SceNgs.cpp
```diff
@@ -61,8 +65,10 @@ EXPORT(int, sceNgsRackSetParamErrorCallback) {
     return UNIMPLEMENTED();
 }
 
-EXPORT(int, sceNgsSystemGetRequiredMemorySize) {
-    return UNIMPLEMENTED();
+EXPORT(int, sceNgsSystemGetRequiredMemorySize, void *params, uint32_t *size) {
+    *size = 1;
```
VelocityRa: ditto

src/emulator/modules/SceNgs/SceNgs.cpp
```diff
@@ -61,8 +65,10 @@ EXPORT(int, sceNgsRackSetParamErrorCallback) {
...
+    LOG_WARN("Stubbed sceNgsSystemGetRequiredMemorySize called (size = 1)");
```
VelocityRa: ditto

src/emulator/modules/SceNgs/SceNgs.cpp
Rinnegatamante: That's the original name of the arg (when added to vita-headers in the near future, it will use that arg naming)

src/emulator/modules/SceNgs/SceNgs.cpp
Rinnegatamante: This got suggested in Discord. (LOG_WARN + return 0). I also find way bad to show an unimplemented warn for stubbed functions. They're different things.

VelocityRa: That's why I suggested to alternatively make a `stubbed("whatever")` logging func (and a macro `STUBBED()` as we do for `unimplemented`), to use for that purpose, if you feel like it.

PR #234, opened 2018-05-10
[WiP] Moved from zip usage to installed games.

Rinnegatamante:
What does this do:

- Emulator can now be opened without giving a vpk via command line, it will show a window where user can select an application to start from installed applications

- Drag'n'dropping a vpk/maidump zip file will cause the emulator to install said application to ux0:/app/TITLEID like on real hardware and then launch the installed app. If the app is already installed, installation process is skipped and the vpk is launched directly

- Fixes translation of savedata0 path making savedata0 partition being accessible.

- app0 access is now faster and applications will also be able to access other applications data files like supposed to be for some homebrews via ux0:app/TITLEID.


src/emulator/host/src/app.cpp
```diff
@@ -32,6 +32,14 @@
 #include <istream>
 #include <iterator>
 
+#ifdef WIN32
```
VelocityRa: Shouldn't have to do this every time we use std::filesystem.
Make a `fs.h` header in `util` directory with just this in it instead.

Rinnegatamante: Done

Rinnegatamante:
Ready for review and merge

src/emulator/host/src/app.cpp
```diff
@@ -97,34 +118,86 @@ bool load_vpk(Ptr<const void> &entry_point, HostState &host, const std::wstring
...
+    output_base_path = host.pref_path + "ux0/app/";
+    std::string title_base_path = output_base_path;
+    if (sfo_path.length() < 20) {
+        output_base_path += host.io.title_id;
+    }
+    title_base_path += host.io.title_id;
+    bool created = fs::create_directory(title_base_path);
+    if (!created) {
+        LOG_INFO("{} already installed, launching application...", host.io.title_id);
+        return true;
```
frangarcj: if already installed there's no savedata0_path filled

Rinnegatamante: Fixed

VelocityRa:
Please refrain from duplicating so much code (talking about redirection code).

Every time I have to change `io.cpp` I end up cleaning and refactoring half of it...

PR #236, opened 2018-05-11
kernel: (Lightweight) Condition Variables

VelocityRa:
Also refactoring of mutex/semaphore code:
- Split some mutex functions to _impl functions because we need to run them from condvar code
- Change naming to have sync object name in the front
- Helper functions to reduce code duplication.
- Correct some returned erorrs
- Change `WaitingThreadData` to accomodate all current sync objects (and couple of future ones)
- Sync logging improvements

and a few other random fixes.

### DIFF
--- a/src/emulator/kernel/include/kernel/state.h
+++ b/src/emulator/kernel/include/kernel/state.h
@@ -46,13 +46,23 @@ typedef std::map<uint32_t, Address> ExportNids;
 
 struct WaitingThreadData {
     ThreadStatePtr thread;
-    std::int32_t lock_count;
-    std::int32_t priority;
+    int32_t priority;
 
-    WaitingThreadData(ThreadStatePtr thread, std::int32_t lock_count, std::int32_t priority)
-        : thread(thread)
-        , lock_count(lock_count)
-        , priority(priority) {}
+    // additional fields for each primitive
+    union {
+        struct { // mutex
+            int32_t lock_count;
+        };
+        struct { // semaphore
+            int32_t signal;
+        };
+        struct { // event flags
+            int32_t wait;
+            int32_t flags;
+        };
+        struct { // condvar
+        };
+    };
 
     bool operator>(const WaitingThreadData &rhs) const {
         return priority > rhs.priority;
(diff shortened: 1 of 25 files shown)


src/emulator/kernel/include/kernel/state.h
```diff
@@ -59,31 +61,75 @@ struct WaitingThreadData {
     }
 };
 
+// Waiting threads for condition variables don't need a lock count (it's binary)
+struct WaitingThreadDataCondvar {
```
frangarcj: I'm using unions inside WaitingThreadData for events. Just a suggestion

VelocityRa: Not sure how that would work. I need to pass this info at _signal_ call site (ex. [here](https://github.com/Vita3K/Vita3K/pull/236/files#diff-c487eaaba136f04f34b9c19061f96493R275)) whereas this struct is created on the _wait_ function.

VelocityRa:
Ready for review.

src/emulator/kernel/include/kernel/state.h
```diff
@@ -46,44 +47,86 @@ typedef std::map<uint32_t, Address> ExportNids;
 
 struct WaitingThreadData {
```
frangarcj: EventFlags have priority too... sorry
```c 
    struct WaitingThreadData {
    ThreadStatePtr thread;
    std::int32_t priority;
    union {
        struct { // mutex
            int32_t lock_count;
        };
        struct { // semaphore
            int32_t signal;
        };
        struct { // event flags
            int32_t wait;
            int32_t flags;
        };
    };

    bool operator>(const WaitingThreadData &rhs) const {
        return priority > rhs.priority;
    }
};
```

src/emulator/kernel/src/thread/sync_primitives.cpp
```diff
@@ -268,5 +333,134 @@ int signal_sema(KernelState &kernel, const char *export_name, SceUID semaid, int
...
+            waiting_thread_data = condvar->waiting_threads.top();
+        }
+        waiting_thread = waiting_thread_data.thread;
+
+        {
+            const std::lock_guard<std::mutex> lock2(waiting_thread->mutex);
+
+            assert(waiting_thread->to_do == ThreadToDo::wait);
+            waiting_thread->to_do = ThreadToDo::run;
+            condvar->waiting_threads.pop();
```
frangarcj: you don't remove target_thread when specific

VelocityRa: Good catch 👍 .
I don't feel like implementing a priority queue now, I will make a couple of helper functions to deal with it until we do.

PR #238, opened 2018-05-12
Added Linux support reference in README

Rinnegatamante:


VelocityRa:
Actually, I think that we should also remove this
image

@bentokun is that right?

bentokun:
Yeah, we have already provided Unicorn for Linux, only SDL2 is required from user side.

PR #246, opened 2018-05-14
Implemented several dialogs related stuffs.

Rinnegatamante:


Rinnegatamante:
it's not stubbed. That's all that function can return.

PR #250, opened 2018-05-14
Link sceKernelWait*Cond to impl. and some logging tweaks

VelocityRa:
Fix something from #236, logging improvements/fixes in sync objects and io.


Rinnegatamante:
LGTM

PR #251, opened 2018-05-15
[mem] Dividing the available pool equally.

Rinnegatamante:
Fixes crashes in vitaGL based homebrews running this specific code: https://github.com/Rinnegatamante/vitaGL/blob/master/source/vitaGL.c#L198-L201


Rinnegatamante:
Why changing it? We still are using a single pool.

PR #252, opened 2018-05-15
vitaGL support

Rinnegatamante:


src/emulator/gxm/src/gxm.cpp
```diff
@@ -388,6 +398,11 @@ void uniform_matrix_4<GLfloat>(GLint location, GLsizei count, GLboolean transpos
     glUniformMatrix4fv(location, count, transpose, value);
 }
 
+template <>
+void uniform_matrix_4<GLint>(GLint location, GLsizei count, GLboolean transpose, const GLint *value) {
+    glUniform4iv(location, count, (GLint *)value);
```
VelocityRa: shouldn't this be something in the likes of `glUniformMatrix*` like the above?

Rinnegatamante: Uniform matrix exists only for floats. This is here just to make the compiler not error out.

PR #256, opened 2018-05-18
Μain refactoring & some random fixes

### LINKED ISSUE #255: Event Flags dialog crashes the emulator
Trying to open the Event Flags debug dialog instantly crashes the emulator.

VelocityRa:
Some needed `main()`, `host` and `imgui` impl. refactoring and a few minor fixes.
Resolves #255.

### Note to devs
Please use `log_hex()` to format hex strings from now on, with `{}` as the format string. It outputs unsigned hex strings in the format of `0xD34DB33F`, with `0` padding according to size of input type.
This is so that we finally have a standardized way to print them.

### DIFF
--- a/src/emulator/gui/include/gui/imgui_impl.h
+++ b/src/emulator/gui/include/gui/imgui_impl.h
@@ -17,27 +17,20 @@
 
 #pragma once
 
-#include <cstdint>
+#include <host/state.h>
 
-static constexpr auto MENUBAR_HEIGHT = 19;
+// clang-format off
+#include <imgui.h>
+#include <gui/imgui_impl_sdl_gl3.h>
+// clang-format on
 
-enum GenericDialogState {
-    UNK_STATE,
-    CONFIRM_STATE,
-    CANCEL_STATE
-};
+#include <glutil/gl.h>
 
-struct HostState;
+namespace imgui {
 
-void DrawMainMenuBar(HostState &host);
-void DrawThreadsDialog(HostState &host);
-void DrawSemaphoresDialog(HostState &host);
-void DrawMutexesDialog(HostState &host);
-void DrawLwMutexesDialog(HostState &host);
-void DrawLwCondvarsDialog(HostState &host);
-void DrawCondvarsDialog(HostState &host);
-void DrawEventFlagsDialog(HostState &host);
-void DrawUI(HostState &host);
-void DrawCommonDialog(HostState &host);
-void DrawGameSelector(HostState &host, bool *is_vpk);
-void DrawReinstallDialog(HostState &host, GenericDialogState *status);
+    void init(WindowPtr window);
+    void draw_begin(WindowPtr window);
+    void draw_main(HostState &host, GLuint TextureID);
+    void draw_end(WindowPtr window);
+
+} // namespace imgui
--- a/src/emulator/host/src/app.cpp
+++ b/src/emulator/host/src/app.cpp
@@ -23,10 +23,10 @@
 #include <host/sfo.h>
 #include <host/state.h>
 #include <host/version.h>
-#include <kernel/thread/thread_functions.h>
 #include <io/state.h>
-#include <util/fs.h>
+#include <kernel/thread/thread_functions.h>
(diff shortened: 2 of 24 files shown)


VelocityRa:
Ready for review.

src/emulator/gui/include/gui/imgui_impl.h
```diff
@@ -0,0 +1,36 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+#pragma once
+
+#include <host/state.h>
```
petmac: Would it be possible to forward declare what we need rather than pull in all the host state?

src/emulator/gui/include/gui/imgui_impl.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include <glutil/gl.h>
+
+// clang-format off
```
petmac: I've seen this in a few places. I'm not saying it's wrong, but why is it needed?

Side note: My preference is to put our headers before 3rd party ones (separated by a newline) as this should encourage ours to be self-contained and not rely on other headers coming first.

src/emulator/host/src/app.cpp
```diff
@@ -235,3 +247,74 @@ bool load_app(Ptr<const void> &entry_point, HostState &host, const std::wstring
...
+
+    return Success;
+}
+
+void no_fb_fallback(HostState &host, GLuint *fb_texture_id) {
+    glGenTextures(1, fb_texture_id);
+    glClearColor(1.0, 0.0, 0.5, 1.0);
+    glClearDepth(1.0f);
+    glViewport(0, 0, host.display.image_size.width, host.display.image_size.height);
+    glMatrixMode(GL_PROJECTION);
```
petmac: These matrix functions aren't in OpenGL 4.1 Core -- we only use shaders so they're not necessary.

src/emulator/main.cpp
```diff
@@ -137,99 +97,31 @@ int main(int argc, char *argv[]) {
...
-        return RunThreadFailed;
-    }
-
-    GLuint TextureID = 0;
-    host.t1 = SDL_GetTicks();
+    GLuint fb_texture_id = 0;
+    host.sdl_ticks = SDL_GetTicks();
     while (handle_events(host)) {
-        if (!TextureID) {
-            glGenTextures(1, &TextureID);
```
petmac: This looks like initialisation code which could just be moved up. I don't think it'll be possible to not have a texture.

src/emulator/gui/include/gui/imgui_impl.h
VelocityRa: It's needed because imgui.h need to be included before any other imgui-related files are included (else you get a compile error) and clang-format changes their order without this.

src/emulator/gui/include/gui/imgui_impl.h
VelocityRa: `WindowPtr` is a `typedef` to `std::shared_ptr<SDL_Window>`, we can't forward declare it in a maintainable way afaik.

src/emulator/gui/include/gui/imgui_impl.h
VelocityRa: And I definitely agree with the side note, I'll change their order.

src/emulator/host/src/app.cpp
VelocityRa: I removed those, `glLoadIdentity` and `glOrtho`.

src/emulator/main.cpp
VelocityRa: This was meant to be a purely-refactoring PR and I'm not sure if that's the case. Surely it's like this for a reason?

petmac: No problem.

PR #258, opened 2018-05-19
Texture cache

petmac:


src/emulator/gxm/src/texture_cache.cpp
```diff
@@ -0,0 +1,128 @@
...
+        } else {
+            // Cache is full. Find least recently used texture.
+            index = find_lru(cache.timestamps, cache.timestamp);
+        }
+        configure = true;
+        upload = true;
+        cache.gxm_textures[index] = gxm_texture;
+    } else {
+        // Texture is cached.
+        index = cached_gxm_texture - cache.gxm_textures.begin();
```
VelocityRa: Can't do ` - gxm_begin;` here?

src/emulator/gxm/src/texture_cache.cpp
```diff
@@ -0,0 +1,128 @@
...
+    const TextureCacheGxmTextures::const_iterator cached_gxm_texture = std::find(gxm_begin, gxm_end, gxm_texture);
+    if (cached_gxm_texture == gxm_end) {
+        // Texture not found in cache.
+        if (cache.used < TextureCacheSize) {
+            // Cache is not full. Add texture to cache.
+            index = cache.used;
+            ++cache.used;
+        } else {
+            // Cache is full. Find least recently used texture.
+            index = find_lru(cache.timestamps, cache.timestamp);
```
VelocityRa: Is this functionality tested? If not, maybe adding a log message would be useful.

src/emulator/gxm/src/texture_cache.cpp
```diff
@@ -0,0 +1,235 @@
...
+}
+
+static const bool operator==(const SceGxmTexture &a, const SceGxmTexture &b) {
+    return memcmp(&a, &b, sizeof(a)) == 0;
+}
+
+static size_t find_lru(const TextureCacheTimestamps &timestamps, TextureCacheTimestamp current_time) {
+    GXM_PROFILE(__func__);
+    
+    uint64_t oldest_age = current_time - timestamps.front();
```
VelocityRa: What't the point of including `current_time` at all when calculating least recently used texture?

Considering `current_time` will be bigger than any given timestamp, isn't it enough to simply search for the entry with the smallest timestamp?

src/emulator/gxm/src/texture_cache.cpp
```diff
@@ -0,0 +1,235 @@
...
+            index = find_lru(cache.timestamps, cache.timestamp);
+        }
+        configure = true;
+        upload = true;
+        cache.gxm_textures[index] = gxm_texture;
+    } else {
+        // Texture is cached.
+        index = cached_gxm_texture - cache.gxm_textures.begin();
+        configure = false;
+        upload = hash != cache.hashes[index];
```
VelocityRa: Change to `upload = (hash != cache.hashes[index]);` for better readability.

src/emulator/gxm/src/texture_cache.cpp
petmac: It was to handle the edge case where the timestamp wraps. It's 64-bit so that _shouldn't_ happen, but you never know.

src/emulator/gxm/src/texture_cache.cpp
petmac: Ah yes! Thanks.

src/emulator/gxm/src/texture_cache.cpp
petmac: Good shout. I "tested" it by reducing the cache size to a small number, but it hasn't had a real world test.

VelocityRa:
@petmac Would it be a good idea to add an option to completely disable this, or just invalidate the cache every frame? Preferably not strictly compile-time so that it's easy to integrate to a config system once we get one.

It will help with finding issues related to the texture cache and not misattributing related bugs.

PR #260, opened 2018-05-22
Move shaders to another repo (submodule)

VelocityRa:
Moved to https://github.com/Vita3K/shaders-db.

Tested on Windows.


timofonic:
Hello.

@VelocityRa Just some naive questions:

- What's the purpose of the shaders? Can they be reused for other projects?
- Would it be possible to use @haasn's [libplacebo](/haasn/libplacebo) with these shaders?
- Why the shaders use non-descriptive names?

Kind regards.

haasn:
> Would it be possible to use @haasn's libplacebo with these shaders?

I have no idea what libplacebo is supposed to do with Vita3K. Maybe I'm missing something, but it just seems completely unrelated.

VelocityRa:
>I probably would have put it in src/external, but I don't mind that much.

@petmac Me too, but current CMake code for deplying/bulding was in `src/emulator` I didn't want to move/touch CMake stuff too much.

PR #267, opened 2018-05-30
Enable vsync for main OpenGL context.

petmac:
Games run on a background thread, so if vsync is disabled, the main thread spins like crazy and burns CPU time on systems with few cores.


petmac:
Pushed branch to main repo rather than fork. Closing.

PR #277, opened 2018-06-10
Add Unicorn as an external project

### LINKED ISSUE #208: Build unicorn ourselves (to get symbols)
Stacktraces from unicorn threads are absolutely useless at the moment, because there are no symbols.
We need to build unicorn ourselves.

For Windows, there's a VS project provided: https://github.com/unicorn-engine/unicorn/tree/master/msvc and CMake apparently offers `INCLUDE_EXTERNAL_MSPROJECT`, so it might be possible.

muemart:
This removes the prebuild Unicorn libraries. Instead, the library is added as a submodule and gets built during the normal build process by using CMake's externalproject_add. It got a bit messy for MSVC because the projects in the repo are for an old version, and I tried to sync the configurations between CMake and the project files.
I haven't tested Linux or MacOS, so would be nice if somebody could. The name for the import library is probably wrong... (EDIT: Travis works, so I guess not)
This would fix #208.


petmac:
Thanks for this. I'll test on macOS before reviewing.

Issue #68, opened 2018-02-15
Switch to Dynarmic ARM Recompiler


0x8080:
https://github.com/MerryMage/dynarmic

This would be recommended to do when possible.

Unicorn is notoriously slow compared to JIT or any other decent recompiler so switching to dynarmic would  result in far better performance the will be greatly needed later on.

(Side Note: YUZU is moving away from unicorn due to performance issues too)

VelocityRa:
We've been talking about this for a few days now (contacted MerryMage as well), so it's definitely planned.

shadowbrony33:
Good to know, also it technically isn't a typo but I should change it to "Switch to a Dynamic ARM Recompiler"

VelocityRa:
Not sure what you mean, unicorn is a"dynamic" (JIT) recompiler too, just slower.

Anyway to add to my previous message, dynarmic isn't compatible with the arch of Vita's CPU yet, else I (or maybe someone else) would have started porting it already. It's in the works though.

shadowbrony33:
Does it just not have all of the opcodes or just doesn't have anything for ARMv7A?

VelocityRa:
There's several differences between ARMv6K and ARMv7A which will need to be implemented, but perhaps the main difference is Thumb-2, which includes a lot of new instructions.

Their semantics for the most part are identical to equivalent ARM instructions (with different encoding) though, so hooking the preexisting ARM emitters for those is what's needed. The decoder for them was already done in the [latest commit](https://github.com/MerryMage/dynarmic/commit/e585e1d49ed65c31edd567510e00508d42decb1c) as you can see.

sunho:
🔥 🦄

sunho:
https://github.com/Vita3K/Vita3K/pull/845

sunho:
We now support dynarmic!

PR #278, opened 2018-06-13
Revert "Add Unicorn as an external project"

petmac:
Reverts Vita3K/Vita3K#277

Crashes on Windows were reported via Discord, so let's revert this and test some more on Windows.


VelocityRa:
I don't think there's a need for this. It should just be fixed in a future commit, no reverts.

d3m3vilurr:
I also same opinion as @VelocityRa don't need revert, just need fix windows build.

muemart:
Found a solution to fix the crashes. The trick is to add `/guard:cf ` as a switch to the compiler. Though it's a bit of a pain to add it in the CMake file, because I can't find a way to set compiler flags with the msbuild command. Any ideas?

VelocityRa:
Apparently the issue happens on the first `svc` instruction, when (before?) the relevant unicorn hook gets called.
Thanks @muemart, I did try it for the `emulator` and `cpu` project but it didn't work, I'm guessing it's needed for all the projects?

In any case, it [looks like](https://msdn.microsoft.com/en-us/library/dn919635.aspx) this has runtime overhead, so maybe we should find a better fix for this. Any ideas?

muemart:
The flag doesn't need to be set for the Vita3K projects, but when compiling unicorn. Sorry if that wasn't clear. Unicorn crashes when calling longjmp (but not every call...), and I found this similar issue [here](https://developercommunity.visualstudio.com/content/problem/115305/longjmp-incompatible-with-guardcf.html) which brought me to that switch. Not sure if there's another fix, but I'm not really sure why that flag fixes the crashes either.

muemart:
After even more trial and error, I found a way to make it work.
Unicorn: Compile with `/Zi` instead of `/ZI`
Vita3K: Disable inlining (`/Ob0`) for cpu, kernel, and modules. In thread.cpp, add a `try{}catch(...){}` around the code inside the call_svc lambda.
I wonder if disabling inlining is too big of a performance hit.

VelocityRa:
Thank you for looking into this more @muemart, but these workarounds are definitely not acceptable. I looked into the issue too as well as using clang instead on Windows yesterday (you can read #development scrollback on our Discord), but no dice.

So, looks like we're gonna have to revert this, it's been a week since master is broken on Windows.

petmac:
Maybe we should accelerate our looking at #68, but I'm aware there's work to be done on Dynarmic itself for that to happen.

VelocityRa:
Not a bad idea... I can try and talk with Mary again, see if anything has changed with ARM-v7 progress.
Although I've been sorta keeping up with its development and I haven't seen much that _directly_ affects us.
I'll look into it.

muemart:
I assume this is a compiler bug, so I'm thinking about opening an issue for MSVC, though I haven't managed to do a minimal repro of the crash.
In the mean time, reverting is probably the best idea unless there's a replacement ready.

PR #280, opened 2018-06-19
Shader fixes

petmac:


petmac:
Good points, I'll clean it up and split non-shader-related changes into a separate PR.

PR #286, opened 2018-07-23
[WIP - DO NOT REVIEW] Dynarmic support

bentokun:
- [x] Add Dynarmic support for Vita3K. Any missing instructions will fallback to Unicorn
- [ ] Add menu to choose between two JIT backend.
- [ ] Add config for logging code, mem access option.
- [ ] Dynarmic VFP FPU enable, read float register

The default thread creation use Dynarmic for now. Feel free to test it, but it's not really useable, as there are some things i still need to figure out.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -40,3 +40,9 @@
 [submodule "src/emulator/shaders-db"]
 	path = src/emulator/shaders-db
 	url = https://github.com/Vita3K/shaders-db.git
+[submodule "src/external/Dynarmic"]
+	path = src/external/Dynarmic
+	url = https://github.com/MerryMage/Dynarmic
+[submodule "src/external/ext-boost"]
+	path = src/external/ext-boost
+	url = https://github.com/MerryMage/ext-boost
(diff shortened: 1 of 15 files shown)


bentokun:
I will squash all of this into one commit later after this is done, so it will not pollute the git history.

MerryMage:
dynarmic comes with a bundled fmtlib; the fmt external *should* not be necessary. (do tell me if this turns out to be incorrect!)

VelocityRa:
Yup and `spdlog` already bundles fmtlib anyway.

[This](
https://github.com/VelocityRa/Vita3K/commit/bad3822cf4182d7db012ad37e4a56134ff36a536#diff-cfec92caf37a4d47d916c167ca006117R42) build script worked fine for me when I was experimenting with dynarmic but @bentokun was having issues IIRC. I only needed to include the dynarmic submodule, not `ext-boost` or `fmtlib`.

This PR is on hold anyway, we decided we're not gonna push for dynarmic yet.
Btw @bentokun we agreed you'd close this and push on another branch, maybe you forgot about it?

bentokun:
Sorry.

PR #288, opened 2018-07-29
Add ccache support

scribam:
Wait for my second commit before to compare compilation times on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,14 @@
 language: cpp
 
+cache: ccache
+
 matrix:
   include:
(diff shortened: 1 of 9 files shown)


scribam:
Travis `osx`:
- Before: 13 min 3 sec
- After: 3 min 31 sec

Travis `linux`:
- Before: 12 min 39 sec
- After: 4 min 12 sec

The second commit about clang-tidy is documented here: https://clang.llvm.org/extra/clang-tidy/checks/modernize-deprecated-headers.html

PR #295, opened 2018-08-08
sceKernelAllocMemBlock: remove assert for SceKernelAllocMemBlockOpt

scribam:
The parameter `SceKernelAllocMemBlockOpt` can be not null when used in `sceKernelAllocMemBlock`, see https://github.com/joshaxey/textquest/blob/3891bcb1cc13c27d599efe18aa34ba49ad62f000/src/graphics.c#L53-L61

This change allows the homebrew TextQuest to go further.

**Before**
```
SceUID export_sceKernelAllocMemBlock(HostState&, SceUID, const char*, const char*, SceKernelMemBlockType, int, emu::SceKernelAllocMemBlockOpt*): Assertion `optp == nullptr' failed.
```
**After**
screenshot from 2018-08-08 19-26-19

### DIFF
--- a/src/emulator/modules/SceSysmem/SceSysmem.cpp
+++ b/src/emulator/modules/SceSysmem/SceSysmem.cpp
@@ -31,7 +31,6 @@ EXPORT(SceUID, sceKernelAllocMemBlock, const char *name, SceKernelMemBlockType t
     assert(name != nullptr);
     assert(type != 0);
     assert(size != 0);
-    assert(optp == nullptr);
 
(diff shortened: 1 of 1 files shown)


VelocityRa:
I wouldn't be surprised if there are other invalid asserts too, but most people of course use Release so they don't trigger and go unnoticed.

I had opened a PR (#144) while ago that checked/logged them on Release (which would help catching things like that) mode but petmac didn't agree it didn't go through. Maybe it's for the best though - it'd definitely generate a lot of log spam.

PR #297, opened 2018-08-09
First steps to SPIR-V based shader recompiler

VelocityRa:
## Description

This replaces our simple GLSL stub generation to a SPIR-V based approach: **GXP -> SPIR-V -> GLSL**, with the first step implemented using [glslang](https://github.com/KhronosGroup/glslang)'s SPIR-V emitter and the last step entirely performed by [SPIRV-Cross](https://github.com/KhronosGroup/SPIRV-Cross).

"Stub" generation in this context means that - like before - minimal shaders are generated, with just uniforms/vertex attribute input declarations and no actual function bodies.

So this is pretty minimal, I expect it to get extended upon a lot once we have actual GXP _instruction decoding_ (will probably happen in a separate project, see "Future Work" below).

## Example Output

image
You can see how it handles all types (scalars, vectors, matrices), structs and arrays (matrices not shown here).

I've enabled this output by default for debugging purposes, you can only disable it at compile-time (since we don't have a config system yet).

## Future Work

1) Handling the below limitations:
    - there can only be one struct instance per declared struct
    - struct _array_ instances don't work
    
    You can see how the second limitation results in weird generated code on [this shader](https://hastebin.com/abuxulediq.cpp).

2) **Emitting vertex _outputs_** (fragment inputs) after understanding how they're encoded in shader binaries, by reversing a few related GXM functions (`sceGxmProgramGetOutputRegisterFormat`, `sceGxmReserveVertexDefaultUniformBuffer`, `sceGxmProgramGetFragmentProgramInputs`).

3) **Emitting actual code in _function bodies_**
    Ever-patient @motoharu-gosuto is working hard on reversing the shader ISA and binary encoding here: https://wiki.henkaku.xyz/vita/SGX543.

    After this effort reaches a good point, we'll start implementing a basic GXP frontend (probably in [psvgxp](https://github.com/motoharu-gosuto/psvgxp)) and get that information over to Vita3k for translation to SPIR-V.

    I'm hoping simple shaders won't be _too_ hard to get working soon™, but we'll see.


src/emulator/renderer/src/load_shaders.cpp
```diff
@@ -5,6 +5,7 @@
 
 #include <gxm/functions.h>
 #include <gxm/types.h>
+#include <shadergen/functions.h>
 #include <util/log.h>
 
 #include <fstream>
```
petmac: Are any includes now unnecessary? Quite a lot of code was deleted.

src/emulator/shadergen/src/shadergen_common.cpp
```diff
@@ -0,0 +1,114 @@
...
+        break;
+    default:
+        category = "Unknown type";
+        break;
+    }
+    LOG_DEBUG("{}: name:{:s} semantic:{} type:{:d} component_count:{} container_index:{}",
+        category, parameter_name_raw(parameter), log_parameter_semantic(parameter), parameter.type, uint8_t(parameter.component_count), log_hex(uint8_t(parameter.container_index)));
+}
+
+const SceGxmProgramParameter *program_parameters(const SceGxmProgram &program) {
```
petmac: I'm pretty sure some of these functions are in gxm library.

src/emulator/gxm/CMakeLists.txt
```diff
@@ -9,5 +9,5 @@ add_library(
 )
 
 target_include_directories(gxm PUBLIC include)
-target_link_libraries(gxm PUBLIC mem rpcs3 threads vita-headers)
+target_link_libraries(gxm PUBLIC mem glslang rpcs3 shadergen threads vita-headers)
```
petmac: Should renderer depend on glslang rather than gxm?

src/emulator/shadergen/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+	STATIC
+	include/shadergen/types.h
+	include/shadergen/functions.h
+	include/shadergen/profile.h
+	src/shadergen_common.cpp
+	src/shadergen_spirv.cpp
+)
+
+target_include_directories(shadergen PUBLIC include)
+target_link_libraries(shadergen PUBLIC gxm util vita-headers)
```
petmac: Gxm dependency seems like it could be private.

src/emulator/shadergen/src/shadergen_common.cpp
VelocityRa: Good catch, didn't see them. Gotta get used to there not being a singular `gxm.cpp` anymore.

src/emulator/renderer/src/load_shaders.cpp
VelocityRa: `gxm/functions.h` was not needed

src/emulator/gxm/CMakeLists.txt
VelocityRa: just removing `glslang` from there worked

PR #298, opened 2018-08-10
Fix game selector being cut off

petmac:


src/emulator/gui/src/ui_main.cpp
```diff
@@ -25,17 +25,11 @@ using namespace std::string_literals;
 
 void DrawGameSelector(HostState &host, AppRunType *run_type) {
     const ImVec4 text_color = ImVec4(255, 255, 0, 255);
+    const ImVec2 display_size = ImGui::GetIO().DisplaySize;
 
-    auto render_game_select_item = [&](const std::string &button_text, const std::string &title_id) {
```
VelocityRa: 🤦‍♂️ sorry about that

PR #304, opened 2018-08-18
[WIP] Implement sceIoGetstatByFd

bentokun:
I uses some ugly black magic, this is a temporary solution until a proper VFS is implemented.
This is not tested on Unix or MacOS yet.


VelocityRa:
Why not keep paths ourselves in a `std::map` or something, instead of using native APIs?

bentokun:
Yeah, it was dumb of me to do so, waste of time.

VelocityRa:
Can you change the code back to doing that? I'd like to prevent such platform specific code when we can afford it and I think it's a better solution anyway. Thanks.

bentokun:
On my way.

bentokun:
Closing for new PR since my local got ruined by rebase.

PR #310, opened 2018-08-24
Boost Integration

VelocityRa:
We mainly need this for config/command-line arg support, but I integrated it in a way that it's both easy for us to add more libraries as needed and easy to build for devs (no system Boost installation needed).

Using a custom Boost subset, made with `bcp`: https://github.com/Vita3K/ext-boost

Currently the ext-boost repo contains:
- Boost.ProgramOptions
- Boost.Filesystem
- Boost.System

No direct usage for the last one yet, but it may be and it was already
a dependency by another lib anyway, so I included it in the "guaranteed"
libraries in the `bcp` command.

TODO:
- [x] AppVeyor
- Travis
  - [x] Linux
  - [x] macOS

### DIFF
--- a/gen.bat
+++ b/gen.bat
@@ -1,3 +1,11 @@
+
+IF "%CI%"=="" (
+	mkdir src/external/boost-build
+	cd src/external/boost
+	b2 -j5 --stagedir=../boost-build stage
+	cd ../../..
+)
+
 mkdir build-windows
 pushd build-windows
 cmake -G "Visual Studio 15 2017 Win64" ..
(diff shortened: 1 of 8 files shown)


VelocityRa:
Ready for review.

gen.bat
```diff
@@ -1,3 +1,11 @@
+
+IF "%CI%"=="" (
+	mkdir src/external/boost-build
+	cd src/external/boost
+	b2 -j5 --stagedir=../boost-build stage
+	cd ../../..
```
VelocityRa: Couldn't get `pushd`/`popd` to work.

PR #312, opened 2018-08-28
Several fixes

frangarcj:
Description in every commit


src/emulator/renderer/src/texture_palette.cpp
```diff
@@ -16,7 +16,7 @@ void palette_texture_to_rgba_4(uint32_t *dst, const uint8_t *src, size_t width,
     for (size_t y = 0; y < height; ++y) {
         uint32_t *const dst_row = &dst[y * width];
         const uint8_t *const src_row = &src[y * stride];
-        for (size_t x = 0; x < width; ++x) {
+        for (size_t x = 0; x < width - 1; ++x) {
```
petmac: Maybe `++x` should be `x += 2` here. What do you think? Then we can revert the `- 1`.

frangarcj: I'll put the x +=2 but leave the condition

petmac: It shouldn't really make a difference, as I think the widths are always even.

PR #322, opened 2018-09-13
Implement vargs and printf functions

pent0:
- Implement a **vargs** class . The **vargs** class contains the layout state and argument layout struct. Each time the **next<T>(CpuState&, MemState&)** function is called, the **vargs** class call *add_arg_to_state* to get the next layout, read the argument and return it. Users can also reset the **vargs** to default state.
- In order to achieve that, i store the layout state after each argument laid out.
- Printf is from [here](https://github.com/mpaland/printf), modified for vita3k usage.

Reviews are welcomed :D


bentokun:
Result with hex format: 
image

src/emulator/modules/SceLibKernel/ScePrintf.h
```diff
@@ -0,0 +1,690 @@
...
+    // evaluate flags
+    flags = 0U;
+    do {
+      switch (*format) {
+        case '0': flags |= FLAGS_ZEROPAD; format++; n = 1U; break;
+        case '-': flags |= FLAGS_LEFT;    format++; n = 1U; break;
+        case '+': flags |= FLAGS_PLUS;    format++; n = 1U; break;
+        case ' ': flags |= FLAGS_SPACE;   format++; n = 1U; break;
+        case '#': flags |= FLAGS_HASH;    format++; n = 1U; break;
+        default :                                   n = 0U; break;
```
elad335: You can remove variable `n` and replace this usage with a simple break. do while (n) can be replaced with while (true).

bentokun: I don't understand. I'm not writing this code, but surely they did this for a reason (that code stay that long in the repo without anyone modify it). The ```break``` can only break the switch, i'm not sure how we can break the while from the switch (maybe im dumb). Do you have an example?

pent0:
@PetMac Please review again

pent0:
@petmac Request review again :)

petmac:
Sorry for not reviewing this yet. I've not had enough sit-down time with a computer to have a good look.

src/emulator/module/include/module/lay_out_args.h
```diff
@@ -98,17 +92,24 @@ constexpr void add_args_to_layout(ArgLayout &head, LayoutArgsState &state) {
...
-        add_args_to_layout<Tail...>(*(&head + 1), state);
+        *(&state + 1) = state;
+        add_args_to_layout<Tail...>(*(&head + 1), *(&state + 1));
     }
 }
 
 template <typename... Args>
-constexpr ArgsLayout<Args...> lay_out() {
+constexpr std::tuple<ArgsLayout<Args...>, LayoutArgsState> lay_out() {
+    std::array<LayoutArgsState, sizeof...(Args)> states = {};
```
petmac: Do we need an array of states? I think we should only need to return the last one, after all the (non-varargs) args have been added.

petmac: I would also add a special case in `add_arg_to_layout` (hidden in diff so I can't comment) to not do anything to the state for varargs type.

src/emulator/module/tests/arg_layout_tests.cpp
```diff
@@ -40,7 +40,7 @@ static std::ostream &operator<<(std::ostream &out, const ArgLayout &layout) {
 }
 
 TEST(lay_out, gpr_overflows_to_stack) {
-    const auto actual = lay_out<int8_t, uint8_t, int16_t, uint16_t, int32_t, uint32_t>();
+    const auto actual = std::get<0>(lay_out<int8_t, uint8_t, int16_t, uint16_t, int32_t, uint32_t>());
```
petmac: I think I would have preferred that `expected` was fixed to be the right type, then the unit tests can test that the final states are correct.

petmac: Please also add a unit test for a typical printf style function, so we check that the vararg does not affect the arg layout. i.e., the format arg is added but the vararg does not take up any extra register or stack space.

petmac:
Review comments added! Overall it's shaping up really nicely, just a few small fixes needed.

Thanks again!

src/emulator/module/tests/arg_layout_tests.cpp
pent0: The varargs does indeeds take extra space as you say. I will remove unneccessary codes (in the lay_out_args)

src/emulator/module/include/module/lay_out_args.h
```diff
@@ -91,24 +89,31 @@ constexpr std::enable_if_t<sizeof...(Args) == 0> add_args_to_layout(ArgLayout &h
...
-    // Add the argument at the head of the list.
-    const std::tuple<ArgLayout, LayoutArgsState> result = add_arg_to_layout<Head>(state);
-    head = std::get<0>(result);
-    state = std::get<1>(result);
-
-    // Recursively add the remaining arguments.
-    if (sizeof...(Tail) > 0) {
-        add_args_to_layout<Tail...>(*(&head + 1), state);
+    // Returns immidiately if the Head is an varargs
+    if constexpr(std::is_same_v<Head, module::vargs>) {
```
VelocityRa: Can't just
```suggestion
    return std::is_same_v<Head, module::vargs>::value;
```
?

src/emulator/module/src/read_arg.cpp
```diff
@@ -0,0 +1,8 @@
+#include <module/read_arg.h>
+#include <module/vargs.h>
+
+// Workaround for old Clang and GCC
+template <>
+module::vargs make_vargs(const LayoutArgsState &state) {
+    return module::vargs {state};
+}
```
VelocityRa: newline

src/emulator/module/include/module/vargs.h
```diff
@@ -0,0 +1,49 @@
...
+    T next(CPUState &cpu, MemState &mem) {
+        const auto state_tuple = add_arg_to_layout<T>(layoutState);
+
+        layoutState = std::move(std::get<1>(state_tuple));
+        currentLayout = std::move(std::get<0>(state_tuple));
+
+        return read<T>(cpu, currentLayout, mem);
+    }
+};
+}
```
VelocityRa: newline

src/emulator/module/include/module/read_arg.h
```diff
@@ -64,10 +68,19 @@ T read(CPUState &cpu, const ArgLayout &arg, const MemState &mem) {
...
-}
+    // Note (bentokun): The else block was intentionally made to workaround evaluation
+    // fault where MSVC evaluates the rest of the function when the Arg type is vargs
+    if constexpr(std::is_same_v<Arg, module::vargs>) {
+        return make_vargs<Arg>(state);
+    } else {
+        const ArmType bridged = read<ArmType>(cpu, args[index], mem);
+        return BridgeTypes<Arg>::arm_to_host(bridged, mem);
+    }
+}
```
VelocityRa: newline

src/emulator/module/include/module/lay_out_args.h
pent0: No. That makes no sense if you apply that with the rest of the function.

VelocityRa: Wtf. I swear the code was completely different when I made the comment. Had `return true` and `return false` below.
¯\_(ツ)_/¯

pent0: Ah, I must have bamboozled you, I remove that function when you write that suggestion ;)

VelocityRa:
@petmac Waiting for your review.

pent0:
You don't need to thanks too many time. Thank you :)

PR #326, opened 2018-09-15
Support syncing of host/guest rendering (for RenderDoc)

VelocityRa:
Allows graphics debuggers like RenderDoc to pick up OpenGL calls from GXM emulation.

Graphics debugger support is still not perfect, since with this approach they also pick up OpenGL calls from the host (main) thread, which messes up captures.
For this to work perfectly we'd need to do all rendering from a single thread, which is not trivial to do.

### DIFF
--- a/src/emulator/main.cpp
+++ b/src/emulator/main.cpp
@@ -126,7 +126,12 @@ int main(int argc, char *argv[]) {
         }
         imgui::draw_end(host.window.get());
 
-        host.display.condvar.notify_all();
+        if (!host.display.sync_rendering)
+            host.display.condvar.notify_all();
+        else {
+            std::unique_lock<std::mutex> lock(host.display.mutex);
+            host.display.condvar.wait(lock);
+        }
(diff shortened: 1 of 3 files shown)


src/emulator/main.cpp
```diff
@@ -126,7 +126,12 @@ int main(int argc, char *argv[]) {
         }
         imgui::draw_end(host.window.get());
 
-        host.display.condvar.notify_all();
+        if (!host.display.sync_rendering)
+            host.display.condvar.notify_all();
+        else {
+            std::unique_lock<std::mutex> lock(host.display.mutex);
+            host.display.condvar.wait(lock);
+        }
```
frangarcj: Have you tried to not disable host.display.condvar when host.display.sync_rendering == true?

frangarcj:
I have to take a look at this

frangarcj:
I've done some test and I think your reverse sync is a good method

Issue #318, opened 2018-09-04
Game selction menu doesn't support unicode characters


wfscans:
![image](https://user-images.githubusercontent.com/20528385/45062179-ca824f80-b075-11e8-8485-c85b8cfe26a6.png)

The above in the sea of ?'s are Japanese game titles.

PR #332, opened 2018-09-17
Prettify GUI

VelocityRa:
## Description

Imgui style tweaks, and support for loading external fonts (I chose [Ruda](https://fonts.google.com/specimen/Ruda) Black).

The latter makes loading unicode ranges (#318) easier.

### Before:
image

### After:
image

image

image
(we have to use the old monospaced font for debug dialogs, so that the columns line up)


HexyFify:
yellow is pretty ugly imo, better to use white on black/grey on black

VelocityRa:
It's not yellow, it's orange, our logo's color.
I do agree it's too much though, I'll try white tomorrow.

VelocityRa:
image

Ready for review.

Issue #324, opened 2018-09-15
Add more command-line options


VelocityRa:
Now that we have a proper config system, adding cmd options is trivial (see [this file](https://github.com/Vita3K/Vita3K/blob/master/src/emulator/host/src/config.cpp)).

There are many settings that could be made run-time options chosen via command-line (or a config file, in the future).

Some of these can be our `LOG_*` constants, like the ones [here](https://github.com/Vita3K/Vita3K/blob/ece26545059071e6ba11896219286fec79169ddf/src/emulator/kernel/src/load_self.cpp#L53-L54), [here](https://github.com/Vita3K/Vita3K/blob/master/src/emulator/renderer/src/uniforms.cpp#L12) and [here](https://github.com/Vita3K/Vita3K/blob/master/src/emulator/renderer/src/sync_state.cpp#L15).

scribam:
@VelocityRa There are plenty of options now (see below), is there something missing you would like to have?
```
$ ./Vita3K --help
Vita3K Command Line Interface
Usage: ./Vita3K [OPTIONS] [content-path]

Positionals:
  content-path TEXT                   Path to the app with a .vpk/.zip extension or folder of content to install & run

Options:
  -h,--help                           Print this help message and exit
  -v,--version                        Print the version string
[Option Group: Input]
  Special options for Vita3K
  Options:
    --firmware TEXT                     Path to the firmware file (.pup extension) to install
  
  Input:
    -z,--console [0]                    Start the emulator in console mode.
    -Z,--app-args TEXT                  Argument for app, use ', ' to separate arguments.
    -a,--load-app-list BOOLEAN [0]      Starts the emulator with load app list.
    -S,--self TEXT [eboot.bin]          Path to the self to run inside Title ID
    -r,--installed-path TEXT:{DEFD20485}
                                        Path to the installed app to run
    -s,--recompile-shader TEXT          Recompile the given PS Vita shader (GXP format) to SPIR_V / GLSL and quit
    -d,--deleted-id TEXT:{DEFD20485}    Title ID of installed app to delete
    --pkg TEXT Needs: --zrif            Path to the app file (.pkg extension) to install
    --zrif TEXT Needs: --pkg            zrif to decode the app (base64 format)
[Option Group: Configuration]
  Modify Vita3K's config.yml file
  Logging:
    -A,--archive-log                    Make a duplicate of the log file with TITLE_ID and Game ID as title
[...]

PR #339, opened 2018-10-07
 host/config: LOG_IMPORTS and LOG_EXPORTS converted to Config settings

bsinky:
Changes two of the constants mentioned in #324. I wanted to get some feedback before getting further into the codebase to tackle the other areas mentioned in that issue. Since I'm not a C++ dev, things may be a little rocky at first.

I appreciate any possible constructive criticism, thanks!

### DIFF
--- a/src/emulator/host/src/config.cpp
+++ b/src/emulator/host/src/config.cpp
@@ -51,7 +51,9 @@ ExitCode init(Config &cfg, int argc, char **argv) {
         po::options_description config_desc("Configuration");
         config_desc.add_options()
             ("lle-modules,m", po::value<std::string>(), "Load given (decrypted) OS modules from disk. Separate by commas to specify multiple modules (no spaces). Full path and extension should not be included, the following are assumed: vs0:sys/external/<name>.suprx\nExample: --lle-modules libscemp4,libngs")
-            ("log-level,l", po::value(&cfg.log_level)->default_value(spdlog::level::trace), "logging level:\nTRACE = 0\nDEBUG = 1\nINFO = 2\nWARN = 3\nERROR = 4\nCRITICAL = 5\nOFF = 6");
+            ("log-level,l", po::value(&cfg.log_level)->default_value(spdlog::level::trace), "logging level:\nTRACE = 0\nDEBUG = 1\nINFO = 2\nWARN = 3\nERROR = 4\nCRITICAL = 5\nOFF = 6")
+            ("log-imports,I", po::bool_switch(&cfg.log_imports), "Log Imports")
+            ("log-exports,E", po::bool_switch(&cfg.log_exports), "Log Exports");
         // clang-format on
 
         // Positional args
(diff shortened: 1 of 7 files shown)


src/emulator/renderer/include/renderer/functions.h
```diff
@@ -23,7 +23,7 @@ bool create(FragmentProgram &fp, State &state, const SceGxmProgram &program, con
 bool create(VertexProgram &vp, State &state, const SceGxmProgram &program, const char *base_path);
 void begin_scene(const RenderTarget &rt);
 void end_scene(Context &context, SceGxmSyncObject *sync_object, size_t width, size_t height, size_t stride_in_pixels, uint32_t *pixels);
-bool sync_state(Context &context, const GxmContextState &state, const MemState &mem, bool enable_texture_cache);
+bool sync_state(Context &context, const GxmContextState &state, const MemState &mem, bool enable_texture_cache, bool log_active_shaders, bool log_uniforms);
```
VelocityRa: Can't pass the `Config` here? I'd rather have this passed everywhere instead of `bool`s or whatever That way we don't have to keep adding args as config options increase.

src/emulator/host/src/config.cpp
```diff
@@ -53,7 +53,9 @@ ExitCode init(Config &cfg, int argc, char **argv) {
             ("lle-modules,m", po::value<std::string>(), "Load given (decrypted) OS modules from disk. Separate by commas to specify multiple modules (no spaces). Full path and extension should not be included, the following are assumed: vs0:sys/external/<name>.suprx\nExample: --lle-modules libscemp4,libngs")
             ("log-level,l", po::value(&cfg.log_level)->default_value(spdlog::level::trace), "logging level:\nTRACE = 0\nDEBUG = 1\nINFO = 2\nWARN = 3\nERROR = 4\nCRITICAL = 5\nOFF = 6")
             ("log-imports,I", po::bool_switch(&cfg.log_imports), "Log Imports")
-            ("log-exports,E", po::bool_switch(&cfg.log_exports), "Log Exports");
+            ("log-exports,E", po::bool_switch(&cfg.log_exports), "Log Exports")
+			("log-active-shaders,S", po::bool_switch(&cfg.log_active_shaders), "Log Active Shaders")
```
VelocityRa: Add two spaces to align it with the rest.

src/emulator/host/src/config.cpp
```diff
@@ -53,7 +53,9 @@ ExitCode init(Config &cfg, int argc, char **argv) {
...
+			("log-uniforms,U", po::bool_switch(&cfg.log_uniforms), "Log Uniforms");
```
VelocityRa: ditto

src/emulator/renderer/include/renderer/functions.h
bsinky: I think I'll have to move Config out of host then?

VelocityRa: Yeah, it'd have to be made into a new project, but I'm not sure it's the best approach.

Whatever, keep it this way with the arguments for each option, if it becomes too much of a hassle we'll change it in the future.
In some ways it's better anyway, since it's more explicit and you know which part of the config it uses.

Issue #342, opened 2018-10-12
Dump args to log


wfscans:
Would it be possible to have it dump the arguments(mainly -m) to the log? I was doing some module testing earlier, and having to add it manually... bleh. I believe this'd be pretty easy to do.

PR #344, opened 2018-10-13
host/config: Log config settings during init

bsinky:
Addresses #342 by logging all current config settings during `config::init`.

Not sure if there's a better way to work with the lle_modules vector, I may have missed an existing utility somewhere to combine them into a comma-separated string.

Also not totally sure if INFO is the right logging level, I could see DEBUG being a good fit as well.

Thanks in advance for the review!

Sample log:

```
[20:30:46.738] |I| [config::init]:  input-installed-id: CRFT00001
[20:30:46.740] |I| [config::init]:  lle-modules: test,test,test,test
[20:30:46.741] |I| [config::init]:  log-level: 2
[20:30:46.741] |I| [config::init]:  log-imports: true
[20:30:46.742] |I| [config::init]:  log-exports: true
[20:30:46.742] |I| [config::init]:  log-active-shaders: true
[20:30:46.742] |I| [config::init]:  log-uniforms: false
```

### DIFF
--- a/src/emulator/host/src/config.cpp
+++ b/src/emulator/host/src/config.cpp
@@ -100,6 +100,22 @@ ExitCode init(Config &cfg, int argc, char **argv) {
 
         logging::set_level(static_cast<spdlog::level::level_enum>(*cfg.log_level));
 
+        LOG_INFO_IF(cfg.vpk_path, "input-vpk-path: {}", *cfg.vpk_path);
+        LOG_INFO_IF(cfg.run_title_id, "input-installed-id: {}", *cfg.run_title_id);
+        if (!cfg.lle_modules.empty()) {
+            auto modules = std::string();
+            for (const auto &mod : cfg.lle_modules) {
+                modules += mod + ",";
+            }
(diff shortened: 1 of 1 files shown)


src/emulator/host/src/config.cpp
```diff
@@ -100,6 +100,22 @@ ExitCode init(Config &cfg, int argc, char **argv) {
 
         logging::set_level(static_cast<spdlog::level::level_enum>(*cfg.log_level));
 
+        LOG_INFO_IF(cfg.vpk_path, "input-vpk-path: {}", *cfg.vpk_path);
+        LOG_INFO_IF(cfg.run_title_id, "input-installed-id: {}", *cfg.run_title_id);
+        if (!cfg.lle_modules.empty()) {
+            auto modules = std::string();
```
VelocityRa: Bit of an unorthodox way to declare/init, just do this which has the same effect:
```cpp
std::string modules;
```

bsinky: Thanks!

PR #345, opened 2018-10-17
ci: Use docker for Linux build

pent0:
~~This use the provided Boost in external source.~~

~~There are some reasons I can't use prebuilt boost from default APT or another source:~~
~~- The Ubuntu version CI use was 14.4, and only Boost 1.54 is max, and the purpose of this originally was to update the Boost issue that was causing path error on some Linux users (discussed on Discord).~~
~~- Somehow prebuilt Boost (from Arch Linux), and from my computer built, don't work with CI, there are link errors, while building with the same GCC package from default APT (7.3.0)~~

~~So what I'm doing here is build the boost in source and use ./b2 install to install it. This hopefully will solve the issue with vpk path argument. Anyone that can successfully make the CI not throwing link errors with prebuilt Boost can pull a PR, and I'm happily to reverse or close the PR if this doesn't solve the path issue or if you can resolved the link issue.~~

**EDIT:** No longer relevant. This PR uses Docker image (Ubuntu 18.04), install Boost 1.65 and build inside the docker. ccache is guranteed.


VelocityRa:
This brings up Travis for Linux to 20 minutes :/

VelocityRa:
Ideally we'd cache it like Citra:
https://travis-ci.org/citra-emu/citra
https://github.com/citra-emu/citra/tree/master/.travis/linux-frozen

pent0:
Can't do a dependencies frozen build, if anyone can, feel free to open another PR after this.

PR #346, opened 2018-10-20
renderer/optimization: Multithreaded texture hashing

VelocityRa:
Should speed up texture hashing by pretty much the amount of cores in the system.

*Note:* "Hashing" is a pretty loose term, since we're still just summing.

### DIFF
--- a/src/emulator/renderer/src/texture_cache.cpp
+++ b/src/emulator/renderer/src/texture_cache.cpp
@@ -10,7 +10,10 @@
 
 #include <algorithm> // find
 #include <cstring> // memcmp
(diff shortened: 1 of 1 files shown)


VelocityRa:
I like how you're still thanking me for PRs :)

Looks like MSVC is the only compiler that supports this... Should I `ifdef` it or close the issue entirely?

petmac:
I think if it provides a real speed up then it's worth it, what do you think?

VelocityRa:
I haven't measured it in practice here (just threw this together in a few mins), but from what I've seen from benchmarks `par_unseq` is a clean xN speed-up.

PR #349, opened 2018-11-05
gui: Provide controls information to the user

TomasKralCZ:
Added an entry to MainMenuBar called "Help" with one entry "Controls", which simply displays the default key mappings.
vitacontrols


VelocityRa:
Looks good, thanks!

Issue #33, opened 2018-02-03
Support rumble


petmac:
Vita does not support rumble but PSTV does.

@Rinnegatamante says to check out `SceCtrlActuator`.

PR #363, opened 2019-01-01
Initial sceCtrlSetActuator implementation.

Rinnegatamante:
Added joystick rumbling support.
Tested with an XInput joystick with vitaQuake (to test it, you just have to wait till player gets damaged on demo playback).

Probably more testing is required with different kind of joysticks to see if the behaviour is correct.

Handles https://github.com/Vita3K/Vita3K/issues/33

### DIFF
--- a/src/emulator/modules/SceCtrl/SceCtrl.cpp
+++ b/src/emulator/modules/SceCtrl/SceCtrl.cpp
@@ -167,8 +167,8 @@ static void add_new_controllers(CtrlState &state) {
                 SDL_Haptic *haptic = SDL_HapticOpenFromJoystick(SDL_GameControllerGetJoystick(controller.get()));
                 SDL_HapticRumbleInit(haptic);
                 const HapticPtr handle(haptic, SDL_HapticClose);
-                state.controllers.insert(GameControllerList::value_type(guid, controller));
-                state.haptics.insert(HapticList::value_type(guid, handle));
+                state.controllers.emplace(guid, controller);
+                state.haptics.emplace(guid, handle);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


src/emulator/modules/SceCtrl/SceCtrl.cpp
```diff
@@ -198,6 +204,24 @@ static int peek_buffer_positive(HostState &host, SceCtrlData *&pad_data) {
     return 0;
 }
 
+static int rumble(HostState &host, const SceCtrlActuator *&pState) {
+    CtrlState &state = host.ctrl;
+    remove_disconnected_controllers(state);
+    add_new_controllers(state);
+
+    for (const HapticList::value_type &haptic : state.haptics) {
```
VelocityRa:
```suggestion
     for (const auto& haptic : state.haptics) {
```

src/emulator/modules/SceCtrl/SceCtrl.cpp
```diff
@@ -162,7 +164,11 @@ static void add_new_controllers(CtrlState &state) {
             const SDL_JoystickGUID guid = SDL_JoystickGetDeviceGUID(joystick_index);
             if (state.controllers.find(guid) == state.controllers.end()) {
                 const GameControllerPtr controller(SDL_GameControllerOpen(joystick_index), SDL_GameControllerClose);
+                SDL_Haptic *haptic = SDL_HapticOpenFromJoystick(SDL_GameControllerGetJoystick(controller.get()));
+                SDL_HapticRumbleInit(haptic);
+                const HapticPtr handle(haptic, SDL_HapticClose);
                 state.controllers.insert(GameControllerList::value_type(guid, controller));
+                state.haptics.insert(HapticList::value_type(guid, handle));
```
VelocityRa:
```suggestion
                 state.haptics.emplace(guid, handle);
```

VelocityRa: Change above code too, there's rarely a reason to use `value_type`.

PR #364, opened 2019-01-01
display/queue: Remove [m]lock.release() for all OS

### LINKED ISSUE #362: sceDisplayWaitVblankStart(); doesn't return in Linux
Homebrew that uses WaitVbalnkstart(); does not work correctly in Linux.
it won't return, therefore it won't show anything(appear as black screen while thread is running)

scribam:
Should fix #362 (with the changes at lines 112-114)

@VelocityRa Do you remember why the preprocessors were added in commit https://github.com/Vita3K/Vita3K/pull/146/commits/44ce9e2e98d819e64c70f024c698d75cee7adae0 (instead of deleting the lines)? Was it because you didn't want to change the behavior on other OS and/or you were only able to test on Windows?

### DIFF
--- a/src/emulator/modules/SceDisplay/SceDisplay.cpp
+++ b/src/emulator/modules/SceDisplay/SceDisplay.cpp
@@ -80,9 +80,6 @@ EXPORT(int, sceDisplayWaitSetFrameBuf) {
     if (!host.display.sync_rendering) {
         std::unique_lock<std::mutex> lock(host.display.mutex);
         host.display.condvar.wait(lock);
-#ifndef WIN32
-        lock.release();
-#endif
     } else
         host.display.condvar.notify_all();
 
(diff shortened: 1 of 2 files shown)


VelocityRa:
Actually it's because we only tested on Windows and macOS (iirc).
So, keep the releases but use `#ifdef __APPLE__` instead.

VelocityRa:
No, the current check is "not Windows".

VelocityRa:
I still thing it's needed since it's _there_, but if want to test, sure.

petmac:
I tested on Mac, and it looks like those `release()` calls introduced a soft-lock anyway! Removing them seems to fix Downwell getting stuck, as well as VitaTester.

So let's please remove the releases on all platforms.

VelocityRa:
🤔 alrighty then.

PR #365, opened 2019-01-01
Fix Downwell regression

### LINKED ISSUE #343: Downwell regression
Downwell crashes on the logo screen after #323 and breaks shaders with #331.

i'm currently trying to find the specific commit that regressed it (~~if i can get it to build~~)

VelocityRa:
- Fixes regression in Downwell and other titles
- Resolves #343
- Update shaders-db to latest master

### DIFF
--- a/src/emulator/kernel/src/load_self.cpp
+++ b/src/emulator/kernel/src/load_self.cpp
@@ -80,7 +80,14 @@ static bool load_var_imports(const uint32_t *nids, const Ptr<uint32_t> *entries,
         const ExportNids::iterator export_address_it = kernel.export_nids.find(nid);
         if (export_address_it != kernel.export_nids.end()) {
             export_address = export_address_it->second;
+
+            if (reloc_entries_count > 0)
+                // 8 is sizeof(EntryFormat1Alt)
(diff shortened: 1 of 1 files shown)


src/emulator/kernel/src/load_self.cpp
```diff
@@ -80,7 +80,14 @@ static bool load_var_imports(const uint32_t *nids, const Ptr<uint32_t> *entries,
         const ExportNids::iterator export_address_it = kernel.export_nids.find(nid);
         if (export_address_it != kernel.export_nids.end()) {
             export_address = export_address_it->second;
+
+            if (reloc_entries_count > 0)
+                // 8 is sizeof(EntryFormat1Alt)
+                if (!relocate(var_reloc_entries, reloc_entries_count * 8, segments, mem, true, export_address))
```
Rinnegatamante: Wouldn't it be better to merge the two ifs into a single one?

VelocityRa: Control flow would be different.
We always want `export_address` to be assigned if first condition succeeds.
Also if we did that it would execute the `else` below when `reloc_entries_count > 0` which would be wrong.

PR #366, opened 2019-01-01
Properly implementing multiple controllers support.

Rinnegatamante:
Before this PR, controllers where not paired to a sceCtrl port resulting in weird behaviours when multiple joysticks where plugged in (or if an application using multiple controllers were used).

Now Vita3K will bind only max 4 controllers and pair them to a sceCtrl port so that a single controller will give input for a single port like it works on retail PSTV.

### DIFF
--- a/src/emulator/modules/SceCtrl/SceCtrl.cpp
+++ b/src/emulator/modules/SceCtrl/SceCtrl.cpp
@@ -146,13 +146,26 @@ static uint8_t float_to_byte(float f) {
     return static_cast<uint8_t>(clamped * 255);
 }
 
+static int reserve_port(CtrlState &state) {
+    int res = -1;
+    for (int i = 0; i < 4; i++) {
+        if (state.free_ports[i]) {
+            state.free_ports[i] = false;
+            res = i + 1;
+            break;
+        }
+    }
+    return res;
+}
+
 static void remove_disconnected_controllers(CtrlState &state) {
-    for (GameControllerList::iterator controller = state.controllers.begin(); controller != state.controllers.end();) {
-        if (SDL_GameControllerGetAttached(controller->second.get())) {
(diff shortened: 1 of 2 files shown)


src/emulator/modules/SceCtrl/SceCtrl.cpp
```diff
@@ -146,35 +146,58 @@ static uint8_t float_to_byte(float f) {
     return static_cast<uint8_t>(clamped * 255);
 }
 
+static int reserve_port(CtrlState &state) {
+    int res = -1;
+    for (int i = 0; i < 4; i++) {
+        if (state.free_ports[i]) {
+            state.free_ports[i] = false;
+            res = i + 1;
```
VelocityRa: Why not just `return i + 1;`?

src/emulator/modules/SceCtrl/SceCtrl.cpp
```diff
@@ -146,35 +146,58 @@ static uint8_t float_to_byte(float f) {
...
+            break;
+        }
+    }
+    return res;
+}
+
 static void remove_disconnected_controllers(CtrlState &state) {
-    for (GameControllerList::iterator controller = state.controllers.begin(); controller != state.controllers.end();) {
-        if (SDL_GameControllerGetAttached(controller->second.get())) {
+    for (ControllerList::iterator controller = state.controllers.begin(); controller != state.controllers.end();) {
```
VelocityRa:
```suggestion
    for (auto controller : controllers) {
```
and change usage accordingly if needed

src/emulator/modules/SceCtrl/SceCtrl.cpp
```diff
@@ -146,35 +146,58 @@ static uint8_t float_to_byte(float f) {
...
+        if (SDL_GameControllerGetAttached(controller->second.controller.get())) {
             ++controller;
```
VelocityRa: Remove this and invert condition.

PR #368, opened 2019-01-05
Implement sceCtrlIsMultiControllerSupported.

Rinnegatamante:


VelocityRa:
Change return type to `bool` and return `true`.

devnoname120:
@VelocityRa I wouldn't do that unless you have looked at the assembly of this function. Some getters can return `<0` on error.

PR #369, opened 2019-01-07
Fix buid for yaml-cpp on cmake

Zangetsu38:
Fix build issue on yaml-cpp after re generate cmake build, caused by issue on gslang with it change variable "CMAKE_INSTALL_PREFIX" with set force "install", causing broken yaml-cpp.
More info on gslang issue here https://github.com/KhronosGroup/glslang/issues/1015

For more understund how it works now with my fix, on first generate gslang set "install" value on "CMAKE_INSTALL_PREFIX" and after my code look if value is by default, and if not re set it with default value forced for can works after all time.

Separate .lib and .exp on lib folder bin folder show only .exe and bin file.

Update yaml-cpp submodule on last master

### DIFF
--- a/src/external/CMakeLists.txt
+++ b/src/external/CMakeLists.txt
@@ -11,6 +11,16 @@ function(check_submodules_present)
 endfunction()
 check_submodules_present()
 
+# Fix a glslang hack
+# See https://github.com/KhronosGroup/glslang/issues/1015 and https://github.com/Vita3K/Vita3K/pull/369 for details
+if(NOT CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+	if(WIN32)
+		set(CMAKE_INSTALL_PREFIX "C:/Program Files/${PROJECT_NAME}" CACHE STRING "Default install prefix is: C:/Program Files" FORCE)
+	elseif(UNIX)
+		set(CMAKE_INSTALL_PREFIX /usr/local CACHE STRING "Default install prefix is: /usr/local" FORCE)
+	endif()
+endif()
+
 option(CAPSTONE_BUILD_SHARED "Build shared library" OFF)
 option(CAPSTONE_BUILD_TESTS "Build tests" OFF)
 option(CAPSTONE_ARM_SUPPORT "ARM support" ON)
@@ -105,6 +115,7 @@ target_include_directories(vita-toolchain INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}
 option(YAML_CPP_BUILD_TESTS "Enable testing" OFF)
 option(YAML_CPP_BUILD_TOOLS "Enable parse tools" OFF)
 option(YAML_CPP_BUILD_CONTRIB "Enable contrib stuff in library" OFF)
+option(YAML_CPP_INSTALL "Enable generation of install target" OFF)
 add_subdirectory(yaml-cpp EXCLUDE_FROM_ALL)
 add_library(yaml INTERFACE)
 target_include_directories(yaml INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}/yaml-cpp/include")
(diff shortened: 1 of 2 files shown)


.gitmodules
```diff
@@ -10,6 +10,7 @@
 [submodule "src/external/glbinding"]
 	path = src/external/glbinding
 	url = https://github.com/cginternals/glbinding.git
+	branch = v3.0
```
pent0: Is this change accident? You haven't mentioned this change in the description

Zangetsu38: I have mentionned it on description, and no, it juste add what branch master already used

pent0: I'm not sure if moving away from master branch is a great idea. May need others to comment on this

Zangetsu38: What ? I move nothing, i juste add branch info mission

pent0: Im sure that those are not default branch ;/

Zangetsu38: No, i have juste added branch corresponding what submodule used

Zangetsu38:
Rebased, and removed part update all submodule for another pr

src/external/CMakeLists.txt
```diff
@@ -105,6 +113,7 @@ target_include_directories(vita-toolchain INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}
 option(YAML_CPP_BUILD_TESTS "Enable testing" OFF)
 option(YAML_CPP_BUILD_TOOLS "Enable parse tools" OFF)
 option(YAML_CPP_BUILD_CONTRIB "Enable contrib stuff in library" OFF)
+option(YAML_CPP_INSTALL "Enable generation of install target" ON)
```
VelocityRa: Why `ON`? We don't need installation.
The only thing that this PR should have is this line with `OFF` and the yaml-cpp submodule update.

VelocityRa: No, it's a static library, we do not need installing.

Zangetsu38: oki i understund

Zangetsu38:
Done changed

VelocityRa:
As I said, the other changes are not needed now. Just that line with OFF.
And make sure you test that the yaml error is indeed gone after re-generating the build files.

Zangetsu38:
No, you have not understund, this eror no relate Yaml, is caused by gslang, and set line is obligatory, i can't change more here, just broken build, is for that, set on or off for it change nothing

Zangetsu38:
@VelocityRa I have improved description information for that
and is for that, use "ON" or "OFF" is same, this error no relate that, so i have just set default value like it used actually on master.

VelocityRa:
Ok but it's hacky. Why not just wait until they fix it themselves? It's not a Vita3K issue.

Zangetsu38:
because i think nver get fixed ? xD, is gslang, i speek with it already for fix that, but i am not think, and for information, is not hack, gslang use same for force "'install" value, is just fix, with waiting it want really fix it, and this issue date to 2017 so eu lol

src/external/CMakeLists.txt
```diff
@@ -11,6 +11,14 @@ function(check_submodules_present)
 endfunction()
 check_submodules_present()
 
+if(NOT CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
```
VelocityRa: Put a comment above this line:
```cmake
# Fix a glslang hack
# See https://github.com/KhronosGroup/glslang/issues/1015 and https://github.com/Vita3K/Vita3K/pull/369 for details
```

Zangetsu38: Done, i am just not sur for option set on OFF, but if that is not needed is good, because actually master using this install ^^

VelocityRa:
Thanks for spending so long on this.

PR #370, opened 2019-01-07
Implement sceKernelIsPSVitaTV

caghandemir:
(#368) returns true, so this should too.
[Here's ](https://github.com/caghandemir/vitaControllerTest) a test program if you want to test this behavior on real hardware.


VelocityRa:
There was never a decision that we should be strictly emulating PS TV and not a PSVita.

So, I feel like this should be a config option (that would ideally affect `sceCtrlIsMultiControllerSupported` and any other relevant function).

Rinnegatamante:
I can work on the switch as bentokun suggested on Discord, maybe it's worth to merge as is for now and deal with both functions in a separated PR dealing with mode switching.

PR #378, opened 2019-01-12
Allocation viewer crash fix

1whatleytay:
This merge prevents a crash when opening the memory allocation viewer. The code that draws the viewer relies on `memory_editor` being null when it should not be drawn, but it was not explicitly set to null.
unknown
This fix ensures that the memory editor is only drawn when it has been opened which prevents the crash.

Thanks to waterflame for pointing this out.

### DIFF
--- a/src/emulator/gui/include/gui/state.h
+++ b/src/emulator/gui/include/gui/state.h
@@ -58,7 +58,7 @@ struct GuiState {
 
     DialogState common_dialog;
     GamesSelector game_selector;
-    MemoryEditor *memory_editor;
+    MemoryEditor *memory_editor = nullptr;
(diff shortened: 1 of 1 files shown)


src/emulator/gui/include/gui/state.h
```diff
@@ -58,7 +58,7 @@ struct GuiState {
 
     DialogState common_dialog;
     GamesSelector game_selector;
-    MemoryEditor *memory_editor;
+    MemoryEditor *memory_editor = nullptr;
```
VelocityRa: Next time use braced initialization please, as we do everywhere else. Or better yet, use smart pointers.
```cpp
    MemoryEditor *memory_editor{};
```

PR #381, opened 2019-01-17
Improvement to sceCtrl.

Rinnegatamante:
- Implemented sceCtrlPeekBufferNegative, sceCtrlReadBufferNegative
- Implemented sceCtrlPeekBufferPositiveExt2, sceCtrlReadBufferPositiveExt2
- Fixed returned value of sceCtrlPeekBufferPositive and sceCtrlReadBufferPositive

### DIFF
--- a/src/emulator/modules/SceCtrl/SceCtrl.cpp
+++ b/src/emulator/modules/SceCtrl/SceCtrl.cpp
@@ -60,6 +60,22 @@ static const KeyBinding key_bindings[] = {
     { SDL_SCANCODE_H, SCE_CTRL_PSBUTTON },
 };
 
+static const KeyBinding key_bindings_ext[] = {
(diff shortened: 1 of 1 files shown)


src/emulator/modules/SceCtrl/SceCtrl.cpp
```diff
@@ -227,7 +259,11 @@ static int peek_buffer_positive(HostState &host, int port, SceCtrlData *&pad_dat
         pad_data->ry = float_to_byte(axes[3]);
     }
 
-    return 0;
+    if (negative) {
+        pad_data->buttons = 0xFFFFFFFF - pad_data->buttons;
```
VelocityRa: no need to fix, just fyi you could do this here (less error prone, works for all sizes)
```suggestion
        pad_data->buttons = ~pad_data->buttons;
```

PR #382, opened 2019-01-18
Update ui_controls_dialog.cpp

wfscans:
Correct typo for R button on the control menu. It's "E" not "R".


pent0:
[Legitimate](https://github.com/Vita3K/Vita3K/blob/master/src/emulator/modules/SceCtrl/SceCtrl.cpp#L55)

PR #384, opened 2019-01-18
Debug hotfix

scribam:
With imgui v1.67, it is now illegal to call `Begin("")` with an empty string. See "Breaking Changes" in https://github.com/ocornut/imgui/releases/tag/v1.67
The debug build asserts without the commit "gui: Add window name for game selector":
```
Vita3K: /home/scribam/CLionProjects/Vita3K/src/external/imgui/imgui.cpp:4746: bool ImGui::Begin(const char*, bool*, ImGuiWindowFlags): Assertion `name != __null && name[0] != '\0'' failed.
```


VelocityRa:
Thanks!

src/external/CMakeLists.txt
```diff
@@ -23,14 +23,17 @@ endif()
 
 option(CAPSTONE_BUILD_SHARED "Build shared library" OFF)
 option(CAPSTONE_BUILD_TESTS "Build tests" OFF)
-option(CAPSTONE_ARM_SUPPORT "ARM support" ON)
```
Zangetsu38: With remove this line if you want on it ?

scribam: Because the default value is ON

scribam: See https://github.com/aquynh/capstone/blob/master/CMakeLists.txt#L40

VelocityRa: because it's on by default
I agree thought, shouldn't have been removed probably

Zangetsu38: yes i know, just me i leave it here ^^

PR #396, opened 2019-02-02
host: Add config file support and custom VFS directory

pent0:
- Add a config system. Config file is "config.yml", in base directory (most of the time where you run the executable)
- Add ability to specify the virtual filesystem directory. To edit this, follow the instruction:
    * exampl01
   * Edit *pref-path* to where you want your virtual filesystem should reside.
   * After installing and running with the *pref-path* variables given, it should look like this: 
   * example1
   
**Note:** This won't move the old game data to the new directory for you. Do it yourself

### DIFF
--- a/src/emulator/host/include/host/config.h
+++ b/src/emulator/host/include/host/config.h
@@ -39,14 +39,30 @@ struct Config {
     bool log_uniforms = false;
     std::vector<std::string> lle_modules;
     bool pstv_mode = false;
+    optional<std::string> pref_path;
 };
 
 namespace config {
 
+/*
+ * \brief Initializes config system from a YML file.
+ * 
+ * \param cfg Config options are returned via this parameter.
+ * \return True on loading success.
+*/
+ExitCode init_from_config_file(Config &cfg, const char *config_file_name);
+
+/*
+ * \brief Save emulator config to a YML file.
+ *
+ * \return True on saving success.
+*/
+ExitCode save_config_to_file(Config &cfg, const char *config_file_name);
+
 /**
   * \brief Initializes config system, parsing command-line args and handling some basic ones:
   *        --help, --version, --log-level
-  * \param cfg Config options are returend via this parameter.
+  * \param cfg Config options are returned via this parameter.
   * \return True on success, false on error.
   */
 ExitCode init(Config &cfg, int argc, char **argv);
--- a/src/emulator/host/src/config.cpp
+++ b/src/emulator/host/src/config.cpp
@@ -27,14 +27,126 @@
 #include <boost/program_options.hpp>
 #include <spdlog/spdlog.h>
 
+#include <yaml-cpp/yaml.h>
+
 #include <exception>
+#include <fstream>
 #include <iostream>
 
 namespace po = boost::program_options;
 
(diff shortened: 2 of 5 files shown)


src/emulator/host/include/host/config.h
```diff
@@ -39,14 +39,30 @@ struct Config {
...
 
 namespace config {
 
+/*
+ * \brief Initializes config system from a YML file.
+ * 
+ * \param cfg Config options are returned via this parameter.
+ * \return True on loading success.
+*/
+ExitCode init_from_config_file(Config &cfg, const char *config_file_name);
```
VelocityRa: rename to `config::deserialize` maybe

src/emulator/host/include/host/config.h
```diff
@@ -39,14 +39,30 @@ struct Config {
...
+
+/*
+ * \brief Save emulator config to a YML file.
+ *
+ * \return True on saving success.
+*/
+ExitCode save_config_to_file(Config &cfg, const char *config_file_name);
```
VelocityRa: rename to `config::serialize` maybe

src/emulator/host/src/config.cpp
```diff
@@ -27,14 +27,126 @@
 #include <boost/program_options.hpp>
 #include <spdlog/spdlog.h>
 
+#include <yaml-cpp/yaml.h>
```
VelocityRa: put this one line above

src/emulator/host/src/config.cpp
```diff
@@ -27,14 +27,126 @@
...
+
+template <typename T>
+void config_file_emit_single(YAML::Emitter &emitter, const char *name, T &val) {
+    emitter << YAML::Key << name << YAML::Value << val;
+}
+
+template <typename T>
+void config_file_emit_optional_single(YAML::Emitter &emitter, const char *name, boost::optional<T> &val) {
+    if (val) {
+        emitter << YAML::Key << name << YAML::Value << val.value();
```
VelocityRa: could re-use `config_file_emit_single` here, if you want

src/emulator/host/include/host/config.h
VelocityRa: Also I'm not sure how useful the `config_file_name` parameter is, maybe remove?

src/emulator/host/include/host/config.h
VelocityRa: ditto about `config_file_name`

src/emulator/host/include/host/config.h
pent0: Yeah, it's better

src/emulator/host/src/config.cpp
pent0: YAML can't serialize boost::optional though

VelocityRa: I thought you could call with `val.value()`.

pent0: Ah, I taken your idea wrong. it's good

Fallenleader:
Thanks guys! Much needed for testers primarily, as the library of games can get thicc.

PR #399, opened 2019-02-03
travis: Logging git push execution

Rinnegatamante:
Travis buildscript for Linux keep failing deploying but we don't have enough info about why this happens.


pent0:
Let's see what went wrong

PR #401, opened 2019-02-03
ui: disassembly + thread viewer

1whatleytay:
This PR adds and fixes a few things:

- Adds a new Disassembly Viewer. It can be accessed from the Debug menu. You can also jump to a segment of memory directly in the Allocation viewer.

diasm_viewer

- Adds a new individual thread viewer. It has information on the thread (its name and id), register info (pc, lr, sp, r0 - r15) and the instruction it is currently executing.

thread_viewer

- I disciplined the memory editor and told it to stop misbehaving. It will now open in the correct place 100% of the time and will never refuse to open. It has been informed that I don't like it very much.

All reviews are welcome! I made some changes to the cpu and disasm modules to make these features possible, but they might not be best practice. If something doesn't look right I'll be happy to fix it.

Tested on macOS and Windows.


src/emulator/cpu/src/cpu.cpp
```diff
@@ -222,7 +218,6 @@ void stop(CPUState &state) {
 
 uint32_t read_reg(CPUState &state, size_t index) {
     assert(index >= 0);
```
VelocityRa: `index` is `size_t` (unsigned) so this check is useless too

src/emulator/gui/include/gui/state.h
```diff
@@ -63,8 +61,15 @@ struct GuiState {
 
     DialogState common_dialog;
     GamesSelector game_selector;
-    std::unique_ptr<MemoryEditor> memory_editor;
-    size_t memory_editor_start, memory_editor_count;
+
+    MemoryEditor memory_editor;
+    size_t memory_editor_start = 0, memory_editor_count = 0;
```
VelocityRa: one statement per line please

src/emulator/gui/include/gui/state.h
```diff
@@ -63,8 +61,15 @@ struct GuiState {
...
+
+    std::string disassembly_arch = "THUMB";
+    char disassembly_address[9] = "00000000", disassembly_count[5] = "100";
```
VelocityRa: ditto

src/emulator/gui/src/ui_allocations_dialog.cpp
```diff
@@ -25,42 +26,50 @@ void DrawAllocationsDialog(HostState &host) {
             continue;
 
         if (ImGui::TreeNode(fmt::format("{}: {}", generation_num, generation_name).c_str())) {
-            int32_t index = -1;
-            uint32_t count = 1;
+            int32_t index = -1, count = 1;
```
VelocityRa: ditto

src/emulator/disasm/src/disasm.cpp
```diff
@@ -46,15 +46,19 @@ bool init(DisasmState &state) {
     return true;
 }
 
-std::string disassemble(DisasmState &state, const uint8_t *code, size_t size, uint64_t address, bool thumb) {
+std::string disassemble(DisasmState &state, const uint8_t *code, size_t size, uint64_t address, bool thumb, uint16_t *insnSize) {
```
VelocityRa: we use macro_case, `insn_size`

1whatleytay:
I did a brief test on windows and it looks like it works fine (no sudden crashes or memory leaks).

src/emulator/gui/src/ui_disassembly_dialog.cpp
```diff
@@ -0,0 +1,112 @@
...
+        host.gui.disassembly.emplace_back("Nothing to disassemble.");
+        return;
+    }
+
+    uint16_t size = 1;
+    uint32_t addr = from;
+
+    for (int a = 0; a < count && size != 0; a++) {
+        // Apparently some THUMB instructions are 4 bytes long, so check all 4 just to be safe.
+        size_t addr_page = addr / KB(4), end_page = (addr + 4) / KB(4);
```
VelocityRa: make separate statements

PR #404, opened 2019-02-04
Shader instruction decoder & recompiler foundation

VelocityRa:
**Edit**: We have and will continue adding more to this before merging, it now implements more things than what is mentioned below.

---

https://github.com/Vita3K/Vita3K/pull/297 introduced a shader recompiler skeleton, this PR provides the foundation for decoding and recompiling actual USSE shader instructions.

It partially implements the `vmov` instruction, but that is not enough to fully recompile basic shaders. Thankfully, now that we have the foundation, implementing instructions should not be too difficult, so I would expect basic shaders to be working soon enough.

I wouldn't open this PR so soon, but there's been some dev interest in helping with this, so I do it in hopes of making that easier and potentially attracting even more help.

Given the latter, I would not suggest testers put any effort in re-testing games _yet_.

Main commit is: https://github.com/Vita3K/Vita3K/commit/4e70b922e8d7c2120e103f4350855008d9865ccd. Could've been split to lots more commits probably, but the hassle ain't worth it.

---

Thanks to @MerryMage for her fancy instruction decoder infrastructure which I used, it makes decoding instructions _that_ much less tedious :)


VelocityRa:
This work is now at the point where all 4 of **Undertale**'s shaders ([PCSE01116.zip](https://github.com/Vita3K/Vita3K/files/2923828/PCSE01116.zip)) are successfully recompiled!

May as well merge.
Other GameMaker games still use a more complicated shader we can't handle right now, obviously expect more work on this. :)

---
To @Vita3K/testers:

**Testing for new games:** I haven't really tested other commercial games with this, _maybe_ we'll start seeing graphics in some titles now, so you can start testing again for graphical output.
There's still plenty of "basic" things we don't handle, so take this assertion with a grain of salt.
Preferably delete your `shaderlog/` folder before starting to test this.

**Testing for regressions:** If you want to test recompilation of shaders in our datatabase, rename or delete the `shaders/` directory and start testing titles that used to work (homebrews, some GameMaker titles). Report on the ones that _do_ work, since we already know that most don't.
Given the above, this is less important at this point. I may ask for more regression testing once we successfully recompile all GameMaker shaders.

PR #405, opened 2019-02-05
util/config: config option to duplicate log with title id/game name as log name

### LINKED ISSUE #402: QOL: Unimportant - Have option for Vita3K to make a copy of the log with the game's title ID for easier compatibility reporting/debugging
Not important, but it would make tracking games easier if V3K either has an option to make a new log file or make a duplicate with the title ID of the game on launch rather than overwriting the existing one.
If someone get's a chance, this would be highly appreciated since I have a large number of games I test.

IllusionMan1212:
This PR fixes #402 

- it creates a copy of the log file with the game's name and title id as the file's name.
- the second log file sink is only added after the game has launched which omits the first few config system lines from logging
- it outputs to a new directory called `logs`
- added a config option to enable this from the command line using -A or --archive-log

### DIFF
--- a/src/emulator/host/include/host/config.h
+++ b/src/emulator/host/include/host/config.h
@@ -40,6 +40,7 @@ struct Config {
     std::vector<std::string> lle_modules;
     bool pstv_mode = false;
     optional<std::string> pref_path;
+    bool duplicate_log = false;
 };
 
 namespace config {
--- a/src/emulator/host/include/host/config.h
+++ b/src/emulator/host/include/host/config.h
@@ -40,7 +40,7 @@ struct Config {
     std::vector<std::string> lle_modules;
     bool pstv_mode = false;
     optional<std::string> pref_path;
-    bool duplicate_log = false;
+    bool archive_log = false;
 };
 
 namespace config {
--- a/src/emulator/host/src/app.cpp
+++ b/src/emulator/host/src/app.cpp
@@ -213,6 +213,13 @@ bool load_app_impl(Ptr<const void> &entry_point, HostState &host, const std::wst
     find_data(host.game_title, host.sfo_handle, "TITLE");
     std::replace(host.game_title.begin(), host.game_title.end(), '\n', ' ');
     find_data(host.io.title_id, host.sfo_handle, "TITLE_ID");
+
+    if (host.cfg.duplicate_log) {
+        if (!fs::exists(host.base_path + "/logs"))
+            fs::create_directory(std::string(host.base_path) + "/logs");
+        logging::add_sink(host.io.title_id, host.game_title);
+    }
+
     std::string category;
     find_data(category, host.sfo_handle, "CATEGORY");
 
--- a/src/emulator/host/src/app.cpp
+++ b/src/emulator/host/src/app.cpp
@@ -232,7 +232,7 @@ bool load_app_impl(Ptr<const void> &entry_point, HostState &host, const std::wst
                 std::cerr << "failed to get working directory" << std::endl;
             }
 #else
-            logging::add_sink(log_name);
+            logging::add_sink(string_utils::utf_to_wide(log_name));
 #endif
         } catch (const spdlog::spdlog_ex &ex) {
             std::cerr << "File log initialization failed: " << ex.what() << std::endl;
--- a/src/emulator/host/src/app.cpp
+++ b/src/emulator/host/src/app.cpp
@@ -214,10 +214,29 @@ bool load_app_impl(Ptr<const void> &entry_point, HostState &host, const std::wst
     std::replace(host.game_title.begin(), host.game_title.end(), '\n', ' ');
     find_data(host.io.title_id, host.sfo_handle, "TITLE_ID");
 
-    if (host.cfg.duplicate_log) {
-        if (!fs::exists(host.base_path + "/logs"))
-            fs::create_directory(std::string(host.base_path) + "/logs");
-        logging::add_sink(host.io.title_id, host.game_title);
+    if (host.cfg.archive_log) {
+        fs::create_directory(std::string(host.base_path) + "/logs");
+        try {
+            std::string game_title = string_utils::remove_special_chars(host.game_title);
+            const auto log_name = fmt::format("{} - [{}].log", game_title, host.io.title_id);
+#ifdef WIN32
+            wchar_t buffer[MAX_PATH];
+            GetModuleFileNameW(NULL, buffer, MAX_PATH);
+            std::string::size_type pos = std::wstring(buffer).find_last_of(L"\\\\");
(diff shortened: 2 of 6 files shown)


src/emulator/util/src/util.cpp
```diff
@@ -110,6 +143,25 @@ std::string wide_to_utf(const std::wstring &str) {
...
+        case '/':
+        case '\\':
+        case ':':
+        case '?':
+        case '"':
+        case '<':
+        case '>':
+        case '|':
+        case '*':
+            c = ' ';
```
pent0: Put a brake and default case in (to get rid of compiler warnings)

src/emulator/host/include/host/config.h
```diff
@@ -40,6 +40,7 @@ struct Config {
     std::vector<std::string> lle_modules;
     bool pstv_mode = false;
     optional<std::string> pref_path;
+    bool duplicate_log = false;
```
VelocityRa: `archive_log`

src/emulator/host/src/app.cpp
```diff
@@ -213,6 +213,13 @@ bool load_app_impl(Ptr<const void> &entry_point, HostState &host, const std::wst
     find_data(host.game_title, host.sfo_handle, "TITLE");
     std::replace(host.game_title.begin(), host.game_title.end(), '\n', ' ');
     find_data(host.io.title_id, host.sfo_handle, "TITLE_ID");
+
+    if (host.cfg.duplicate_log) {
+        if (!fs::exists(host.base_path + "/logs"))
```
VelocityRa: remove this line, not needed

src/emulator/host/src/config.cpp
```diff
@@ -95,6 +95,7 @@ ExitCode serialize(Config &cfg) {
     config_file_emit_single(emitter, "log-active-shaders", cfg.log_active_shaders);
     config_file_emit_single(emitter, "log-uniforms", cfg.log_uniforms);
     config_file_emit_single(emitter, "pstv-mode", cfg.pstv_mode);
+    config_file_emit_single(emitter, "duplicate-log", cfg.duplicate_log);
```
VelocityRa: ditto

src/emulator/host/src/config.cpp
```diff
@@ -125,15 +126,16 @@ ExitCode deserialize(Config &cfg) {
     get_yaml_value(config_node, "log-active-shaders", &cfg.log_active_shaders, false);
     get_yaml_value(config_node, "log-uniforms", &cfg.log_uniforms, false);
     get_yaml_value(config_node, "pstv-mode", &cfg.pstv_mode, false);
-    
+    get_yaml_value(config_node, "duplicate-log", &cfg.duplicate_log, false);
```
VelocityRa: ditto, also delete added space below

src/emulator/host/src/config.cpp
```diff
@@ -166,7 +168,8 @@ ExitCode init(Config &cfg, int argc, char **argv) {
             ("log-imports,I", po::bool_switch(&cfg.log_imports), "Log Imports")
             ("log-exports,E", po::bool_switch(&cfg.log_exports), "Log Exports")
             ("log-active-shaders,S", po::bool_switch(&cfg.log_active_shaders), "Log Active Shaders")
-            ("log-uniforms,U", po::bool_switch(&cfg.log_uniforms), "Log Uniforms");
+            ("log-uniforms,U", po::bool_switch(&cfg.log_uniforms), "Log Uniforms")
+            ("duplicate-log,D", po::bool_switch(&cfg.duplicate_log), "Makes a duplicate of the log file with TITLE_ID and Game ID as title");
```
VelocityRa: ditto

src/emulator/util/src/util.cpp
```diff
@@ -79,6 +79,39 @@ void set_level(spdlog::level::level_enum log_level) {
...
+#else
+        sinks.push_back(std::make_shared<spdlog::sinks::basic_file_sink_mt>(log_name, true));
+#endif
+    } catch (const spdlog::spdlog_ex &ex) {
+        std::cerr << "File log initialization failed: " << ex.what() << std::endl;
+    }
+
+    spdlog::set_default_logger(std::make_shared<spdlog::logger>("vita3k logger", begin(sinks), end(sinks)));
+
+    spdlog::set_error_handler([](const std::string &msg) {
```
VelocityRa: I don't think anything below this is needed, it was done in `init`

src/emulator/util/src/util.cpp
```diff
@@ -110,6 +143,25 @@ std::string wide_to_utf(const std::wstring &str) {
     return myconv.to_bytes(str);
 }
 
+std::string remove_special_chars(std::string str) {
```
VelocityRa: are you sure this is all of them? normally you use a whitelist for sanitizing, not a blacklist

src/emulator/util/src/util.cpp
```diff
@@ -79,6 +79,13 @@ void set_level(spdlog::level::level_enum log_level) {
     spdlog::set_level(log_level);
 }
 
+void add_sink(std::wstring log_path) {
+    sinks.push_back(std::make_shared<spdlog::sinks::basic_file_sink_mt>(log_path, true));
+
+    spdlog::set_default_logger(std::make_shared<spdlog::logger>("vita3k logger", begin(sinks), end(sinks)));
+    spdlog::set_pattern("%^[%H:%M:%S.%e] |%L| [%!]: %v%$");
```
VelocityRa: is this needed?

IllusionMan1212: Yes, both lines are needed, if the first is removed the log gets created but nothing is written to it, and if the second is removed the log will be unformatted

VelocityRa: ok, then making another function `logging::set_pattern` that just does this, and use it in init too, or make the pattern string a `static const char* LOG_PATTERN` and use it in init too.

IllusionMan1212:
Updated PR

src/emulator/util/src/util.cpp
```diff
@@ -71,14 +72,21 @@ void init() {
...
+    spdlog::set_pattern(LOG_PATTERN);
     spdlog::flush_on(spdlog::level::debug);
 }
 
 void set_level(spdlog::level::level_enum log_level) {
     spdlog::set_level(log_level);
 }
 
+void add_sink(std::wstring log_path) {
+    sinks.push_back(std::make_shared<spdlog::sinks::basic_file_sink_mt>(log_path, true));
```
VelocityRa: macOS build fails because `basic_file_sink_mt` accepts `std::string` and not `std::wstring`
do an `#ifdef _WIN32` with the above line there, and `utf16_to_utf8(log_path)` instead of `log_path` for the `#else`
need to find a neater way to deal with this stuff...

PR #412, opened 2019-02-26
cmake: Add some Info.plist entries.

MaddTheSane:
This adds a `CFBundleIdentifier` and a `CFBundleName`.


VelocityRa:
Could you prepend the commit message with:
`cmake:`
?

MaddTheSane:
Done

VelocityRa:
That's the Issue title :)
I meant the commit's name, it matters more because it will remain in our git history.
One way to do it is via `git commit --amend`, change it in the text editor (vim by default) and then `git push -f`.

MaddTheSane:
Ah, will do.

VelocityRa:
Got no macOS experience, so I'd rather not merge it immediately, let's wait for a review from @petmac.

PR #426, opened 2019-03-12
io: bugfixes and refactoring.

Rinnegatamante:


src/emulator/io/include/io/functions.h
```diff
@@ -31,7 +31,7 @@ struct SceIoDirent;
 void init_device_paths(IOState &io);
 
 bool init(IOState &io, const char *pref_path, const char *base_path);
-SceUID open_file(IOState &io, const std::string &path_, int flags, const char *pref_path, const char *export_name);
+SceUID open_file(IOState &io, std::string &path_, int flags, const char *pref_path, const char *export_name);
```
frangarcj: That const should not be dropped

src/emulator/io/src/io.cpp
```diff
@@ -254,7 +254,8 @@ void translate_path(std::string &path, VitaIoDevice &device, const IOState::Devi
     }
 }
 
-SceUID open_file(IOState &io, const std::string &path, int flags, const char *pref_path, const char *export_name) {
+SceUID open_file(IOState &io, std::string &path, int flags, const char *pref_path, const char *export_name) {
+    trim_leading_slash(path);
```
frangarcj: Trim is done over translated_path

src/emulator/io/src/io.cpp
```diff
@@ -658,7 +660,7 @@ int open_dir(IOState &io, const char *path, const char *pref_path, const char *e
 
     Directory dir_node;
     dir_node.dir_handle = dir;
-    dir_node.path = path;
+    dir_node.path = translated_path.c_str();
```
frangarcj: This is only used for stat I think. It is over path or translated_path?

Rinnegatamante:
Ready for review:
- Added trimming of "." too.
- Properly making trimming slash function to work (before it did nothing due to the function not taking a string reference but a string object as param.
- Properly statting file in read_dir (Before it used to fill only an attribute in the stat struct.
- Fixed a bug in stat_file for Windows that made so that directory received both IFREG and IFDIR attributes.

This fixes EDuke32 IO issues making it fully playable on the emulator.

https://cdn.discordapp.com/attachments/408917186254733317/555114416421470291/unknown.png

PR #427, opened 2019-03-13
Gui: add app_ver on game list

Zangetsu38:
Add app_ver on game list and inside log.
Like you see here:
I have set version before title game, same tittle id, because game name longer is random.
I have think add it like rpcs3 for can know what version game is used, i have see few time, diferent version can get different boot statut
image
image

Update few submodule.
gslang to last tag ->7.11.3113
imgui to last tag ->1.69
last master shader_db

### DIFF
--- a/src/emulator/host/src/app.cpp
+++ b/src/emulator/host/src/app.cpp
@@ -214,6 +214,7 @@ bool load_app_impl(Ptr<const void> &entry_point, HostState &host, const std::wst
 
     load_sfo(host.sfo_handle, params);
 
+    find_data(host.game_version, host.sfo_handle, "APP_VER");
     find_data(host.game_title, host.sfo_handle, "TITLE");
     std::replace(host.game_title.begin(), host.game_title.end(), '\n', ' ');
     find_data(host.io.title_id, host.sfo_handle, "TITLE_ID");
@@ -248,6 +249,7 @@ bool load_app_impl(Ptr<const void> &entry_point, HostState &host, const std::wst
 
     LOG_INFO("Title: {}", host.game_title);
     LOG_INFO("Serial: {}", host.io.title_id);
+    LOG_INFO("Version: {}", host.game_version);
     LOG_INFO("Category: {}", category);
 
     init_device_paths(host.io);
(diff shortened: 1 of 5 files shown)


Rinnegatamante:
Touch state is supposed to be not enabled at boot time.

Zangetsu38:
hmm i see, i have just tested one thging and all game have request this mode, and on yçour pr you have just revert one partt on this commit have breoken it
https://github.com/Vita3K/Vita3K/commit/4823c1700e417dc57155f4fcf5d192dce33b41e4#diff-44121ac517e8912761c1f3aeebd51889
with add on thjing for log, i have wanted removed this like you but i have thinked is not good solution.
But if you said is good like that oki

src/emulator/host/src/host.cpp
```diff
@@ -235,9 +235,11 @@ bool init(HostState &state, Config cfg) {
                 if (vfs::read_app_file(params, state.pref_path, state.io.title_id, "sce_sys/param.sfo")) {
                     SfoFile sfo_handle;
                     load_sfo(sfo_handle, params);
+                    find_data(state.game_version, sfo_handle, "APP_VER");
```
VelocityRa: nit: put this line below TITLE, for consistency

src/emulator/host/src/app.cpp
```diff
@@ -217,6 +217,8 @@ bool load_app_impl(Ptr<const void> &entry_point, HostState &host, const std::wst
     find_data(host.game_title, host.sfo_handle, "TITLE");
     std::replace(host.game_title.begin(), host.game_title.end(), '\n', ' ');
     find_data(host.io.title_id, host.sfo_handle, "TITLE_ID");
+    find_data(host.game_version, host.sfo_handle, "APP_VER");
+    std::replace(host.game_version.begin(), host.game_version.end(), '\n', ' ');
```
VelocityRa: should not be needed, I doubt `APP_VER` can contain newlines. test it.

src/emulator/host/src/host.cpp
```diff
@@ -235,9 +235,11 @@ bool init(HostState &state, Config cfg) {
...
+                    std::replace(state.game_version.begin(), state.game_version.end(), '\n', ' ');
```
VelocityRa: should not be needed, I doubt `APP_VER` can contain newlines. test it.

src/emulator/host/src/host.cpp
Zangetsu38: oki, i have just set it before because on game list and on sfo is writed before title

PR #428, opened 2019-03-13
touch: Masking only reportNum when sampling is disabled.

Rinnegatamante:


Zangetsu38:
Test on NumptyPhysics, it is not fixed with that, you line added in end with touch mode broken it :/

Rinnegatamante:
Touch in Numpty Physics is supposed not to work given the code. Have you tested on real hardware if touch actually works on it? Vita has all samplings disabled by default. Feel free to write a sample case with vitasdk and you'll see it is like i say.

Zangetsu38:
i have see that on doc, but yes, i know is right, but for Numpty Physics touch is enable on psvita just looçk video for see that, touch works normally with draw on screen

Rinnegatamante:
Touch doesn't work on Numpty Physics on real hardware. Tested on my own. I don't get what video should i see.

src/emulator/modules/SceTouch/SceTouch.cpp
```diff
@@ -29,36 +29,38 @@ static int peek_touch(HostState &host, SceUInt32 port, SceTouchData *pData) {
...
+        };
+
+        const SceFVector2 touch_pos_viewport = {
+            (touch_pos_drawable.x - host.viewport_pos.x) / host.viewport_size.x,
+            (touch_pos_drawable.y - host.viewport_pos.y) / host.viewport_size.y
+        };
+
+        if ((touch_pos_viewport.x >= 0) && (touch_pos_viewport.y >= 0) && (touch_pos_viewport.x < 1) && (touch_pos_viewport.y < 1)) {
+            pData->report[pData->reportNum].x = static_cast<uint16_t>(touch_pos_viewport.x * 1920);
+            pData->report[pData->reportNum].y = static_cast<uint16_t>(touch_pos_viewport.y * 1088);
```
VelocityRa: `1088` is not a typo right?

Zangetsu38: i have think same after see that, no idea if error for want said 1080*

Rinnegatamante: Not a typo, Vita screen is 960x544. Touchscreen to handle it returns samples into the res x2 range (so 1920x1088).

PR #430, opened 2019-03-15
Host cleanup

petmac:
I'm not finished yet, but my laptop battery is dying and I wanted to check I hadn't broken the Windows or Linux builds.


VelocityRa:
* Don't really need multiple formatting commits, just squash to one imo
* Some commits titles don't have a `prefix: `
* Why move discrete GPU exports to emulator? It's a host thing and in this case it also makes `main.cpp` messier.

petmac:
Thanks for reading

> * Don't really need multiple formatting commits, just squash to one imo
Good point 🤦‍♂️ I'll squash.

> * Some commits titles don't have a `prefix: `
These tended to be a straight move from host to emulator, and I wasn't sure how to prefix a change that touches both. Happy to fix, just let me know what you prefer.

> * Why move discrete GPU exports to emulator? It's a host thing and in this case it also makes `main.cpp` messier.
Don't worry too much about the emulator target getting messy. My next job is a cleanup of that, when I move app state out of HostState and clean up whatever else I can.

I see "host" as the emulated game's view of the host machine, so I don't think the discrete GPU exports go there. Maybe in emulator because it creates the first OpenGL context? 🤷‍♂️ (I moved init out of host because modules never call it.)

VelocityRa:
> These tended to be a straight move from host to emulator, and I wasn't sure how to prefix a change that touches both. Happy to fix, just let me know what you prefer.

I'd squash the "Move" commits to one and name it something like
`host: Move .../.../... to emulator` (mention them in the description instead if they don't fit)

> Don't worry too much about the emulator target getting messy. My next job is a cleanup of that, when I move app state out of HostState and clean up whatever else I can.

Ok, but I'm not sure about the code structure decision of moving all these files to `src/emulator`.
Perhaps if we moved Vita3K.ico/png/rc, resource.h outside of that directory it would be cleaner.

> I see "host" as the emulated game's view of the host machine, so I don't think the discrete GPU exports go there. Maybe in emulator because it creates the first OpenGL context? 🤷‍♂️ (I moved init out of host because modules never call it.)

I'm still not sure I understand this, wouldn't the emulated game's view of the host machine just be `HostState`? Either all the operations on that state would be "implementation details" and therefore not be part of "the game's view", or... what? I can't think of a definitive criterion of which of these operations would need to meet to belong in the host and not the emulator.

I see you picked [these](https://github.com/petmac/Vita3K/tree/host-cleanup/src/emulator/host/src) but they seem arbitrary to me. The SFO code which you [split](https://github.com/petmac/Vita3K/blob/host-cleanup/src/emulator/sfo.cpp) seems especially puzzling... even if it makes sense in some way, is it worth sacrificing readability by splitting logic like that?

petmac:
> I'd squash the "Move" commits to one and name it something like
> `host: Move .../.../... to emulator` (mention them in the description instead if they don't fit)

OK, sounds good. I might have to separate some of the commits due to the order I did the work in, but I'll try.

> Ok, but I'm not sure about the code structure decision of moving all these files to `src/emulator`.
> Perhaps if we moved Vita3K.ico/png/rc, resource.h outside of that directory it would be cleaner.

I think if something is only used from `emulator` it should live there, or in a library which is used by there. It shouldn't be in host, as that's for things used by modules (i.e., the emulated game.)

> I'm still not sure I understand this, wouldn't the emulated game's view of the host machine just be `HostState`? Either all the operations on that state would be "implementation details" and therefore not be part of "the game's view", or... what? I can't think of a definitive criterion of which of these operations would need to meet to belong in the host and not the emulator.

Similarly, if something is used by modules it can go in host, if there's no better place. I would like host to be small if possible, since every module depends on it.

I agree that currently the only thing in host should be the HostState, and any necessary functions for dealing with it. Which I think is nearly where I'm at now.

I would like to slim down `HostState`, maybe by storing the per-module state structs by pointer rather than value. I'd just like to fix the "every module depends on the definitions for every other module" problem that `host` currently has, as it increases compile times unnecessarily.

> I see you picked [these](https://github.com/petmac/Vita3K/tree/host-cleanup/src/emulator/host/src) but they seem arbitrary to me. The SFO code which you [split](https://github.com/petmac/Vita3K/blob/host-cleanup/src/emulator/sfo.cpp) seems especially puzzling... even if it makes sense in some way, is it worth sacrificing readability by splitting logic like that?

I left some SFO code in host, because `get_data()` is called by `sceAppMgrAppParamGetString()`. The rest was only used by `emulator` so it got moved there. It's likely that the contents of `host/sfo.h` and `host/app_util.h` belong elsewhere.

It's likely both got dumped in host because we didn't have a better place to put them.

VelocityRa:
Ok, fair enough.

petmac:
I'm merging this, as no one else has commented and the branch runs the risk of diverging from master too much.

PR #433, opened 2019-03-17
Gui: Add can hide gui after boot game.

Zangetsu38:
detected with test rtc, gui hide all time up part on screen for game, i have so thinked can add possibility for can hide gui in boot game directly without press G

### DIFF
--- a/src/emulator/gui/src/main_menubar.cpp
+++ b/src/emulator/gui/src/main_menubar.cpp
@@ -52,6 +52,10 @@ static void draw_config_menu(HostState &host) {
             }
             ImGui::EndMenu();
         }
+        if (ImGui::BeginMenu("GUI")) {
+            ImGui::MenuItem("Enable", nullptr, &host.cfg.show_gui);
+            ImGui::EndMenu();
+        }
         ImGui::PopStyleColor();
         ImGui::EndMenu();
     }
--- a/src/emulator/host/include/host/config.h
+++ b/src/emulator/host/include/host/config.h
@@ -40,6 +40,7 @@ struct Config {
     bool log_uniforms = false;
     std::vector<std::string> lle_modules;
     bool pstv_mode = false;
+    bool show_gui = false;
     optional<std::string> pref_path;
     bool archive_log = false;
 };
--- a/src/emulator/host/src/app.cpp
+++ b/src/emulator/host/src/app.cpp
@@ -305,6 +305,13 @@ ExitCode load_app(Ptr<const void> &entry_point, HostState &host, const std::wstr
         return ModuleLoadFailed;
     }
 
+    if (!host.cfg.show_gui) {
+        auto &imgui_render = host.display.imgui_render;
+
+        bool old_imgui_render = imgui_render.load();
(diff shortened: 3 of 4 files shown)


src/emulator/gui/src/main_menubar.cpp
```diff
@@ -52,6 +52,15 @@ static void draw_config_menu(HostState &host) {
             }
             ImGui::EndMenu();
         }
+        if (ImGui::BeginMenu("Gui State")) {
+            if (ImGui::MenuItem("Hide Gui", nullptr, !host.cfg.gui_state)) {
```
Rinnegatamante: Use a checkbox instead of two different voices.

Zangetsu38: ah yes

Zangetsu38: Done :)

src/emulator/host/src/app.cpp
```diff
@@ -305,6 +305,14 @@ ExitCode load_app(Ptr<const void> &entry_point, HostState &host, const std::wstr
         return ModuleLoadFailed;
     }
 
+    // Gui State
+    if (!host.cfg.gui_state) {
+        auto &display = host.display;
```
VelocityRa: may as well do
```suggestion
        auto &imgui_render = host.display.imgui_render;
```

src/emulator/host/src/config.cpp
```diff
@@ -126,6 +127,7 @@ ExitCode deserialize(Config &cfg) {
     get_yaml_value(config_node, "log-active-shaders", &cfg.log_active_shaders, false);
     get_yaml_value(config_node, "log-uniforms", &cfg.log_uniforms, false);
     get_yaml_value(config_node, "pstv-mode", &cfg.pstv_mode, false);
+    get_yaml_value(config_node, "show-gui", &cfg.show_gui, false);
```
VelocityRa:
```suggestion
    get_yaml_value(config_node, "show-gui", &cfg.show_gui, true);
```

src/emulator/host/include/host/config.h
```diff
@@ -40,6 +40,7 @@ struct Config {
     bool log_uniforms = false;
     std::vector<std::string> lle_modules;
     bool pstv_mode = false;
+    bool show_gui = false;
```
VelocityRa:
```suggestion
    bool show_gui = true;
```

src/emulator/host/src/config.cpp
Zangetsu38: ah for that no, i want leave it on false by default, for all user is better hide gui in defaut setting

src/emulator/host/include/host/config.h
Zangetsu38: same for here, like i have said, gui hide up part game, so hide it by defaut is normal

src/emulator/host/src/config.cpp
VelocityRa: Aight I guess it's fine if we can still enable it from the settings.

PR #434, opened 2019-03-18
Some GXM implementations and a few other fixes

VelocityRa:


src/emulator/host/src/app.cpp
```diff
@@ -309,7 +309,8 @@ ExitCode load_app(Ptr<const void> &entry_point, HostState &host, const std::wstr
         auto &imgui_render = host.display.imgui_render;
 
         bool old_imgui_render = imgui_render.load();
-        while (!imgui_render.compare_exchange_weak(old_imgui_render, !old_imgui_render));
+        while (!imgui_render.compare_exchange_weak(old_imgui_render, !old_imgui_render))
+            ;
```
Zangetsu38: Sorry, for here, and i have see, i have set one think no needed here you can remove While, with set just like that so 
>!imgui_render.compare_exchange_weak(old_imgui_render, !old_imgui_render);

pent0: The while was intentional, it was supposed to lock the current thread until the conditional is satisfied.

Zangetsu38: Oh i see i have think is needed only for loop 
 With press g

src/emulator/modules/SceGxm/SceGxm.cpp
```diff
@@ -858,56 +858,61 @@ EXPORT(int, sceGxmProgramIsSpriteCoordUsed) {
...
+    uint64_t parameter_offset = program->parameters_offset;
+
+    if (parameter_offset > 0)
+        parameter_offset += (uint64_t)&program->parameters_offset;
+    return (uint32_t)((uint64_t)parameter - parameter_offset) >> 4;
 }
 
-EXPORT(Ptr<void>, sceGxmProgramParameterGetName, Ptr<const SceGxmProgramParameter> parameter) {
-    return Ptr<void>(parameter.address() + parameter.get(host.mem)->name_offset);
+EXPORT(Ptr<const char>, sceGxmProgramParameterGetName, Ptr<const SceGxmProgramParameter> parameter) {
```
VelocityRa: @petmac Can this be rewritten to use raw C pointers? I'm not sure what the recent (?) bridging change was about.

petmac: I don't think so, because there's no safe way to convert back from a raw pointer to an address (Ptr).

VelocityRa: I see, thanks.

PR #441, opened 2019-03-22
gui: Improvements in game selector sorter.

Rinnegatamante:
- Made the Title sorter non case-dependant.
- Added descendant/ascendant sorting for every sort mode.

### DIFF
--- a/src/emulator/util/src/util.cpp
+++ b/src/emulator/util/src/util.cpp
@@ -170,6 +170,13 @@ std::u16string utf8_to_utf16(const std::string &str) {
 }
 #endif
 
+std::string toupper(const std::string &s) {
+    std::string r = s;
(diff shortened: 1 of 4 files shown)


src/emulator/util/src/util.cpp
```diff
@@ -170,6 +170,13 @@ std::u16string utf8_to_utf16(const std::string &str) {
 }
 #endif
 
+std::string toupper(std::string s) {
```
pent0:
```suggestion
std::string toupper(const std::string &s) {
```

Passing it as reference, since you did make a copy of ``s`` below

pent0: Do same for header

pent0:
Tested, lgtm

PR #443, opened 2019-03-23
gdbstub: Basic Implementation

1whatleytay:
This PR adds a gdb stub for debugging with GDB and IDA. With the latest commit, stability and speed has greatly improved, and I think it's ready to be tried out. Here's how you use it.

Uncomment this line in `src/emulator/CMakeLists.txt` and you're set. The GDB server will start on port 2159 the next time you start the game.
Screen Shot 2019-04-04 at 10 19 42 AM
With GDB, you should be able to connect with `target remote 127.0.0.1:2159` or with the GDB Remote Debugger option on IDA.

The server isn't fully complete (not every action is supported yet), but here's what you can do with it now:

- Connect to a running game or have the game wait for a debugger to connect.
- Read registers and memory.
- Continue and step through execution.
- Place and delete **breakpoints** (yay!)
- Switch and view threads (although not tested, main thread is usually selected by default).

breakpoints-1
Screen Shot 2019-04-05 at 12 52 30 PM

Tested on macOS and Windows. Let me know if you find any bugs. Reviews and feedback are welcome!

Big thanks to jfhs, xyz and friends for helping me solve and understand all the problems I encountered.


src/emulator/cpu/src/cpu.cpp
```diff
@@ -261,6 +291,22 @@ uint32_t read_lr(CPUState &state) {
     return value;
 }
 
+uint32_t read_fps(CPUState &state) {
```
VelocityRa:
```suggestion
uint32_t read_fpscr(CPUState &state) {
```

src/emulator/gdbstub/src/gdb.cpp
```diff
@@ -0,0 +1,597 @@
...
+std::string content_string(PacketCommand &command) {
+    return std::string(command.content_start, static_cast<unsigned long>(command.content_length));
+}
+
+// Hexes
+std::string be_hex(uint32_t value) {
+    return fmt::format("{:0>8x}", htonl(value));
+}
+
+std::string to_hex(uint32_t value) {
```
VelocityRa: Use [`this`](https://github.com/Vita3K/Vita3K/blob/master/src/emulator/util/include/util/log.h#L59) instead (add the `>` thing in there, too).
Because it's generic and there's no `8` it might not work. Test it, and if it doesn't remove the template and make multiple versions for u32, u64 and whatever else we use.
Or if you feel fancy you can probably still make that compile time with templates, but it's not trivial and there's no real need.

src/emulator/kernel/src/thread/thread.cpp
```diff
@@ -208,6 +212,8 @@ bool run_thread(ThreadState &thread, bool callback) {
             break;
         }
     }
+
+    if (true) { }
```
VelocityRa: What's this for? Add a comment if it's not an accident.

src/emulator/gdbstub/src/gdb.cpp
```diff
@@ -0,0 +1,597 @@
...
+// Sockets
+#include <arpa/inet.h>
+#include <netinet/in.h>
+#include <netinet/tcp.h>
+#include <sys/socket.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#define LOG_GDB_EVERYTHING
+//#define LOG_GDB_DEBUG
```
VelocityRa: Could make this `#define LOG_GDB_DEBUG 0` and check with `#if` below, would get rid of the comment.

src/emulator/gdbstub/src/gdb.cpp
```diff
@@ -0,0 +1,597 @@
...
+#define LOG_GDB_DEBUG
+#endif
+
+#ifdef LOG_GDB_DEBUG
+#define LOG_GDB LOG_INFO
+#else
+#define LOG_GDB(a, ...)
+#endif
+
+// Based off RPCS3's implementation.
```
VelocityRa: Might want to credit jfhs here.

src/emulator/kernel/src/thread/thread.cpp
1whatleytay: Nope definitely an accident! Was trying to give some ground for a breakpoint to stop at but I never used it.

src/emulator/gdbstub/src/gdb.cpp
1whatleytay: I would, but I would have to remove the `0x` in that method too (gdb already knows its expecting raw hexes). Maybe I could add another method? One that spits out raw hexes and another that adds the `0x` at the beginning?

src/emulator/gdbstub/src/gdb.cpp
1whatleytay: Yeah, he and xyz were super helpful.

src/emulator/gdbstub/src/gdb.cpp
VelocityRa: Ah ok. No worries then, use your own function.

1whatleytay:
I added the option to wait for a debugger to connect as a definition instead of an option in the config file because I didn't have access to the ConfigState in `thread_function()`.

Is it important enough that I should try to get it to work anyway? It'd probably be as simple as passing the ConfigState around a bit, but I wasn't in the mood to figure that out at the time.

VelocityRa:
It's fine, though we normally don't use CMake vars for this sort of thing, but do [this](https://github.com/Vita3K/Vita3K/blob/master/src/emulator/kernel/src/load_self.cpp#L56-L57) instead.

VelocityRa:
You committed some unrelated things :)

1whatleytay:
Yeah, I told clang-format to do its thing and it found some changes in some different files if that's what you're talking about. I thought I might as well push them. Do you want me to undo them?

P.S. I think it's about done. Just want to check with JFHS on one thing and then I think the PR is good to go.

src/emulator/CMakeLists.txt
```diff
@@ -49,13 +49,17 @@ endmacro(configure_boost)
 
 configure_boost()
 
+# -DUSE_GDBSTUB will compile Vita3K with a GDB stub.
```
VelocityRa: this

src/emulator/kernel/src/thread/thread.cpp
```diff
@@ -208,6 +227,9 @@ bool run_thread(ThreadState &thread, bool callback) {
             break;
         }
     }
+
+    if (true) {
```
VelocityRa: this

src/emulator/modules/SceGxm/SceGxm.cpp
```diff
@@ -1134,7 +1134,7 @@ EXPORT(void, sceGxmSetTwoSidedEnable, SceGxmContext *context, SceGxmTwoSidedMode
 }
 
 EXPORT(int, sceGxmSetUniformDataF, void *uniformBuffer, const SceGxmProgramParameter *parameter, unsigned int componentOffset, unsigned int componentCount, const float *sourceData) {
-//    assert(uniformBuffer != nullptr);
+    //    assert(uniformBuffer != nullptr);
```
VelocityRa: this

src/emulator/shader/src/usse_translator_entry.cpp
```diff
@@ -36,7 +36,7 @@ using USSEMatcher = shader::decoder::Matcher<Visitor, uint64_t>;
 template <typename V>
 boost::optional<const USSEMatcher<V> &> DecodeUSSE(uint64_t instruction) {
     static const std::vector<USSEMatcher<V>> table = {
-// clang-format off
+    // clang-format off
```
VelocityRa: this

1whatleytay:
Yeah, the first one is my bad (don't remember doing it, must be a ghost). The other two I'm pretty sure were clang-format.

src/emulator/host/include/host/state.h
```diff
@@ -29,7 +30,6 @@
 #include <nids/types.h>
 #include <np/state.h>
 #include <renderer/state.h>
-#include <gdbstub/state.h>
```
1whatleytay: clang-format also did this in an attempt to make it all nice and alphabetical, but it ended up breaking the windows build (state.h needs windows.h to be included before) it. I'll address all of these things. Thanks for letting me know :D

src/emulator/gdbstub/include/gdbstub/state.h
```diff
@@ -0,0 +1,53 @@
...
+
+    std::string last_reply = "";
+
+    bool extended_mode = false;
+    SceUID current_thread = 0;
+
+    bool is_running = false;
+};
+
+#endif
```
pent0: Newline

pent0: New line

src/emulator/mem/include/mem/mem.h
```diff
@@ -37,6 +37,7 @@ struct MemState {
     Allocated allocated_pages;
     std::mutex generation_mutex;
     GenerationNames generation_names;
+    std::map<Address, uint32_t> breakpoints;
```
pent0: It feels wrong to put breakpoints in here. It doesn't belong there. Perhaps put it in host or kernel ? Or put it in GDB struct, and do memory modification there.

src/emulator/CMakeLists.txt
```diff
@@ -49,13 +49,16 @@ endmacro(configure_boost)
 
 configure_boost()
 
+# add_definitions(-DUSE_GDBSTUB)
```
pent0: Making this an CMake option and put it in CMake config file (```config_file``` ?) would be cool. I would avoid doing this to be honest.

src/emulator/CMakeLists.txt
```diff
@@ -49,13 +49,16 @@ endmacro(configure_boost)
...
+
 add_subdirectory(audio)
 add_subdirectory(cpu)
 add_subdirectory(crypto)
 add_subdirectory(ctrl)
 add_subdirectory(dialog)
 add_subdirectory(disasm)
 add_subdirectory(glutil)
+add_subdirectory(gdbstub)
```
pent0: Put this before ```glutil```. Everything is in order and it should be, PetMac wants this to prevent conflicts in the future.

a b c **d** e f g h i j k **l** m n o p q r s t u v w x y z

pent0:
Please also fix the conflicts. Other than that, yeah, this is really cool, thank you again :D

src/emulator/gdbstub/src/gdb.cpp
```diff
@@ -0,0 +1,652 @@
...
+    return "";
+}
+
+std::string cmd_unimplemented(HostState &state, PacketCommand &command) {
+    LOG_GDB("GDB Server: Unimplemented Packet. {}", content_string(command));
+
+    return "";
+}
+
+PacketFunctionBundle functions[] = {
```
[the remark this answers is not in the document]
1whatleytay: And maybe const?

src/emulator/mem/include/mem/mem.h
1whatleytay: I think I put it in there so I could access it from a CPUState object, but the CPU doesn't even care anymore. I think I'll put it in the Kernel struct.

1whatleytay:
Will do! Thanks for reviewing! 😄

src/emulator/gdbstub/src/gdb.cpp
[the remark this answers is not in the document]
pent0: If it works :)

PR #444, opened 2019-03-24
modules/SceLibKernel: Some sceClib improvements.

Rinnegatamante:
- Implemented sceClibMemcmp, sceClibMemmove and sceClibStrcmp.
- Made sceClibMemcpy and sceClibMemcpy_safe compliant to standard C functions.

### DIFF
--- a/src/emulator/modules/SceLibKernel/SceLibKernel.cpp
+++ b/src/emulator/modules/SceLibKernel/SceLibKernel.cpp
@@ -88,30 +88,31 @@ EXPORT(int, sceClibMemchr) {
     return UNIMPLEMENTED();
 }
 
-EXPORT(int, sceClibMemcmp) {
-    return UNIMPLEMENTED();
+EXPORT(int, sceClibMemcmp, const void *ptr1, const void *ptr2, SceSize num) {
+    return memcmp(ptr1, ptr2, num);
 }
 
 EXPORT(int, sceClibMemcmpConstTime) {
     return UNIMPLEMENTED();
 }
 
-EXPORT(int, sceClibMemcpy, void *destination, const void *source, SceSize num) {
-    memcpy(destination, source, num);
-    return 0;
+EXPORT(Ptr<void>, sceClibMemcpy, Ptr<void> destination, const void *source, SceSize num) {
+    memcpy(destination.get(host.mem), source, num);
+    return destination;
 }
 
 EXPORT(int, sceClibMemcpyChk) {
     return UNIMPLEMENTED();
 }
 
-EXPORT(int, sceClibMemcpy_safe, void *destination, const void *source, SceSize num) {
-    memcpy(destination, source, num);
-    return 0;
+EXPORT(Ptr<void>, sceClibMemcpy_safe, Ptr<void> destination, const void *source, SceSize num) {
+    memcpy(destination.get(host.mem), source, num);
+    return destination;
 }
 
-EXPORT(int, sceClibMemmove) {
-    return UNIMPLEMENTED();
+EXPORT(Ptr<void>, sceClibMemmove, Ptr<void> destination, const void *source, SceSize num) {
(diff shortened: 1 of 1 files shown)


src/emulator/modules/SceLibKernel/SceLibKernel.cpp
```diff
@@ -88,30 +88,31 @@ EXPORT(int, sceClibMemchr) {
...
 }
 
 EXPORT(int, sceClibMemcmpConstTime) {
     return UNIMPLEMENTED();
 }
 
-EXPORT(int, sceClibMemcpy, void *destination, const void *source, SceSize num) {
-    memcpy(destination, source, num);
-    return 0;
+EXPORT(Ptr<void>, sceClibMemcpy, Ptr<void> destination, const void *source, SceSize num) {
```
VelocityRa: why is source not a `Ptr<>`? you sure it works?

src/emulator/modules/SceLibKernel/SceLibKernel.cpp
```diff
@@ -88,30 +88,31 @@ EXPORT(int, sceClibMemchr) {
...
 }
 
 EXPORT(int, sceClibMemcpyChk) {
     return UNIMPLEMENTED();
 }
 
-EXPORT(int, sceClibMemcpy_safe, void *destination, const void *source, SceSize num) {
-    memcpy(destination, source, num);
-    return 0;
+EXPORT(Ptr<void>, sceClibMemcpy_safe, Ptr<void> destination, const void *source, SceSize num) {
```
VelocityRa: ditto

src/emulator/modules/SceLibKernel/SceLibKernel.cpp
```diff
@@ -227,8 +228,8 @@ EXPORT(Ptr<char>, sceClibStrchr, const char *str, int c) {
     return Ptr<char>(res, host.mem);
 }
 
-EXPORT(int, sceClibStrcmp) {
-    return UNIMPLEMENTED();
+EXPORT(int, sceClibStrcmp, const char *string1, const char *string2) {
```
VelocityRa: ditto for `string1` and `string2`

src/emulator/modules/SceLibKernel/SceLibKernel.cpp
Rinnegatamante: Original code wasn't a Ptr too.

src/emulator/modules/SceLibKernel/SceLibKernel.cpp
Rinnegatamante: ditto

src/emulator/modules/SceLibKernel/SceLibKernel.cpp
Rinnegatamante: ditto

src/emulator/modules/SceLibKernel/SceLibKernel.cpp
pent0: Those functions are all wrong. Not sure why they originally work. Use ptr<>, it represents guest's 32 bit pointer. Using those is basically 64-bit integer.

Rinnegatamante: There are a lot of functions using direct C pointers in the whole emulator codebase. I can recall @petmac a long time ago saying it being now safe after a PR due to changes in bridging.

pent0: Sorry I didn't remember. These are avail since forever. [Here](https://github.com/Vita3K/Vita3K/blob/master/src/emulator/module/include/module/bridge_types.h#L34)

pent0: Unless we need the guest address, we can safely use this

PR #448, opened 2019-03-30
modules: Add SceFios2Kernel module and stub function.

Zangetsu38:
- Register module SceFios2Kernel  with stub function know


VelocityRa:
Not sure we should be adding these manually.
This is what [gen-modules](https://github.com/Vita3K/Vita3K/tree/master/src/gen-modules) is for, we should run it on the new NIDs.

petmac:
> Not sure we should be adding these manually.
> This is what [gen-modules](https://github.com/Vita3K/Vita3K/tree/master/src/gen-modules) is for, we should run it on the new NIDs so that everything is added, not just `SceFios2Kernel`.

That is true. Give me a shout @Zangetsu38 if you need help using it. This Wiki page should be helpful too:
https://github.com/Vita3K/Vita3K/wiki/Module-generation

PR #450, opened 2019-04-03
Upgrade SDL from 2.0.7 to 2.0.9

petmac:
The main purpose of this was to attempt to work around OpenGL bugs in macOS 10.14.4. I don't think it was a success, but we may as well update SDL anyway, to take any further fixes.

I took the Mac version from https://www.libsdl.org/download-2.0.php rather than pull it out of the Homebrew package. One thing I noticed was the code signing information disappeared. It would be helpful if another Mac user can verify it works OK on their system.


1whatleytay:
On it, captain PetMac!

1whatleytay:
Can confirm (after solving my boost issues), I am able to build and run Vita Doom on macOS 10.14.3. I don't think the update breaks anything, afaik.
Screen Shot 2019-04-10 at 8 44 04 AM

PR #451, opened 2019-04-09
boost: Load Boost with FindBoost.cmake when possible

1whatleytay:
This PR gives the option for users to load Boost through find_package. This allows mac users (like me :D) to load from Homebrew, etc.

I mainly did this because my compiler can't build boost 1.67 anymore, but I hope this helps or saves someone time.


1whatleytay:
Yeah, I think that would be a good idea. I don't want to force developers to have brew on their computer though (even if everyone already has it), so maybe you could pass an option to gen-macos.sh that skips building boost?

That way you don't need to `brew install boost` to build Vita3K but you can use it if you have it.

1whatleytay:
Reminder to future desgroup to make find_package `QUIET` as suggested by @pent0.

pent0:
Okey dokey

PR #452, opened 2019-04-11
emulator: Throw error dialogue when failed to create OpenGL context

0x8080:


1whatleytay:
LGTM. Will have to pull out my HP from 8 years ago to test if it works though.

1whatleytay:
Should probably open an issue in the future to print out, not only the OpenGL version, but also driver info. My Bootcamp drivers had a bunch of issues, so it would help debugging for driver specific issues.

0x8080:
yeh probably

Issue #321, opened 2018-09-09
Build issues wrt boost 


yoshi314:
When attempting a build i get the following

```
CMake Error at /usr/share/cmake/Modules/FindBoost.cmake:2044 (message):
  Unable to find the requested Boost libraries.

  Boost version: 1.67.0

  Boost include path: /home/yoshi/src/console/vita/Vita3K/src/external/boost

  Could not find the following static Boost libraries:

          boost_filesystem
          boost_program_options

  Some (but not all) of the required Boost libraries were found.  You may
  need to install these additional Boost libraries.  Alternatively, set
  BOOST_LIBRARYDIR to the directory containing Boost libraries or BOOST_ROOT
  to the location of Boost.
Call Stack (most recent call first):
  src/emulator/CMakeLists.txt:37 (find_package)
  src/emulator/CMakeLists.txt:50 (configure_boost)


Boost library is not found
```

Do i have to build the local submodule copy of boost first? My system has boost 1.65.0 , but it seems that's too old.

VelocityRa:
Hello, I just noticed we lack the proper build scripts for linux since we added boost.
Please run these commands and try again:
https://github.com/Vita3K/Vita3K/blob/master/gen.sh#L6-L15

VelocityRa:
Did it work?

yoshi314:
I'll let you know in a few hours.

yoshi314:
I am experiencing some issues building boost with the instructions provided. I'll post the error log later.

yoshi314:
i'm trying right now, i'll see what happens.

yoshi314:
```
/gen-linux.sh 
+ [[ -z '' ]]
+ mkdir -p src/external/boost-build
+ cd src/external/boost
+ chmod +x tools/build/src/engine/build.sh
+ sh bootstrap.sh
Building Boost.Build engine with toolset gcc... tools/build/src/engine/bin.linuxx86_64/b2
Detecting Python version... 3.6
Detecting Python root... /usr/lib/python-exec/python3.6/../../..
Unicode/ICU support for Boost.Regex?... /usr
Generating Boost.Build configuration in project-config.jam...

Bootstrapping is done. To build, run:

    ./b2
    
To adjust configuration, edit 'project-config.jam'.
Further information:

   - Command line help:
     ./b2 --help
     
   - Getting started guide: 
     http://www.boost.org/more/getting_started/unix-variants.html
     
   - Boost.Build documentation:
[...]

PR #454, opened 2019-04-13
gen-linux.sh: Add ignore-site-config for boost

xyzz:
This fixes the build on gentoo and specifically this issue https://github.com/Vita3K/Vita3K/issues/321#issuecomment-450370966

Since the idea is to build an independent copy of boost there's no reason to include the OS site-config


pent0:
Thanks

PR #458, opened 2019-04-21
Neater search bar

IllusionMan1212:
image


1whatleytay:
O_O
Screen Shot 2019-04-21 at 12 42 52 PM

src/emulator/gui/src/game_selector.cpp
```diff
@@ -142,6 +141,13 @@ void draw_game_selector(HostState &host) {
                 break;
             }
         }
+        ImGui::PushStyleColor(ImGuiCol_Text, GUI_COLOR_SEARCH_BAR_TEXT);
+        ImGui::PushStyleColor(ImGuiCol_FrameBg, GUI_COLOR_SEARCH_BAR_BG);
+        ImGui::SameLine(display_size.x - 515);
```
pent0: I don't really accept this, sorry, i know this also subtract padding but it also contains hardcoded text width. It may breaks with other font in the future.

Use ImGui::CalcTextSize to get text size, than get the position by:
```cpp
display_size.x - (ImGui::CalcTextSize("Search").x + padding + text_box_width)
```

PR #460, opened 2019-04-21
gui: Option To Add a Background Image

1whatleytay:
As per request, config.yml accepts the `background-image` and `background-alpha` options.

`background-image` is an optional path to some picture on your hardrive. It changes your launch page's background for some nice tasty customization. It should stretch to fit your screen, just pretend that it is in the next screenshot.
Screen Shot 2019-04-21 at 12 05 34 PM

`background-alpha` is an optional decimal that lets you see the image more or less. The higher the value, the harder it is to see. Kinda confusing imo but at least it works.

### DIFF
--- a/src/emulator/glutil/include/glutil/object.h
+++ b/src/emulator/glutil/include/glutil/object.h
@@ -24,20 +24,25 @@
 
 class GLObject {
 public:
-    using Deleter = std::function<void(GLuint)>;
+    using SingularDeleter = std::function<void(GLuint)>;
+    using AggregateDeleter = std::function<void(int, const GLuint *)>;
 
     GLObject() = default;
     ~GLObject();
 
-    bool init(GLuint name, Deleter deleter);
+    bool init(GLuint name, AggregateDeleter aggregate_deleter);
+    bool init(GLuint name, SingularDeleter singular_deleter);
     GLuint get() const;
     operator GLuint() const;
+    operator bool() const;
 
 private:
+    bool init(GLuint name);
     const GLObject &operator=(const GLObject &) = delete;
 
     GLuint name = 0;
-    Deleter deleter = nullptr;
+    AggregateDeleter aggregate_deleter = nullptr;
+    SingularDeleter singular_deleter = nullptr;
 };
 
 using SharedGLObject = std::shared_ptr<GLObject>;
--- a/src/emulator/gui/include/gui/state.h
+++ b/src/emulator/gui/include/gui/state.h
@@ -90,6 +90,8 @@ struct State {
     char disassembly_count[5] = "100";
     std::vector<std::string> disassembly;
 
+    uint32_t background_texture = 0;
+
     SceUID thread_watch_index = -1;
 
     // imgui
--- a/src/emulator/gui/include/gui/state.h
+++ b/src/emulator/gui/include/gui/state.h
@@ -22,6 +22,8 @@
 #include <imgui.h>
 #include <imgui_memory_editor.h>
 
+#include <glutil/object.h>
+
 namespace gui {
 
 enum SelectorState {
@@ -90,7 +92,7 @@ struct State {
     char disassembly_count[5] = "100";
     std::vector<std::string> disassembly;
 
-    uint32_t background_texture = 0;
+    GLObject background_texture;
 
     SceUID thread_watch_index = -1;
 
--- a/src/emulator/gui/src/gui.cpp
+++ b/src/emulator/gui/src/gui.cpp
@@ -138,8 +138,11 @@ static void init_background(State &gui, const std::string &image_path) {
         return;
     }
 
-    glGenTextures(1, &gui.background_texture);
-    glBindTexture(GL_TEXTURE_2D, gui.background_texture);
+    GLuint texture;
+    glGenTextures(1, &texture);
+    gui.background_texture.init(texture, glDeleteTextures);
+
+    glBindTexture(GL_TEXTURE_2D, gui.background_texture.get());
     glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA, width, height, 0, GL_RGBA, GL_UNSIGNED_BYTE, data);
     glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_REPEAT);
     glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_S, GL_REPEAT);
--- a/src/emulator/gui/src/gui.cpp
+++ b/src/emulator/gui/src/gui.cpp
@@ -29,6 +29,9 @@
 
 #include <SDL_video.h>
 
+#define STB_IMAGE_IMPLEMENTATION
+#include <stb_image.h>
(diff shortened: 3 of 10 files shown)


src/emulator/gui/include/gui/state.h
```diff
@@ -90,6 +90,8 @@ struct State {
     char disassembly_count[5] = "100";
     std::vector<std::string> disassembly;
 
+    uint32_t background_texture = 0;
```
1whatleytay: Just wanted to point out that this isn't a special texture object or even a `GLuint`.

src/emulator/gui/src/gui.cpp
```diff
@@ -120,12 +123,40 @@ static void init_font(State &gui) {
...
 
+static void init_background(State &gui, const std::string &image_path) {
+    if (!std::ifstream(image_path).good()) {
+        LOG_INFO("Invalid background file path {}.", image_path);
+        return;
+    }
+
+    int32_t width = 0;
+    int32_t height = 0;
+    stbi_uc *data = stbi_load(image_path.c_str(), &width, &height, nullptr, 3);
```
pent0: STBI_rgb_alpha. Waste some more GPU memory to acheive godly effect on some image

src/emulator/gui/src/gui.cpp
```diff
@@ -120,12 +123,40 @@ static void init_font(State &gui) {
...
+
+    if (!data) {
+        LOG_INFO("Could not load background from {}.", image_path);
+        return;
+    }
+
+    glGenTextures(1, &gui.background_texture);
+    glBindTexture(GL_TEXTURE_2D, gui.background_texture);
+    glTexImage2D(GL_TEXTURE_2D, 0, GL_RGB, width, height, 0, GL_RGB, GL_UNSIGNED_BYTE, data);
```
pent0: GL_RGBA too!

Zangetsu38:
No relate it, but if you can found how implement game icon ^^

1whatleytay:
Yeah, now that we have stb included might as well do other things with it. Will look into it.

1whatleytay:
If you don't think we need a glDeleteTextures call that's cool. I just wanted to play it safe.

pent0:
Everything in Vita3K should properly cleanup itself with RAII. Wraps it in a [Resource<GLint>](https://github.com/Vita3K/Vita3K/blob/master/src/emulator/util/include/util/resource.h)

VelocityRa:
@1whatleytay Yup you should do that. More specifically use [GLObject](https://github.com/Vita3K/Vita3K/blob/master/src/emulator/glutil/include/glutil/object.h).
Will need some template magic to deal with the fact that `glDeleteTextures` accepts 2 arguments (first one being '1') and it it currently [doesn't do](https://github.com/Vita3K/Vita3K/blob/master/src/emulator/glutil/src/object.cpp#L24) that, or if you're not very experienced with them maybe you can figure out a simpler way to do it, like a deriving class where you override the deleter.

VelocityRa:
Or maybe we should just have a separate class for it like so (rough idea):
```cpp
class GLTexture
{
public:
    GLTexture(const std::string& name, int width, int height) :
        m_id(LoadTexture(name.c_str(), width, height)) {
    }

    ~GLTexture() {
        if(m_id)
            glDeleteTextures(1, &m_id);
        m_id = 0;
    }

    GLuint id() const {
        return m_id;
    }

private:
    GLTexture(const GLTexture&);
    GLTexture& operator=(const GLTexture&);

    GLuint m_id;
};
```
in which you'd also put your `init_background` configuration GL calls so it's more reusable.

VelocityRa:
...or here's [RPCS3's texture class](https://github.com/RPCS3/rpcs3/blob/12dc3c18723788024fd62a8dd6dd0f24d830d479/rpcs3/Emu/RSX/GL/GLHelpers.h#L1332-L1820). Maybe we ought to simply steal theirs? :)

pent0:
Or just do [this](https://godbolt.org/z/5F-Fpk)?

VelocityRa:
@pent0 This two-argument delete call happens often for GL objects, so maybe it's better to have a way to do it in `GLObject` itself.
Also consider `GLTexture` which would abstract the calls needed to configure the texture as I said (you _could_ make an `init` function or something if you don't like OOP, but we already have classes for GL stuff so why not).

1whatleytay:
What's wrong with
```
    GLuint texture;
    glGenTextures(1, &texture);
    gui.background_texture.init(texture, [](GLuint tex) { glDeleteTextures(1, &tex); });
```

1whatleytay:
Ehh!? My bad, I'm a bit tired today. Is this okay?

VelocityRa:
@1whatleytay I don't think so, please read above.

1whatleytay:
Okay, how about this? (see amend)
Would rather use the existing GLObject class than make another one specifically for textures. I'm sure it will be helpful, but I'm not sure if this is the right PR to put it under imo.

VelocityRa:
Ok, we can refactor later if need be.
Could you change ShaderDeleter/Deleter names to something like SingularDeleter/AggregateDeleter though?

1whatleytay:
No problem. Amended again.

src/emulator/glutil/include/glutil/object.h
```diff
@@ -24,20 +24,25 @@
...
     operator GLuint() const;
+    operator bool() const;
 
 private:
+    bool init(GLuint name);
     const GLObject &operator=(const GLObject &) = delete;
 
     GLuint name = 0;
-    Deleter deleter = nullptr;
+    AggregateDeleter deleter = nullptr;
```
VelocityRa:
```suggestion
    AggregateDeleter aggregate_deleter = nullptr;
```

src/emulator/glutil/include/glutil/object.h
```diff
@@ -24,20 +24,25 @@
...
+    SingularDeleter shaderDeleter = nullptr;
```
VelocityRa:
```suggestion
    SingularDeleter singular_deleter = nullptr;
```

src/emulator/glutil/include/glutil/object.h
VelocityRa: This was not renamed

1whatleytay: AHHHHH I'm sorry T_T

petmac:
Apologies for missing the review boat on this one. Looks like the changes to `GLObject` to add aggregate deleters are more or less what `GLObjectArray` is for. It's OK to have an array with 1 thing in it.

VelocityRa:
Damn. Forgot about `GLObjectArray`.

VelocityRa:
Maybe we leave the changes as they are and refactor existing code to use `GLObject` when only 1 object is required?
Some current usages of `GLObjectArray` one are for 1 object, and seeing, ie. `context.element_buffer[0]` all over in the code is misleading that more than one such objects exist (I remember being briefly confused about this in the past). We could enforce this by disallowing `GLObjectArray` to be used with 1 object.

PR #463, opened 2019-04-22
kernel/load_self: Add another check for when the segment shouldn't be loaded

xyzz:
As we only have up to 4 segments in the structure, don't attempt to place any more than that.


src/emulator/kernel/src/load_self.cpp
```diff
@@ -458,15 +458,19 @@ SceUID load_self(Ptr<const void> &entry_point, KernelState &kernel, MemState &me
...
+        // Skip non-loadable segments
         auto it = segment_reloc_info.find(segment_index);
-        if (it != segment_reloc_info.end())
-            segment.vaddr = it->second.addr;
-        else
-            segment.vaddr = 0;
+        if (it == segment_reloc_info.end())
+            continue;
 
+        if (segment_index >= 4) {
```
petmac: Is there a named constant we can use? If not, can we add one?

xyzz: Sure

PR #464, opened 2019-04-22
audio: Fix use after free in mix_out_port

xyzz:
`output` becomes invalidated after calling `pop()`


src/emulator/audio/src/audio.cpp
```diff
@@ -49,8 +49,9 @@ static void mix_out_port(uint8_t *stream, uint8_t *temp_buffer, int len, AudioOu
             output.buf += bytes_to_put;
             output.len_bytes -= bytes_to_put;
             if (output.len_bytes <= 0) {
+                SceUID thread = output.thread;
```
pent0:
```suggestion
                const SceUID thread = output.thread;
```

PR #465, opened 2019-04-22
Add .gitattributes file.

totlmstr:
Quick and simple commit.

This is mostly to ensure line endings when grabbing files for cross-compilation.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,17 @@
+# Properly detect files as C++ on Github
+*.h			linguist-language=cpp
+*.inc			linguist-language=cpp
+*.frag			linguist-language=GLSL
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -0,0 +1,17 @@
+# Properly detect files as C++ on Github
+*.h			linguist-language=cpp
+*.inc			linguist-language=cpp
+*.frag			linguist-language=cpp
```
pent0:
```suggestion
*.frag			linguist-language=GLSL
```

totlmstr: Done. Also did so for `.vert` files.

pent0:
Won't hurt I guess.

PR #466, opened 2019-04-23
cpu: Don't map the NULL page into unicorn

xyzz:
This allows us to gracefully handle invalid memory access by a game
instead of crashing the whole thing.


xyzz:
While this fixes the (possibly the most common) NULL page issue, by allowing us to gracefully handle invalid memory access and not crash the whole emulator, the same issue is present if a game accesses any invalid pointer.

The proper way (IMO) to do it would be to change memory mapping code to only map memory into unicorn as it's mapped, instead of currently mapping the whole 4G address space. However, this is non trivial as there are multiple unicorns running and you need to update everybody's memory map as they aren't shared... For now, I believe this change still is beneficial for the case of debugging games that crash on null derefs.

petmac:
Thanks for this.

petmac:
I agree with your comments FYI, and you are correct about why we don’t currently only map as needed.

PR #468, opened 2019-04-25
Fix various warnings

pent0:
We should write code that produce less warnings.

### DIFF
--- a/src/emulator/app.cpp
+++ b/src/emulator/app.cpp
@@ -240,9 +240,9 @@ bool install_vpk(Ptr<const void> &entry_point, HostState &host, const std::wstri
     return true;
 }
 
-bool load_app_impl(Ptr<const void> &entry_point, HostState &host, const std::wstring &path, AppRunType run_type) {
+static bool load_app_impl(Ptr<const void> &entry_point, HostState &host, const std::wstring &path, AppRunType run_type) {
     if (path.empty())
-        return InvalidApplicationPath;
+        return false;
 
     if (run_type == AppRunType::Vpk) {
         if (!install_vpk(entry_point, host, path)) {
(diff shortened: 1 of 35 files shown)


src/emulator/renderer/src/texture.cpp
```diff
@@ -38,8 +38,8 @@ void configure_bound_texture(const SceGxmTexture &gxm_texture) {
     glTexParameteriv(GL_TEXTURE_2D, GL_TEXTURE_SWIZZLE_RGBA, swizzle);
 
     const GLenum internal_format = translate_internal_format(fmt);
-    const unsigned int width = gxm::get_width(&gxm_texture);
-    const unsigned int height = gxm::get_height(&gxm_texture);
+    const unsigned int width = static_cast<decltype(width)>(gxm::get_width(&gxm_texture));
```
VelocityRa:
```suggestion
    const auto width = static_cast<uint32_t>(gxm::get_width(&gxm_texture));
```
same in 3 cases below
better avoid `decltype` when we can

src/emulator/app.cpp
```diff
@@ -370,7 +370,9 @@ ExitCode run_app(HostState &host, Ptr<const void> &entry_point) {
         ::call_import(host, cpu, nid, main_thread_id);
     };
 
-    const SceUID main_thread_id = create_thread(entry_point, host.kernel, host.mem, host.io.title_id.c_str(), SCE_KERNEL_DEFAULT_PRIORITY_USER, SCE_KERNEL_STACK_SIZE_USER_MAIN, call_import, false);
+    const SceUID main_thread_id = create_thread(entry_point, host.kernel, host.mem, host.io.title_id.c_str(), SCE_KERNEL_DEFAULT_PRIORITY_USER, static_cast<int>(SCE_KERNEL_STACK_SIZE_USER_MAIN),
+        call_import, false);
```
VelocityRa: weird line break, did you run clang-format?

src/emulator/screen_render.cpp
```diff
@@ -84,7 +84,8 @@ void gl_screen_renderer::render(const HostState &host) {
 
     glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
 
-    glViewport(host.viewport_pos.x, host.viewport_pos.y, host.viewport_size.x, host.viewport_size.y);
+    glViewport(static_cast<gl::GLint>(host.viewport_pos.x), static_cast<gl::GLint>(host.viewport_pos.y), static_cast<gl::GLsizei>(host.viewport_size.x), 
+        static_cast<gl::GLsizei>(host.viewport_size.y));
```
VelocityRa: ditto

src/emulator/app.cpp
pent0: Run now.

PR #473, opened 2019-05-03
main: Refactor working paths with a new Root class using Boost.Filesystem.

totlmstr:
In the current code, there are three different locations where the program gets the current working path.

* [spdlog initialization](https://github.com/Vita3K/Vita3K/blob/2a45507596910d63b4c5a125647f3a8209243274/src/emulator/util/src/util.cpp#L37)
*  [SDL initialization](https://github.com/Vita3K/Vita3K/blob/2a45507596910d63b4c5a125647f3a8209243274/src/emulator/init.cpp#L104)
* [and one when creating a new sink for spdlog](https://github.com/Vita3K/Vita3K/blob/2a45507596910d63b4c5a125647f3a8209243274/src/emulator/app.cpp#L267)

In a worst-case scenario, all three could be different, and the logs and the program would be in different places.

This code refactors the entire sequence of events to [which is recommended by SDL](https://github.com/Vita3K/Vita3K/blob/2a45507596910d63b4c5a125647f3a8209243274/src/external/sdl/windows/include/SDL_filesystem.h#L41): initialize the paths at the start. This also has the side benefit of testing Boost, SDL, and spdlog all in one. On the other hand, this bumps the *entire* repo's `CMAKE_CXX_STANDARD` to C++17.


src/emulator/app.cpp
```diff
@@ -265,28 +262,11 @@ static bool load_app_impl(Ptr<const void> &entry_point, HostState &host, const s
...
-            }
-#else
-            logging::add_sink(string_utils::utf_to_wide(log_name));
-#endif
-        } catch (const spdlog::spdlog_ex &ex) {
-            std::cerr << "File log initialization failed: " << ex.what() << std::endl;
-        }
+        const fs::path log_directory{ host.base_path + "/logs" };
+        fs::create_directory(log_directory);
+        const auto log_name{ log_directory / (host.game_title + " - [" + host.io.title_id + "].log") };
```
IllusionMan1212: you need to sanitize `host.game_title` with `string_utils::remove_special_chars` for the logging to work correctly

totlmstr: Done.

I initially assumed that Boost.Filesystem removes all the special characters from the filename but it [doesn't apparently remove all of the ones mentioned in `remove_special_chars`](https://www.boost.org/doc/libs/1_70_0/libs/filesystem/doc/reference.html#Operating-system-examples) from my searching.

totlmstr:
I feel as if I should have put in a needs testing line somewhere.

In future commits, this line will be added. *doh*

VelocityRa:
Or you can interactive rebase (`git rebase -i`) and rename them @totlmstr.

totlmstr:
Just did an interactive rebase as suggested by @VelocityRa, since it was pretty simple to do. @pent0

src/emulator/init.cpp
```diff
@@ -118,11 +105,11 @@ bool init(HostState &state, Config cfg) {
     if (state.cfg.wait_for_debugger) {
         state.kernel.wait_for_debugger = state.cfg.wait_for_debugger.value();
     }
-    state.base_path = base_path.get();
+    state.base_path = root_paths.get_base_path_string();
 
     // If configuration already provides preference path
-    if (state.cfg.pref_path.empty()) {
-        state.pref_path = pref_path.get();
+    if (!state.cfg.pref_path.empty()) {
```
Zangetsu38: wrong here, remove "!", like that your remplace value inside pref_path if user use external folder.

totlmstr: Done. That was an spot-check failure on my part. *doh*

VelocityRa:
This will be merged soon, fyi @petmac this will have conflicts with your refactoring.

totlmstr:
Just to leave a few notes, since the last commit of the refactoring is rather large:

* in `config.cpp`, several lines had to be rearranged in order to capture the command line. Specifically, `lle-modules` had to be pushed down

* This also adds in a few flags so that now a `config.yml` (or whatever name is given)  can now be loaded to Vita3K

Some additional work was done in `io` for safety reasons since Boost.Filesystem is sensitive to paths that don't exist.

src/emulator/gui/src/settings_dialog.cpp
```diff
@@ -70,10 +94,21 @@ void draw_settings_dialog(HostState &host) {
             ImGui::SetTooltip("Uncheck the box to disable texture cache.");
         ImGui::Spacing();
         ImGui::PushItemWidth(400);
-        ImGui::InputTextWithHint("Set emulated system storage folder \n(Reboot after change for apply)", "Write your path folder here", &host.cfg.pref_path);
+        ImGui::InputTextWithHint("Set emulated system storage folder.", "Write your path folder here", &host.cfg.pref_path);
```
pent0: Redo your changes here

totlmstr: Made it a tooltip for the Apply button instead.

And, since it was bugging me, moved some of the other long lines to tooltips as well.

PR #476, opened 2019-05-11
Gui/settings: fix initialize value for news config.

Zangetsu38:
fix issue with get wrong default value on news config file created.

### DIFF
--- a/src/emulator/config.cpp
+++ b/src/emulator/config.cpp
@@ -143,7 +143,7 @@ static bool deserialize(Config &cfg) {
     get_yaml_value(config_node, "sys-button", &cfg.sys_button, static_cast<int>(SCE_SYSTEM_PARAM_ENTER_BUTTON_CROSS));
     get_yaml_value(config_node, "sys-lang", &cfg.sys_lang, static_cast<int>(SCE_SYSTEM_PARAM_LANG_ENGLISH_US));
     get_yaml_value(config_node, "background-image", &cfg.background_image, std::string{});
-    get_yaml_value(config_node, "background-alpha", &cfg.background_alpha, 0.000f);
+    get_yaml_value(config_node, "background-alpha", &cfg.background_alpha, 0.300f);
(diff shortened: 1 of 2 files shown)


src/emulator/host/include/host/config.h
```diff
@@ -40,11 +40,11 @@ struct Config {
...
-    std::string pref_path;
+    std::string pref_path = {};
     bool archive_log = false;
     bool texture_cache = true;
     optional<bool> wait_for_debugger;
-    std::string background_image;
-    float background_alpha;
-    int icon_size;
+    std::string background_image = {};
+    float background_alpha = 0.000f;
```
pent0: Set default alpha to something that people can see (in mid of not visible and totally visible, 0.5f or 129.0f? I'm not sure)

src/emulator/config.cpp
```diff
@@ -182,12 +182,16 @@ InitResult init(Config &cfg, int argc, char **argv) {
         po::options_description config_desc("Configuration");
         config_desc.add_options()
             ("archive-log,A", po::bool_switch(&cfg.archive_log), "Makes a duplicate of the log file with TITLE_ID and Game ID as title")
+            ("background-alpha,a", po::value(&cfg.background_alpha), "Background Alpha")
```
pent0: I don't think people will specifiy these stuffs in command line, also they doesn't have debug value. Delete those changes (except log_uniforms).

src/emulator/host/include/host/config.h
Zangetsu38: value is already set by default on 0.000f, and this value is for full transparency, but yes, you have right, for game i recommande 0.300 for see game title correctly

pent0: Yeah

src/emulator/config.cpp
Zangetsu38: oki, i have delete

src/emulator/host/include/host/config.h
Zangetsu38: fixed so

PR #478, opened 2019-05-14
Use unicorn as submodule in Unix

dracc:
Spiritual successor of #277 by @muemart which was reverted in #278 because Windows builds were broken.
As no proper fix for the Windows build problems has been found, this interim solution of using a submodule on the platforms in which it works seemed better in my opinion.

As I only have access to Linux, building has not been tested in neither Windows nor macOS. Those ought to be unaffected by these changes though.

### DIFF
--- a/src/external/unicorn-build.sh
+++ b/src/external/unicorn-build.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+set -ex
+
+UNICORN_ARCHS=arm UNICORN_DEBUG=no UNICORN_SHARED=no ./make.sh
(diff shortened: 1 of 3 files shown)


src/external/unicorn-build.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
pent0: Putting this script in external root doesn't look OK to me. Maybe put it in a scripts folder somewhere? It just feel out of place

petmac: I think I agree, but I also don't know of a better place to put it. I'm happy with it here for now, until we think of somewhere better.

dracc: I agree that the positioning of this file is suboptimal. I couldn't figure out a better spot though. :(

PR #485, opened 2019-05-21
shader/translator: Correct num field length of components in vbw

pent0:


pent0:
Need more investigation

PR #490, opened 2019-05-24
Attempt to fix main audio underflow bug

petmac:
When a game attempts to output audio, I put it to the port's stream immediately rather than deferring that to the audio callback.

This avoids underflow when the game's thread doesn't wake up fast enough to output more audio.

Additionally, I only put the thread to sleep if the audio isn't going to finish playing in the next audio callback. This means the calling thread generally won't stop if submitting audio in small chunks.

I then increased the SDL chunk size, as it was popping a bit on my Mac.

### DIFF
--- a/src/emulator/audio/include/audio/state.h
+++ b/src/emulator/audio/include/audio/state.h
@@ -21,7 +21,6 @@
 #include <map>
 #include <memory>
 #include <mutex>
-#include <queue>
 #include <vector>
 
 #include <psp2/types.h>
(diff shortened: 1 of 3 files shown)


src/emulator/audio/src/audio.cpp
```diff
@@ -103,7 +92,7 @@ bool init(AudioState &state, ResumeAudioThread resume_thread) {
     desired.freq = 48000;
     desired.format = AUDIO_S16LSB;
     desired.channels = 2;
-    desired.samples = 512;
+    desired.samples = 1024;
```
VelocityRa: This should ideally be configurable right? Unrelated to this PR, just an observation.

I remember I had to do this for decaf (https://github.com/decaf-emu/decaf-emu/pull/397) since games behaved differently on different hardware and there's a trade-off between audio latency/stuttering.

src/emulator/audio/include/audio/state.h
```diff
@@ -31,28 +30,19 @@
...
-    SceUID thread = -1;
-};
-
 struct ReadOnlyAudioOutPortState {
     int len_bytes = 0;
-};
-
-struct AudioCallbackOutPortState {
-    AudioStreamPtr stream;
+    int freq = 0;
```
Rinnegatamante: Is this used anywhere?

Zangetsu38:
look in audio.cpp you have added space empty in 37/48 line ^^
run format fix that
But good works again :)

src/emulator/audio/src/audio.cpp
petmac: This is a good shout. I split that change into a separate commit and added a note to the commit message.

src/emulator/audio/include/audio/state.h
petmac: No, thanks for noticing it! I put it in when generating sine waves for debugging, and forgot to take it out.

petmac:
> look in audio.cpp you have added space empty in 37/48 line ^^
> run format fix that

Thanks, fixed.

Rinnegatamante:
when merging, remember to change commit name by including "audio:" at the start of the commit name.

PR #498, opened 2019-05-29
gui/renderer: Shader debugging utilities

VelocityRa:
image

### DIFF
--- a/src/emulator/gui/src/shader_editor_dialog.cpp
+++ b/src/emulator/gui/src/shader_editor_dialog.cpp
@@ -1,0 +1,144 @@
+// Vita3K emulator project
+// Copyright (C) 2018 Vita3K team
+//
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation; either version 2 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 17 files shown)


src/emulator/gui/include/gui/functions.h
```diff
@@ -42,3 +43,17 @@ void draw_game_selector(HostState &host);
 void draw_reinstall_dialog(HostState &host, GenericDialogState *status);
 
 } // namespace gui
+
+// Extensions to ImGUi
+namespace ImGui {
+
+static auto vector_getter = [](void *vec, int idx, const char **out_text) {
```
pent0: What's the point of this being a lambda.

VelocityRa: oh yea it was originally inside a func, forgot to make it a normal func when it was moved out

src/emulator/gui/src/shader_editor_dialog.cpp
```diff
@@ -0,0 +1,165 @@
...
+#include <stdexcept>
+
+#define EXTERNAL_DISASM_ENABLE 0
+#define EXTERNAL_DISASM_PATH ""
+#define EXTERNAL_DISASM_ARGS ""
+
+namespace util {
+
+template <size_t MaxOutSize>
+std::string exec(std::string cmd) {
```
pent0:
```suggestion
std::string exec(const std::string &cmd) {
```

I want to put this in utils. Don't want cross-platform code hanging around in gui.

pent0: You can still put this here, im just making suggestion

VelocityRa: yea fair enough was just being lazy tbh

PR #499, opened 2019-05-29
fixed the loading of the font

BogdanTheGeek:
Passed the base_path of the app to init_font in order to make the fonts load when the app is loaded from any location.

### DIFF
--- a/src/emulator/gui/src/gui.cpp
+++ b/src/emulator/gui/src/gui.cpp
@@ -114,13 +114,13 @@ static GLuint load_texture(int32_t width, int32_t height, const unsigned char *d
     return texture;
 }
 
-static void init_font(State &gui) {
-    const auto DATA_PATH = "data";
-    const auto FONT_PATH = "fonts";
+static void init_font(HostState &host) {
+    const auto DATA_DIRNAME = "data";
+    const auto FONT_DIRNAME = "fonts";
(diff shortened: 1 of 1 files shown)


src/emulator/gui/src/gui.cpp
```diff
@@ -114,8 +114,8 @@ static GLuint load_texture(int32_t width, int32_t height, const unsigned char *d
     return texture;
 }
 
-static void init_font(State &gui) {
-    const auto DATA_PATH = "data";
+static void init_font(HostState &host) {
+    const auto DATA_PATH = host.base_path + "data";
```
VelocityRa: `DATA_PATH` is poorly named, it was supposed to be the name of the data directory
please rename this to `DATA_DIRNAME`, same for `FONT_PATH` below

so remove this change and add the base path below like so:
```cpp
    fs::path font_dir = fs::path(host.base_path) /= fs::path(DATA_PATH) /= FONT_PATH;
```

Ideally we should have a `data_path` in `host` like we have `base_path` and `pref_path` now, but that's not in the scope of this PR I suppose.

BogdanTheGeek: Ok, I did that

BogdanTheGeek: sorry for not using the correct naming the first time.

PR #501, opened 2019-05-31
Multi-touch support for enabled devices

BogdanTheGeek:


src/emulator/gui/src/controls_dialog.cpp
```diff
@@ -25,7 +25,7 @@ namespace gui {
 
 void draw_controls_dialog(HostState &host) {
     float width = ImGui::GetWindowWidth() / 1.35f;
-    float height = ImGui::GetWindowHeight() / 1.35f;
+    float height = ImGui::GetWindowHeight() / 1.25f;
```
pent0: Is there any explaination for this

src/emulator/modules/SceTouch/SceTouch.cpp
```diff
@@ -24,14 +24,95 @@
 
 // TODO Move elsewhere.
 static uint64_t timestamp;
+SDL_TouchFingerEvent finger_buffer[8];
+uint8_t finger_count = 0;
+long int finger_id_ref;
+auto touchscreen_port = SCE_TOUCH_PORT_FRONT;
+
+SceTouchData recover_touch_events() {
```
pent0: For function that stays on this cpp file and not accessible on other cpp files, use ``static`` modifier

src/emulator/modules/SceTouch/SceTouch.cpp
```diff
@@ -24,14 +24,95 @@
...
+    SceTouchData touch_data;
+    touch_data.timeStamp = timestamp;
+
+    for (int i = 0; i < finger_count; i++) {
```
pent0: This can create a signed/unsigned comparison warning.

```suggestion
    for (uint8_t i = 0; i < finger_count; i++) {
```

src/emulator/modules/SceTouch/SceTouch.cpp
```diff
@@ -24,14 +24,95 @@
...
+            }
+        }
+        break;
+    }
+    }
+
+    return 0;
+}
+
+bool registered_touch() {
```
pent0: ditto

src/emulator/modules/SceTouch/SceTouch.cpp
```diff
@@ -64,6 +149,10 @@ static int peek_touch(HostState &host, SceUInt32 port, SceTouchData *pData, SceU
         }
     }
 
+    else if (registered_touch() == true && port == touchscreen_port) {
```
pent0: Weird newline

src/emulator/modules/SceTouch/SceTouch.h
```diff
@@ -17,8 +17,12 @@
 
 #pragma once
 
+#include <SDL.h>
 #include <module/module.h>
 
+int handle_touch_event(SDL_TouchFingerEvent finger);
```
pent0: Try to pass the parameter as reference. and foward declare it. That way SDL include is unneccessary

```suggestion
struct SDL_TouchFingerEvent;
typedef SDL_TouchFingerEvent SDL_TouchFingerEvent;

int handle_touch_event(SDL_TouchFingerEvent &finger);
```

Same with source. This will save some times btw (if you dont pass it as reference it will copy the data)

pent0:
Also remember to rename your commit message to our preference

src/emulator/gui/src/controls_dialog.cpp
BogdanTheGeek: i made the list longer when i added the toggle so i had to make the window taller

pent0: Right

src/emulator/modules/SceTouch/SceTouch.cpp
BogdanTheGeek: so it should be static SceTouchData recover_touch_events() ?

pent0: Yes that's correct. Google static linkage if you want to find out why i suggest it

src/emulator/modules/SceTouch/SceTouch.cpp
BogdanTheGeek: whats ditto aprat from a pokemon?

pent0: like upper, said before. ditto clones other in some way right?

1whatleytay: I think he means make it `static`.

src/emulator/modules/SceTouch/SceTouch.h
BogdanTheGeek: i don't quite understand this, i get the passing by reference, but not what you did with the structs.
Should i just copy it?

pent0: That's foward declaring, basically you declare the existence of a struct, than filling the struct info later somewhere (in the cpp file for example). If you can foward declare it, you can remove the SDL.h include. You should avoid unneccessary inclusion when you can.

 Sure i still suggest passing by reference

pent0: And sure, try to copy it, and remove including SDL.h in SceTouch.h if it fails, than we will only pass by reference :p

BogdanTheGeek: by just copy it i meant your suggested change

src/emulator/modules/SceTouch/SceTouch.cpp
BogdanTheGeek: peek_touch is always polling so only recover the touches if any touches are registered. also its polling on both the back and front touch panel so i filter it out with port == touchscreen_port so it doesn't touch both simultaneously

pent0: By new line i mean why this exist

image

BogdanTheGeek: i think i have to run format.sh again

BogdanTheGeek: its actually still there, no idea why

src/emulator/modules/SceTouch/SceTouch.cpp
```diff
@@ -19,19 +19,101 @@
...
 #include <psp2/touch.h>
 
+#include <SDL.h>
 #include <SDL_mouse.h>
 #include <SDL_video.h>
 
 // TODO Move elsewhere.
 static uint64_t timestamp;
+SDL_TouchFingerEvent finger_buffer[8];
+uint8_t finger_count = 0;
```
Rinnegatamante: Make those statics too like timestamp var

src/emulator/modules/SceTouch/SceTouch.cpp
```diff
@@ -55,16 +137,22 @@ static int peek_touch(HostState &host, SceUInt32 port, SceTouchData *pData, SceU
 
         if ((touch_pos_viewport.x >= 0) && (touch_pos_viewport.y >= 0) && (touch_pos_viewport.x < 1) && (touch_pos_viewport.y < 1)) {
             pData->report[pData->reportNum].x = static_cast<uint16_t>(touch_pos_viewport.x * 1920);
-            pData->report[pData->reportNum].y = static_cast<uint16_t>(touch_pos_viewport.y * 1088);
+            if (port == SCE_TOUCH_PORT_FRONT) {
+                pData->report[pData->reportNum].y = static_cast<uint16_t>(touch_pos_viewport.y * 1088);
+            } else {
+                pData->report[pData->reportNum].y = static_cast<uint16_t>(108 + touch_pos_viewport.y * (889 - 108));
```
Rinnegatamante: Why 889 - 108? Also what's that 108 value?

src/emulator/modules/SceTouch/SceTouch.cpp
```diff
@@ -55,16 +137,22 @@ static int peek_touch(HostState &host, SceUInt32 port, SceTouchData *pData, SceU
...
+            }
             ++pData->reportNum;
         }
 
         if (!host.ctrl.touch_mode[port]) {
             pData->reportNum = 0;
         }
+    } else if (registered_touch() == true && port == touchscreen_port) {
+        pData[0] = recover_touch_events();
```
Rinnegatamante: i would prefer this function to have a SceTouchData* as argument instead of returning a SceTouchData. But not really required.

src/emulator/modules/SceTouch/SceTouch.cpp
BogdanTheGeek: i left it as 889 - 108 too be easier to read, but i realise now thats an extra calculation each cycle. 108 is minAaY and 889 is maxAaY for the back panel. If you just use the front display sizes the back touchscreen is offset as it gets closer to the edges.

pent0: No extra cycle, the compiler will be smart enough to calculate it at compiletime

BogdanTheGeek: good to know, i did change it though before i read this

Rinnegatamante: parhentesis should be removed if you're going to remove the 889 - 108. Btw it was just curiosity in why we had this 889 and 108 (they looked like some magic random numbers to me), not really a change request.

BogdanTheGeek: aghhh...forgot the parenthesis ill change them now and push

BogdanTheGeek: done

BogdanTheGeek:
Anything else to do before merging ?

PR #487, opened 2019-05-23
shader: implement bitwise-test

1whatleytay:
Some small changes that fix test instructions.

It also adds support for bitwise operations for test instructions, so that's nice.

PR #502, opened 2019-06-01
Momo

pent0:
Original PR description (#487):

```
Some small changes that fix test instructions.

It also adds support for bitwise operations for test instructions, so that's nice.
```

Additions:
- Add an anti-crash around the problem of coord type in vert vs frag has different type. This instead force all coord to be vec4. The frag will only copy what is neccessary in the vec4 passed.
- Rewrite shader bank to a flat array. This makes thing easier
- Implement INDEX and INDEXED bank.
- Correct VMAD2, VTST, VPCK (handling src2) implementation
- Implement CMOV.
- Correctly apply modifiers on all load case.
- Simplify load and store.
- Use spv::Function* instead of spv::Block* for dealing with translated block (aka subroutine), so it's more friendlier (with both driver compiler and spirv-cross)

Known regressions:
- Downwell color looks like Blue channel got swapped for red

Working:
- Downwell (results of INDEXED bank being implemented)

image

- Swapquest (Fog effect)

image

- A virus name tom (Splash)

image

- Run Sackboy! Run (Splash, thanks @IllusionMan1212)

image

- Danganronpa 2 (Thanks for testing @Zangetsu38)

image


src/emulator/gxm/src/gxp.cpp
```diff
@@ -103,13 +103,8 @@ std::string parameter_name_raw(const SceGxmProgramParameter &parameter) {
...
 std::string parameter_name(const SceGxmProgramParameter &parameter) {
     auto full_name = gxp::parameter_name_raw(parameter);
-    auto struct_idx = full_name.find('.');
-    bool is_struct_field = struct_idx != std::string::npos;
-
-    if (is_struct_field)
-        return full_name.substr(struct_idx + 1, full_name.length());
-    else
-        return full_name;
+    std::replace(full_name.begin(), full_name.end(), '.', '_');
```
pent0: This is a simple sanitize. We need to be more careful in the future though.

Like array of struct. We can't only do this.

pent0:
Sorry again, i push a quick fix

Issue #329, opened 2018-09-16
Minimize platform-specific code


VelocityRa:
Since we now have access to `boost::filesystem`, we could convert some platform-specific `io` code to use it.

imWatchful:
I believe this has been done over the years, and the rest of the platform specific io code is mostly unavoidable

PR #505, opened 2019-06-01
io: Refactoring with Classes (#329)

totlmstr:
This PR implements ~~mark two of~~ io refactoring. `io` now uses two classes (one for directories and one for files) as the baseline for interacting with files, combined with [using Better Enums](https://github.com/aantron/better-enums) for manipulation of paths. As a result, *any* Vita path can now be constructed and translated.

~~This is a Work-in-Progress because directories still need to be refactored (traversing a directory WILL cause errors), but testing is needed.~~

### DIFF
--- a/src/emulator/io/include/io/device.h
+++ b/src/emulator/io/include/io/device.h
@@ -1,0 +1,126 @@
+// Vita3K emulator project
+// Copyright (C) 2018 Vita3K team
+//
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation; either version 2 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+#pragma once
+
+#include <io/VitaIoDevice.h>
+
+#include <util/fs.h>
+
+#include <string>
+
+namespace device {
+
+/**
+  * \brief Get a Vita device from a given path.
+  * \param path The input path to be tested.
+  * \return The path's root as an enumeration. Otherwise, VitaIoDevice::_INVALID.
+  */
+inline VitaIoDevice get_device(const std::string &path) {
+    if (path.empty())
+        return VitaIoDevice::_INVALID;
+
+    const auto colon = path.find_first_of(':');
+    if (colon == std::string::npos)
+        return VitaIoDevice::_INVALID;
+
+    const auto p = path.substr(0, colon);
+    const auto dev = VitaIoDevice::_from_string_nocase_nothrow(p.c_str());
+    if (VitaIoDevice::_is_valid(dev))
+        return VitaIoDevice::_from_string(p.c_str());
+
+    return VitaIoDevice::_INVALID;
+}
+
+/**
+  * \brief Get a valid Vita device as a string.
+  * \param dev The input Vita device needed.
+  * \param with_colon Output the string appended with a colon (default: false)
+  * \return A string version of the Vita device.
+  */
+inline std::string get_device_string(const VitaIoDevice dev, const bool with_colon = false) {
+    return with_colon ? std::string(dev._to_string()).append(":") : dev._to_string();
+}
+
+/**
+  * \brief Check if the device is a valid output path.
+  * \param device Input device to be checked.
+  * \return True if valid, False otherwise.
+  */
+inline bool is_valid_output_path(const VitaIoDevice device) {
+    return !(device == VitaIoDevice::savedata0 || device == VitaIoDevice::savedata1 || device == VitaIoDevice::app0
+        || device == VitaIoDevice::_INVALID || device == VitaIoDevice::addcont0 || device == VitaIoDevice::tty0
+        || device == VitaIoDevice::tty1);
+}
+
+/**
+  * \brief Check if the device string is valid.
+  * \param device Input device to be checked.
+  * \return True if valid, False otherwise.
+  */
+inline bool is_valid_output_path(const std::string &device) {
+    return !(device == (+VitaIoDevice::savedata0)._to_string() || device == (+VitaIoDevice::savedata1)._to_string() || device == (+VitaIoDevice::app0)._to_string()
+        || device == (+VitaIoDevice::_INVALID)._to_string() || device == (+VitaIoDevice::addcont0)._to_string() || device == (+VitaIoDevice::tty0)._to_string()
+        || device == (+VitaIoDevice::tty1)._to_string());
+}
+
+/**
+  * \brief Construct a normalized path (optionally with an extension) to be outputted onto the Vita.
+  * \param dev The input Vita device.
+  * \param path The Vita location needed.
+  * \param ext The extension of the file (optional).
+  * \return An std::string of a real Vita translated path.
+  */
+std::string construct_normalized_path(VitaIoDevice dev, const std::string &path, const std::string &ext = "");
+
+/**
+  * \brief Remove a device string from the path, and optionally prepend it with a different string.
+  * \param path Input path to be modified.
+  * \param device The device to be removed.
+  * \param mod_path The new path to prepend the path (optional).
+  * \return The string without the device, normalized.
+  */
+std::string remove_device_from_path(const std::string &path, VitaIoDevice device, const std::string &mod_path = "");
+
+/**
+  * \brief Used for removing an extra device from a path. (Note: Not intended to be a permanent solution.)
+  * \param path The input Vita path.
+  * \param device The device to be removed.
+  * \return New path without the duplicate device.
+  */
+inline std::string remove_duplicate_device(const std::string &path, VitaIoDevice &device) {
+    auto cur_path = remove_device_from_path(path, device);
+    if (get_device(cur_path) != VitaIoDevice::_INVALID) {
+        device = get_device(cur_path);
+        return cur_path;
+    }
+
+    return path;
+}
+
+/**
+  * \brief Construct the emulated Vita path (optionally with an extension).
+  * \param dev The input Vita device used.
+  * \param path The path of the file.
+  * \param base_path The main output for the file.
+  * \param ext The extension of the file (optional).
+  * \return A complete Boost.Filesystem path normalized.
+  */
+inline fs::path construct_emulated_path(const VitaIoDevice dev, const fs::path &path, const std::string &base_path, const std::string &ext = "") {
+    return fs_utils::construct_file_name(base_path, get_device_string(dev, false), path, ext);
+}
+} // namespace device
--- a/src/emulator/io/include/io/filesystem.h
+++ b/src/emulator/io/include/io/filesystem.h
@@ -1,0 +1,74 @@
+// Vita3K emulator project
+// Copyright (C) 2018 Vita3K team
+//
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation; either version 2 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+#pragma once
+
+#include <util/fs.h>
+
+#ifdef WIN32
+#include <util/string_utils.h>
+#endif
+
+#include <dirent.h>
+
+typedef std::shared_ptr<FILE> FilePtr;
+
+// For opening Boost.Filesystem files, Boost returns wide strings for Windows, normal strings for other OS
+// Dirent and FILE only accept and return wide char strings for Windows, and normal for other OS
+#ifdef WIN32
+const wchar_t *translate_open_mode(const int flags);
+
+inline FilePtr create_shared_file(const fs::path &path, const int open_mode) {
+    const auto file = _wfopen(path.generic_path().wstring().c_str(), translate_open_mode(open_mode));
+    return file ? FilePtr(file, std::fclose) : FilePtr();
+}
+
+typedef std::shared_ptr<_WDIR> DirPtr;
+
+inline DirPtr create_shared_dir(const fs::path &path) {
+    return DirPtr(_wopendir(path.generic_path().wstring().c_str()), _wclosedir);
+}
+
+inline std::string get_file_in_dir(const _wdirent *file) {
+    return string_utils::wide_to_utf(file->d_name);
+}
+
+inline _wdirent *get_system_dir_ptr(const DirPtr &dir) {
+    return _wreaddir(dir.get());
+}
+#else
+const char *translate_open_mode(const int flags);
+
+inline FilePtr create_shared_file(const fs::path &path, const int open_mode) {
+    const auto file = fopen(path.generic_path().string().c_str(), translate_open_mode(open_mode));
+    return file ? FilePtr(file, std::fclose) : FilePtr();
+}
+
+typedef std::shared_ptr<DIR> DirPtr;
+
+inline DirPtr create_shared_dir(const fs::path &path) {
+    return DirPtr(opendir(path.generic_path().string().c_str()), closedir);
+}
+
+inline std::string get_file_in_dir(dirent *file) {
+    return std::string(file->d_name);
+}
+
+inline dirent *get_system_dir_ptr(DirPtr dir) {
+    return readdir(dir.get());
+}
+#endif
--- a/src/emulator/io/include/io/state.h
+++ b/src/emulator/io/include/io/state.h
@@ -17,51 +17,83 @@
 
 #pragma once
 
-#include <miniz.h>
+#include <io/filesystem.h>
+#include <io/util.h>
 
-#include <io/types.h>
-
-#include <cstdio>
 #include <map>
-#include <memory>
 
-#ifdef WIN32
-struct _WDIR;
-#else
-#include <dirent.h>
-#endif
+// Class for all needed information to access files on Vita3K.
+class FileStats : public VitaStats {
+    // Shared file pointer
+    FilePtr wrapped_file;
 
-typedef std::shared_ptr<FILE> FilePtr;
-typedef std::shared_ptr<mz_zip_archive> ZipPtr;
-typedef std::shared_ptr<mz_zip_reader_extract_iter_state> ZipFilePtr;
+public:
+    // Constructor used for files
+    // Based on https://codereview.stackexchange.com/questions/4679/
+    explicit FileStats(const char *vita, const std::string &t, const fs::path &file, const int open) {
+        wrapped_file = create_shared_file(file, open);
 
-#ifdef _WIN32
-typedef std::shared_ptr<_WDIR> DirPtr;
-#else
-typedef std::shared_ptr<DIR> DirPtr;
-#endif
+        file_info.vita_loc = vita;
+        file_info.translated = t;
+        file_info.sys_loc = file;
+        file_info.open_mode = open;
+        file_info.file_mode = SCE_SO_IFREG | SCE_SO_IROTH;
+        file_info.access_mode = SCE_S_IFREG;
+    }
 
-enum TtyType : std::uint8_t {
-    TTY_IN = 0b01,
-    TTY_OUT = 0b10,
-    TTY_INOUT = TTY_IN | TTY_OUT
+    bool is_regular_file() const {
+        return file_info.file_mode & SCE_SO_IFREG;
+    }
+
+    // Check if the file is writable
+    bool can_write_file() const {
+        if (!is_regular_file())
+            return false;
+
+        return can_write(file_info.open_mode);
+    }
+
+    // File operations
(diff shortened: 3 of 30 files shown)


totlmstr:
Refactoring is now finished.

Needs testing!

pent0:
Tell me when this got ready to review. If we let it here too long there might be more merge conflicts

totlmstr:
@pent0 PR is now ready for review. Testing gave the green light.

pent0:
Nice document, I will give it more look later

src/emulator/io/include/io/state.h
```diff
@@ -41,27 +42,177 @@ typedef std::shared_ptr<_WDIR> DirPtr;
...
+#ifdef WIN32
+    _wdirent *get_dir_ptr() const {
+        return _wreaddir(dir_ptr.get());
+    }
+
+    const wchar_t *get_char_path() const {
+        return dir_info.sys_loc.generic_path().wstring().c_str();
+    }
+#else
+    dirent *get_dir_ptr() const {
```
pent0: This broke the abstraction? Consider making interface and implementations model

src/emulator/io/include/io/state.h
```diff
@@ -17,13 +17,14 @@
...
+#include <io/vfs.h>
 
-#include <io/types.h>
+#include <util/fs.h>
 
-#include <cstdio>
+#include <dirent.h>
 #include <map>
-#include <memory>
+#include <miniz.h>
```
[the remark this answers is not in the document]
totlmstr: It turns out that `miniz.h` is used over in `host` for `load_self`, but the shared pointers for the zip are only used in `interface` at emulator. When looking over it, I feel as if the `load_self` function itself should be broken down into smaller functions and `miniz` removed from `host` as a dependency for cleanliness, but that is outside the scope of this PR.

totlmstr:
Please review again, @pent0.

src/emulator/io/include/io/state.h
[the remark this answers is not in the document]
pent0: Okay

src/emulator/io/include/io/state.h
pent0: point still stands

src/emulator/io/include/io/io.h
```diff
@@ -17,27 +17,8 @@
...
-
-constexpr const char *get_device_string(VitaIoDevice dev, bool with_colon = false);
-bool read_file(VitaIoDevice device, FileBuffer &buf, const std::string &pref_path, const fs::path &vfs_file_path);
-bool read_app_file(FileBuffer &buf, const std::string &pref_path, const std::string &title_id, const fs::path &vfs_file_path);
-} // namespace vfs
+constexpr int SCE_ERROR_ERRNO_ENOENT = 0x80010002; // Associated file or directory does not exist
+constexpr int SCE_ERROR_ERRNO_EEXIST = 0x80010011; // File exists
+constexpr int SCE_ERROR_ERRNO_EMFILE = 0x80010018; // Too many files are open
+constexpr int SCE_ERROR_ERRNO_EBADFD = 0x80010051; // File descriptor is invalid for this operation
+constexpr int SCE_ERROR_UNKNOWN = 0xFFFFFFFF; // Unknown Error
```
Rinnegatamante: does this error really exists?

src/emulator/modules/SceAppUtil/SceAppUtil.cpp
```diff
@@ -149,40 +150,37 @@ EXPORT(int, sceAppUtilResetCookieWebBrowser) {
...
-            close_file(host.io, fd, export_name);
-            break;
         case SCE_APPUTIL_SAVEDATA_DATA_SAVE_MODE_DIRECTORY:
-            create_dir(host.io, file_path.c_str(), 0777, host.pref_path.c_str(), export_name);
+            create_dir(host.io, file_path, 0777, host.pref_path, export_name);
             break;
+        case SCE_APPUTIL_SAVEDATA_DATA_SAVE_MODE_FILE:
         default:
-            fd = open_file(host.io, file_path, SCE_O_WRONLY | SCE_O_CREAT, host.pref_path.c_str(), export_name);
+            fd = open_file(host.io, file_path, SCE_O_WRONLY, host.pref_path, export_name);
```
Rinnegatamante: why SCE_O_CREAT disappeared?

src/emulator/modules/SceAppUtil/SceAppUtil.cpp
```diff
@@ -211,28 +206,19 @@ EXPORT(int, sceAppUtilSaveDataMount) {
 }
 
 EXPORT(int, sceAppUtilSaveDataSlotCreate, unsigned int slotId, SceAppUtilSaveDataSlotParam *param, SceAppUtilSaveDataMountPoint *mountPoint) {
-    std::string slot_path = "savedata0:/SlotParam_";
-    slot_path += std::to_string(slotId);
-    slot_path += ".bin";
-    SceUID fd = open_file(host.io, slot_path, SCE_O_WRONLY | SCE_O_CREAT, host.pref_path.c_str(), export_name);
+    const auto fd = open_file(host.io, construct_slotparam_path(slotId), SCE_O_CREAT, host.pref_path, export_name);
```
Rinnegatamante: ditto but for WRITE.

src/emulator/io/src/state_functions.cpp
```diff
@@ -0,0 +1,64 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+#include <io/state.h>
+
+SceOff FileStats::read(void *input_data, const int element_size, const SceSize element_count) const {
+    if (!fs::exists(file_info.sys_loc))
```
1whatleytay: I feel like it would be more appropriate to check the file pointer, since you created that right off the bat.

src/emulator/io/src/state_functions.cpp
```diff
@@ -0,0 +1,64 @@
...
+        return -1;
+
+    return fread(input_data, element_size, element_count, wrapped_file.get());
+}
+
+SceOff FileStats::write(const void *data, const SceSize size, const int count) const {
+    if (!can_write_file())
+        return -1;
+
+    if (!fs::exists(file_info.sys_loc.parent_path()))
```
1whatleytay: Why is this different from the checks in read and seek?

src/emulator/modules/SceAppUtil/SceAppUtil.cpp
```diff
@@ -149,40 +150,37 @@ EXPORT(int, sceAppUtilResetCookieWebBrowser) {
     return UNIMPLEMENTED();
 }
 
-EXPORT(int, sceAppUtilSaveDataDataRemove, emu::SceAppUtilSaveDataFileSlot *slot, emu::SceAppUtilSaveDataRemoveItem *files, unsigned int fileNum, SceAppUtilSaveDataMountPoint *mountPoint) {
+const char *construct_savedata0_path(const char *data, const char *ext = "") {
+    return device::construct_normalized_path(VitaIoDevice::savedata0, data, ext);
+}
+
+const char *construct_slotparam_path(const unsigned int data) {
+    return construct_savedata0_path(std::string("SlotParam_" + std::to_string(data)).c_str(), "bin");
```
1whatleytay: Calling `c_str` on a string that's about to go out of scope is undefined behaviour, and I **think** that's what's happening here.

src/emulator/modules/SceLibKernel/SceLibKernel.cpp
```diff
@@ -357,26 +351,26 @@ EXPORT(int, sceIoDevctlAsync) {
 }
 
 EXPORT(int, sceIoDopen, const char *dir) {
-    return open_dir(host.io, dir, host.pref_path.c_str(), export_name);
+    return open_dir(host.io, dir, host.pref_path, export_name);
 }
 
-EXPORT(int, sceIoDread, SceUID fd, emu::SceIoDirent *dir) {
```
1whatleytay: I think `emu::` is correct here.

src/emulator/modules/SceLibKernel/SceLibKernel.cpp
```diff
@@ -1092,36 +1086,36 @@ EXPORT(int, sceKernelGetTimerTime) {
...
         }
 
         if (seek_file(file, 0, SCE_SEEK_SET, host.io, export_name) < 0) {
-            error_val = RET_ERROR(size);
+            error_val = RET_ERROR(static_cast<int>(size));
             return false;
         }
 
-        char *data = new char[size];
+        auto *data = new char[static_cast<int>(size) + 1]; // null-terminated char array
```
1whatleytay: Why cast here?

src/emulator/io/include/io/device.h
```diff
@@ -0,0 +1,141 @@
...
+#include <string>
+
+namespace device {
+
+/**
+  * \brief Get a Vita device from a given path.
+  * \param path The input path to be tested.
+  * \return The path's root as an enumeration. Otherwise, VitaIoDevice::_INVALID.
+  */
+inline VitaIoDevice get_device(const std::string &path) {
```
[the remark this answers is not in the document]
totlmstr: See above resolved with pent0.

Some functions were (later) moved for the compiler.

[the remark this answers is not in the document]
1whatleytay: Ahh, are you talking about https://github.com/Vita3K/Vita3K/pull/505#discussion_r305556235? Why only some and not all?

src/emulator/modules/SceAppUtil/SceAppUtil.cpp
totlmstr: That is actually resolved with [this](https://github.com/totlmstr/Vita3K/blob/io-paths/src/emulator/util/src/util.cpp#L161) function mentioned in `device::construct_normalized_path`.

src/emulator/modules/SceLibKernel/SceLibKernel.cpp
totlmstr: I removed the `emu::` here in favor of `vita-headers`' definition of SceIoDirent.

src/emulator/modules/SceLibKernel/SceLibKernel.cpp
totlmstr: Partially for logging consistency.

Other lines for size were already casted to int for RET_ERROR, so it would make sense to have this line with size as casted to int as well.

src/emulator/io/include/io/device.h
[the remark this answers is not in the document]
totlmstr: Part lessening clutter, part class definition.

I treated the `device` namespace as a pseudo-class for `VitaIoDevice` (like member functions). This pretty much meant having short, compiler-friendly, fast functions as `inline`, and having them as such will be beneficial in the long run since they are faster (e.g., having `get_device` as inline will be really helpful in the long run due to all of the device path logic it does).

src/emulator/io/include/io/filesystem.h
```diff
@@ -0,0 +1,73 @@
...
+#include <util/string_utils.h>
+#endif
+
+#include <dirent.h>
+
+typedef std::shared_ptr<FILE> FilePtr;
+
+// For opening Boost.Filesystem files, Boost returns wide strings for Windows, normal strings for other OS
+#ifdef WIN32
+const wchar_t *translate_open_mode(const int flags);
```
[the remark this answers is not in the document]
totlmstr: While I do think that this might be a better solution for the char situation we have here, this would be better suited for a different PR.

src/emulator/modules/SceLibKernel/SceLibKernel.cpp
1whatleytay: I don't think you should. The only difference between the `emu::` struct and vita-header's struct is that the pointer `d_private` is of correct size (4 bytes instead of 8).

src/emulator/io/include/io/device.h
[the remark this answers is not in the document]
1whatleytay: I mean, the cost of calling the function is going to be just a few cycles, plus I don't know how often they are really going to be called. I don't think they really need to be inline imo. Up to you though. If you think it's good, it's good.

src/emulator/io/include/io/filesystem.h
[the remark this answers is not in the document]
1whatleytay: Do you want me to make a pr to your branch and change them? I don't think it's that much work.

src/emulator/modules/SceAppUtil/SceAppUtil.cpp
1whatleytay: Nice. Is the result being freed later though?

src/emulator/modules/SceLibKernel/SceLibKernel.cpp
1whatleytay: Okay :)

1whatleytay:
Ahhh! I wish there was just a commit with all the fixes I could look at. I'll review tomorrow, but in the meantime can you fix conflicts so we can get appveyor running?

src/emulator/modules/SceAppUtil/SceAppUtil.cpp
totlmstr: I ended up remodifying the function because it was being freed. Now, it is a `vector<char>` (which is recommended).

src/emulator/modules/SceLibKernel/SceLibKernel.cpp
totlmstr: Confirmed that up on Discord (wanted a second opinion), so I reverted the changes.

totlmstr:
[Despite being correct in the implementation](https://github.com/Vita3K/Vita3K/commit/7967ed20086f4568a9d706215ef5fcbdce24e703#diff-ce79b55e74ed0c96f7346b1396e79040R160) with `vector<char>`, I was getting unreliable results in further testing with the normalized path (garbage results, even though AppVeyor compiled it correctly and the path there was fine). As such, I removed the conversion in favor of just keeping the path as a string.

For a simple idea on how to create paths for `io`, [the now-rebased `context.cpp` provides one idea](https://github.com/Vita3K/Vita3K/pull/505/commits/660c43c75504e6648288eaa2f703d9e012d52829#diff-0d5d80e10f400f4abb7494fcb5c8fce5R291).

One more review from both @1whatleytay and @Rinnegatamante should be enough to merge and close this PR.

src/emulator/io/src/io.cpp
```diff
@@ -390,333 +288,236 @@ SceOff seek_file(SceUID fd, SceOff offset, int whence, IOState &io, const char *
...
-        }
-
-        if (S_ISREG(sb.st_mode)) {
-            statp->st_mode |= SCE_S_IFREG;
-        } else if (S_ISDIR(sb.st_mode)) {
-            statp->st_mode |= SCE_S_IFDIR;
-        }
-
-        statp->st_size = sb.st_size;
+    struct stat sb;
```
1whatleytay: Why `struct` in C++? T_T

src/emulator/io/src/io.cpp
```diff
@@ -390,333 +288,236 @@ SceOff seek_file(SceUID fd, SceOff offset, int whence, IOState &io, const char *
...
     __RtcTicksToPspTime(&statp->st_mtime, last_modification_time_ticks);
     __RtcTicksToPspTime(&statp->st_ctime, creation_time_ticks);
 
     return 0;
 }
 
-int stat_file_by_fd(IOState &io, const int fd, SceIoStat *statp, const char *pref_path, uint64_t base_tick, const char *export_name) {
-    assert(statp != NULL);
+int stat_file_by_fd(IOState &io, const SceUID fd, SceIoStat *statp, const std::string &pref_path, SceUInt64 base_tick, const char *export_name) {
+    assert(statp != nullptr);
```
1whatleytay: I would probably write just `assert(!statp);` but if the original writer was doing this to be more clear I understand.

src/emulator/io/src/state_functions.cpp
```diff
@@ -0,0 +1,61 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+#include <io/state.h>
+
+SceOff FileStats::read(void *input_data, const int element_size, const SceSize element_count) const {
+    if (!get_file_pointer())
```
1whatleytay: Why use the get method when you have access to `wrapped_file` directly?

src/emulator/modules/SceAppUtil/SceAppUtil.cpp
```diff
@@ -149,40 +150,37 @@ EXPORT(int, sceAppUtilResetCookieWebBrowser) {
     return UNIMPLEMENTED();
 }
 
-EXPORT(int, sceAppUtilSaveDataDataRemove, emu::SceAppUtilSaveDataFileSlot *slot, emu::SceAppUtilSaveDataRemoveItem *files, unsigned int fileNum, SceAppUtilSaveDataMountPoint *mountPoint) {
+static std::string construct_savedata0_path(const std::string &data, const char *ext = "") {
+    return device::construct_normalized_path(VitaIoDevice::savedata0, data, ext);
+}
+
+static std::string construct_slotparam_path(const unsigned int data) {
+    return construct_savedata0_path(std::string("SlotParam_" + std::to_string(data)), "bin");
```
1whatleytay: Is `std::string` here redundant?

src/emulator/modules/SceAppUtil/SceAppUtil.cpp
```diff
@@ -149,40 +150,37 @@ EXPORT(int, sceAppUtilResetCookieWebBrowser) {
...
+}
+
+EXPORT(int, sceAppUtilSaveDataDataRemove, SceAppUtilSaveDataFileSlot *slot, emu::SceAppUtilSaveDataRemoveItem *files, unsigned int fileNum, SceAppUtilSaveDataMountPoint *mountPoint) {
```
1whatleytay: Ahh! `emu::` missing here too! You can use psp2 header if the struct does not contain a pointer or a struct that contains a pointer (because the sizes will be wrong). SceAppUtilSaveDataFileSlot has a pointer :/

src/emulator/modules/SceAppUtil/SceAppUtil.cpp
```diff
@@ -149,40 +150,37 @@ EXPORT(int, sceAppUtilResetCookieWebBrowser) {
...
-        slot_path += std::to_string(slot->id);
-        slot_path += ".bin";
-        remove_file(host.io, slot_path.c_str(), host.pref_path.c_str(), export_name);
+        remove_file(host.io, construct_slotparam_path(slot->id).c_str(), host.pref_path, export_name);
     }
     return 0;
 }
 
-EXPORT(int, sceAppUtilSaveDataDataSave, emu::SceAppUtilSaveDataFileSlot *slot, emu::SceAppUtilSaveDataFile *files, unsigned int fileNum, SceAppUtilSaveDataMountPoint *mountPoint, SceSize *requiredSizeKB) {
+EXPORT(int, sceAppUtilSaveDataDataSave, SceAppUtilSaveDataFileSlot *slot, emu::SceAppUtilSaveDataFile *files, unsigned int fileNum, SceAppUtilSaveDataMountPoint *mountPoint, SceSize *requiredSizeKB) {
```
1whatleytay: Here too.

src/emulator/np/src/trophy/trp_parser.cpp
```diff
@@ -19,6 +19,8 @@
 #include <np/trophy/trp_parser.h>
 #include <util/bytes.h>
 
+#include <cstring>
```
1whatleytay: Why is this header included?

pent0:
Could you also take #548 . It's ok if you decline though, but this amy suit for this PR

src/emulator/np/src/trophy/trp_parser.cpp
totlmstr: `np` had a compile error (related to strnicmp, iirc) and wanted this include. It wasn't previously included before, so it was somewhat unexpected.

1whatleytay: Weird. Okay.

src/emulator/io/src/io.cpp
totlmstr: This line is a carry-over from before my PR.

[IEEE specification for getting file status in POSIX](http://pubs.opengroup.org/onlinepubs/009695399/functions/stat.html).

src/emulator/io/src/state_functions.cpp
```diff
@@ -0,0 +1,61 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+#include <io/state.h>
+
+SceOff FileStats::read(void *input_data, const int element_size, const SceSize element_count) const {
+    if (!wrapped_file.get())
```
1whatleytay: You don't even need to use `.get()` :)

src/emulator/io/src/io.cpp
1whatleytay: That's fair but I'd rather you changed it now that you've found it. T_T

src/emulator/io/src/io.cpp
```diff
@@ -357,30 +257,28 @@ int write_file(SceUID fd, const void *data, SceSize size, const IOState &io, con
...
+        const auto written = file->second.write(data, 1, size);
+        LOG_TRACE("{}: Writing to fd: {}, size: {}", export_name, log_hex(fd), size);
+        return static_cast<int>(written);
     }
 
     return IO_ERROR(SCE_ERROR_ERRNO_EBADFD);
 }
 
-SceOff seek_file(SceUID fd, SceOff offset, int whence, IOState &io, const char *export_name) {
-    assert((whence == SCE_SEEK_SET) || (whence == SCE_SEEK_CUR) || (whence == SCE_SEEK_END));
```
Rinnegatamante: Quite sure when whence is incorrect, a different error than SCE_ERROR_ERRNO_EBADFD is returned.

pent0:
Nice job.

PR #511, opened 2019-06-05
modules: Add new NIDs

1whatleytay:
It has been a well known fact that Vita3K has not had that many NIDs (which is absolutely horrible). However, this PR aims to fix that.

I have told robots to do the following:
 - Update vita-headers to the most recent version.
 - Run gen-modules and weave the new changes into our current implementations.
 - Create an `emu::SceGxmTexture` because the old one got totally broken.

I would appreciate any feedback you would like to give my robots. They worked hard to get this pr done- but they occasionally miss some things. Dropwell at least boots to the menu, so I don't think they broke anything.


src/emulator/renderer/src/texture_cache.cpp
```diff
@@ -94,10 +94,14 @@ void cache_and_bind_texture(TextureCacheState &cache, const SceGxmTexture &gxm_t
     const TextureCacheHash hash = hash_texture_data(gxm_texture, mem);
 
     // Try to find GXM texture in cache.
-    const TextureCacheGxmTextures::const_iterator gxm_begin = cache.gxm_textures.cbegin();
-    const TextureCacheGxmTextures::const_iterator gxm_end = gxm_begin + cache.used;
-    const TextureCacheGxmTextures::const_iterator cached_gxm_texture = std::find(gxm_begin, gxm_end, gxm_texture);
-    if (cached_gxm_texture == gxm_end) {
+    size_t cached_gxm_texture_index = -1;
+    for (size_t a = 0; a < cache.used; a++) {
```
pent0: Reason for changing from std::find

1whatleytay: Explained in #developement but for some reason the == operator got deleted. I couldn't compare both structs directly, so it failed to compile.

pent0:
Thank you very much :)

PR #514, opened 2019-06-05
Import call logging blacklist & optimization

VelocityRa:
- Blacklist some commonly used NIDs that spam the log too much (debug-only feature, irrelevant to users)
- A potential perf improvement to import calling, though unlikely because of branch prediction.
- Linux build hotfix

### DIFF
--- a/src/emulator/host/src/host.cpp
+++ b/src/emulator/host/src/host.cpp
@@ -24,7 +24,13 @@
 #include <util/lock_and_find.h>
 #include <util/log.h>
 
-static bool LOG_IMPORT_CALLS = false;
+static
+#ifdef NDEBUG // Leave it as non-constexpr on Debug so that we can enable/disable it at will via set_log_import_calls
+    constexpr
+#endif
+    bool LOG_IMPORT_CALLS
+    = false;
+
 static constexpr bool LOG_UNK_NIDS_ALWAYS = false;
 
 #define NID(name, nid) extern const ImportFn import_##name;
@@ -91,6 +97,9 @@ void call_import(HostState &host, CPUState &cpu, uint32_t nid, SceUID thread_id)
     }
 }
 
-void log_import_calls(bool enabled) {
+#ifndef NDEBUG
+// Import logging is really slow and this allows us to enable/disable it from whenever we please, for easy in debugging
+void set_log_import_calls(bool enabled) {
     LOG_IMPORT_CALLS = enabled;
 }
+#endif
--- a/src/emulator/host/src/host.cpp
+++ b/src/emulator/host/src/host.cpp
@@ -24,7 +24,8 @@
 #include <util/lock_and_find.h>
 #include <util/log.h>
 
-static
+#include <unordered_set>
+
 #ifdef NDEBUG // Leave it as non-constexpr on Debug so that we can enable/disable it at will via set_log_import_calls
     constexpr
 #endif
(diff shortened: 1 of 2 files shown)


src/emulator/host/src/host.cpp
```diff
@@ -23,8 +23,16 @@
...
 #include <util/log.h>
+#include <util/preprocessor.h>
+
+#include <unordered_set>
+
+#ifdef NDEBUG // Leave it as non-constexpr on Debug so that we can enable/disable it at will via set_log_import_calls
+    constexpr
+#endif
+    bool LOG_IMPORT_CALLS
+    = true;
```
VelocityRa: clang-format is confused

src/emulator/host/src/host.cpp
```diff
@@ -23,8 +23,16 @@
...
 #include <util/log.h>
+#include <util/preprocessor.h>
+
+#include <unordered_set>
+
+#ifdef NDEBUG // Leave it as non-constexpr on Debug so that we can enable/disable it at will via set_log_import_calls
+    constexpr
+#endif
+    bool LOG_IMPORT_CALLS
+    = false;
```
VelocityRa: clang-format got confused here

VelocityRa:
Fixed typo, ready for review.

src/emulator/host/src/host.cpp
```diff
@@ -58,15 +65,27 @@ static Address resolve_export(KernelState &kernel, uint32_t nid) {
     return export_address->second;
 }
 
+void log_import_call(char emulation_level, uint32_t nid, SceUID thread_id, const std::unordered_set<uint32_t> &nid_blacklist) {
```
pent0:
```suggestion
static void log_import_call(const char emulation_level, const uint32_t nid, const SceUID thread_id, const std::unordered_set<uint32_t> &nid_blacklist) {
```

src/emulator/host/src/host.cpp
```diff
@@ -82,15 +101,19 @@ void call_import(HostState &host, CPUState &cpu, uint32_t nid, SceUID thread_id)
...
         const ThreadStatePtr thread = lock_and_find(thread_id, host.kernel.threads, host.kernel.mutex);
         const std::lock_guard<std::mutex> lock(thread->mutex);
         write_pc(*thread->cpu, export_pc);
     }
 }
 
-void log_import_calls(bool enabled) {
+#ifndef NDEBUG
+// Import logging is really slow and this allows us to enable/disable it from whenever we please, for easy in debugging
+void set_log_import_calls(bool enabled) {
```
pent0: Dunno if this should be static or not hehe

src/emulator/host/src/host.cpp
VelocityRa: we usually don't make fundamental value types `const` but yea I'll make it `static`

src/emulator/host/src/host.cpp
VelocityRa: probably not, it's exported in the header since we want to use it from anywhere (and `static` funcs in header are weird)

pent0: Okay, I was just not sure

src/emulator/host/src/host.cpp
pent0: TIL 😲

PR #519, opened 2019-06-07
Discord Rich Presence

IllusionMan1212:
Tested on Windows and works fine, @1whatleytay tested an older commit on mac and reported that it works fine, need testers for linux and since this is a newer commit, mac testers are appreciated too.

- While on game selector
image

- Ingame
Untitled


src/emulator/discord.cpp
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	void shutdown()
+	{
+		Discord_Shutdown();
+	}
+
+	void update_presence(const std::string& state, const std::string& details, bool reset_timer)
+	{
+		DiscordRichPresence discordPresence = {};
```
1whatleytay: Remember to adapt our style... Shouldn't this be discord_presence? etc.

src/emulator/gui/src/gui.cpp
```diff
@@ -21,6 +21,9 @@
 
 #include <gui/imgui_impl_sdl_gl3.h>
 
+#ifdef NDEBUG
+#include <discord.h>
+#endif
```
1whatleytay: Think a comment here is necessary? I don't know how many people would be confused about this- but still.

src/emulator/host/include/host/config.h
```diff
@@ -53,4 +53,5 @@ struct Config {
     std::string background_image = {};
     float background_alpha = 0.300f;
     int icon_size = 64;
+    bool discord_rich_presence = false;
```
1whatleytay: Think this could be optional?

src/emulator/main.cpp
```diff
@@ -155,6 +159,9 @@ int main(int argc, char *argv[]) {
 
         set_window_title(host);
     }
+#ifdef NDEBUG
+    discord::shutdown();
```
1whatleytay: I think it's about time for a host_destroy. Just organizes these destroy functions into a nice clean place. I was going to do it in a GDB pr soon anyway. I want to hear what you guys think.

src/external/discord-rpc/include/discord_register.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
```
1whatleytay: Did you try adding this as a module? This might just be github glitching out, but it looks like a hard copy to me.

src/emulator/host/include/host/config.h
Zangetsu38: Yes moreb set in on tue by default

src/emulator/discord.cpp
IllusionMan1212: copy-pasta, forgot about it.
also, should i credit rpcs3 for this?

src/external/discord-rpc/include/discord_register.h
IllusionMan1212: I'm not really sure what you mean here, could you elaborate please?

src/emulator/host/include/host/config.h
IllusionMan1212: sure. i'm not sure having it on by default is a good idea zang, but if the devs are ok with it then ok

VelocityRa: have it on by default

src/emulator/config.cpp
```diff
@@ -167,6 +168,7 @@ static ExitCode parse(Config &cfg, fs::path load_path, const std::string &root_p
     get_yaml_value(config_node, "background-alpha", &cfg.background_alpha, 0.300f);
     get_yaml_value(config_node, "log-level", &cfg.log_level, static_cast<int>(spdlog::level::trace));
     get_yaml_value(config_node, "pref-path", &cfg.pref_path, root_pref_path);
+    get_yaml_value(config_node, "discord-rich-presence", &cfg.discord_rich_presence, false);
```
Zangetsu38: Is good set both on true

src/emulator/main.cpp
VelocityRa: not sure if discord and gdb deinit are host things though, they have no relation to the actual `host` library do they? at least not discord idk about gdb

src/external/discord-rpc/include/discord_register.h
VelocityRa: he meant submodule: https://github.com/discordapp/discord-rpc

src/emulator/gui/src/gui.cpp
```diff
@@ -278,7 +281,14 @@ void draw_ui(HostState &host) {
 
     if (host.gui.configuration_menu.settings_dialog)
         draw_settings_dialog(host);
-
+#ifdef NDEBUG
```
VelocityRa: unrelated to gui, discard changes in this file and find somewhere else to put this like app.cpp

src/emulator/main.cpp
VelocityRa: disregard, I didn't realize we initialized HostState in `emulator/init.cpp` (for some reason)

src/external/discord-rpc/include/discord_register.h
VelocityRa: and yes, do use it instead of the hard-copy

src/emulator/discord.cpp
1whatleytay: I did the same for my GDB stub. Go ahead if you'd like :)

src/emulator/gui/src/gui.cpp
1whatleytay: About the NDEBUG stuff. It's necessary but its kinda confusing. I kinda wish there was another definition that would enable discord (like USE_DISCORD added release). It would make the intention clearer- but its not a big deal. Just throwing stuff around lol.

VelocityRa: Yea, @IllusionMan1212 try making it a CMake option like [gdb stuff](https://github.com/Vita3K/Vita3K/blob/master/CMakeLists.txt#L8) and you'd need to do [this](to use it as a define in C++).
You can then use the `CI` CMake var (see line 36 there for example) to enable it on CI builds.

src/emulator/shader/src/usse_translator_entry.cpp
```diff
@@ -37,7 +37,7 @@ using USSEMatcher = shader::decoder::Matcher<Visitor, uint64_t>;
 template <typename V>
 boost::optional<const USSEMatcher<V> &> DecodeUSSE(uint64_t instruction) {
     static const std::vector<USSEMatcher<V>> table = {
-    // clang-format off
+// clang-format off
```
VelocityRa: remove this change

src/emulator/host/include/host/config.h
```diff
@@ -53,4 +53,5 @@ struct Config {
     std::string background_image = {};
     float background_alpha = 0.300f;
     int icon_size = 64;
+    optional<bool> discord_rich_presence = true;
```
VelocityRa: why optional?

src/external/CMakeLists.txt
```diff
@@ -145,6 +145,16 @@ target_link_libraries(yaml INTERFACE yaml-cpp)
 add_library(rpcs3 INTERFACE)
 target_include_directories(rpcs3 INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}/rpcs3/include")
 
+add_library(discord-rpc INTERFACE)
+target_include_directories(discord-rpc INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}/discord-rpc/include")
+if(APPLE)
+	target_link_libraries(discord-rpc INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}/discord-rpc-libs/libdiscord-rpc-mac.a")
```
VelocityRa: Rename directory `discord-rpc-libs` -> `discord-rpc-lib` as is the convention.

src/emulator/shader/src/usse_translator_entry.cpp
1whatleytay: Why the heck does clang format keep on changing that >:(

src/emulator/host/include/host/config.h
1whatleytay: Why not? Would rather have more flexibility than less.

VelocityRa: What's the point of making it `optional` and directly initializing it with a value?

IllusionMan1212: The initialization is my bad, des asked for it to be optional and so i just did it. should i remove it?

1whatleytay: That is a very good point.

Then drop the ` = true` and check if the optional is empty [here](https://github.com/Vita3K/Vita3K/pull/519/files#diff-64550198e4afbca5fefcafa71ec8889fR341) and give it a default value if it is empty?

... or you can just make it mandatory. Up to you.

IllusionMan1212:
updated

src/emulator/CMakeLists.txt
```diff
@@ -49,6 +49,10 @@ if (USE_GDBSTUB)
 	add_definitions(-DUSE_GDBSTUB)
 endif()
 
+if (USE_DISCORD_RICH_PRESENCE)
+	add_definitions(-DDISCORD_RICH_PRESENCE)
```
VelocityRa:
```suggestion
	add_definitions(-DUSE_DISCORD_RICH_PRESENCE)
```
kind of nitpicky but gdbstub is done the same way

src/emulator/main.cpp
```diff
@@ -111,6 +115,14 @@ int main(int argc, char *argv[]) {
             glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
             gui::draw_begin(host);
 
+#ifdef USE_DISCORD_RICH_PRESENCE
+            if (host.cfg.discord_rich_presence) {
+                discord::initialize();
```
1whatleytay: This isn't a request to change anything yet, just curious. Is it supposed to initialize discord every frame like this? Or just update presence?

HolyMcDiver:
Still looking for Linux testers? I can ask a cousin of mine to test some homebrew with kubuntu.

src/emulator/main.cpp
IllusionMan1212: It's supposed to initialize on every frame because I call `shutdown()` when the condition is false. Maybe you have a better way of doing this?

VelocityRa: There's many ways but the simplest would be to keep a `discord_rich_presence_old` before the loop and then after the loop compare it with the new one. If it goes from off to on, shutdown and the inverse.
"Idle" would not work in that specific case (unless you do the check inside the loop and then toggle one of the two).

Maybe it's better to make a `discord::update_init_status` with the cfg.discord_rich_presence as an arg or something and  handle that logic in there though to not pollute main.

src/emulator/app/src/discord.cpp
```diff
@@ -0,0 +1,44 @@
+#ifdef USE_DISCORD_RICH_PRESENCE
+#include <app/discord.h>
+#include <discord_register.h>
+#include <discord_rpc.h>
+
+// Credits to the RPCS3 Project
+
+namespace discord {
+void initialize(const std::string &application_id) {
```
VelocityRa:
```suggestion
void init(const std::string &application_id) {
```
to fit the rest of our modules

IllusionMan1212:
Changes done

1whatleytay:
LGTM

src/emulator/app/src/app_init.cpp
```diff
@@ -174,6 +178,12 @@ bool init(HostState &state, Config cfg, const Root &root_paths) {
         return false;
     }
 
+#ifdef USE_DISCORD_RICH_PRESENCE
+    discord::init();
+    if (cfg.discord_rich_presence) {
+        discord::update_presence("");
```
VelocityRa:
```suggestion
        discord::update_presence();
```
you already have a default argument
but whatever it's fine

PR #522, opened 2019-06-10
Texture dumping

VelocityRa:
Dumps textures in `shaderlog/<TITLE_ID>` as they are used.

Exposed in the config (config file + cmd arg) and in the Debug settings dialog.

Some games change textures every frame and the way the dumper is implemented, it dumps _all_ variations as long as the texture contents are different. So you might want to enable this at runtime from the Debug settings whenever you want to get the textures, and then and disable it (but you might miss some that appeared before and are not used anymore).

It only works with implemented texture formats.

#### Alone with You:
<details>
 <summary>Screenshot</summary>

image
(up until the intro screen)
</details>

####  Darius Burst: Chronicle Saviours
<details>
 <summary>Screenshot</summary>

image
(up until the main menu)
</details>

#### Downwell
<details>
 <summary>Screenshot</summary>

image
</details>

---

### Testers needed
'Danganronpa: Trigger Havoc' crashes my driver, needs more testing to find a pattern or (unlikely) whether it's a driver bug (could be if it only happens on Nvidia).

### DIFF
--- a/src/emulator/renderer/src/texture.cpp
+++ b/src/emulator/renderer/src/texture.cpp
@@ -9,6 +9,11 @@
 #include <mem/ptr.h>
 #include <util/log.h>
 
+#define STB_IMAGE_WRITE_IMPLEMENTATION
+#include "stb_image_write.h"
+
+#include <set>
+
 namespace renderer {
 namespace texture {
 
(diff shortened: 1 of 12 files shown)


src/emulator/renderer/src/texture.cpp
```diff
@@ -82,5 +87,50 @@ void upload_bound_texture(const emu::SceGxmTexture &gxm_texture, const MemState
...
+    const TextureCacheHash hash = hash_texture_data(gxm_texture, mem);
+
+    if (g_dumped_hashes.find(hash) != g_dumped_hashes.end())
+        return;
+
+    g_dumped_hashes.insert(hash);
+
+    const size_t width = gxm::get_width(&gxm_texture);
+    const size_t height = gxm::get_height(&gxm_texture);
+    ;
```
pent0: ??

VelocityRa: i did this with no sleep man cut me some slack

pent0: go to sleep!!!!

image

VelocityRa: lmao

wfscans:
Tested against windows AMD / linux. Both crash on DR, but at different points. 

[vita3k_linux.log](https://github.com/Vita3K/Vita3K/files/3274760/vita3k_linux.log)
[vita3k_win.log](https://github.com/Vita3K/Vita3K/files/3274761/vita3k_win.log)

Linux seems to get further by actually creating a folder, but crashes with `Vita3k: malloc.c3722 _int_malloc: Assertion '(unsigned long) (size) >= (unsigned long) (nb)' failed`

PR #527, opened 2019-06-22
external: Replace glbinding by glad

### LINKED ISSUE #489: Get rid of glbinding in favour of glad
It's causing annoying build issues, binary bloat, and longer compile times (especially glbinding-aux).

scribam:
Close #489 

~~Warning: the release build of glad doesn't include pre/post callbacks so I didn't replace the before_callback/after_callback functions by a "glad" equivalent. I am not sure if I should use the debug build to have these callbacks, specially if it can affect performance as mentioned in https://github.com/Dav1dde/glad/issues/132. Any advice?~~


src/emulator/renderer/src/compile_program.cpp
```diff
@@ -35,7 +35,7 @@ static SharedGLObject compile_glsl(GLenum type, const std::string &source) {
         LOG_ERROR("{}", log.data());
     }
 
-    GLboolean is_compiled = GL_FALSE;
+	GLint is_compiled = GL_FALSE;
```
pent0: Spacing weird

src/emulator/renderer/src/renderer.cpp
```diff
@@ -112,10 +109,7 @@ bool create(Context &context, SDL_Window *window) {
     context.gl = GLContextPtr(SDL_GL_CreateContext(window), SDL_GL_DeleteContext);
     assert(context.gl != nullptr);
 
-    const glbinding::GetProcAddress get_proc_address = [](const char *name) {
-        return reinterpret_cast<ProcAddress>(SDL_GL_GetProcAddress(name));
-    };
-    Binding::initialize(get_proc_address, false);
+	gladLoadGLLoader((GLADloadproc)SDL_GL_GetProcAddress);
```
pent0: Is this tab?

pent0:
Will leave the ideas of debug GLAD to another devs. Maybe discuss here or on Discord

petmac:
Is this better than SDL_opengl.h? The reason to use glbinding was for profiling and error checking, which I expect glad doesn’t do.

Also, I’m confused why all the GLenums changed to GLints. That didn’t seem right.

scribam:
@pent0 My bad, I forgot to run clang-format before to push. It should be ok now.

VelocityRa:
The error checks should definitely be kept for Debug mode.
You could make it a compile-time switch to turn them off optionally if they're that slow.

scribam:
@petmac glTexParameteriv's third argument type is `const GLint * params` (see https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glTexParameter.xhtml) that's why I changed the type of the swizzle variable here https://github.com/Vita3K/Vita3K/blob/master/src/emulator/renderer/src/texture.cpp#L29 and to all translate_swizzle functions. glbinding casts the GLenum to a GLint which is why it wasn't an issue:
```
void glTexParameteriv(GLenum target, GLenum pname, const GLenum * params)
{
    glTexParameteriv(target, pname, reinterpret_cast<const GLint *>(params));
}
```
I can revert the code related to the GLenums to do something like that if you prefer:
```
const GLint *const swizzle = reinterpret_cast<const GLint *>(translate_swizzle(fmt));
```

PR #528, opened 2019-06-24
gdbstub: tune up

1whatleytay:
This PR does a couple of things:

- Adds support for writing to registers or memory through GDB.
- Thread IDs are apparently hexes so the stub no longer interprets them as decimals.
- GDB *should* also catch unicorn errors now.
- Made the configuration option no longer optional for two reasons: 
  - Everything else isn't optional anymore so it kinda looks ugly.
  - I was editing the wrong config file and nothing was changing and I got scared and changed it.
- Makes a proper app:destroy() method to deinitialize the app.
  - In my honest opinion, I think this is a good thing. It separates all the code that terminates stuff from main and gives it a nice cozy home.
  - I understand if you think this shouldn't be included, either because it's not necessary or this isn't the pr to do it (or any other reason).
  - Tell me if you want this to go and I'll remove it.

Before the PR is merged, I need to:
- [x] test writing to memory.
- [x] test writing to registers.

I wanted to test catching unicorn errors but I had trouble getting one to work, even with Waterflame's helpful list (I NEED THE GAMES). It should be able to catch them afaik but I'm not sure how well it would work.

### DIFF
--- a/src/emulator/gdbstub/src/gdb.cpp
+++ b/src/emulator/gdbstub/src/gdb.cpp
@@ -298,7 +298,7 @@ static std::string cmd_write_registers(HostState &state, PacketCommand &command)
         return "";
     CPUState &cpu = *state.kernel.threads[state.gdb.current_thread]->cpu.get();
 
-    std::string content = content_string(command).substr(1);
+    const std::string content = content_string(command).substr(1);
 
     for (uint32_t a = 0; a < content.size() / 8; a++) {
         uint32_t value = parse_hex(content.substr(a * 8, 8));
@@ -313,7 +313,7 @@ static std::string cmd_read_register(HostState &state, PacketCommand &command) {
         return "";
     CPUState &cpu = *state.kernel.threads[state.gdb.current_thread]->cpu.get();
 
-    std::string content = content_string(command);
+    const std::string content = content_string(command);
     int32_t reg = parse_hex(content.substr(1, content.size() - 1));
 
     return be_hex(fetch_reg(cpu, static_cast<uint32_t>(reg)));
@@ -324,7 +324,7 @@ static std::string cmd_write_register(HostState &state, PacketCommand &command) 
         return "";
     CPUState &cpu = *state.kernel.threads[state.gdb.current_thread]->cpu.get();
 
-    std::string content = content_string(command);
+    const std::string content = content_string(command);
     uint32_t equal_index = content.find('=');
     int32_t reg = parse_hex(content.substr(1, equal_index - 1));
     uint32_t value = parse_hex(content.substr(equal_index + 1));
@@ -334,15 +334,15 @@ static std::string cmd_write_register(HostState &state, PacketCommand &command) 
 }
 
 static bool check_memory_region(Address address, Address length, MemState &mem) {
-    auto page_first = static_cast<uint32_t>(address / KB(4));
-    auto page_length = static_cast<uint32_t>(length + KB(4)) / KB(4);
+    const auto page_first = static_cast<uint32_t>(address / KB(4));
+    const auto page_length = static_cast<uint32_t>(length + KB(4)) / KB(4);
 
-    auto pages_begin = mem.allocated_pages.begin();
-    auto range_end = pages_begin + page_first + page_length;
+    const auto pages_begin = mem.allocated_pages.begin();
+    const auto range_end = pages_begin + page_first + page_length;
 
-    bool memory_none = std::find(pages_begin + page_first, range_end, 0) != range_end;
-    bool memory_null = std::find(pages_begin + page_first, range_end, 1) != range_end;
-    bool memory_safe = !(memory_none || memory_null);
+    const bool memory_none = std::find(pages_begin + page_first, range_end, 0) != range_end;
+    const bool memory_null = std::find(pages_begin + page_first, range_end, 1) != range_end;
+    const bool memory_safe = !(memory_none || memory_null);
 
     if (!memory_safe)
         LOG_GDB("Accessing unsafe memory page. {} - {}", page_first, page_length);
@@ -351,13 +351,13 @@ static bool check_memory_region(Address address, Address length, MemState &mem) 
 }
 
 static std::string cmd_read_memory(HostState &state, PacketCommand &command) {
-    std::string content = content_string(command);
-    size_t pos = content.find(',');
+    const std::string content = content_string(command);
+    const size_t pos = content.find(',');
 
-    std::string first = content.substr(1, pos - 1);
-    std::string second = content.substr(pos + 1);
-    uint32_t address = parse_hex(first);
-    uint32_t length = parse_hex(second);
+    const std::string first = content.substr(1, pos - 1);
+    const std::string second = content.substr(pos + 1);
+    const uint32_t address = parse_hex(first);
+    const uint32_t length = parse_hex(second);
 
     if (!check_memory_region(address, length, state.mem))
         return "EAA";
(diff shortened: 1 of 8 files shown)


1whatleytay:
Anyone know what the whole "discord is not a namespace" deal is with the appveyor build?

VelocityRa:
main uses discord functions but does not include `app/discord.h`

1whatleytay:
Yikes. It's fixed.

To modify registers correctly you might have to `set endian big` in gdb console. Otherwise it appears to work.

src/emulator/gdbstub/src/gdb.cpp
```diff
@@ -308,30 +371,76 @@ static std::string cmd_read_memory(HostState &state, PacketCommand &command) {
...
+}
+
+// server_next() might not be able to tell the difference between the end of the packet ($) and 0x24 ($).
+// Thus, cmd_write_binary is disabled.
+static std::string cmd_write_binary(HostState &state, PacketCommand &command) {
+    const std::string content = content_string(command);
+    const size_t pos_first = content.find(',');
+    const size_t pos_second = content.find(':');
+
+    const
```
pent0: if you have read crayon shinchan vol 23 page 15 then you will know that it's irrelevant to what I'm saying. This is some weird things

src/emulator/gdbstub/src/gdb.cpp
```diff
@@ -379,6 +488,20 @@ static std::string cmd_continue_supported(HostState &state, PacketCommand &comma
     return "vCont;c;C;s;S;t;r";
 }
 
+static std::string cmd_thread_alive(HostState &state, PacketCommand &command) {
+    const std::string content = content_string(command);
+    const int32_t thread_id = parse_hex(content.substr(1));
+
+    // Assuming a thread is removed from the map when it closes or is killed.
+    for (const auto &pair : state.kernel.threads) {
```
pent0: Should use find(key), if this is a map

```suggestion
    if (state.kernel.threads.find(thread_id) != state.kernel.threads.end()) return "OK";
```

src/emulator/gdbstub/src/gdb.cpp
```diff
@@ -258,46 +293,74 @@ static std::string cmd_read_registers(HostState &state, PacketCommand &command)
...
+    modify_reg(cpu, reg, value);
+
+    return "OK";
 }
 
-static std::string cmd_read_memory(HostState &state, PacketCommand &command) {
-    std::string content = content_string(command);
-    size_t pos = content.find(',');
+static bool check_memory_region(Address address, Address length, MemState &mem) {
+    const auto page_first = static_cast<uint32_t>(address / KB(4));
```
pent0: KB(4) can be an constexpr, but it doesn't hurt really. Optional.

src/emulator/gdbstub/src/gdb.cpp
```diff
@@ -258,46 +293,74 @@ static std::string cmd_read_registers(HostState &state, PacketCommand &command)
     return stream.str();
 }
 
-static std::string cmd_write_registers(HostState &state, PacketCommand &command) { return ""; }
+static std::string cmd_write_registers(HostState &state, PacketCommand &command) {
+    if (state.gdb.current_thread == -1)
+        return "";
+    CPUState &cpu = *state.kernel.threads[state.gdb.current_thread]->cpu.get();
```
pent0: This is not safe if the current_thread id doesn't point to a legitimate thread. Better make a null check

src/emulator/gdbstub/src/gdb.cpp
```diff
@@ -258,46 +293,74 @@ static std::string cmd_read_registers(HostState &state, PacketCommand &command)
...
     int32_t reg = parse_hex(content.substr(1, content.size() - 1));
 
     return be_hex(fetch_reg(cpu, static_cast<uint32_t>(reg)));
 }
 
 static std::string cmd_write_register(HostState &state, PacketCommand &command) {
-    return "";
+    if (state.gdb.current_thread == -1)
+        return "";
+    CPUState &cpu = *state.kernel.threads[state.gdb.current_thread]->cpu.get();
```
pent0: Same, as above.

src/emulator/gdbstub/src/gdb.cpp
```diff
@@ -258,46 +293,74 @@ static std::string cmd_read_registers(HostState &state, PacketCommand &command)
...
-    std::string first = content.substr(1, pos - 1);
-    std::string second = content.substr(pos + 1, content.size() - pos);
-    uint32_t address = parse_hex(first);
-    uint32_t length = parse_hex(second);
+    const auto pages_begin = mem.allocated_pages.begin();
+    const auto range_end = pages_begin + page_first + page_length;
+
+    const bool memory_none = std::find(pages_begin + page_first, range_end, 0) != range_end;
+    const bool memory_null = std::find(pages_begin + page_first, range_end, 1) != range_end;
+    const bool memory_safe = !(memory_none || memory_null);
```
pent0: Really informative, however I don't know if I can take extra class tomorrow or i drop it since the wind is too large and when i go to sleep there might be a chance of rain so hard that i will sleep past where i have to go to class.

My grandfather is the leader of a secert guild that teach kong fu technique. One day he passed the technique on me and suddenly I am being chased by bad guys, so I have to transfer school a lot. I don't have many friends, but when I enter Inoba High School there is this girl named Kazuki who befriended me. We have fun a lot. but turns out she was an enemy too, she was stabbing me from the back. When I was being chased by here, i run and hit another girl named Mina who fell in love with me. Kazuki then stop chasing and being a love rival. But during that time, another girl steps in. Her name is Trish and she came with a bunch of weird Italians who seek me to help them find the treasure of a secert Italian gangster leader. From there, all things are messed up. What should I do

src/emulator/gdbstub/src/gdb.cpp
1whatleytay: pos_comma, pos_colon?

pent0: I don't really understand but why the rest of the line is below?

pent0: image

1whatleytay: I blame it on CLion.

Also I looked up Volume 23 of Crayon Shin-Chan and I am severely disappointed.
Screen Shot 2019-07-03 at 2 00 04 PM

src/emulator/gdbstub/src/gdb.cpp
1whatleytay: Yep, sounds good.

src/emulator/gdbstub/src/gdb.cpp
1whatleytay: This kills the whole point of `constexpr`, but how about `MemState::page_size`? I forgot that existed.

pent0: Oh yeah. I forgot that. go with it

src/emulator/gdbstub/src/gdb.cpp
pent0: It wasn't popular on the East. I was just taking some random manga while my brain is high

image

Here is the page though (in a language you may not know)

src/emulator/gdbstub/src/gdb.cpp
1whatleytay: I mean a check won't hurt, but gdb should be picking a valid thread anyway.

I will do it.

pent0: I thought you can explicitly specify some random thread ID (in some case users doing extremely obscure thing). If it's too overload than drop, I'm really sorry.

src/emulator/gdbstub/src/gdb.cpp
1whatleytay: Dude, that situation sounds tight. Hope you find a way out of it soon.

If the bad guys are still chasing you I would suggest talking to the Move Deleter in Lilycove City.

src/emulator/gdbstub/src/gdb.cpp
1whatleytay: You could probably do that, and it's better to be safe than sorry. I can imagine someone accidentally typing an extra 1 at the end during an important debugging session and then crying their eyes out when everything suddenly crashes.

It's just that GDB should know all threads that exist, so it should stop them? I have no idea if that's the case or not though.

src/emulator/gdbstub/src/gdb.cpp
pent0: I afraid I'm gonna get danger since Trish and the gang can't protect me all the time, and Kazuki will stop loving me because I'm not cool ass anymore, and so do Mina, she will try to kill me again.  This is really tight. 

Maybe I will get the treasure and go home first, than drive my helicopter to Lilycove city with gon from hunter x hunter, which is the show i never see but the name keeps popping up on my twitter.com page

pent0: Why did i wrote this

1whatleytay: It was fun.

1whatleytay:
Conflicts are fixed! Can someone review?

src/emulator/main.cpp
```diff
@@ -28,9 +28,7 @@
 #include <util/log.h>
 #include <util/string_utils.h>
 
-#ifdef USE_GDBSTUB
-#include <gdbstub/functions.h>
-#endif
+#include <SDL.h>
```
pent0: Is this a leftover from rebase or needed

1whatleytay: Rebase 😦  Fixed in new commit.

PR #530, opened 2019-06-26
Major Leaguer

pent0:
- Correct repeat jump for VDP: Analyzing 2 shaders in Super Mario Wars, I found out that without SMLSI to adjust repeat amount, src1 has repeat jump of 4, whether dest has repeat jump of 1.
- Correct custom increment mode for single-issue instruction.

More to come, this PR tends to correct all faults and missings in Super Mario Wars and other VitaGL games. Maybe mega

TODO:
- [ ] Implement SOP2 (for fixed alu op)


pent0:
Close for riichi

PR #521, opened 2019-06-10
Add app module

petmac:
I'm going to try and clean up the emulator dependencies over a series of PRs rather than a giant one.

The end goal is to have `main()` control/instantiate `gui`, which controls/owns `app`, which controls/owns `host`.

<table>
<tr><th>emulator</th></tr>
<tr><td>
<table>
<tr><th>gui</th></tr>
<tr><td>
<table>
<tr><th>app</th></tr>
<tr><td>
<table>
<tr><th>host</th></tr>
<tr><td>
<table>
<tr><td>audio</td><td>gxm</td><td>kernel</td><td>...</td></tr>
</table>
</td></tr>
</table>
</td></tr>
</table>
</td></tr>
</table>
</td></tr>
</table>

At the moment there is a dependency cycle involving `gui`, which I will attempt to break in a later PR.

PR #531, opened 2019-06-26
emulator: Create Config namespace

totlmstr:
This PR originally started by tackling a glaring problem: [in init config, the configuration from the command line may end up overwriting the loaded config file in the base path](https://github.com/Vita3K/Vita3K/blob/master/src/emulator/app/src/app_config.cpp#L194).

After some cleanup in `app_config.cpp`, I realized that the Config is rather spread out across the program, and it would be much better if instead it was instead its own library. This was partially inspired by PR #521, and makes the program much cleaner.

Another side effect is moving `version.h` and `version.cpp.in` files. The odd thing about these files is that they are requested by both `host` and `app`, but only used once in each, but is called first in this new module `config`. As a result, the `version.h` and its companion `version.cpp.in` are now located in the `config` namespace.

This is still a draft because some changes are still needed before review.


src/emulator/gui/src/settings_dialog.cpp
```diff
@@ -92,13 +93,38 @@ using namespace list;
...
+            }
+        }
+        if (ImGui::IsItemHovered())
+            ImGui::SetTooltip("Reset Vita3K's configuration to the default values. \nAfter pressing, restart Vita3K to fully apply. \nWARNING: This cannot be undone.");
+        ImGui::EndTabItem();
+    } else {
+        ImGui::PopStyleColor();
+    }
+
+    // Modules
```
Zangetsu38: don't change this, keep Core, i have set this for can add in futuru Other backend cpu

src/emulator/gui/src/settings_dialog.cpp
```diff
@@ -149,6 +175,7 @@ void draw_settings_dialog(HostState &host) {
         if (ImGui::IsItemHovered())
             ImGui::SetTooltip("This is the button that is used as 'Confirm' in game dialogs. \nSome games don't use this and get default confirmation button.");
         ImGui::RadioButton("Circle", &host.cfg.sys_button, 0);
+        ImGui::SameLine();
```
Zangetsu38: same, revert it, i have tgester both and other dev have prefered this

src/emulator/host/CMakeLists.txt
```diff
@@ -1,20 +1,15 @@
...
-version.cpp
+	host
+	STATIC
+	include/host/functions.h
+	include/host/import_fn.h
+	include/host/sfo.h
+	include/host/state.h
+	include/host/app_util.h
+	src/host.cpp
+	src/sfo.cpp
```
Zangetsu38: Clang format ?

src/emulator/config/src/config.cpp
```diff
@@ -199,30 +199,31 @@ ExitCode init_config(Config &cfg, int argc, char **argv, const Root &root_paths)
...
         generic_desc.add_options()
             ("help,h", "print this usage message")
             ("version,v", "print version string");
 
         po::options_description input_desc("Input");
         input_desc.add_options()
-            ("input-vpk-path,i", po::value(&cfg.vpk_path), "path of app in .vpk format to install & run")
-            ("input-installed-id,r", po::value(&cfg.run_title_id), "title ID of installed app to run")
-            ("recompile-shader,s", po::value(&cfg.recompile_shader_path), "Recompile the given PS Vita shader (GXP format) to SPIR-V / GLSL and quit");
+            ("input-vpk-path,i", po::value(&command_line.vpk_path), "path of app in .vpk format to install & run")
```
Zangetsu38: keep cfg, command line name not have sense

src/emulator/config/src/config.cpp
```diff
@@ -255,24 +256,36 @@ ExitCode init_config(Config &cfg, int argc, char **argv, const Root &root_paths)
...
             return QuitRequested;
         }
-        if (var_map.count("recompile-shader"))
+        if (var_map.count("recompile-shader")) {
             return QuitRequested;
-
-        if (cfg.load_config || cfg.config_path != root_paths.get_base_path()) {
-            if (cfg.config_path.empty()) {
-                cfg.config_path = root_paths.get_base_path();
+        }
```
Zangetsu38: remove {} don't needed  here

src/emulator/gui/src/settings_dialog.cpp
```diff
@@ -92,13 +93,38 @@ using namespace list;
...
     ImGui::PushStyleColor(ImGuiCol_Text, GUI_COLOR_TEXT_MENUBAR_OPTIONS);
     ImGui::Begin("Settings", &host.gui.configuration_menu.settings_dialog, ImGuiWindowFlags_AlwaysAutoResize);
-    ImGuiTabBarFlags settings_tab_flags = ImGuiTabBarFlags_None;
+    const auto settings_tab_flags = ImGuiTabBarFlags_None;
     ImGui::BeginTabBar("SettingsTabBar", settings_tab_flags);
 
-    // Core
-    if (ImGui::BeginTabItem("Core")) {
+    // General
+    if (ImGui::BeginTabItem("General")) {
```
Zangetsu38: Remove this, and set this inside "Emulator" tab, don't need touch tab.

src/emulator/config/src/config.cpp
totlmstr: I originally added the braces for readability rather than for need (and because it looks cleaner when with the other Boost.Optional). This may change in a different commit, but if nothing comes out of it, I'll delete it.

src/emulator/config/include/config/config_func.h
```diff
@@ -30,7 +30,7 @@ namespace app {
   * \param cfg Config operations to save.
   * \param output_path The location to save the configuration file.
   * \return Success on saving the config file, otherwise Error.
- */
+  */
```
Zangetsu38: ? ^^

src/emulator/config/include/config/config.h
```diff
@@ -43,15 +42,34 @@ struct Config {
...
-};
+    optional<bool> wait_for_debugger;
+
+    // void merge_config(const Config &cfg, const std::string &new_pref_path = {}) const {
+    //     this = std::move(cfg);
+    //     this->pref_path = new_pref_path;
+    // }
+
+    // Optimal based on: http://stackoverflow.com/a/24477023
+    void merge_lle(std::vector<std::string> new_lle) const {
```
pent0: Maybe make an function in util? Maybe useful in the future

pent0:
Nice idea btw, I like moving version.h to it. Maybe rename it to something more generic would be cool

src/emulator/config/include/config/config.h
totlmstr: Done, and made it templated. It was a pretty quick and good idea, anyway.

For convenience, it is over in `util/vector_utils.h`.

src/emulator/gui/src/settings_dialog.cpp
```diff
@@ -197,6 +198,23 @@ void draw_settings_dialog(HostState &host) {
         }
         if (ImGui::IsItemHovered())
             ImGui::SetTooltip("After pressing, restart Vita3K to fully apply changes.");
+        ImGui::SameLine();
+        if (ImGui::Button("Reset Configuration")) {
+            host.cfg = Config{};
+
+            LOG_INFO("Reset Vita3K configuration and config file to default values.");
+            if (host.cfg.pref_path != host.pref_path) {
```
Zangetsu38:
```suggestion
            if (host.cfg.pref_path != host.pref_path && change_pref_location(host.cfg.pref_path, host.pref_path)) {
```

Zangetsu38: If is possible like that ?

src/emulator/config/include/config/config.h
```diff
@@ -43,15 +40,24 @@ struct Config {
...
+    int icon_size = 64;
     bool archive_log = false;
     bool texture_cache = true;
-    optional<bool> wait_for_debugger;
+    int sys_button = 1;
+    int sys_lang = 1;
     std::string background_image = {};
     float background_alpha = 0.300f;
-    int icon_size = 64;
+    int log_level = 0;
```
Zangetsu38: why reorder all this ?

totlmstr: Consistency. In `config.cpp` it is ordered as it is now in `config.h`.

Zangetsu38: ah, i see

totlmstr:
Renaming branch and focus of PR due to the scope change of the latest commits.

PR #537, opened 2019-07-07
Functions needed to get dynarecs working on Vita3k

frangarcj:


src/emulator/modules/SceSysmem/SceSysmem.cpp
```diff
@@ -44,8 +44,22 @@ EXPORT(SceUID, sceKernelAllocMemBlock, const char *name, SceKernelMemBlockType t
...
-EXPORT(int, sceKernelAllocMemBlockForVM) {
-    return UNIMPLEMENTED();
+EXPORT(int, sceKernelAllocMemBlockForVM, const char *name, int size) {
+    MemState &mem = host.mem;
+    assert(name != nullptr);
+    assert(size != 0);
+    
+    const Ptr<void> address(alloc(mem, size, name));
+    if (!address) {
+        return SCE_KERNEL_ERROR_NO_MEMORY;
```
Rinnegatamante: Shouldn't this be return RET_ERROR(SCE_KERNEL_ERROR_NO_MEMORY); ?

src/emulator/modules/SceSysmem/SceSysmem.cpp
```diff
@@ -81,8 +95,18 @@ EXPORT(int, sceKernelFreeMemBlock, SceUID uid) {
     return SCE_KERNEL_OK;
 }
 
-EXPORT(int, sceKernelFreeMemBlockForVM) {
-    return UNIMPLEMENTED();
+EXPORT(int, sceKernelFreeMemBlockForVM, SceUID uid) {
+    assert(uid >= 0);
+    
+    KernelState *const state = &host.kernel;
+    const Blocks::const_iterator block = state->blocks.find(uid);
```
pent0: Since you are pushing vm_blocks explicitly, is it better to search in vm_blocks.
Else just delete the vm_blocks map completely. It has no usage, and just increase memory usage

frangarcj: The new map is needed for future functions: open, close and sync domain. I'll make the find change.

Issue #455, opened 2019-04-13
Drawing from a thread different from the one that's created the context


xyzz:
If the game's created a context (sceGxmCreateContext) from one thread, but draws or does any gl-related operation such as create shader, etc, on another, Vita3k crashes (on linux nvidia drivers, possible that it just doesn't draw anything on other drivers).

The reason is that vita3k calls SDL_GL_CreateContext in its implementation of sceGxmCreateContext which assigns the OpenGL context to the current thread, however when trying to draw from a different thread, there is no gl context.

Here's an example change to a homebrew which works on vita but crashes on vita3k https://github.com/xyzz/VitaTester/commit/e51f2ed5e33e086dad568ab00ece6341e760a85a

Also, this behavior is seen in a commercial game "BreakQuest: Extra Evolution".

I've tried working around it like so https://github.com/xyzz/Vita3K/commit/08b88634af134bb4346f7d74a34f956080a86caa but SDL_GL_MakeCurrent just returns a generic error "Unable to make GL context current".

xyzz:
After adding a https://github.com/xyzz/Vita3K/commit/86eaf9c9cdb93302474acecb505a968930089cbe the modified vitatester works fine, though it's probably still not the best solution to this problem

petmac:
Thanks @xyzz! Historically I assumed GXM contexts were only usable from a single thread, and if we found otherwise we would cross that bridge when we came to it.

I had planned to move all OpenGL calls to the main thread using a command queue system, to make debugging with apitrace easier, and I think that would fix this too.

Or we should just bite the bullet and move to Vulkan... 🤧

Rinnegatamante:
@petmac Gxm allows multithreaded rendering even if i think there are not a lot of titles actually using that feature.

pent0:
Hope this is solved with newest master. Please test again.

PR #538, opened 2019-07-09
poc: non stop rendering on one thread

pent0:
This is trying to solve the problem when a context is created in one thread but is used for rendering in other thread. 

Each GXM context should has a responsible OpenGL context, that's necessary. There was two solution to the problem upper that i think of before:
- Using a GL Scope: This is just costly, because context switching is expensive.
- Detect when the GXM context is used in other thread and bind it to that thread. For eg when it called BeginScene in other thread, we set the opengl context responsibles with the GXM context in current thread. However, how do we free the OpenGL context that was used in previous thread. It's not easy.

This brings me to a solution of doing OpenGL operations only in one thread. This make control of set/unset OpenGL context much easier.

- Command are pushed into command queue, along with pointer to an integer which will report the return code when the request is finished. For sync operation, the thread will wait for the conditional variable and also wait for the request code to change.
- A OpenGL thread processing it, and notify all threads when a request is done. 

I don't know how this make debugging graphics easier though. Open for discussion, if it's welcomed then I will make a full PR

### DIFF
--- a/src/emulator/renderer/CMakeLists.txt
+++ b/src/emulator/renderer/CMakeLists.txt
@@ -1,6 +1,7 @@
 add_library(
 	renderer
 	STATIC
+	include/renderer/commands.h
 	include/renderer/functions.h
 	include/renderer/state.h
 	include/renderer/texture_cache_state.h
@@ -8,6 +9,7 @@ add_library(
 	src/attribute_formats.cpp
 	src/compile_program.cpp
 	src/draw.cpp
+	src/driver_thread.cpp
 	src/functions.h
 	src/load_shaders.cpp
 	src/renderer.cpp
@@ -21,5 +23,5 @@ add_library(
 )
 
 target_include_directories(renderer PUBLIC include)
-target_link_libraries(renderer PUBLIC crypto shader glutil vita-headers)
+target_link_libraries(renderer PUBLIC crypto shader glutil threads vita-headers)
 target_link_libraries(renderer PRIVATE sdl2)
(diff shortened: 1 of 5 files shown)


pent0:
You might interested @petmac

petmac:
I am interested, thank you @pent0.

OpenGL debuggers will still get confused as they don't seem to handle multiple contexts and threads well.

https://renderdoc.org/docs/behind_scenes/opengl_support.html?highlight=thread#multiple-contexts-multithreading

What I would attempt instead would be to only ever use a single OpenGL instance (created on the "main" UI thread) and perform all OpenGL calls from that thread. Obviously keeping OpenGL state in sync with GXM state, and restoring state for IMGUI when done processing GXM commands, will be a pain.

pent0:
> Obviously keeping OpenGL state in sync with GXM state, ... will be a pain.

This is one of the reason I wouldn't attempt it. For example, two thread creating a GXM context. One thread might issue a BeginScene command when other thread would issue a Draw command. Restoring the state would be a pain.

I still want the one context like you describe in #455, but I don't know restoring the state would drop the performance though. Also there is the sync object, I didn't think about it carefully.

pent0:
Whatever we do there is still something that cost in the way ...... agh

- Using one context: State has to be switched frequently, we can do it but cost performance (FBO switching is costly if the internet tells me right).
- Using multiple contexts in multiple thread: issue #455 
- Using multiple contexts and switch it frequent in one thread, do OpenGL operations on that thread: Avoid point 1 and point 2 (only has to switch the OpenGL context) but hard to debug.
- Vulkan, do operations on one thread: Avoid the state machine performance but takes time and knowledge :p

pent0:
Okay, I think of a way to use only one context, and would serve for deferred context later. Use a command queue for a context, when EndScene is called, the command queue will be submitted to the OpenGL thread ? That way it would satisfy what we want. I will go with it

PR #540, opened 2019-07-10
Wait for lock instead of stop CPU and wait later

pent0:
Unicorn startup is slow, so better not stop it.

Improve framerate in debug build for Undertale on some places (like start screen). However the more glyph it draws, the more framerate it drop (to 6fps on my i5, and previously 2fps). I think it's fault somewhere in the rendering system?


pent0:
This?

image

pent0:
App stops so why wait, would that cause deadlock?

frangarcj:
Free unicorn resources. I think it's ok

PR #543, opened 2019-07-11
Riichi 

pent0:
The PR tries to bring all graphics operations to one thread. This surpass many problems like we discussed before:
- Context is created and used on different thread.
- Hard to debug. It should be gone with this PR.

The approach:
- The main core is to use a command queue system.
- Command is demand, in this context, is a graphic request from a client.
- In order to use the GXM API, we need to create a context. There are two kind of that: immediate (?) context and deferred context. Deferred context allows us to record draw commands and let that command list to be used by the immediate context. Currently our implementation makes it hard to do that also.
- Now, each context has a command queue. A command queue is a list of command. Note that command queue can be reusable. When EndScene is called, the command queue will be pushed to the main queue that actually do graphics processing.
- The command queue system also allows us to implement more backend easily.

Some more info for future:
- ExecuteCommandList: This function linked another command list to the current end of the caller's command list.


pent0:
Finished ideal:

Main: 
- PA Hearbeat is used to calculate average scene per frame, so any dev using the emulator is recommend to call heartbeat function when ending a frame.
- Vita3k will try to do rendering with ImGui in main thread, and match a game rendering as close as possible.
- Move components that can be reused: texture cache, shader cache, etc...

Seperation:
- OpenGL: Main thread doing GXM operations. When EndScene is finished, jobs are pushed to CPU. Data that doesn't persists after a scene, including uniform and vertex, are copied and later passed to GPU.
- D3D and Vulkan: Device context or device generally are used only in main thread. The thread doing GXM operations will build up D3D/Vulkan command list/buffer directly, and later queue it for execution. It is an optimization from OpenGL, we can build API command list directly instead of going through another layer.

I like D3D for its tools (can help us debugging shader also?), so I will try to also fiddle with it also in pr

VelocityRa:
Renderdoc has great GL support doesn't it? You sure it's worth it to have a separate backend (with its own bugs/maintainance) just for maybe a few added debug features?

VelocityRa:
If you're gonna write a new backend, why not go for Vulkan? [V-EZ](https://github.com/GPUOpen-LibrariesAndSDKs/V-EZ) can make it quite a lot less complicated. It's worth a look at least.

pent0:
Shader debugging

pent0:
I can use v-ez, looks like directx 11

pent0:
I have to note if you want to try it:
- This has not been merged to master yet. It's on my Riichi PR.
- The TV and text certainly is render good with OpenGL 4.1. If you want effect like TV static or non broken graphics, we are sad to announce you need to have OpenGL 4.3 upper or support these extensions:
    * Fragment Shader Interlock
    * Framebuffer Fetch (intel only, this runs fast)
   * Texture barrier
- The console will log out information about these (wherever your GPU support these extension).

Degerz:
For anyone interested as to why framebuffer fetch is limited to Intel hardware according to their command reference [documentation](https://01.org/sites/default/files/documentation/intel-gfx-prm-osrc-skl-vol02a-commandreference-instructions.pdf) for Gen 9, there's a graphics state register called "3DSTATE_PS_BLEND" on page 143 and in DWord #1/bit #30 the setting is named "Has Writeable RT". 
I assume enabling that bit in the register pretty much allows Intel hardware to bypass the general memory accesses and also allows their hardware to directly supply the values from the pixel shaders to their regular raster paths as implied in relation Andrew Lauritzen's presentation on the Gen 9 graphics architecture.

src/emulator/main.cpp
```diff
@@ -146,19 +147,20 @@ int main(int argc, char *argv[]) {
 
     while (handle_events(host)) {
         gl_renderer.render(host);
+
+        // Driver acto!
+        renderer::process_batches(*host.renderer.get(), host.features, host.mem, host.cfg, host.base_path.c_str(),
+            host.io.title_id.c_str());
+
         gui::draw_begin(gui, host);
+        //gui::draw_display(gui, host.display, host.mem);
```
VelocityRa: ?

src/emulator/app/src/app_init.cpp
```diff
@@ -152,11 +219,73 @@ bool init(HostState &state, Config cfg, const Root &root_paths) {
...
+    if (state.features.direct_fragcolor) {
+        LOG_INFO("Your GPU support direct access to last fragment color. Your performance with programmable blending game will be optimized.");
+    } else if (state.features.support_shader_interlock) {
+        LOG_INFO("Your GPU supports shader interlock, some games use programmable blending will have better performance.");
+    } else if (state.features.support_texture_barrier) {
+        LOG_INFO("Your GPU only supports texture barrier, performance may not be good on programmable blending games.");
+        LOG_INFO("Consider updating to GPU that has shader interlock.");
+    } else {
+        LOG_INFO("Your GPU doesn't support extensions that make programmable blending possible. Some games may have broken graphics.");
+        LOG_INFO("Consider updating your GPU.");
```
VelocityRa:
```suggestion
        LOG_WARN("Consider updating your graphics drivers or upgrading your GPU.");
```

src/emulator/app/src/app_init.cpp
```diff
@@ -152,11 +219,73 @@ bool init(HostState &state, Config cfg, const Root &root_paths) {
...
+        auto find_result = check_extensions.find(extension);
+
+        if (find_result != check_extensions.end()) {
+            *find_result->second = true;
+            check_extensions.erase(find_result);
+        }
+    }
+
+    if (state.features.direct_fragcolor) {
+        LOG_INFO("Your GPU support direct access to last fragment color. Your performance with programmable blending game will be optimized.");
```
VelocityRa:
```suggestion
        LOG_INFO("Your GPU supports direct access to last fragment color. Your performance with programmable blending games will be optimized.");
```

src/emulator/app/src/app_init.cpp
```diff
@@ -152,11 +219,73 @@ bool init(HostState &state, Config cfg, const Root &root_paths) {
...
+    } else if (state.features.support_shader_interlock) {
+        LOG_INFO("Your GPU supports shader interlock, some games use programmable blending will have better performance.");
```
VelocityRa:
```suggestion
        LOG_INFO("Your GPU supports shader interlock, some games that use programmable blending will have better performance.");
```

src/emulator/app/src/app_init.cpp
```diff
@@ -152,11 +219,73 @@ bool init(HostState &state, Config cfg, const Root &root_paths) {
...
+    } else if (state.features.support_texture_barrier) {
+        LOG_INFO("Your GPU only supports texture barrier, performance may not be good on programmable blending games.");
+        LOG_INFO("Consider updating to GPU that has shader interlock.");
```
VelocityRa:
```suggestion
        LOG_WARN("Consider updating to GPU that has shader interlock.");
```

src/emulator/modules/SceGxm/SceGxm.cpp
```diff
@@ -172,7 +192,7 @@ EXPORT(int, sceGxmColorSurfaceInit, emu::SceGxmColorSurface *surface, SceGxmColo
     assert(surface != nullptr);
     assert(colorFormat == SCE_GXM_COLOR_FORMAT_A8B8G8R8);
     assert(surfaceType == SCE_GXM_COLOR_SURFACE_LINEAR);
-    assert(scaleMode == SCE_GXM_COLOR_SURFACE_SCALE_NONE);
+    //assert(scaleMode == SCE_GXM_COLOR_SURFACE_SCALE_NONE);
```
VelocityRa: leftover I assume

src/emulator/modules/SceGxm/SceGxm.cpp
```diff
@@ -441,25 +521,41 @@ EXPORT(int, sceGxmEndCommandList) {
 EXPORT(int, sceGxmEndScene, SceGxmContext *context, const emu::SceGxmNotification *vertexNotification, const emu::SceGxmNotification *fragmentNotification) {
     const MemState &mem = host.mem;
     assert(context != nullptr);
-    assert(vertexNotification == nullptr);
+    //assert(vertexNotification == nullptr);
```
VelocityRa: ditto

src/emulator/main.cpp
pent0: I remember deleted it. Weird

PR #544, opened 2019-07-11
audio: Added initial microphone support.

Rinnegatamante:


src/emulator/audio/src/audio.cpp
```diff
@@ -100,10 +100,19 @@ bool init(AudioState &state, ResumeAudioThread resume_thread) {
...
     }
 
+    SDL_PauseAudio(0);
+
+    SDL_AudioSpec received = {};
+    desired.freq = 16000;
+    desired.format = AUDIO_S16LSB;
+    desired.channels = 1;
+    desired.samples = 256;
+	desired.callback = nullptr;
```
VelocityRa: indentation

src/emulator/audio/src/audio.cpp
```diff
@@ -100,10 +100,19 @@ bool init(AudioState &state, ResumeAudioThread resume_thread) {
...
+	desired.userdata = nullptr;
```
VelocityRa: ditto

Rinnegatamante:
Should be ready for review. Audio is still not perfect (there's some delay for the recording and it seems audio stutters a bit) but it's an initial implementation at least i guess. If anyone has any idea what could cause those two issues, i can try to solve them.

src/emulator/modules/SceAudioIn/SceAudioIn.cpp
```diff
@@ -29,28 +35,92 @@ EXPORT(int, sceAudioInGetMicGain) {
...
+        return RET_ERROR(SCE_AUDIO_IN_ERROR_PORT_FULL);
+    }
+    if (param != SCE_AUDIO_IN_PARAM_FORMAT_S16_MONO) {
+        return RET_ERROR(SCE_AUDIO_IN_ERROR_INVALID_PORT_PARAM);
+    }
+    if (portType != SCE_AUDIO_IN_PORT_TYPE_VOICE && portType != SCE_AUDIO_IN_PORT_TYPE_RAW) {
+        return RET_ERROR(SCE_AUDIO_IN_ERROR_INVALID_PORT_TYPE);
+    }
+    if (portType == SCE_AUDIO_IN_PORT_TYPE_VOICE) {
+        if (freq != 16000) {
```
pent0: it's more great if these constants can be constexpr to be more understandable.

Rinnegatamante: They're just samplerate value/buf size (they're not alias for something strange). How would it make it more understandable?

pent0: DEFAULT_FREQUENCY? or something like that? Either way, It doesn't seems you want to do it, so let's just ignore

Rinnegatamante: It's not really a default frequency. You can start it with 48000 too as samplerate for a RAW port.

pent0: I don't really understand the code, so I was just throwing an example. Naming it would make the code understandable for future contribs who want to look at it, so feel free to modify it on your will. LGTM still

PR #546, opened 2019-07-25
audio: Start audio thread after temp buffer is allocated

petmac:
Fixes race condition where audio thread tries to access buffer before it is allocated.


1whatleytay:
Simple fix!

PR #550, opened 2019-07-29
Enable/disable depth,stencil based on depth stencil surface

pent0:
On every begin of scene.

Should fix quake


pent0:
Closing this for proper PR. I forgot

PR #552, opened 2019-07-29
Enable discord_rpc by default for release build

Zangetsu38:
- Set discord_rpc by default on ON for all compiler, and disable it only for debug build.


VelocityRa:
Why build Discord RPC support on Debug? This is useless.

Zangetsu38:
what ? i said exactly inverse, i enable it only for release build, debug build don't works and fail compile with discord rpc

VelocityRa:
Ok, but figure out a way to do it without uglyfying the ifdefs even more.
Maybe add this [here](https://github.com/Vita3K/Vita3K/blob/f4b10dee2f5e75f8f9e248369cf9a4f664e4584f/src/external/CMakeLists.txt#L156)
```cmake
target_compile_definitions(discord-rpc PUBLIC DISCORD_RPC=defined(USE_DISCORD_RICH_PRESENCE)&&defined(NDEBUG))
```
and change the ifdefs to `#ifdef DISCORD_RPC`.

Zangetsu38:
pushed news commit with finally change why i have all time wanted change, with rename emulator to vita3k,  for get like rpcs3/yuzu etc

VelocityRa:
I don't see why this is necessary, `emulator` fits the rest of our module naming scheme perfectly, and we don't have to do what someone else does anyway.
Please remove so I can merge.

Zangetsu38:
i have not set this for remove :/, and no relate other, i juste already using that on my build, is very better, i don't very like use emulator name, is very poor :/ 
So i have just set for more clear

VelocityRa:
So your arguments are
* rpcs3 (a different project, with a very different file/code structure) does it
* you use it on one of your shitty frankestein builds
* `is very better`
* you don't like it because ` is very poor :/`

I've already explained my reasoning, I'm not going to discuss this further.

Zangetsu38:
:( i just said use vita3k name for project name corresponding exe, because here is emulator, and for me is very strange that all :/ is is not because otherhave, but juste because all other see like yuzu pcsx2 and all other have exactly same, using name emulator for project exe, but here is emulator, i have all time founded that strange.
And i said juste is better for want said more clear, on VS i see vita3k project ok i understund, just that :/

Zangetsu38:
But so here i want option by @petmac and other also, if really is not wanted i revert, but juste is very more clear like that, screen added for more compare and see

petmac:
I'm content for the emulator target to be renamed, but it doesn't really provide much value. All the devs will have to fix their Xcode schemes/VS user settings and stuff to point to the new target.

I think it should be a separate Pull Request though.

Zangetsu38:
i see, i split

Zangetsu38:
ready to merge for only fix link and discord rpc

VelocityRa:
If you're gonna make such change remove the directory entirely, put `external/` in the project root, and `gen-modules` + `native-tool` in a `tools/` subdir.

Zangetsu38:
ok; i set this

Zangetsu38:
rebased and conflict fixed, ready to merge

src/emulator/app/include/app/discord.h
```diff
@@ -1,6 +1,6 @@
 #pragma once
 
-#ifdef USE_DISCORD_RICH_PRESENCE
+#if DISCORD_RPC
```
1whatleytay: What is `DISCORD_RPC`?

Zangetsu38: ? this code is give by velocity, all is right and ready to merge

CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
 option(USE_GDBSTUB "Build Vita3K with GDB Debugger." OFF)
-option(USE_DISCORD_RICH_PRESENCE "Build Vita3K with Discord Rich Presence" OFF)
+option(USE_DISCORD_RICH_PRESENCE "Build Vita3K with Discord Rich Presence" ON)
```
1whatleytay: Ahh I get it. But now this option does nothing.

Zangetsu38: no, this option is used, and tested, if exemple dev want compile release but disable discord rpc ^^ so by default is now on, but you can disable this option if you want :)

1whatleytay: You're right again! Damn.

Zangetsu38:
Is same, like I have see with vélocity and testing works only with if

PR #558, opened 2019-08-02
Orange?

Ristellise:
Context:  
DiscordPTB_2019-08-02_20-57-19-646

The Clear color for `renderer.cpp` is RGB(247,198,0) as asked by velocity to screen pick it. For `screen_render.cpp`, the color is RGB(233,127,2) which is the "orange" color in the logo that is above the dirty magenta.

### DIFF
--- a/src/emulator/app/src/screen_render.cpp
+++ b/src/emulator/app/src/screen_render.cpp
@@ -74,7 +74,7 @@ bool gl_screen_renderer::init(const std::string &base_path) {
     );
     glEnableVertexAttribArray(uvAttrib);
 
-    glClearColor(0.968627450f, 0.776470588f, 0.0f, 1.0f);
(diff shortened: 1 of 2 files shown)


Ristellise:
Done. Changed the color to be the same.

pent0:
Don't. It's to difference between a frame has been drawn vs only clear screen has happen

VelocityRa:
Which one is more frequent? That should be the orange and the other should be the blue you had.

pent0:
I am just saying that the color should be different. Tbh i dont know which one is frequent

VelocityRa:
@Ristellise Can you do some testing to determine that?

Ristellise:
I can't really do testing because I don't really have a Vita. But fron the typical pictures, I see the light blue is more common than the screen render darker blue.

VelocityRa:
Ok, changed it based on that then.

VelocityRa:
👍

PR #559, opened 2019-08-02
travis: Use gcc docker image

scribam:
Replace docker image `ubuntu:18.04` by `gcc:8`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,9 +29,7 @@ matrix:
           tags: false
       script: bash $TRAVIS_BUILD_DIR/.travis/macos/build.sh
     - os: linux
-      sudo: required
       services: docker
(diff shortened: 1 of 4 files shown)


1whatleytay:
Travis is having some strlen error. You might have to change some part of the code to get it working with gcc too?

VelocityRa:
Any reason for this change?

scribam:
- It makes the file `.travis/linux/docker.sh` compiler-agnostic
- It will be easier to modify the compiler if wanted by changing only the docker image
- The `apt install` step should be faster because gcc & related packages are not downloaded (41.2 MB vs 250MB) but the gcc docker image is bigger than ubuntu (408 MB vs 27 MB)

PR #561, opened 2019-08-03
dialog: Disable sceTouch and sceCtrl polling when common dialog is running.

Rinnegatamante:


src/emulator/touch/src/touch.cpp
```diff
@@ -17,6 +17,7 @@
 
 #include <touch/touch.h>
```
VelocityRa: you'll need the include here now

src/emulator/touch/include/touch/touch.h
```diff
@@ -17,10 +17,10 @@
 
 #pragma once
 
-#include <ctrl/state.h>
+#include <host/state.h>
```
VelocityRa: forward declare, no need for the include.
```suggestion
struct HostState;
```

PR #563, opened 2019-08-04
travis: Create GitHub releases

### LINKED ISSUE #413: Deploying macOs/Linux builds as releases
According to GitHub rules, the way we're actually deploying nightlies for macOs and Linux is not compliant to how a repository is meant to be and we could be subject to warnings from GitHub itself.

Some related info:
https://help.github.com/en/articles/distributing-large-binaries
https://help.github.com/en/articles/what-is-my-disk-quota

scribam:
Fix #413, see example here: https://github.com/scribam/Vita3K/releases

Before merge, the deploy api_key need to be setup. You can either use an environment variable (like I did) named `GITHUB_OAUTH_TOKEN` in Travis from a user which has access to the `Vita3K/Vita3K` repository (`scribam/Vita3K` in my case) or, an other solution would be to setup a secure api_key like this:
```
deploy:
  provider: releases
  api_key:
    secure: YOUR_API_KEY_ENCRYPTED
```

Documentation from Travis: https://docs.travis-ci.com/user/deployment/releases/


pent0:
Can we use the secure key currently on master? If not @petmac should be able to add environment variable.

Rinnegatamante:
That secure key should work fine i think.

pent0:
Let's give it everything we got... Sorry, we didn't confirm anything yet, so you have to use the secure key. Hope it's fine.

pent0:
No one against this, so let's merge.

PR #565, opened 2019-08-06
host: removed assert for unknown encryption.

Ristellise:
Pretty self explanatory. Removed Assert.


Ristellise:
I think I'll leave this closed. I wanna tackle implementing encryption.

Ristellise:
Yeah, I was wondering why VS was formatting it that way XD

PR is closed cuz I wanted to get Decryption working. I have an idea based on PS3 self structs i read.

PR #568, opened 2019-08-11
Vulkan GUI

1whatleytay:
This PR aims to add some basic Vulkan functionality. All I want this PR to do at the moment is support Vita3K's gui and provide a decent groundwork for the game renderer.

Screen Shot 2019-08-19 at 4 31 05 PM

To build Vita3K with Vulkan support, go into root CMakeLists.txt and change the `USE_VULKAN` option to `ON`. You will need the Vulkan SDK for the build to succeed. You can now change the backend in your configuration file. Change the `backend` option to `Vulkan` to enable the Vulkan backend. You may have to add the option to your config file. 

Some problems I want to fix before go-go time:

- [x] Full ImGui functionality (graphics and custom images, callbacks, etc.)
- [x] The ability to compile without the Vulkan SDK.
- [x] Fix resizing issues.
- [x] Fix no game drawing in OpenGL.
- [ ] Some testing on different platform to make sure that Vulkan is able to initialize on all vendors.


pent0:
Just make a cmake option for building with Vulkan. When Vulkan renderer backend is stable enough we will remove it and try to include sdk with CI.

pent0:
Well you have done it... Turn it off.. And dont include VMA if vulkan cmake option is off

1whatleytay:
I'm not sure if its that simple. We have vulkan headers everywhere right? So we're have to delete those when the sdk is not there. Then the renderer definitions will be garbage so those will have to be skipped too by preprocessor probably. Then linker will have issues in the switch statements throughout the renderer, we'd have to ifdef those out too.

So now we have a ridiculous amount of `ifdefs` everywhere Vulkan is being used. There has to be a better way. Or maybe I'm overthinking it...

Oops I forgot to turn it off. I'll do it tomorrow.

src/emulator/renderer/src/vulkan/renderer.cpp
```diff
@@ -0,0 +1,574 @@
...
+const static std::vector<const char *> instance_layers = {
+#ifndef NDEBUG
+    "VK_LAYER_LUNARG_standard_validation",
+#endif
+};
+
+const static std::vector<const char *> instance_extensions = {
+    "VK_KHR_surface",
+#ifdef __APPLE__
+    "VK_MVK_macos_surface",
```
Degerz: Just a small recommendation but `VK_MVK_macos_surface` has been depreciated in favour of `VK_EXT_metal_surface` as stated in the MoltenVK runtime [guide](https://github.com/google/angle/blob/master/src/libANGLE/renderer/d3d/d3d11/Renderer11.cpp) so prefer to use that extension instead.

Degerz:
Things to keep in mind for MoltenVK for future reference:

Metal currently does not support a primitive topology known as triangle fans. A workaround would be is to manually generate an index buffer and to transform the triangle fan into a list of triangles. Google's ANGLE project generates an index buffer on the CPU side and then copies it onto the GPU in their D3D11 [backend](https://github.com/google/angle/blob/master/src/libANGLE/renderer/d3d/d3d11/Renderer11.cpp) for their WebGL implementation. Another way to manually generate an index buffer is to use a compute shader instead to fill the buffer rather than generating the indices on the CPU which can potentially save some CPU overhead which is often the bottleneck of emulation. Need a benchmark to see which approach is better in this instance. Also, does the Vita support primitive restart as well ? 

VK_KHR_sampler_mirror_clamp_to_edge is available in MoltenVK on macOS devices so this extension can be enabled automatically without much fuss. 

Vulkan does not directly expose the concept of a "texture barrier" but it does have "resource barriers/renderpasses/subpasses/input attachments" all of which can be used to implement the same functionality [according](https://community.khronos.org/t/is-programmable-blending-with-order-independent-transparency-possible-with-vulkan/6756/7 ) to an AMD engineer. In Metal, it [appears](https://developer.apple.com/documentation/metal/mtlrendercommandencoder/1515866-texturebarrier) as though that Apple has already depreciated the textureBarrier() function since macOS Mojave. It looks as if it was [superceded](
https://developer.apple.com/documentation/metal/mtlrendercommandencoder/3019524-memorybarrier) by the memoryBarrier method. I'm not sure if MoltenVK will be able to handle this case exactly the way you intend but do open an issue in the MoltenVK repository if you experience a problem.

src/emulator/renderer/src/vulkan/renderer.cpp
1whatleytay: Ah I had no idea! Thanks a lot. I think SDL threw an error message earlier asking me to activate this one. It also doesn't appear in my `vulkaninfo` listing... Let me try.

1whatleytay: You are [absolutely right](https://github.com/KhronosGroup/MoltenVK/blob/master/Docs/MoltenVK_Runtime_UserGuide.md)! But I don't seem to have `VK_EXT_metal_surface`... Do I have to enable it or do something special? My Vulkan SDK version happens to be 1.1.108 and I snatched it from LunarG distribution site.

Degerz: Try grabbing the latest version of MoltenVK from the repository and make sure your macOS device is up to date as well ?

Degerz: Yeap, it looks like they only added just [last](https://github.com/KhronosGroup/MoltenVK/commit/aa6a10570a4ce0077120d49dd24c70fd5f5d8481#diff-82d68bf9683270d8d20a27e6da6ad9a3) month in MoltenVK.

1whatleytay: EH! I'm actually 3 updates behind at macOS 10.14.3- that's probably why. I've been holding off updating because 10.14.4 SDK broke OpenGL support and it would stop me from working on this project. I have no idea if it was fixed in .5 or .6.

Just wondering, if I do update and I only use VK_EXT_metal_surface, does that mean Vita3K Vulkan will drop support for macOS versions earlier than Mojave?

1whatleytay: Oh- or not. Maybe its just an SDK thing. Phew.

Degerz: Generally, it depends on what extensions you're going to need that'll bump up the macOS version requirements. For an example, if you're going to require fragment shader interlock which is only available in Metal 2 then your app will need macOS Mojave at minimum with devices that also support that extension as well. 

Apple has a good track record of supporting older devices on Metal anyways so any devices that worked with the initial release of Metal will also work in the upcoming latest Metal 3.0 which will be available in macOS Catalina. If I were a MoltenVK user then I would just upgrade without worrying too much for now since even Metal 3.0 will give you just about everything you would want. 

Hypothetically, as a macOS developer I'd be more concerned if my device supported [raster order groups](https://developer.apple.com/documentation/metal/gpu_features/understanding_gpu_family_4/about_raster_order_groups) or not since it's only available on AMD Vega/Navi GPUs and it'll have a profound impact on emulation as well.

Zangetsu38:
for build fail is normal, for can build myself i have fixed 2 things missing

1whatleytay:
OpenGL renderer seems to be completely broken in this PR. Absolutely no graphics for me anymore.

src/emulator/renderer/src/vulkan/renderer.cpp
Degerz: Sorry to bother again but here are some notes below on how to enable `VK_EXT_metal_surface`.

> You can enable the `VK_EXT_metal_surface` extension by defining the `VK_USE_PLATFORM_METAL_EXT` guard macro in your compiler build settings. See the description of the `mvk_vulkan.h` file below for a convenient way to enable this extension automatically.

> The `mvk_vulkan.h` header file automatically enables the `VK_USE_PLATFORM_METAL_EXT` 
   build setting and `VK_EXT_metal_surface` *Vulkan* extension

.Did you include the MoltenVK header file or use the guard macro in your project yet because I don't see them in this PR ?

1whatleytay: I can guarantee you that I absolutely did not do that. I'll do that in a bit. I made an a special function that checks if you have it in your extension list though :/

I'll include it. Thanks for the heads up.

VelocityRa:
> Also, does the Vita support primitive restart as well?

It does not, afaik.

Degerz:
@VelocityRa Makes things a whole lot easier to emulate the triangle fan primitive topology on MoltenVK (because Metal doesn't support triangle fans) since you don't have to generate a new index buffer for every draw call!

pent0:
😍😍😍 great work. Opengl back to normal again?

vita3k/app/src/app_init.cpp
```diff
@@ -168,37 +105,28 @@ bool init(HostState &state, Config cfg, const Root &root_paths) {
...
-    SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 4);
-    SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_CORE);
-
-    int choosen_minor_version = 0;
-
-    for (int i = 0; i < sizeof(accept_gl_version) / sizeof(int); i++) {
-        SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, accept_gl_version[i]);
+    renderer::Backend backend;
+#ifdef USE_VULKAN
+    if (state.cfg.backend == "Vulkan")
```
pent0: This generally shouldn't belong here. Move it to config.cpp

vita3k/app/src/app_init.cpp
```diff
@@ -168,37 +105,28 @@ bool init(HostState &state, Config cfg, const Root &root_paths) {
...
+#endif
+        backend = renderer::Backend::OpenGL;
 
-        state.glcontext = GLContextPtr(SDL_GL_CreateContext(state.window.get()), SDL_GL_DeleteContext);
-        if (state.glcontext) {
-            choosen_minor_version = accept_gl_version[i];
-            break;
-        }
+    SDL_WindowFlags window_type;
+    switch (backend) {
```
pent0: Default case missing! In this case you should tell bad configuration happens. I want to keep this switch if we ever have future backend, so add a failure on default

vita3k/app/src/app_init.cpp
```diff
@@ -208,104 +136,26 @@ bool init(HostState &state, Config cfg, const Root &root_paths) {
...
-    if (state.cfg.overwrite_config)
-        config::serialize_config(state.cfg, state.cfg.config_path);
-
-    if (!renderer::init(state.renderer, renderer::Backend::OpenGL)) {
+    if (renderer::init(state.window, state.renderer, backend)) {
+        state.renderer->features.hardware_flip = state.cfg.hardware_flip;
+        update_viewport(state);
+        return true;
+    } else {
+        switch (backend) {
```
pent0: Ditto

vita3k/config/src/config.cpp
```diff
@@ -133,6 +133,7 @@ static ExitCode parse(Config &cfg, const fs::path &load_path, const std::string
     get_yaml_value(config_node, "color-surface-debug", &cfg.color_surface_debug, false);
     get_yaml_value(config_node, "hardware-flip", &cfg.hardware_flip, false);
     get_yaml_value(config_node, "performance-overlay", &cfg.performance_overlay, false);
+    get_yaml_value(config_node, "backend", &cfg.backend, std::string("OpenGL"));
```
pent0: Love the string config. People can edit the backend easier in config

vita3k/gui/src/imgui_impl_sdl_vulkan.cpp
```diff
@@ -0,0 +1,880 @@
...
+    uint64_t index_offset = 0;
+
+    for (uint32_t a = 0; a < draw_data->CmdListsCount; a++) {
+        ImDrawList *draw_list = draw_data->CmdLists[a];
+
+        for (const auto &cmd : draw_list->CmdBuffer) {
+            if (cmd.UserCallback) {
+                cmd.UserCallback(draw_list, &cmd);
+            } else {
+//                vk::Rect2D scissor_rect(
```
pent0: Must do scissor test.

vita3k/gui/src/imgui_impl_sdl_vulkan.cpp
```diff
@@ -0,0 +1,880 @@
...
+    state.gui_vulkan.command_buffer.reset(vk::CommandBufferResetFlags());
+
+    vk::CommandBufferBeginInfo begin_info(
+        vk::CommandBufferUsageFlags(), // No Flags
+        nullptr // Inheritance
+    );
+
+    state.gui_vulkan.command_buffer.begin(begin_info);
+
+    float matrix[] = {
```
pent0: Const

vita3k/gui/src/imgui_impl_sdl_vulkan.cpp
```diff
@@ -0,0 +1,880 @@
...
+        vk::AccessFlagBits(), // Will be read by the shader.
+        vk::ImageLayout::eTransferDstOptimal, // Old Layout
+        vk::ImageLayout::eShaderReadOnlyOptimal, // New Layout
+        VK_QUEUE_FAMILY_IGNORED, VK_QUEUE_FAMILY_IGNORED, // Transfer to General Queue (but I don't right now)
+        texture->image, // Image
+        base_subresource_range // Subresource Range
+    );
+
+    transfer_buffer.pipelineBarrier(
+        vk::PipelineStageFlagBits::eTransfer, vk::PipelineStageFlagBits::eBottomOfPipe, // Transfer -> Fragment Shader Stage
```
pent0: You done it... I assume it work on your Mac. Change your comment btw

vita3k/gui/src/imgui_impl_sdl_vulkan.cpp
```diff
@@ -0,0 +1,880 @@
...
+    ImGui_ImplSdlVulkan_DeleteTexture(renderer, state.gui_vulkan.font_texture);
+
+    state.device.destroy(state.gui_vulkan.pipeline);
+    state.device.destroy(state.gui_vulkan.pipeline_layout);
+
+    state.device.destroy(state.gui_vulkan.sampler);
+    for (const vk::Framebuffer &framebuffer : state.gui_vulkan.framebuffers) state.device.destroy(framebuffer);
+    state.device.destroy(state.gui_vulkan.renderpass);
+}
+IMGUI_API bool ImGui_ImplSdlVulkan_CreateDeviceObjects(renderer::State *renderer) {
```
pent0: spacing

vita3k/gui/src/imgui_impl_sdl_vulkan.cpp
```diff
@@ -0,0 +1,880 @@
...
+
+    renderer::vulkan::free_command_buffer(state, renderer::vulkan::CommandType::General, state.gui_vulkan.command_buffer);
+
+    ImGui_ImplSdlVulkan_DeleteTexture(renderer, state.gui_vulkan.font_texture);
+
+    state.device.destroy(state.gui_vulkan.pipeline);
+    state.device.destroy(state.gui_vulkan.pipeline_layout);
+
+    state.device.destroy(state.gui_vulkan.sampler);
+    for (const vk::Framebuffer &framebuffer : state.gui_vulkan.framebuffers) state.device.destroy(framebuffer);
```
pent0: who does this?

vita3k/renderer/include/renderer/gl/state.h
```diff
@@ -0,0 +1,50 @@
...
+
+namespace renderer::gl {
+struct GLState : public renderer::State {
+    GLContextPtr context;
+
+    ShaderCache fragment_shader_cache;
+    ShaderCache vertex_shader_cache;
+    ProgramCache program_cache;
+
+    struct {
```
pent0: What is the reason for this ImGui state existence.. Bring it else where, like ImGuiState with pointer to the render state

vita3k/app/src/app_init.cpp
1whatleytay: Yep. I tried some weird approach in new commit. Is this okay?

vita3k/app/src/app_init.cpp
1whatleytay: Done.

vita3k/app/src/app_init.cpp
1whatleytay: Done!

vita3k/config/src/config.cpp
1whatleytay: Yep. Zang wanted it to be a number. In new commit it is kinda both...

vita3k/gui/src/imgui_impl_sdl_vulkan.cpp
1whatleytay: Are you talking about the `const &`?

1whatleytay:
It took me a while, but the gui anon structs are gone and the code looks much cleaner thanks to that (imo).

1whatleytay:
Build is passing. I think I got everything.

vita3k/config/include/config/config.h
```diff
@@ -57,5 +57,5 @@ struct Config {
     bool color_surface_debug = false;
     bool hardware_flip = false;
     bool performance_overlay = false;
-    std::string backend = {};
+    int backend = -1;
```
Zangetsu38: Why -1 ?

1whatleytay: So command line options won't overwrite the config file options if nothing is specified.

Zangetsu38: set 0 works

1whatleytay: Are you sure you're getting the Vulkan gui and not the OpenGL gui that looks exactly the same? The only difference afaik is the messages in log that give you the GL version, and maybe some graphical differences.

Zangetsu38: and personnaly, after very workis with kd, you can keep std:string wtih write OpenGL, i have already done works gui for works like that

pent0: Leave it as string for user to easier to config. Leave it all lowercase

Zangetsu38: yes ^^ anjd i prepare pr after for implement select backend render

1whatleytay: From my experience it needs an empty value or else po is going to overwrite it. -1 is the empty value for now. Config file can't have -1 so its safe but po can. @Zangetsu38 I'm pretty sure you're just getting the OpenGL gui.

I'll make it lowercase.

1whatleytay: How about std::tolower so the user can enter any combination they please?

Zangetsu38: i get vulkan and opengl with switch in my setting dialog :)

pent0: 😍😍😍😍😍😍😍😍😍😍😍😍😍😍

vita3k/gui/src/imgui_impl_sdl_vulkan.cpp
```diff
@@ -0,0 +1,882 @@
...
+    }
+
+    state.command_buffer.endRenderPass();
+
+    vk::ImageMemoryBarrier color_attachment_present_barrier(
+        vk::AccessFlagBits::eColorAttachmentWrite, // From rendering
+        vk::AccessFlagBits::eMemoryRead, // To readable format for presenting
+        vk::ImageLayout::eColorAttachmentOptimal,
+        vk::ImageLayout::ePresentSrcKHR,
+        VK_QUEUE_FAMILY_IGNORED, VK_QUEUE_FAMILY_IGNORED, // No Transfer
```
1whatleytay: Build is passing again :)

This is one thing we might want to get cleaned up. It seems to work without proper queue families on amd, nvidia and moltenvk. Is it okay to leave it like this?

pent0: Sure, your choice. We can modify later if things turn bad.

CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 
 option(USE_GDBSTUB "Build Vita3K with GDB Debugger." OFF)
 option(USE_DISCORD_RICH_PRESENCE "Build Vita3K with Discord Rich Presence" ON)
+option(USE_VULKAN "Build Vita3K with Vulkan backend." OFF)
```
Zangetsu38: why is off ?

vita3k/app/src/app_init.cpp
```diff
@@ -168,37 +111,24 @@ bool init(HostState &state, Config cfg, const Root &root_paths) {
...
+    switch (backend) {
+    case renderer::Backend::OpenGL:
+        window_type = SDL_WINDOW_OPENGL;
+        break;
+#ifdef USE_VULKAN
+    case renderer::Backend::Vulkan:
+        window_type = SDL_WINDOW_VULKAN;
+        break;
+#endif
+    default:
```
pent0: Where is my break!!! It still stops but would be nice to have.

Zangetsu38:
after test, get regression lot code is wrong now
image
spam log on vulkan, and also, if i set vulkan, it rename it to Vulkan, so after i reboot it restore to OpenGL, you have bad set few things inside config

CMakeLists.txt
1whatleytay: App Veyor doesn't have the Vulkan SDK installed and it won't build.

vita3k/app/src/app_init.cpp
1whatleytay: Dammit, it ran away again!

vita3k/app/include/app/functions.h
```diff
@@ -39,7 +41,7 @@ enum class AppRunType {
 };
 
 bool init(HostState &state, Config cfg, const Root &root_paths);
-void destory(HostState &state);
+void destroy(HostState &host, ImGui_State *imgui);
```
pent0: Forware decalrable. Please do and remove the include

vita3k/app/src/app_init.cpp
```diff
@@ -168,37 +112,25 @@ bool init(HostState &state, Config cfg, const Root &root_paths) {
...
-        3, // OpenGL 4.3
-        2, // OpenGL 4.2
-        1 // OpenGL 4.1
-    };
-
-    SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 4);
-    SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_CORE);
-
-    int choosen_minor_version = 0;
+    auto backend = static_cast<renderer::Backend>(state.cfg.backend);
```
pent0: Can this be modified? If not add const

vita3k/app/src/app_init.cpp
```diff
@@ -208,107 +140,41 @@ bool init(HostState &state, Config cfg, const Root &root_paths) {
...
+            break;
+#endif
+        default:
+            error_dialog(fmt::format("Unknown backend {}.", state.cfg.backend));
+            break;
+        }
         return false;
     }
-
-    return true;
```
pent0: Should return persists here?

vita3k/config/src/config.cpp
```diff
@@ -35,6 +35,13 @@
 namespace po = boost::program_options;
 
 namespace config {
+static std::string to_lower(const std::string &text) {
```
pent0: Util has that? No then put in util module

vita3k/config/src/config.cpp
```diff
@@ -318,6 +347,16 @@ ExitCode init_config(Config &cfg, int argc, char **argv, const Root &root_paths)
         po::store(po::command_line_parser(argc, argv).options(all).positional(positional).run(), var_map);
         po::notify(var_map);
 
+#ifdef USE_VULKAN
+        if (to_lower(backend) == "vulkan")
+            command_line.backend = 1;
+        else
+#endif
+            if (to_lower(backend) == "opengl")
```
pent0: Fucked up

vita3k/gui/include/gui/imgui_impl_sdl_vulkan.h
```diff
@@ -0,0 +1,56 @@
...
+    VmaAllocation transformation_allocation = VK_NULL_HANDLE;
+    vk::Buffer transformation_buffer;
+
+    vk::CommandBuffer command_buffer;
+
+    vk::Semaphore image_acquired_semaphore;
+    vk::Semaphore render_complete_semaphore;
+
+    inline renderer::vulkan::VulkanState &get_renderer() {
+        return dynamic_cast<renderer::vulkan::VulkanState &>(*renderer);
```
pent0: static if you can. Dynamic cast is not really good, but not really bad either

vita3k/gui/src/imgui_impl_sdl_vulkan.cpp
```diff
@@ -0,0 +1,879 @@
...
+    state.command_buffer.bindPipeline(vk::PipelineBindPoint::eGraphics, state.pipeline);
+    state.command_buffer.bindDescriptorSets(
+        vk::PipelineBindPoint::eGraphics, // Bind Point
+        state.pipeline_layout, // Layout
+        0, // First Set
+        1, &state.matrix_set, // Sets
+        0, nullptr // Dynamic Offsets
+    );
+
+    //    vk::Viewport viewport(0, 0, draw_data->DisplaySize.x, draw_data->DisplaySize.y, 0.0f, 1.0f);
```
pent0: !

vita3k/config/src/config.cpp
1whatleytay: You mean the indenting? Clang-format did this so I though I'd trust it :/

I agree it looks ugly... I'll try to figure something out.

vita3k/gui/include/gui/imgui_impl_sdl_vulkan.h
1whatleytay: I'll make it static (although I like the whole get style like this like in SceGxmProgram). What do you want me to use for cast instead?

pent0: No i mean static cast

vita3k/app/src/app_init.cpp
1whatleytay: I don't think it will be reached 😅

pent0: Compiler will remember that

1whatleytay: Okay, I'll add it just in case.

1whatleytay:
I think I got everything. Is this okay?

vita3k/config/include/config/config.h
```diff
@@ -57,5 +57,5 @@ struct Config {
     bool color_surface_debug = false;
     bool hardware_flip = false;
     bool performance_overlay = false;
-
+    int backend = 0;
```
Zangetsu38: forget, not remove news line here

vita3k/config/src/config.cpp
```diff
@@ -249,6 +266,8 @@ void merge_configs(Config &lhs, const Config &rhs, const std::string &new_pref_p
         lhs.hardware_flip = rhs.hardware_flip;
     if (lhs.performance_overlay != rhs.performance_overlay && (!init || rhs.performance_overlay))
         lhs.performance_overlay = rhs.performance_overlay;
+    if (lhs.backend != rhs.backend && (!init || rhs.backend != -1))
```
Zangetsu38:
```suggestion
    if (lhs.backend != rhs.backend && (!init || rhs.backend != 0))
```
i no understund this -1 is totaly wrong, is "0"

CMakeLists.txt
Zangetsu38: for this you need serach solution, rpcs3 no have any probleme

vita3k/config/src/config.cpp
Zangetsu38:
```suggestion
    if (lhs.backend != rhs.backend && (!init || rhs.backend != static_cast<int>(0)))
```
forget static cast and is 0 no -1

1whatleytay: No, that's wrong!

5583755dba065d80218d0a3634c8d6195d6543d1v2_hq

This works, and changing it will break the config.yml file and overwrite it 100% of the time with an empty command line. I am sure of this. `-1` is the empty option, the config.yml file will never have it, but the command line might use it to say that you didn't specify the backend via `-B backend`.

If you want to specify another method to tackle this, go ahead- but changing this value to 0 will just create issues.

... sorry I've been playing too much danganronpa.

vita3k/config/include/config/config.h
1whatleytay: Is this really necessary? Most structs like `HostState` and `GuiState`, etc. don't have new lines are the end of definitions.

Zangetsu38: all struct i have see have this news line hmm

CMakeLists.txt
1whatleytay: Vulkan isn't useful enough right now for the trouble of setting up sdks everywhere. Not everyone will need it- or can use it right now.

I was thinking of doing this in another pr.

vita3k/config/src/config.cpp
Zangetsu38: hmm if you said ^^

CMakeLists.txt
Zangetsu38: yes, for that i am right ^^

vita3k/config/include/config/config.h
1whatleytay: Here are my examples:
https://github.com/Vita3K/Vita3K/blob/master/vita3k/host/include/host/state.h#L92
https://github.com/Vita3K/Vita3K/blob/master/vita3k/gui/include/gui/state.h#L130
https://github.com/Vita3K/Vita3K/blob/master/vita3k/kernel/include/kernel/state.h#L201
etc.

I think this newline was a mistake?

vita3k/config/src/config.cpp
pent0: https://www.who.int/features/qa/gaming-disorder/en/

vita3k/shader/include/shader/spirv_recompiler.h
```diff
@@ -1,11 +1,12 @@
...
 #include <gxm/types.h>
 #include <psp2/gxm.h>
 #include <shader/usse_translator_types.h>
 #include <shader/usse_types.h>
 
+#include <features/state.h>
+
 #include <string>
 #include <utility>
 #include <vector>
```
pent0: Shuichi not wrong image wastes my 4g, though this pr is awesome anyway. Thanks.

vita3k/config/include/config/config.h
```diff
@@ -57,5 +57,5 @@ struct Config {
     bool color_surface_debug = false;
     bool hardware_flip = false;
     bool performance_overlay = false;
-
+    std::string backend = {};
```
Zangetsu38: change here
```suggestion
    std::string backend_render = "OpenGL";
```

vita3k/config/src/config.cpp
```diff
@@ -249,6 +251,8 @@ void merge_configs(Config &lhs, const Config &rhs, const std::string &new_pref_p
         lhs.hardware_flip = rhs.hardware_flip;
     if (lhs.performance_overlay != rhs.performance_overlay && (!init || rhs.performance_overlay))
         lhs.performance_overlay = rhs.performance_overlay;
+    if (lhs.backend != rhs.backend && (!init || !rhs.backend.empty()))
```
Zangetsu38: change here also by
```suggestion
      if (lhs.backend_render != rhs.backend_render && (!init || rhs.backend_render != static_cast<std::string>("OpenGL")))
```

vita3k/app/src/app_init.cpp
```diff
@@ -168,37 +113,28 @@ bool init(HostState &state, Config cfg, const Root &root_paths) {
...
-        2, // OpenGL 4.2
-        1 // OpenGL 4.1
-    };
-
-    SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 4);
-    SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_CORE);
+    renderer::Backend backend = renderer::Backend::OpenGL;
 
-    int choosen_minor_version = 0;
+    if (string_utils::toupper(state.cfg.backend) == "VULKAN")
```
Zangetsu38: forget add "#ifdef USE_VULKAN" ?

vita3k/config/include/config/config.h
1whatleytay: I'm not doing that. It'll screw up the command line and you wouldn't be able to switch to Vulkan vis config.yml. This needs (should) to be empty by default!

vita3k/config/src/config.cpp
1whatleytay: I don't like OpenGL being the empty option.

vita3k/app/src/app_init.cpp
1whatleytay: Thank you!

vita3k/config/include/config/config.h
Zangetsu38: no, sorry, no set empty, no have any sense, is obligatory one things writed, is OpenGL or Vulkan, but never empty, and no broken command line is same all other

vita3k/config/src/config.cpp
Zangetsu38: set OpenGL include config and set like that here

vita3k/config/include/config/config.h
1whatleytay: We set the default [here](https://github.com/Vita3K/Vita3K/pull/568/files#diff-a6a8977717cb5a4e9908c4f9b0256552R135). app_init.cpp makes sure that you select either. I really don't think we need this.

Zangetsu38: yes, i know, just is more sense and more good writed, default value is also need for create file no forget, for not set like i have myself forget set before and get icon size on 0 on create config lol set value is needed

Zangetsu38: i have all time right haha, exactly like i have explain, is fail config file, keep backend entry empty, write "OpenGL" here
image

PR #571, opened 2019-08-15
Fix sceIoLseek32

yousifd:
The recent io refactor (97183ffc1bbc1876afabeabf698e83ba7f9cc75b) mistakingly changed the offset type from int to SceOff which is defined as SceInt64 which causes issues when calling the 32 bit version of sceIoLseek32 but works fine on sceIoLseek (64 bit version).

Function prototype https://docs.vitasdk.org/group__SceFcntlUser.html#gaca489d7dcc95d9ee8f85248601c2102c

### DIFF
--- a/src/emulator/modules/SceIofilemgr/SceIofilemgr.cpp
+++ b/src/emulator/modules/SceIofilemgr/SceIofilemgr.cpp
@@ -207,7 +207,7 @@ EXPORT(int, sceIoGetstatByFdAsync) {
     return UNIMPLEMENTED();
 }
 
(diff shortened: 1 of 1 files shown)


pent0:
Thanks. Sorry for this weird request, can you change it to int32_t. Just want to be sure the param is always 4 bytes even in future

pent0:
I know the code around does use "int" frequently, but i hope we can agree to change them in future to int32_t

pent0:
Sorry, I'm too hesitate to merge this. If it fails because of type not exist, open a PR include "cstdint" header. I'm so sorry.

PR #575, opened 2019-08-24
external: Force Windows to use Vita3K's SDL2 include directory

totlmstr:
Quick commit.

`find_package(SDL2 REQUIRED)` on Windows may point the include directories to an incorrect location (e.g., in an MSYS2 environment, it may get MSYS2's SDL2, which *cannot* be built on Windows directly).

This commit forces CMake to use Vita3K's own SDL2 include directory on Windows instead.


1whatleytay:
Okay! This is good. Can you tackle the conflict and then we can merge?

totlmstr:
Apologies for that delay.

Anyway, here's that rebase. @1whatleytay

PR #578, opened 2019-08-29
Nurses in Resident Evil

pent0:
Correct the viewport formula in flipped case. I think this is correct, if it đoesnt work with both steins gate and zero i will close


1whatleytay:
Well if it's been tested merge time. Changes are good.

pent0:
Not really, I will wait for test.

pent0:
I will dooo

PR #579, opened 2019-09-06
external: update SPIRV-Cross submodule

1whatleytay:
This should fix the SPIRV-Cross module issues... Updated it to latest on Vita3K/SPIRV-Cross.


mailwl:
sorry, fixed

PR #580, opened 2019-09-08
Shadow

pent0:
- Use a memory barrier between draw call on shader interlock, since write of last draw call wasnt visible to new draw call. Interlock only works between invocations in a draw calls. Fix artifacts on RTX.
- Prepare for Vulkan!:
    * [x] Basic default uniform buffer generation
    * [x] Do nice and proper uniform buffer alignment as specified in GXM library
    * [x] Custom uniform buffer generation
    * [x] Implement uniform buffer update on OpenGL
- For Quake: Plan to add clip planes. (TODO)

I only plan to do this....


vita3k/renderer/src/gl/renderer.cpp
```diff
@@ -275,26 +275,14 @@ bool create(WindowPtr &window, std::unique_ptr<State> &state) {
 
     // Workaround driver bugs
     const std::string gpu_name = reinterpret_cast<const GLchar *>(glGetString(GL_RENDERER));
```
Zangetsu38: You can also delete this 2 line above :)

pent0:
The PR introduce a OpenGL requirements move to 4.3 by using SSBO to implement writeable uniform buffer, which is used by quite some AAA games.

If anyone agrees, I will split up the uniform buffer part and try to merge it first. This will not move the requirements and speed up Vulkan implementation, since this PR will be on hold for more decision if SSBO will be added.

VelocityRa:
Don't let old hardware & unmaintained drivers hold you back.
4.3 is more than an acceptable minimum in 2019.

petmac:
Mac only supports 4.1 😬 
https://support.apple.com/en-us/HT202823

VelocityRa:
@petmac macOS also doesn't/won't support OpenGL at all eventually.
And desgroup is working on a Vulkan backend anyway so at least it'll probably work with that.

1whatleytay:
I think it might be flat out impossible to emulate gxm with only OpenGL 4.1... but I'd rather have games work crappily than not work at all.

I just started working on Vulkan again yesterday, it will probably be a while until it runs any games. Still trying to get gl_screen_renderer out of the way. I'm up for adding this as an option in the meantime (and just discarding stores)!

Is `use_ubo` a selection between UBO and SSBO or between UBO and nothing?

pent0:
The second one. If use UBO is not enable, the shadergen will emits separate uniforms variables from debug symbols. Else it will just emit a chunk of vec4 as uniform buffer.

vita3k/renderer/src/renderer.cpp
```diff
@@ -184,4 +184,19 @@ void set_uniform(State &state, Context *ctx, const bool is_vertex_uniform, const
         break;
     }
 }
+
+void set_uniform_buffer(State &state, Context *ctx, const bool is_vertex_uniform, const int block_number, const std::uint16_t block_size, const void *data) {
+    switch (state.current_backend) {
+    default: {
```
1whatleytay: For both Vulkan and OpenGL?

vita3k/util/include/util/bytes.h
```diff
@@ -42,3 +45,5 @@ T network_to_host_order(T val) {
 
     return byte_swap(val);
 }
+
+void float_to_half(const float *src, std::uint16_t *dest, const int total);
```
1whatleytay: New line.

1whatleytay:
good luck :)

pent0:
niceu

PR #581, opened 2019-09-09
[WIP] Add choice backend render in gpu tab.

Zangetsu38:
- Add backend renderer choice inside gpu tab on setting dialog.

# Result : 
image

### DIFF
--- a/vita3k/config/src/config.cpp
+++ b/vita3k/config/src/config.cpp
@@ -252,8 +252,14 @@ void merge_configs(Config &lhs, const Config &rhs, const std::string &new_pref_p
         lhs.hardware_flip = rhs.hardware_flip;
     if (lhs.performance_overlay != rhs.performance_overlay && (!init || rhs.performance_overlay))
         lhs.performance_overlay = rhs.performance_overlay;
-    if (lhs.backend_renderer != rhs.backend_renderer && (!init || rhs.backend_renderer != std::string("OpenGL")))
-        lhs.backend_renderer = rhs.backend_renderer;
+    if (lhs.backend_renderer != rhs.backend_renderer) {
+#ifdef USE_VULKAN
+        if (!init || rhs.backend_renderer != std::string("OpenGL"))
+#else
+        if (!init || lhs.backend_renderer != std::string("OpenGL"))
+#endif
+            lhs.backend_renderer = rhs.backend_renderer;
+    }
 
     // Not stored in config file
     if (init) {
@@ -289,7 +295,9 @@ ExitCode init_config(Config &cfg, int argc, char **argv, const Root &root_paths)
         po::options_description config_desc("Configuration");
         config_desc.add_options()
             ("archive-log,A", po::bool_switch(&command_line.archive_log), "Makes a duplicate of the log file with TITLE_ID and Game ID as title")
+#ifdef USE_VULKAN
             ("backend-renderer,B", po::value(&command_line.backend_renderer), "Renderer backend to use, either \"OpenGL\" or \"Vulkan\"")
+#endif
             ("color-surface-debug, C", po::bool_switch(&command_line.color_surface_debug),"Save color surfaces")
             ("config-location,c", po::value<fs::path>(&command_line.config_path), "Get a configuration file from a given location. If a filename is given, it must end with \".yml\", otherwise it will be assumed to be a directory. \nDefault: <Vita3K folder>/config.yml")
             ("keep-config,w", po::bool_switch(&command_line.overwrite_config)->default_value(true), "Do not modify the configuration file after loading.")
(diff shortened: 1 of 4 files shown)


vita3k/config/src/config.cpp
```diff
@@ -336,8 +344,10 @@ ExitCode init_config(Config &cfg, int argc, char **argv, const Root &root_paths)
             cfg.recompile_shader_path = std::move(command_line.recompile_shader_path);
             return QuitRequested;
         }
-        if (var_map.count("backend-renderer"))
-            cfg.backend_renderer = command_line.backend_renderer;
+#ifdef USE_VULKAN
+        if (var_map.count("backend-renderer") && string_utils::toupper(command_line.backend_renderer) == "OPENGL")
+            cfg.backend_renderer = std::string("OpenGL");
```
1whatleytay: Why did we change how this worked? (I liked the USE_VULKAN though.)

vita3k/config/src/config.cpp
```diff
@@ -252,8 +252,14 @@ void merge_configs(Config &lhs, const Config &rhs, const std::string &new_pref_p
...
         lhs.performance_overlay = rhs.performance_overlay;
-    if (lhs.backend_renderer != rhs.backend_renderer && (!init || rhs.backend_renderer != std::string("OpenGL")))
-        lhs.backend_renderer = rhs.backend_renderer;
+    if (lhs.backend_renderer != rhs.backend_renderer) {
+#ifdef USE_VULKAN
+        if (!init || rhs.backend_renderer != std::string("OpenGL"))
+#else
+        if (!init || lhs.backend_renderer != std::string("OpenGL"))
+#endif
+            lhs.backend_renderer = rhs.backend_renderer;
```
1whatleytay: Why is this different. This whole thing should be wrapped in a USE_VULKAN statement I think. It's not clear why changing `lhs` to `rhs` in a vulkan build will help accomplish the goal.

vita3k/host/include/host/state.h
```diff
@@ -67,6 +67,7 @@ struct HostState {
     uint32_t ms_per_frame = 0;
     bool should_update_window_title = false;
     WindowPtr window;
+    renderer::Backend backend_renderer;
```
1whatleytay: You can access the backend_renderer field via `renderer->backend`. Having a copy in host is kinda unnecessary and opens up the possibility of getting out of sync.

vita3k/gui/src/settings_dialog.cpp
```diff
@@ -155,6 +155,15 @@ void draw_settings_dialog(GuiState &gui, HostState &host) {
     ImGui::PushStyleColor(ImGuiCol_Text, GUI_COLOR_TEXT_MENUBAR_OPTIONS);
     if (ImGui::BeginTabItem("GPU")) {
         ImGui::PopStyleColor();
+#ifdef USE_VULKAN
+        ImGui::Combo("Backend Render (Reboot for apply)", reinterpret_cast<int *>(&host.backend_renderer), "OpenGL\0Vulkan\0");
```
1whatleytay: Only change `Config`'s `backend_renderer` field. If you need a variable to hold the Combo's state, make one in `gui::State`.

vita3k/gui/src/settings_dialog.cpp
```diff
@@ -155,6 +155,15 @@ void draw_settings_dialog(GuiState &gui, HostState &host) {
...
+        if (ImGui::IsItemHovered())
+            ImGui::SetTooltip("Select your preferred backend render.");
+        if (static_cast<int>(host.backend_renderer) == 0 && host.cfg.backend_renderer != "OpenGL")
```
1whatleytay: This would be better as switch cases I think.

```cpp
switch (backend_renderer) {
case renderer::Backend::OpenGL:
	...
	break;
case renderer::Backend::Vulkan:
	...
	break;
}
```

vita3k/host/include/host/state.h
Zangetsu38: i can't, use it causing crash, bercause it said i can't initialise value by app_init

vita3k/gui/src/settings_dialog.cpp
Zangetsu38: yes, i know, just i am not fan using case juste for 2 value ^^

1whatleytay: We will probably have more renderers in the future...

vita3k/host/include/host/state.h
1whatleytay: Keep `renderer->backend` the same for the entire execution. Don't change it- the only place that should change it is `renderer::init`.

vita3k/gui/src/settings_dialog.cpp
Zangetsu38: hmm yes, hmm exept dx12 no idea what ^^

vita3k/host/include/host/state.h
Zangetsu38: yes i know, but i canb't using it is my problem, it is impossible affect for setting anhd for init, both causing crash

vita3k/config/src/config.cpp
Zangetsu38: for here is very special, my test have shown, it return all time on ogl, so i have founded special things can get works

1whatleytay:
Going down the pr list. Is this going to be finished? :D

Zangetsu38:
ah yes for it, hmm i close it for now lol

PR #585, opened 2019-09-17
Try to fix segment fault on macOS.

orklann:
1. Calling mmap() before calling mprotect() i
2. Changing PROT_NONE to PROT_READ | PROT_WRITE for mmap()

### DIFF
--- a/vita3k/host/src/load_self.cpp
+++ b/vita3k/host/src/load_self.cpp
@@ -42,6 +42,7 @@
 #include <iomanip>
 #include <iostream>
 #include <fstream>
+#include <sys/mman.h>
 
 #define ET_SCE_EXEC 0xFE00
 
--- a/vita3k/host/src/load_self.cpp
+++ b/vita3k/host/src/load_self.cpp
@@ -42,7 +42,6 @@
 #include <iomanip>
 #include <iostream>
 #include <fstream>
-#include <sys/mman.h>
 
 #define ET_SCE_EXEC 0xFE00
 
--- a/vita3k/mem/src/mem.cpp
+++ b/vita3k/mem/src/mem.cpp
@@ -81,7 +81,11 @@ bool init(MemState &state) {
 #else
     // http://man7.org/linux/man-pages/man2/mmap.2.html
     void *const addr = nullptr;
+#ifdef __APPLE__
     const int prot = PROT_READ | PROT_WRITE;
+#else
+    const int prot = PROT_NONE;
+#endif
     const int flags = MAP_PRIVATE | MAP_ANONYMOUS;
     const int fd = 0;
     const off_t offset = 0;
--- a/vita3k/mem/src/mem.cpp
+++ b/vita3k/mem/src/mem.cpp
@@ -43,8 +43,7 @@ static void delete_memory(uint8_t *memory) {
 }
 
 static void alloc_inner(MemState &state, Address address, size_t page_count, Allocated::iterator block, const char *name) {
-    //uint8_t *const memory = &state.memory[address];
-    uint8_t * memory = &state.memory[address];
+    uint8_t *const memory = &state.memory[address];
     const size_t aligned_size = page_count * state.page_size;
 
     const Generation generation = ++state.generation;
--- a/vita3k/mem/src/mem.cpp
+++ b/vita3k/mem/src/mem.cpp
@@ -57,10 +57,6 @@ static void alloc_inner(MemState &state, Address address, size_t page_count, All
     const void *const ret = VirtualAlloc(memory, aligned_size, MEM_COMMIT, PAGE_READWRITE);
     LOG_CRITICAL_IF(!ret, "VirtualAlloc failed: {}", log_hex(GetLastError()));
 #else
-    void *addr = mmap(NULL, aligned_size, PROT_READ | PROT_WRITE, MAP_ANON | MAP_PRIVATE, -1, 0);
-    if (addr == MAP_FAILED)
-        printf("mmap failed\n");
-    memory = (uint8_t *)addr;
     mprotect(memory, aligned_size, PROT_READ | PROT_WRITE);
 #endif
     std::memset(memory, 0, aligned_size);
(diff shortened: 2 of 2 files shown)


vita3k/host/src/load_self.cpp
```diff
@@ -42,6 +42,7 @@
 #include <iomanip>
 #include <iostream>
 #include <fstream>
+#include <sys/mman.h>
```
pent0: ?

pent0: Don't include unnecessary stuff

vita3k/mem/src/mem.cpp
```diff
@@ -56,6 +57,10 @@ static void alloc_inner(MemState &state, Address address, size_t page_count, All
     const void *const ret = VirtualAlloc(memory, aligned_size, MEM_COMMIT, PAGE_READWRITE);
     LOG_CRITICAL_IF(!ret, "VirtualAlloc failed: {}", log_hex(GetLastError()));
 #else
+    void *addr = mmap(NULL, aligned_size, PROT_READ | PROT_WRITE, MAP_ANON | MAP_PRIVATE, -1, 0);
```
pent0: I would like to remove the added code and try again. I think the PROT_READ | PROT_WRITE added lower will make it work eventually

pent0: This is also allocating new pages. We already reserved some.

orklann: > This is also allocating new pages. We already reserved some.

I think `void *addr = mmap(memory, aligned_size, PROT_READ | PROT_WRITE, MAP_ANON | MAP_PRIVATE, -1, 0);` should work too, so that it does not allocate new pages?

vita3k/host/src/load_self.cpp
orklann: yeah, sorry, I did not notice I include this header, just forgot to remove it after debug. It does not matter to remove it.

vita3k/mem/src/mem.cpp
pent0: I am not confident about the behavior of that call again. mprotect should do the job. 

If you do it like that, it's like reserve pages at fixed address, which by somehow works on Mac although we already reserved that region.

Just try to remove the new code I tell you.

1whatleytay: On my computer, [this](https://github.com/Vita3K/Vita3K/pull/585/files#diff-98fe52b4d9c7ab8753d68aaaa13e765bR89) change does the job.

orklann: Yep, just change to `PROT_READ | PROT_WRITE` will do the job, I am finally testing the result, and will make a new commit.

orklann: Ok, have made a new commit.

vita3k/mem/src/mem.cpp
```diff
@@ -43,7 +43,8 @@ static void delete_memory(uint8_t *memory) {
 }
 
 static void alloc_inner(MemState &state, Address address, size_t page_count, Allocated::iterator block, const char *name) {
-    uint8_t *const memory = &state.memory[address];
+    //uint8_t *const memory = &state.memory[address];
+    uint8_t * memory = &state.memory[address];
```
1whatleytay: I'm not sure if we need this either. I was okay with the header change btw!

orklann: Yep, this should not be changed. :D

orklann: Hey, May I ask what to put in commit message in this case. I don't know how to name this commit message, XD. Any suggestion?

orklann: OK, just make a third commit.

1whatleytay: Awesome. We usually name our commits like this `module: what commit does`. I would've probably named the whole thing something like `mem: fix broken protects on macos`.

We'll probably squash all 3 commits before merging if that's okay :)

orklann: OK, that sounds great.

pent0:
Don't like this at all, giving too much power for 4GB but Mac like it so I guess we should do this for now, if you can confirm it work.

pent0:
For future visitors: Mac don't like PROT_NONE. [Mac OS document](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/mmap.2.html) doesn't even list PROT_NONE as supported, so I think at least we can't gurantee stability from PROT_NONE.

orklann:
> Sorry again, can I ask for this change to Mac specific? Use a macro on mac build.

I just make a fourth commit which make this PR macOS specific.

pent0:
Thank you very much. If @1whatleytay has no objection (errno himself on discord), i will merge.

orklann:
OK. Happy to help.

1whatleytay:
```
     16 EBUSY Resource busy.  An attempt to use a system resource which was in
             use at the time in a manner which would have conflicted with the
             request.
```

... what?

1whatleytay:
If this doesn't look solvable by error code, merge away. I don't know anything about unix error codes.

pent0:
0x16=22=EINVAL. It's not 16. Invalid arguments. Nothing seems wrong so I guess Mac throws a tantrum from undefined behavior with PROT_NONE idk. Please merge if you agree, got to study

1whatleytay:
No that's wrong! 
52b83349eec52ca77e827fe78a068d1dba65fcf5r1-2048-1152v2_uhq

I screwed up the hexes but the error is different.
`
     [EINVAL]           addr is not a multiple of the page size.
`

Something's up with the address. I think we should fix the multiple issue instead of giving prot access to everything...

1whatleytay:
```
    uintptr_t pointer_diff = reinterpret_cast<uintptr_t>(memory) % getpagesize();

#ifdef WIN32
    const void *const ret = VirtualAlloc(memory, aligned_size, MEM_COMMIT, PAGE_READWRITE);
    LOG_CRITICAL_IF(!ret, "VirtualAlloc failed: {}", log_hex(GetLastError()));
#else
    mprotect(memory - pointer_diff, aligned_size + pointer_diff, PROT_READ | PROT_WRITE);
#endif
    std::memset(memory - pointer_diff, 0, aligned_size + pointer_diff);
```

This forcefully aligns memory for mem protect and adds a bit of extra size just to be safe. Seems to work for me.
Screen Shot 2019-09-17 at 12 38 16 PM

1whatleytay:
Closing since issue is fixed. :)

Issue #584, opened 2019-09-16
Mac version crash on macOS 10.14.6 after loading a game


orklann:
The Mac version runs well, but after loading `Alone With You`, it crashes, here is the logging message from the terminal.

```
/Users/aaron/Downloads/Vita3K.app/Contents/MacOS/Vita3K
[18:27:08.427] |I| [init_config]: backend-renderer: OpenGL
[18:27:08.427] |I| [init_config]: hardware-flip: false
[18:27:08.428] |I| [init_config]: log-level: 0
[18:27:08.428] |I| [init_config]: log-imports: false
[18:27:08.428] |I| [init_config]: log-exports: false
[18:27:08.428] |I| [init_config]: log-active-shaders: false
[18:27:08.428] |I| [init_config]: log-uniforms: false
[18:27:08.428] |I| [main]: Vita3K v0.1 100-daff95f
[18:27:10.528] |I| [create]: Swap interval = 1
[18:27:10.537] |I| [create]: GPU = Intel(R) Iris(TM) Plus Graphics 640
[18:27:10.537] |I| [create]: GL_VERSION = 4.1 INTEL-12.10.12
[18:27:10.537] |I| [create]: GL_SHADING_LANGUAGE_VERSION = 4.10
[18:27:10.537] |I| [create]: Your GPU doesn't support extensions that make programmable blending possible. Some games may have broken graphics.
[18:27:10.537] |W| [create]: Consider updating your graphics drivers or upgrading your GPU.
[18:27:10.745] |W| [init_icons]: Default icon not found for title , .
[18:27:11.886] |I| [load_app_impl]: Title: Alone With You
[18:27:11.886] |I| [load_app_impl]: Serial: PCSB00964
[18:27:11.886] |I| [load_app_impl]: Version: 01.00
[18:27:11.886] |I| [load_app_impl]: Category: gd
[18:27:11.906] |I| [load_self]: Loading symbols for SELF: app0:sce_module/libc.suprx
[18:27:11.907] |D| [load_var_exports]: 	NID 0x936C8A78 (SYSLYB) at 0x48314
[18:27:11.907] |D| [load_var_exports]: 	NID 0x6C2224BA (NID_MODULE_INFO) at 0x3D4DC
[18:27:11.907] |W| [load_var_imports]: 	NID NOT FOUND 0x93B8AA67 (__stack_chk_guard) at 0x4163C, setting to stub value 0xDEADBEEF
[...]

Issue #582, opened 2019-09-11
Assert on debug build


mailwl:
assert in https://github.com/Vita3K/Vita3K/blob/master/vita3k/config/src/config.cpp#L293

unneeded space between long and short options "color-surface-debug, C"

1whatleytay:
That causes an assert!? I saw it when I was adding config options but I didn't think it would cause an issue. Thanks for the heads up!

mailwl:
Ah, yes, assert itself here: https://github.com/Vita3K/ext-boost/blob/6b780f14b64013e601fb8ada0626855a39de07ee/libs/program_options/src/options_description.cpp#L177

1whatleytay:
Fixed in new PR.

PR #586, opened 2019-09-18
Memory, Gui and Config tuneup

### LINKED ISSUE #576: Regular files should not be detected as games.
macOS creates ".DS_Store" files in every folder I open with Finder. They're usually invisible to see and can be frustrating to remove.

Vita3K seems to interpret regular files in ux0/app as actual games (which it shouldn't) and tries to load them anyway. This tends to show up on my game selection screen:

Screen Shot 2019-08-28 at 6 55 35 PM

This could probably be fixed by adding a check if each game is a directory via [`fs::is_directory`](https://www.boost.org/doc/libs/1_63_0/libs/filesystem/doc/reference.html#is_directory) before loading it.

1whatleytay:
Fixes #576 (sorry `N/A` was really bugging me), #582 and #584.

We know mem size is going to be aligned (although I don't think it needs to be) because it's a product of `state.page_size`. I think velo told me to use align up, I think he meant align down. I copied boost's implementation.


vita3k/mem/src/mem.cpp
```diff
@@ -43,7 +43,14 @@ static void delete_memory(uint8_t *memory) {
 }
 
 static void alloc_inner(MemState &state, Address address, size_t page_count, Allocated::iterator block, const char *name) {
-    uint8_t *const memory = &state.memory[address];
+    Address alignment;
```
petmac: I think the address should already be aligned, so this should have no effect.

vita3k/mem/src/mem.cpp
```diff
@@ -43,7 +43,14 @@ static void delete_memory(uint8_t *memory) {
...
+#ifdef WIN32
+    alignment = 4096;
+#else
+    alignment = getpagesize();
+#endif
```
petmac: The MemState should have the correct page size already.

vita3k/mem/src/mem.cpp
```diff
@@ -56,7 +63,7 @@ static void alloc_inner(MemState &state, Address address, size_t page_count, All
     const void *const ret = VirtualAlloc(memory, aligned_size, MEM_COMMIT, PAGE_READWRITE);
     LOG_CRITICAL_IF(!ret, "VirtualAlloc failed: {}", log_hex(GetLastError()));
 #else
-    mprotect(memory, aligned_size, PROT_READ | PROT_WRITE);
+    assert(mprotect(memory, aligned_size + 1, PROT_READ | PROT_WRITE) == 0);
```
petmac: The mprotect call will get discarded in release builds. Also, what is the + 1 for?

vita3k/mem/src/mem.cpp
1whatleytay: Does `state.page_size` have the ps vita page size or the host's page size? I thought it was the vita's...

vita3k/mem/src/mem.cpp
1whatleytay: Nope :/

Some elfs like "Alone with You" seem to have an odd `p_vaddr` value that are not aligned to 4096.

vita3k/mem/src/mem.cpp
1whatleytay: You're right! I've been bamboozled by assert again!

As for the + 1... It seems `aligned_size` wasn't enough for some reason after I aligned memory. I would have a bad access somewhere in unicorn or `load_elf`.

The +1 was just a guess... problems are gone. I don't know why it needs more space!

1whatleytay:
Fixed evil assert.

vita3k/config/src/config.cpp
```diff
@@ -290,7 +290,7 @@ ExitCode init_config(Config &cfg, int argc, char **argv, const Root &root_paths)
         config_desc.add_options()
             ("archive-log,A", po::bool_switch(&command_line.archive_log), "Makes a duplicate of the log file with TITLE_ID and Game ID as title")
             ("backend-renderer,B", po::value(&command_line.backend_renderer), "Renderer backend to use, either \"OpenGL\" or \"Vulkan\"")
-            ("color-surface-debug, C", po::bool_switch(&command_line.color_surface_debug),"Save color surfaces")
+            ("color-surface-debug,C", po::bool_switch(&command_line.color_surface_debug), "Save color surfaces")
```
Zangetsu38: oups :D

1whatleytay: What oops?

Zangetsu38: my fault lol, i have added this line, forget format :D

1whatleytay: NP XD

Should be fixed now.

vita3k/mem/src/mem.cpp
pent0: > As for the + 1... It seems aligned_size wasn't enough for some reason after I aligned memory. I would have a bad access somewhere in unicorn or load_elf.

Wdym by not enough? The aligned size is something like 4095 or something? Or Unicorn won't work with 4096?

vita3k/mem/src/mem.cpp
```diff
@@ -43,7 +43,14 @@ static void delete_memory(uint8_t *memory) {
 }
 
 static void alloc_inner(MemState &state, Address address, size_t page_count, Allocated::iterator block, const char *name) {
-    uint8_t *const memory = &state.memory[address];
+    Address alignment;
+#ifdef WIN32
+    alignment = 4096;
```
pent0:
```cpp
    int get_host_page_size() {
#if PLATFORM(WIN32)
        SYSTEM_INFO system_info = {};
        GetSystemInfo(&system_info);

        return system_info.dwPageSize;
#else
        return sysconf(_SC_PAGESIZE);
#endif
    }
```

I should forbid something like this. Also, use 4096. Then round them up to whatever Windows/Linux page size want.
The reason I tell you to do this is because the total of page was ordered to alloc was Vita page, and Vita page size is 4096. So the size Vita wants to allocate is 4096 * page_count. Then round them up to whatever Windows/Linux want,

vita3k/mem/src/mem.cpp
petmac: At the moment it's host, but after @pent0's comment below, `state.page_size` should probably be `max(4096, host page size)` instead of just the host page size.

pent0: Yep, that's should be right.

vita3k/mem/src/mem.cpp
1whatleytay: I'm not sure, I can check the bad access address. I think I was getting some bad accesses towards the end of the memory block. + 1 probably adds an entire other page to the mprotect call...

I'm not sure what to do... By default aligned_size is aligned, right? Should I add back what I subtracted from the pointer to the size?

vita3k/mem/src/mem.cpp
1whatleytay: You should!

Any reason for `sysconf` over `getpagesize`? Just curious.

pent0: Pretty sure I and petmac copy code from stackoverflow.

vita3k/mem/src/mem.cpp
1whatleytay: Wait a minute, wouldn't that be essentially the same as locking it to 4096?

If host size is under 4096, host size will be set to 4096.
If host size is above 4096, this assert in MemState init will go off.
```
assert(state.page_size >= 4096); // Limit imposed by Unicorn.
```

pent0: Well, what if there is Unicorn 2.0 and it doesn't limit page size to 4096? Seriously don't throw away the step!

1whatleytay: I'll do it anyway 😅

pent0: Merci boku naruto uchiha

pent0:
I told desgroup to remove the +1. That should do it

PR #587, opened 2019-09-29
Implement Dual

1whatleytay:
4a7392c162d48d007501152456d0c3a8

This PR makes gxmfun work better. Should be tested a bit because I'm guessing on a use of the .type field of TexCoordInfo (I bet some or even most games will be broken... I need to get those shaders though).


1whatleytay:
Yessir! On it.

vita3k/shader/src/usse_program_analyzer.cpp
```diff
@@ -53,6 +53,8 @@ std::uint8_t get_predicate(const std::uint64_t inst) {
     }
 
     // VMAD normal version, predicates only occupied two bits
+    case 0b00100:
+    case 0b00101:
     case 0b00000:
         return ((inst >> 32) & ~0xFCFFFFFF) >> 24;
```
pent0: Correctly handle the condition here please. Its shorten and not normal predicate value

1whatleytay: Thanks for reminding me! Note that it is actually 2 bits long.

1whatleytay: How about this? https://github.com/Vita3K/Vita3K/pull/587/files#diff-87a524ae6144f7f6b50e6263d12ef5daR58

pent0: Why not

1whatleytay:
If something breaks we can patch it out later. Plus we will have another shader with different type behaviour, which will help us understand the real reason for the field.

PR #589, opened 2019-10-10
Fix warnings when compiling with Clang

petmac:
Part of an effort to get the emulator compiling clean with no warnings.

There are still a few left so I'll update this PR with more fixes once I figure the code out better.

### DIFF
--- a/vita3k/renderer/src/gl/draw.cpp
+++ b/vita3k/renderer/src/gl/draw.cpp
@@ -66,8 +66,6 @@ void draw(GLState &renderer, GLContext &context, GxmContextState &state, const F
     const SceGxmProgram &fragment_gxp_program = *state.fragment_program.get(mem)->program.get(mem);
 
     if (log_active_shaders) {
-        const SceGxmProgram &vertex_gxp_program = *state.vertex_program.get(mem)->program.get(mem);
-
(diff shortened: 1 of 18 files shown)


vita3k/renderer/src/gl/draw.cpp
```diff
@@ -95,15 +92,11 @@ void draw(GLState &renderer, GLContext &context, GxmContextState &state, const F
     for (auto &vertex_uniform : context.vertex_set_requests) {
         gl::set_uniform(program_id, vertex_program_gxp, vertex_gl_statics, mem, vertex_uniform.parameter, vertex_uniform.data,
             log_uniforms);
-
-        delete vertex_uniform.data;
```
pent0: On Discord I have already prove this to be valid. Though currently the data passed in command buffers doesn't allow a vector to be stored (since we only have uint8_t bufffer which size of 0x80 for each command buffer to store data). Should be same with stream data and fragment data, until I can rework to use vector.

vita3k/util/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ add_library(
 	util
 	STATIC
 	include/util/align.h
+    include/util/bytes.h
```
pent0: Space/tab mismatch.

PR #604, opened 2019-10-27
avcdec, avplayer and yuv

1whatleytay:
The goal of this pr is to implement video decoding for vita. I use FFmpeg for video decoding :)

The submodule is an interesting case though. [PPSSPP's](https://github.com/hrydgard/ppsspp-ffmpeg) library was a bit outdated and was some 600MB. I don't like the size, especially since we aren't using half the targets it supports. I made my own build by forking PPSSPP's fork and updating the sources. I also got rid of any non x64 target (considering unicorn is locked at x86_64). It's [~60MB now](https://github.com/1whatleytay/ffmpeg-core). If you want to move it on to the Vita3K's organization I'd be up for that.

- [x] Implement shutdown decoder functions (flush, stop, destroy)
- [x] Fix that one "paletted format is unsupported" error message that keeps popping up (fix is staged on my windows partition).
- [x] Fix that nullptr passing thing that some games do (P4G). I think it wants you to allocate something, I have addressed it before but I haven't done it in this branch.
  - I don't think it matters.
- [x] I want to implement SceTimer for cutscene frame limiting for some games.
  - Not tested.
- [x] Maybe tackle SceMP4?
  - I'm not doing it, I haven't found a way for it to work with a demuxer yet too...
- [x] Maybe tackle SceAvPlayer?
  - Some bugs require you to view videos glitchy for a bit or at x2 speed...
- [x] Figure out how audio works with FFmpeg and SceAvcdec.
  - It doesn't.
- [x] Better YUV -> RGB decode. FFmpeg has a library for that.
  - sws_scale?
- [x] Fix the build problems. I'm just assuming they'll be there.
- [x] Rebase on master.
- [x] Fix texture hashing.
- [x] Swap submodules (printf isn't on Vita3K/printf and ffmpeg-core has new commit hash at Vita3K/ffmpeg-core).
- [x] GXM memcpy and stutter fix experiment.
  - didn't need it, apparently i just screwed up ffmpeg sizes :/


VelocityRa:
Forked to https://github.com/Vita3K/ffmpeg-core

vita3k/modules/CMakeLists.txt
```diff
@@ -170,3 +170,4 @@ add_library(modules STATIC
 )
 target_include_directories(modules PUBLIC include)
 target_link_libraries(modules PUBLIC module)
```
Zangetsu38:
```suggestion
target_link_libraries(modules PUBLIC module ffmpeg)
```
no need 2 line here

1whatleytay: You are right.

1whatleytay:
Alright, almost everything is done I think. I am going to mark this for review.

Would appreciate any suggestions, I'm open for refactoring suggestions (if someone wants me to move things into another module so we don't have to link ffmpeg to all modules, don't know if its worth the work).

Still investigating some issues... @Zangetsu38 has said Disgaea 3 plays video but has no audio. Some games go black when playing a video. For the majority of cases I think it works though.

1whatleytay:
D3 crash fixed by locking the kernel, seems like d3 creates the audio thread that decodes audio and then calls AddSource(), which created a race condition.

1whatleytay:
Screen Shot 2019-11-25 at 10 28 47 PM
Implemented a bit of SceJpeg so velocity ultra can show its art now on loading screen.

PR #606, opened 2019-10-27
Update SDL

frangarcj:


1whatleytay:
We gotta make sure it works with older versions of macOS too :)

Also if we're going to update headers, might as well update windows and linux binaries too.

frangarcj:
it's just 2.0.9 for 2.0.10. Code says is for the same mac os version.

The difference is that the binary is code signed.

Can you do Windows or leave it for another pr?

VelocityRa:
Why is this not a submodule again?

1whatleytay:
I'd be up for making this a submodule :D 
Would make it easy to swap out in the future.

I can do windows when I get home.

Zangetsu38:
you can squash all in one commit, no need few commit for any step ^^, set inside one "remove master sdl for move on external :)

VelocityRa:
It's fine, I'll squash when merging.

PR #607, opened 2019-10-28
io fixes

IllusionMan1212:
fixes the crash on downwell when accessing a weird path

- Properly check for device
- Add missing devices
- Check if path is not empty before calling front()
- Properly read directory and properly stat files and folders
- Properly report last modification time on files and folders


vita3k/io/src/io.cpp
```diff
@@ -482,13 +490,15 @@ SceUID read_dir(IOState &io, const SceUID fd, emu::SceIoDirent *dent, const std:
             if (fs::is_regular_file(cur_path)) {
                 if (stat_file(io, file_path.c_str(), &dent->d_stat, pref_path, base_tick, export_name) < 0)
                     return IO_ERROR(SCE_ERROR_ERRNO_EMFILE);
+                else
+                    return 1; // move to the next file
             }
             if (fs::is_directory(cur_path)) {
-                const auto dir_fd = open_dir(io, file_path.c_str(), pref_path, export_name);
-                return read_dir(io, dir_fd, dent, pref_path, base_tick, export_name);
+                stat_file(io, file_path.c_str(), &dent->d_stat, pref_path, base_tick, export_name);
```
VelocityRa: it looks like this identical line is executed if either `is_regular_file` or `is_directory` so why check?

vita3k/io/src/io.cpp
```diff
@@ -482,13 +490,15 @@ SceUID read_dir(IOState &io, const SceUID fd, emu::SceIoDirent *dent, const std:
...
+                return 1;
```
VelocityRa: not sure what `return 1;` means, the func's supposed to return a `SceUID` isn't it?
(not your fault since it's existing code but it'd be nice to explain or clean it up)

vita3k/io/src/io.cpp
IllusionMan1212: huh, you're right. i had the old code commented while testing it so i didn't notice lol

vita3k/io/src/io.cpp
IllusionMan1212: `SceIoDread` returns an int.
this is what vitasdk says:
```
= 0 - No more directory entries left
> 0 - More directory entries to go
< 0 - Error
```

VelocityRa: Ok, would be cool to have an enum for this.

vita3k/io/src/io.cpp
```diff
@@ -164,7 +166,8 @@ static std::string translate_path(const char *path, VitaIoDevice &device, const
     case +VitaIoDevice::ur0:
     case +VitaIoDevice::ux0:
     case +VitaIoDevice::vd0:
-    case +VitaIoDevice::vs0: {
+    case +VitaIoDevice::vs0:
+    case +VitaIoDevice::xmc0: {
```
1whatleytay: `default:` ?

PR #610, opened 2019-11-04
config: Generate Config members using higher-order macros

totlmstr:
The config file has practically been the bane of me ever since I've been contributing for Vita3K, and for several reasons:

* it's ugly
* it's unmanageable in the long run with no less than *four* places that need to be updated if a new variable is added
* did I mention it's ugly?
* a lot of debugging for me has been focused on this exact one area more than I can admit
* it's also *ugly*

The running idea with this PR is that the YAML already has all the variables we need, so we can just use it like a state variable, and output it when it is updated.

Problem: how do we update the `yaml_node` with class members?

Solution: [use higher order macros](https://journal.stuffwithstuff.com/2012/01/24/higher-order-macros-in-c/). Essentially, we pass a macro within a macro to generate variables, members, and even comparisons and updates.

Now, only one area of code in one base file (`config.h`) needs to be updated. Sinply add in the type of the variable, the name of the variable, and the default value, and everything will be generated automatically (note that you may need to clean the Config solution for this to work properly). Additionally, a file based on default values will be created on program start as a precaution.

---

~~Edit: It turns out Travis' Mac builder is [disagreeing with the macros I have](https://travis-ci.org/Vita3K/Vita3K/jobs/606936123#L1797), which was honestly a shot in the dark for me on whether it *could* work. An additional edit is needed for Mac (Linux is fine).~~ Xcode is really strict with the macros, since they generated errors instead of warnings for a bad concatenation. So, it was a pretty simple fix.

---

~~This PR requires additional testing before merging.~~


1whatleytay:
THANK YOU SO MUCH :pray:

pent0:
This is so much likeable and high fashion, thank you very much 😍😍😍😍😍😍😍😍😍🥰🥰🥰🤯😍🤯🤯😍🥰🥰😍🥰😍🥰🥰😍🥰😍🥰

vita3k/config/include/config/config.h
```diff
@@ -17,47 +17,49 @@
...
+#include <psp2/system_param.h>
+#include <spdlog/spdlog.h>
 
-#include <boost/optional.hpp>
+// clang-format off
+// Singular options produced in config file
+// Order is code(option_type, option_name, default_value)
+// When adding in a new macro for generation, ALL options must be stated.
+#define CONFIG_INDIVIDUAL(code)                                                                 \
+    code(bool, log_imports, false)                                                              \
```
pent0: Put the name of config in "". I dobt like the underscore, and i doubt "" will bite

totlmstr: You mean like this?

`code(bool, "log_imports", false)`

pent0: Yeah, also i like to keep the name like before ( log-imports )

totlmstr: Those requests will actually be a problem.

e.g., note that the macro `code(bool, "log-imports", false) becomes `YAML_MEMBER(bool, "log-imports", false)` which becomes `bool "log-imports" = false`, which doesn't make sense in C++ and fails equally elsewhere in the PR. Additionally, dashes in names (without quotes) are not valid variables, and it is too difficult/unknown at least from my searching to find out how to replace "-" with "_" (and it is better to make the names uniform). An `std::map<std::string, std::variant>` would do the trick (and is the easier option to get to what you want, and we are in C++17) and is not too hard to implement (relatively speaking), but doing it will be pretty drastic for Vita3K, since it is currently getting the variables as members and not as a map/dictionary *all over the place*.

So, it is either getting the members from a dictionary-like object, or keep them as they are and use the member variables instead. If you want to do the dictionary route, I wouldn't mind; it will take some time, though.

totlmstr:
I took another page out of that link on higher order macros above for enums and [implemented it into these changes here](https://github.com/Vita3K/Vita3K/pull/610/files#diff-a6a8977717cb5a4e9908c4f9b0256552R217).

It's freaking beautiful, if I say so myself. :^)

PR #612, opened 2019-11-07
sceAppUtil: implement sceAppUtilSaveDataGetQuota

IllusionMan1212:
Steins;Gate fails to check for available space to create the system data.

(credits to @Zangetsu38  for testing it and posting the log)
steinsgate

and it progresses to menu with this
image

---

### DIFF
--- a/vita3k/io/src/io.cpp
+++ b/vita3k/io/src/io.cpp
@@ -71,6 +71,15 @@ bool read_app_file(FileBuffer &buf, const std::string &pref_path, const std::str
     return read_file(VitaIoDevice::ux0, buf, pref_path, fs::path("app") / title_id / vfs_file_path);
 }
 
+SpaceInfo get_space_info(const VitaIoDevice device, const std::string &vfs_path, const std::string &pref_path) {
+    SpaceInfo space_info;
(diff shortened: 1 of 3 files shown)


pent0:
This need further confirmation about the size. sceAppMgr is an internal module, so not neccessary to implement it.

Zangetsu38:
Oh good catch

vita3k/io/src/io.cpp
```diff
@@ -71,6 +71,15 @@ bool read_app_file(FileBuffer &buf, const std::string &pref_path, const std::str
     return read_file(VitaIoDevice::ux0, buf, pref_path, fs::path("app") / title_id / vfs_file_path);
 }
 
+SpaceInfo get_space_info(const VitaIoDevice device, const std::string &vfs_path, const std::string &pref_path) {
+    SpaceInfo space_info;
+    const auto host_path = device::construct_emulated_path(device, vfs_path, pref_path);
+    space_info.max_capacity = fs::space(host_path).capacity;
+    space_info.used = fs::space(host_path).capacity - fs::space(host_path).available;
```
pent0: Fill free first, then subtract max capacity with free that already called to get. I think its just more efficient, the effect is probablly not seenable though.

PR #617, opened 2019-11-11
[TESTER REQUEST, WIP] module: load firmware modules.

Zangetsu38:
- Enable using fw module like ft2 and pvf
Request tester with have see pvf unimplemented function on game and test it.


Zangetsu38:
not correct for load pre loaded module ? on your compit is totaly hacky, here is correct so with load module like loading other libc and other is exactly same libpvf and ft2

VelocityRa:
Many reasons.

PR #618, opened 2019-11-12
ScePvf LLE compatibility (relocation & loader fixes)

VelocityRa:


vita3k/kernel/src/relocation.cpp
```diff
@@ -417,18 +421,34 @@ bool relocate(const void *entries, uint32_t size, const SegmentInfosForReloc &se
...
+            g_offset += format2_entry->offset;
+            g_saddr = (format2_entry->symbol_segment == 0xf) ? 0 : symbol_seg_start;
+            g_addend = format2_entry->addend;
+            g_type2 = 0;
+
+            const auto s = g_saddr;
+            const auto a = g_addend;
+            const auto p = g_addr + g_offset;
+
+            // Warn because it's kind of untested
```
frangarcj: I'd not warm because it is as much tested as the others

VelocityRa: Well, at least some of the others _have_ to work. With this one, the game I tested work just the same, even without doing the `relocate_entry` (just from moving the offset) so we truly don't know.
But anyway, It's not very useful I suppose, I'll remove it.

PR #623, opened 2019-12-02
game context menu: Add manuals popup

Zangetsu38:
- little rework code for can using host.io.title_id already here

- Add Manual item in context menu
image

- Open on Manual dialog with can show your owner languague(automatic, using config sys_lang)
# Result: 
image

### DIFF
--- a/vita3k/gui/include/gui/functions.h
+++ b/vita3k/gui/include/gui/functions.h
@@ -36,14 +36,15 @@ enum GenericDialogState {
     CANCEL_STATE
 };
 
-void delete_game(GuiState &gui, HostState &host, const std::string &title_id);
-void game_context_menu(GuiState &gui, HostState &host, bool *selected, const std::string &title_id);
+void delete_game(GuiState &gui, HostState &host);
+void game_context_menu(GuiState &gui, HostState &host, bool *selected);
 void get_game_titles(GuiState &gui, HostState &host);
 void get_modules_list(GuiState &gui, HostState &host);
 void init(GuiState &gui, HostState &host);
 void init_background(GuiState &gui, const std::string &image_path);
 void init_icons(GuiState &gui, HostState &host);
-void load_game_background(GuiState &gui, HostState &host, const std::string &title_id);
+void load_game_background(GuiState &gui, HostState &host);
+void load_manual(GuiState &gui, HostState &host, const std::string &image_name);
 bool refresh_game_list(GuiState &gui, HostState &host);
 
 void draw_begin(GuiState &gui, HostState &host);
--- a/vita3k/gui/src/game_context_menu.cpp
+++ b/vita3k/gui/src/game_context_menu.cpp
@@ -18,31 +18,26 @@
 #include "private.h"
 
 #include <gui/functions.h>
-#include <gui/state.h>
-
-#include <host/state.h>
 
 #include <util/log.h>
 #include <util/string_utils.h>
 
+#include <SDL_keycode.h>
+
 namespace gui {
 
-static bool delete_game_popup = false;
-static bool delete_savedata_popup = false;
-static bool check_savedata_and_shaderlog = false;
-
-void delete_game(GuiState &gui, HostState &host, const std::string &title_id) {
-    const fs::path game_path{ fs::path(host.pref_path) / "ux0/app" };
+void delete_game(GuiState &gui, HostState &host) {
+    const fs::path game_path{ fs::path(host.pref_path) / "ux0/app" / host.io.title_id };
 
     for (const auto &game : gui.game_selector.games) {
         const auto games_index = std::find_if(gui.game_selector.games.begin(), gui.game_selector.games.end(), [&game](const Game &g) {
             return g.app_ver == game.app_ver && g.title == game.title && g.title_id == game.title_id;
         });
 
-        if (title_id == game.title_id) {
-            LOG_INFO_IF(fs::remove_all(game_path / game.title_id), "Successfully deleted '{} [{}]'.", game.title_id, game.title);
+        if (host.io.title_id == game.title_id) {
+            LOG_INFO_IF(fs::remove_all(game_path), "Successfully deleted '{} [{}]'.", game.title_id, game.title);
 
-            if (!fs::exists(game_path / game.title_id)) {
+            if (!fs::exists(game_path)) {
                 gui.delete_game_background = true;
                 gui.delete_game_icon = true;
                 gui.game_selector.games.erase(games_index);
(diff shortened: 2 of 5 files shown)


Zangetsu38:
very reworks code, i thinks is perfect now, ready for review

vita3k/gui/src/game_context_menu.cpp
```diff
@@ -164,6 +192,51 @@ void game_context_menu(GuiState &gui, HostState &host, bool *selected, const std
                 ImGui::EndPopup();
             } else
                 ImGui::PopStyleColor();
+
+            // Manual Popup
+            if (manuals_popup)
+                ImGui::OpenPopup("Manual");
+            const auto display_size = ImGui::GetIO().DisplaySize;
+            ImGui::SetNextWindowPos(ImVec2(-5.0f, -5.0f), ImGuiCond_Always);
+            ImGui::SetNextWindowSize(ImVec2(display_size.x + 10.0f, display_size.y + 10.0f), ImGuiCond_Always);
```
1whatleytay: I would make the `-5.0f` and `10.0f` constexpr and global. Their own variables basically.

vita3k/gui/include/gui/functions.h
```diff
@@ -36,14 +36,15 @@ enum GenericDialogState {
...
+void delete_game(GuiState &gui, HostState &host);
+void game_context_menu(GuiState &gui, HostState &host, bool *selected);
 void get_game_titles(GuiState &gui, HostState &host);
 void get_modules_list(GuiState &gui, HostState &host);
 void init(GuiState &gui, HostState &host);
 void init_background(GuiState &gui, const std::string &image_path);
 void init_icons(GuiState &gui, HostState &host);
-void load_game_background(GuiState &gui, HostState &host, const std::string &title_id);
+void load_game_background(GuiState &gui, HostState &host);
+void load_manuals(GuiState &gui, HostState &host, const std::string &image_name);
```
1whatleytay: Keep the string parameter but just pass `host.io.title_id`.

vita3k/gui/src/game_context_menu.cpp
```diff
@@ -164,6 +192,51 @@ void game_context_menu(GuiState &gui, HostState &host, bool *selected, const std
...
+                if (gui.current_manual)
+                    ImGui::GetWindowDrawList()->AddImage(reinterpret_cast<void *>(gui.current_manual), ImVec2(0, 0), display_size);
+                else
+                    manuals_popup = false;
+
+                ImGui::SetCursorPos(ImVec2(display_size.x - 40.0f, 20.0f));
+                if (ImGui::Button("X", ImVec2(30.0f, 0.f)))
+                    manuals_popup = false;
+
+                ImGui::SetCursorPos(ImVec2(display_size.x / 2.0f - 90.0f, display_size.y - 40.0f));
```
1whatleytay: Ditto, constexpr and global is good.

vita3k/gui/src/game_context_menu.cpp
```diff
@@ -18,28 +18,28 @@
...
+static auto delete_savedata_popup = false;
+static auto check_savedata_and_shaderlog = false;
+static auto manuals_popup = false;
+static auto current_manual = 0;
+static std::vector<std::string> manuals_list;
 
-void delete_game(GuiState &gui, HostState &host, const std::string &title_id) {
-    const fs::path game_path{ fs::path(host.pref_path) / "ux0/app" };
+void delete_game(GuiState &gui, HostState &host) {
+    static const fs::path game_path{ fs::path(host.pref_path) / "ux0/app" };
```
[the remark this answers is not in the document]
Zangetsu38: is good, is const value

vita3k/gui/include/gui/functions.h
Zangetsu38: don't needed, is good like that

vita3k/gui/src/game_context_menu.cpp
Zangetsu38: no, only for one things so not needed

vita3k/gui/src/game_context_menu.cpp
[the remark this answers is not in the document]
1whatleytay: The const here is actually reassigned every time delete_game is called. It just can't be changed later in the function. `static const` in a function will make sure it is only assigned once during runtime. This is bad if pref_path changes between the first delete_game call and the last.

Plus it does basically nothing in your favor. I really don't like using static in functions.

vita3k/gui/include/gui/functions.h
1whatleytay: What if someone needs to load an image of the next game? Or a newly added game? They would  have to store `host.io.title_id` in some temp variable, set it again to their preferred game then call the function and restore it from the temp variable. Just pass `host.io.title_id` instead of hard wiring every function to use it. It leaves options open.

vita3k/gui/src/game_context_menu.cpp
1whatleytay: I don't mind if you don't do this, but consider a person looking over your code trying to increase the left padding or something. They can't tell which constants do what.

Like what does 40.0f do here? I'm not sure... If you had all the constants named like `static constexpr float LEFT_PADDING = 40.0f;` then its super easy to tell.

vita3k/gui/src/game_context_menu.cpp
```diff
@@ -18,28 +18,28 @@
...
 
-static bool delete_game_popup = false;
-static bool delete_savedata_popup = false;
-static bool check_savedata_and_shaderlog = false;
+static auto delete_game_popup = false;
+static auto delete_savedata_popup = false;
+static auto check_savedata_and_shaderlog = false;
+static auto manuals_popup = false;
+static auto current_manual = 0;
+static std::vector<std::string> manuals_list;
```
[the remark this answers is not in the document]
Zangetsu38: i have used same ohter things set by other dev in master, don't need spam not needed value on gui state for nothings ^^

vita3k/gui/src/game_context_menu.cpp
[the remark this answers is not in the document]
Zangetsu38: no, like i said, this value is app folder, not game folder ^^ so static is good not worry

vita3k/gui/src/game_context_menu.cpp
```diff
@@ -52,12 +52,32 @@ void delete_game(GuiState &gui, HostState &host, const std::string &title_id) {
     }
 }
 
-void game_context_menu(GuiState &gui, HostState &host, bool *selected, const std::string &title_id) {
+void get_manuals_list(GuiState &gui, HostState &host) {
+    current_manual = 0;
+    gui.manual.clear();
+    manuals_list.clear();
+
+    auto manuals_path{ fs::path(host.pref_path) / "ux0/app" / host.io.title_id / "sce_sys/manual/" };
```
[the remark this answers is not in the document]
Zangetsu38: for me use it is just perfect

vita3k/gui/src/game_context_menu.cpp
Zangetsu38: i have set this for can get fullscreen and not get empty space in left/right border
not sur need create const value just for one call 🤔

vita3k/gui/src/game_context_menu.cpp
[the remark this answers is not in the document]
1whatleytay: You're using pref path though- that could totally change over a config option in the future. Actually, you might've already implemented it.

vita3k/gui/src/game_context_menu.cpp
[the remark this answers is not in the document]
1whatleytay: for others maybe not

vita3k/gui/include/gui/functions.h
Zangetsu38: like i said, don't have any problem, value is updated and perfecly works for all

vita3k/gui/src/game_context_menu.cpp
1whatleytay: There are only advantages to constexpr- but I see what you mean. If you don't want to do it that's alright imo

vita3k/gui/src/game_context_menu.cpp
Zangetsu38: hmm i see ^^ but normally you can easy see what is exacly ^^

1whatleytay: That might be true but names make it easier XD

vita3k/gui/include/gui/functions.h
1whatleytay: Maybe for now but someone in the future might have to revert this- I really don't see why you can't just pass it.

vita3k/gui/src/game_context_menu.cpp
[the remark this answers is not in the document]
Zangetsu38: ah yes, you have' right for pref path lol, but i have tested change pref path and works good like that, but you have right

vita3k/gui/src/game_context_menu.cpp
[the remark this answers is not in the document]
Zangetsu38: i have tested, all, and all works normally, not worry ^^

vita3k/gui/src/game_context_menu.cpp
Zangetsu38: yes ^^

vita3k/gui/src/game_context_menu.cpp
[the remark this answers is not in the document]
Zangetsu38: pushed restore

vita3k/gui/src/game_context_menu.cpp
Zangetsu38: yes, but i keep like that, i don't like create value for call it just one time haha

vita3k/gui/include/gui/functions.h
Zangetsu38: because is not needed, string title id is used only if you don't have value outside, but here is already exist on io.title id, so not needed

vita3k/gui/src/game_context_menu.cpp
```diff
@@ -52,12 +52,37 @@ void delete_game(GuiState &gui, HostState &host, const std::string &title_id) {
...
+            const fs::path manual = *m;
+            if (m->path().extension() == ext)
+                manuals_list.push_back({ manual.filename().replace_extension().string() });
+        }
+    }
+
+    if (manuals_list.empty())
+        return false;
+
+    return true;
```
1whatleytay:
```
return !manuals_list.empty();
```

vita3k/gui/include/gui/functions.h
1whatleytay: Same deal is keeping me from approving. This is just a downgrade imo.

vita3k/gui/src/game_context_menu.cpp
1whatleytay: Keeping me from approving.

vita3k/gui/src/game_context_menu.cpp
Zangetsu38: ahhh lol ok

Issue #620, opened 2019-11-19
Integrate SPIR-V shader validation


VelocityRa:
Currently recompiled shaders are usually not valid SPIR-V, they would not pass `spirv-val`.
They only work because SPIRV-Cross does not validate them and can accept very wrong shaders.

The solution for this is to programmatically invoke the validator from SPIRV-Tools every time we recompile a shader.

*Edit:* We can invoke glslang's wrapper, no need to import SPIRV-Tools https://github.com/KhronosGroup/glslang/blob/ea69a2982986cd668ed15287cdcb7e96a637d92b/SPIRV/SpvTools.h#L70

GamingDucking:
SPIR-V didn't exactly went to plan in my opinion and it's currently "deprecated" as of the latest bulid.

Macdu:
Actually, SPIR-V is the only backend used by the Vulkan renderer, but it is marked as deprecated for opengl. Moreover, this PR is about making valid Spir-V code. Vulkan Validation layers check for spirV validity. So as part of making the Vulkan renderer I also fixed all spirV validation issues.

PR #629, opened 2020-01-02
shader: optimize and verify spirv

1whatleytay:
This is a small pr that adds SPIRV-Tools optimizations and validation.

I'm not sure if performance is increased (I want to know though). I am sure that optimization fails completely every blue moon, so I created an option to disable it. Validation is also run on generated SPIRV  so debugging can be performed to eliminate issues.

If this PR stays open for a while I might try to do some validation fixing.

### DIFF
--- a/vita3k/shader/src/spirv_recompiler.cpp
+++ b/vita3k/shader/src/spirv_recompiler.cpp
@@ -30,6 +30,7 @@
 
 #include <SPIRV/SpvBuilder.h>
 #include <SPIRV/disassemble.h>
+#include <spirv-tools/optimizer.hpp>
 #include <boost/optional.hpp>
 #include <spirv_glsl.hpp>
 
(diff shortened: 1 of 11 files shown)


VelocityRa (dismissed):

vita3k/shader/src/spirv_recompiler.cpp
```diff
@@ -1029,13 +1030,44 @@ static spv::Function *make_vert_finalize_function(spv::Builder &b, const SpirvSh
...
+    spvDiagnosticDestroy(diagnostic);
+    spvContextDestroy(validator);
+
+    spvtools::OptimizerOptions optimizer_options;
+    optimizer_options.set_run_validator(false);
+
+    spvtools::Optimizer optimizer(target);
+
+    optimizer.RegisterPerformancePasses();
+    optimizer.Run(input.data(), input.size(), &result, optimizer_options);
```
sunho, line 1056: This one returns error! You can return the raw input if error occur to make it not crash.

sunho:
Fix https://github.com/Vita3K/Vita3K/issues/620

Zangetsu38:
crazy obselet and move soon on glsl so i close :)

PR #632, opened 2020-01-09
Better unicorn error reporting.

1whatleytay:
Tells us specific error names plus register values.

velo was talking about tracking down which module the error might've happened in
that might be cool to add

### DIFF
--- a/vita3k/cpu/src/cpu.cpp
+++ b/vita3k/cpu/src/cpu.cpp
@@ -151,6 +151,25 @@ static void enable_vfp_fpu(uc_engine *uc) {
     assert(err == UC_ERR_OK);
 }
 
+static void log_error_details(CPUState &state, uc_err code) {
+    // I don't especially want the time logged for every line, but I also want it to print to the log file...
+    LOG_ERROR("Unicorn error {}. {}", log_hex(code), uc_strerror(code));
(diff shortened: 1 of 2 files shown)


vita3k/cpu/src/cpu.cpp
```diff
@@ -219,10 +238,11 @@ int run(CPUState &state, bool callback) {
     err = uc_emu_start(state.uc.get(), pc, state.entry_point & 0xfffffffe, 0, 0);
 
     if (err != UC_ERR_OK) {
-        std::uint32_t error_pc = read_pc(state);
-        uint32_t lr = read_lr(state);
-        LOG_CRITICAL("Unicorn error {} at: start PC: {} error PC {} LR: {}",
-            log_hex(err), log_hex(pc), log_hex(error_pc), log_hex(lr));
+//        std::uint32_t error_pc = read_pc(state);
```
VelocityRa: you can delete these no?

1whatleytay: Right, amended.

1whatleytay:
Will merge after travis :D

VelocityRa:
No need to wait for CI if you're sure it builds btw.

PR #634, opened 2020-01-13
implemented savedata dialog as well as other things

IllusionMan1212:


vita3k/gui/src/common_dialog.cpp
```diff
@@ -22,6 +22,9 @@
 #include <util/string_utils.h>
 
 #include <SDL.h>
+#include <gui/imgui_impl_sdl.h>
```
VelocityRa: put this below the util/string_utils include and stb_image here

PR #637, opened 2020-01-15
kernel/sysmem: Implement sceKernelGetMemBlockInfoByAddr

frangarcj:


frangarcj:
> Run the format script before pushing any code please.

I have format script enabled in visual studio 2019. What am I missing?

VelocityRa:
In Visual Studio you can only run clang-format per-file manually.
The format _scripts_ are the ones these ones in the root of the project, like this for windows: https://github.com/Vita3K/Vita3K/blob/master/format.bat

If you have clang-format in your PATH (you can do that by installing LLVM and adding its /bin to the PATH env var) then you can just run `format` in a cmd in the vita3k dir and it will run clang-format for the entire codebase. That's what we suggest everyone does before opening PRs.

Zangetsu38:
No créate commit just for this ^^ édit master commit and push so here squash

PR #645, opened 2020-01-22
renderer/gl: check status instead of info length

1whatleytay:
I was debugging on bootcamp again and noticed it was printing an empty error message for every shader it recompiled. This pr should fix this again.

### DIFF
--- a/vita3k/renderer/src/gl/compile_program.cpp
+++ b/vita3k/renderer/src/gl/compile_program.cpp
@@ -26,11 +26,9 @@ static SharedGLObject compile_glsl(GLenum type, const std::string &source) {
     glCompileShader(shader->get());
 
     GLint log_length = 0;
-    GLint status = 0;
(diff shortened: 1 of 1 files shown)


1whatleytay:
Ahh. Well specifically the issue is that intel driver returns `1` as info log length (one null character). So maybe `> 1`?

1whatleytay:
BAM, done.

VelocityRa:
lol

@1whatleytay please add a comment explaining Intel's stupidity
Or check 1 only on intel drivers. Do we have [driver caps](https://github.com/RPCS3/rpcs3/blob/c1180d76dd31510bb7f5d269b652de10d2261fda/rpcs3/Emu/RSX/GL/GLHelpers.h#L119-L122)? I don't remember. If we don't, whatever leave it as-is whatever.

1whatleytay:
How about this? :)

PR #649, opened 2020-02-01
Remove vita-headers

1whatleytay:
For discussion and testing on removing vita-headers.

Some reasons for removing vita-headers:
 - We can't use a bunch of vita-headers structs because of the host/target pointer size (64-bit host vs 32-bit guest).
 - Because of that we have to redeclare lots of the struct in `namespace emu`. This is bad because:
   - Redeclaring the contents of the header defeats the purpose of including the header at all.
   - Causes confusion between emu and global versions of structs. Some common questions include:
     - Is there an emu version of XXX struct?
     - Do I need to make an emu version of XXX struct?
     - Does YYY struct/function use the emu or global version of XXX struct?
     - Do I need to cast between emu and global here?
 - vita-headers license is supposed to let us copy everything. Copying everything would also let us have more control over the structs themselves (for example, get methods for width/height of SceGxmTexture).

I need...
 - Feedback on organization (like util/types.h). There's probably a good and logical way to organize vita-headers structs, especially those that are shared across multiple modules- but I couldn't think of it.
 - To drop comments that I forgot to remove.
 - To erase all the remaining `namespace emu`s that have survived my wrath. **OK**


vita3k/config/include/config/config.h
```diff
@@ -17,9 +17,39 @@
 
 #pragma once
 
-#include <psp2/system_param.h>
 #include <spdlog/spdlog.h>
 
+// temporary enum copy, should be moved to some header like util/types.h before merge
```
Zangetsu38: i confirm, here is wrong haha, and this relate apputil, go just use like rpcs3 ans set in inside include corresponding :)

vita3k/util/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ add_library(
 	include/util/preprocessor.h
 	include/util/resource.h
 	include/util/string_utils.h
+    include/util/types.h
```
Zangetsu38: fail format ^^

1whatleytay:
Amended.

vita3k/util/include/util/types.h
```diff
@@ -0,0 +1,261 @@
...
+
+typedef int SceUID;
+
+typedef int ScePID;
+#define PSP2_KERNEL_PROCESS_ID_SELF 0 //!< Current running process ID is always 0
+
+typedef char* SceName;
+#define PSP2_UID_NAMELEN 31 //!< Maximum length for kernel object names
+
+enum SceSystemParamLang {
```
Zangetsu38: not fan with thhis place :/, why not using include corresponding apputil directly ?  util is diferent apputil module

1whatleytay: Is there an apputil cmake module? I kinda agree though, maybe a separate header for sysparams.

1whatleytay:
Please don't break anything please don't break anything please please please please

PR #652, opened 2020-02-07
Replace Some Assert With Possible Fix 

Co2333:
nothing much just replaced some assert that may be able to fix at runtime like surface buffer if we alloc them when they don't exist.


Zangetsu38:
Lot need change, you need squash this 3 commit on one, and need run clang format script

Co2333:
okay wait a min

Zangetsu38:
Don't need close pr, just squash all and force push

Co2333:
after a force push, it is not possible to just reopen it but recreate one xD

Zangetsu38:
Yes for that no close pr just force push

Issue #644, opened 2020-01-21
Add a new button to open Vita3K partitions folder in File Explorer


Rinnegatamante:
Most likely the best place would be as a new button for File submenu. It should open up %appdata%/Vita3K/Vita3K (or whatever Emulated System Storage Folder is set to) in file explorer (eg. explorer.exe for Windows). This would be handy when an user have to install sa0 fonts and/or game data files for homebrews.

IllusionMan1212:
Should we close?

1whatleytay:
Yeah lets close it.

PR #659, opened 2020-02-18
gui: added a button to open pref path

IllusionMan1212:
Closes issue #644


Zangetsu38:
oh, i have forget implement this, but i thinks is better set in inside setting in same place pref path setting 🤔

IllusionMan1212:
> oh, i have forget implement this, but i thinks is better set in inside setting in same place pref path setting 🤔

rinne wants it in File menu 🤷‍♂

PR #660, opened 2020-02-19
Refactor CMake for Boost.Build and C++17

totlmstr:
The short explanation of this PR is that this allows CMake to automate the building process of Boost and allows for compiling to C++14/C++17 for Vita3K.

The long explanation is Travis.

---

Travis, despite being an excellent CI, does not yet support C++17 for Mac. Specifically, the version that is needed, macOS 10.15 "Catalina", is [not yet available](https://docs.travis-ci.com/user/reference/osx/#macos-version) at the time of this writing. Additionally, casual searching led me to find out that Apple *completely skipped* the `experimental` headers, so Travis is essentially stuck at C++11 for now.

I initially had the idea of updating all parts of Vita3K to C++17, but that obviously does not work with the above. As a result, the bulk of my time and the first part of this PR was making [this CMake file which can build, compile, and set relevant Boost variables all by itself](https://gist.github.com/totlmstr/f034130b77625be18ff1f2864b3ad255). The gist I linked is the heavily annotated one that I am committing, so if you need any more information, see that link.

The other part of this PR is making sure Vita3K is C++11 compliant. The reasoning behind this is the fact that users who are the most likely to use Boost are those who are in C++11 mode. ~~You can *probably* build it as it is now to test it without Boost, but in order for this to be a proper PR, it needs to also pass C++11 compliance (and a PR for `ext-boost`), and I am *roughly* halfway there on that.~~ Now, when you are compiling this program, it properly reports what C++ standard it is on (which I previously been having problems with). Edit: Now that Travis is able to compile Vita3K, users should now be able to compile with or without Boost. Commits after this will be maintenance or cleaning up.

Another thing: the reason why this adds CLI11 is that an analysis of Vita3K shows that Boost.Filesystem and Boost.Optional are needed throughout Vita3K and Boost.Program Options is needed in exactly one place: `config`. Personally, it doesn't make sense to have this hindrance when switching to C++14/C++17, and Boost.Program Options relies too heavily on other Boost libraries, but then there's CLI11 which is self-contained (funnily enough, CLI11 *started out* as a Boost.Program Options extension).


isJuhn:
I think you can build C++17 stuff on 10.14 as long as it targets 10.15 or newer, so should be fine for CI. Don't know if we wanna up the macOS requirement to 10.15 though. It'd be really nice be able to use std::filesystem imo. And doing so we might be able to phase out boost entirely.

1whatleytay:
Afaik Travis CI only supports 10.14. C++17 FS is still kinda out of the question :(

petmac:
Be careful also if upgrading from Xcode 9.4.1 to a newer version.

Last time I looked, upgrading Xcode updated the macOS SDK, which broke OpenGL support for us.

.gitmodules
```diff
@@ -84,3 +81,7 @@
 [submodule "external/ffmpeg"]
 	path = external/ffmpeg
 	url = https://github.com/Vita3K/ffmpeg-core.git
+[submodule "external/boost"]
+	path = external/boost
+	url = https://github.com/totlmstr/ext-boost.git
```
1whatleytay: Let's fork this!

Margen67:
GitHub Actions has macOS 10.15: https://github.com/actions/virtual-environments/blob/master/images/macos/macos-10.15-Readme.md

.gitmodules
totlmstr: Note that the branch being checked out there is `boost-trim`. The branch name is not seen here, [but they are indeed very different](https://github.com/totlmstr/ext-boost/compare/509da24bc9cf868c9029297c5b54426e7ede33ae...c9ee09ddddab89c7f35d1f6d9f493d846c1f39cd).

scribam:
@totlmstr Is this PR still in progress? Could it be possible to have a separate PR for the replacement of Boost ProgramOptions by CLI11? Or can I reuse the commits here to do so?
CLI11 contains features I need.

totlmstr:
> @totlmstr Is this PR still in progress? Could it be possible to have a separate PR for the replacement of Boost ProgramOptions by CLI11? Or can I reuse the commits here to do so?
> CLI11 contains features I need.

I have been a bit busy as of late, but I am still working on this PR.

However, I did notice your PR #725. To be frank, a CLI11 refactor was not a focus of this PR, since it was mostly a "one and done" kind of thing. Your PR might not be merged now, but I can just simply rebase it onto yours when it is complete.

External repo stuff ended up delaying this PR (`std::u16string`, specifically), but I'm currently scrapping it due to me needing to do more research and the rather long time this PR has been around.

Zangetsu38:
Totaly outdated, request rebase and fix conflict

Zangetsu38:
@totlmstr you keep it or you want close ?

totlmstr:
@Zangetsu38 I'll close it. It is simply too old now, and as such would be too difficult to rebase and update as is for me, in my honest opinion.

At a later date, I'll try to see if I can cherry-pick from this.

PR #666, opened 2020-03-02
AppVeyor: Upgrade to Visual Studio 2019, Boost 1.71, other changes

Margen67:
Format comments, remove useless ones.
Delete white space.
Use proper multi-line for cmd commands.
Remove verbosity since normal is the default.
Change slash to \ to be consistent with Windows.
Change Release to $env:CONFIGURATION

gen-windows.bat: Change default to Visual Studio 2019

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,38 +1,34 @@
 version: '{build}'
 
-image: Visual Studio 2017
+image: Visual Studio 2019
 
 environment:
   global:
-    BOOST_ROOT: C:\Libraries\boost_1_67_0
-    BOOST_INCLUDEDIR: C:\Libraries\boost_1_67_0\boost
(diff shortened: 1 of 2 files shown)


totlmstr:
I should mention that when I finish #660, this PR's changes are nullified.

Appveyor is happy without Boost in my build, and users will be able to optionally build with Boost.

1whatleytay:
Ahh that's nice :D 

How close is #660 to done? If it'll take some time we can just merge in the meantime.

totlmstr:
At the current rate I am going, I can probably finish it next week due to [a FAT16 error which seems easier to parse than it looks](https://travis-ci.org/Vita3K/Vita3K/jobs/656564403#L647) and me being somewhat busy currently. After this, there should not be any more demanding errors (I have to do a slow crawl upwards on Vita3K).

I do, however, think it will be better just to merge this PR. There are some changes in the YML file that I can use from here.

Issue #673, opened 2020-04-04
Alert when user launches a game dumped with Vitamin


1whatleytay:
Lots of people like running games dumped with Vitamin. Vitamin almost never works, so it might save some time if we add a check in the software to tell people that their game isn't going to work before they launch it.

Usually `sce_sys/steroid.spurx` is an indicator of vitamin action, maybe a popup would be nice? Something nice and clear.

1whatleytay:
Resolved as of #674

PR #674, opened 2020-04-04
Alert users of vitamin dump

KorewaWatchful:
Issue #673 Alerts users when a vitamin dump is launched.
Tested with a real vitamin dump. 
capturevitamin

### DIFF
--- a/vita3k/host/src/load_self.cpp
+++ b/vita3k/host/src/load_self.cpp
@@ -16,7 +16,7 @@
 // 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
 #include <host/load_self.h>
-
+#include <app/functions.h>
 #include <config/state.h>
 #include <kernel/relocation.h>
 #include <kernel/state.h>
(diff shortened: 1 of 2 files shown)


vita3k/host/src/load_self.cpp
```diff
@@ -342,6 +342,11 @@ SceUID load_self(Ptr<const void> &entry_point, KernelState &kernel, MemState &me
         LOG_CRITICAL("SELF {} header type {} is not supported.", self_path, self_header.header_type);
         return -1;
     }
+    
+    if (self_path == "app0:sce_module/steroid.suprx") {
+        LOG_CRITICAL("You're trying to load a vitamin dump. It is not supported.");
```
1whatleytay, line 347: Let's do a popup if we can! [`error_dialog`](https://github.com/Vita3K/Vita3K/blob/master/vita3k/app/include/app/functions.h#L45) is good for this.

KorewaWatchful: Done.

vita3k/host/src/load_self.cpp
```diff
@@ -342,7 +342,12 @@ SceUID load_self(Ptr<const void> &entry_point, KernelState &kernel, MemState &me
         LOG_CRITICAL("SELF {} header type {} is not supported.", self_path, self_header.header_type);
         return -1;
     }
-
+    
+    if (self_path == "app0:sce_module/steroid.suprx") {
+        app::error_dialog("You're trying to load a vitamin dump. It is not supported.", nullptr);
```
totlmstr, line 347: Seems unnecessary to place this fix here.

Why not in `interface.cpp` instead?

1whatleytay:
Yeah sorry maybe we should move it into something in app...

We have a circular dependancy now where `host` depends on `app` and `app` depends on `host`.

1whatleytay:
Optimistic merging.

ichisadashioko:
I think we should just warn the user but not prevent them from adding vitamin dump. Otherwise, user will have to manually unzip and copy it to the `rom` directory.

PR #686, opened 2020-04-23
Update build instructions

ichisadashioko:
- Update build instruction so that we can build locally on Linux and Mac OS.
- Fix missing headers for `ext-boost`.

This will help anyone who is interested in the project able to build locally on Linux and Mac OS.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@
 .vs
 .vscode
 build
+# generated on macOS
+.DS_Store
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
 option(USE_GDBSTUB "Build Vita3K with GDB Debugger." OFF)
-option(USE_DISCORD_RICH_PRESENCE "Build Vita3K with Discord Rich Presence" ON)
+option(USE_DISCORD_RICH_PRESENCE "Build Vita3K with Discord Rich Presence" OFF)
 option(USE_VULKAN "Build Vita3K with Vulkan backend." OFF)
 
 find_program(CCACHE_PROGRAM ccache)
--- a/building.md
+++ b/building.md
@@ -1,0 +1,105 @@
+# Build Vita3K
+
+## Windows (Visual Studio)
+
+- Install Visual Studio 2017 or 2019 and choose to install `Desktop development with C++`. You will get compiler and `cmake` required for building.
+
+Example for Visual Studio 2019:
+
+![Required tools for VS 2019](https://i.imgur.com/bkY15Oh.png)
+
+- Install `git` to `clone` the project. Download and install `git` from [here](https://git-scm.com).
+
+- Clone this repo.
+
+```cmd
+git clone --recursive https://github.com/Vita3K/Vita3K
+cd Vita3K
+```
+
+- Open `Developer Command Prompt` for your respective Visual Studio version. `cmake` and other `C/C++` toolchain only available on `Developer Command Prompt`. 
+
+Example for Visual Studio 2019:
+
+![Developer Command Prompt for VS 2019](https://i.imgur.com/w6Umx1S.png)
+
+- Run `gen-windows.bat` to create Visual Studio project in `build-windows` directory.
+
+```cmd
+.\gen-windows
+```
+
+- Open the project generated in the `build-windows` directory.
+
+- Build and run it.
+
+## macOS (Xcode)
+
+- Install Xcode at App Store.
+
+- Install [`brew`](https://brew.sh).
+
+- Install dependencies with `brew`.
+
+```sh
+brew install git cmake
+```
+
+- Clone this repo.
+
+```sh
+git clone --recursive https://github.com/Vita3K/Vita3K
+cd Vita3K
+```
+
+- Generate Xcode project.
+
+```
+./gen-macos.sh
+```
+
+- Open Xcode project `vita3k.xcodeproj` generated in `build-macos` directory.
+
+- When prompted to create schemes, create one for the `vita3k` target only. The project builds many targets, so it will make your life easier if you create schemes as needed.
+
+- Build.
+
+## Linux
+
+### Ubuntu/Debian
+
+- Install dependencies.
+
+```sh
+sudo apt install git cmake libsdl2-dev pkg-config libgtk-3-dev
+```
+
+- Clone this repo.
+
+```sh
+git clone --recursive https://github.com/Vita3K/Vita3K
+cd Vita3K
+```
+
+- Build the project.
+
+```sh
+./gen-linux.sh
+cd build-linux
+make -j$(nproc)
+```
+
+__Note:__ If Unicorn can't find Python, use `make UNICORN_QEMU_FLAGS="--python=/usr/bin/python2"` or similar to point it to your installation. `make` works after you do this once.
+
+## Note
+
+- After cloning or checking out a branch, you should always update submodules.
+
+`git submodule update --init --recursive`
+
+- If `boost` failed to build, you can opt out for system `boost` package (Linux and macOS only).
+
+```sh
+brew install boost # for macOS
+sudo apt install libboost-all-dev # for Ubuntu/Debian
+```
--- a/gen-linux.sh
+++ b/gen-linux.sh
@@ -7,12 +7,11 @@ if [[ -z "${CI}" ]]; then
 	mkdir -p external/boost-build
 	cd external/boost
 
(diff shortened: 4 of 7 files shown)


CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
 option(USE_GDBSTUB "Build Vita3K with GDB Debugger." OFF)
-option(USE_DISCORD_RICH_PRESENCE "Build Vita3K with Discord Rich Presence" ON)
+option(USE_DISCORD_RICH_PRESENCE "Build Vita3K with Discord Rich Presence" OFF)
```
KorewaWatchful, line 9: I'm not sure that you should default this option to OFF, as most people use windows where this is functional.

building.md
```diff
@@ -0,0 +1,86 @@
...
+- When prompted to create schemes, create one for the `emulator` target only. The project builds many targets, so it will make your life easier if you create schemes as needed.
+
+- Build.
+
+### Linux (Debian)
+
+- Install dependencies.
+
+```sh
+sudo apt install git cmake libboost-all-dev libsdl2-dev
```
KorewaWatchful, line 62: I think this should be more than just Debian, as it's possible to build this on other distros. Instead of giving a debian command-line example, maybe just list the dependencies like: cmake, boost, sdl2, gtk3, pkg-config. (you missed the last 2)

building.md
```diff
@@ -0,0 +1,86 @@
...
+make -j$(nproc)
+```
+
+#### Note: If Unicorn can't find Python, use `make UNICORN_QEMU_FLAGS="--python=/usr/bin/python2"` or similar to point it to your installation. `make` works after you do this once.
+
+## Note
+
+- After cloning or checking out a branch, you should always update submodules.
+
+`git submodule update --init --recursive`
```
KorewaWatchful, line 86: Since this applies to everything and you should do this before building, maybe put this on the top?

CMakeLists.txt
ichisadashioko: For building the first time, I think it is important that people should able to build without trouble. Failing to build without knowing why is quite frustrating.

building.md
ichisadashioko: It is because we give instruction for installing dependencies with `apt`. For other package managers, if anyone is more familiar with it, we should let them put in the instructions. Yeah, I think we should have sub-sections for `Linux`.

About dependencies, I will re-check them and update. Thanks!

building.md
ichisadashioko: I also think this is important but putting it first doesn't make much sense. This should be applied when someone is actively changing sub-modules or generating untracked files in sub-modules.

It just not make much sense for putting it on top. I think we should have a `CONTRIBUTING.md` later and move technical notes to there.

ichisadashioko: `git clone --recursive` will get the job done for one-time visitors.

CMakeLists.txt
1whatleytay: I agree that this should be off. It was off initially too.

building.md
```diff
@@ -0,0 +1,86 @@
+# Build Vita3K
+
+## Windows (Visual Studio)
+
+- Install Visual Studio 2017 or 2019 and choose to install `Desktop development with C++` (as shown in this [image](https://i.imgur.com/bkY15Oh.png)). You will get compiler and `cmake` required for building.
```
1whatleytay, line 5: Maybe we should have an `images` folder. We are already using other images in README.

building.md
```diff
@@ -0,0 +1,86 @@
...
+
+- Generate Xcode project.
+
+```
+./gen-macos.sh
+```
+
+- Open Xcode project `Vita3K.xcodeproj` generated in `build-macos` directory.
+
+- When prompted to create schemes, create one for the `emulator` target only. The project builds many targets, so it will make your life easier if you create schemes as needed.
```
1whatleytay, line 53: It's called `vita3k` now. We should change this in our wiki too- oops.

gen-linux.sh
```diff
@@ -1,23 +1,5 @@
...
-	# Non-Windows needs to build Boost.Build first
-	chmod +x tools/build/src/engine/build.sh
-	sh bootstrap.sh
-
-	# Build our Boost subset
-	./b2 --ignore-site-config -j5 --build-dir=../boost-build --stagedir=../boost-build stage
-	cd ../..
-fi
-
-# Generate project files
```
1whatleytay, line 19: I don't think this is the strat. The best solution is to just fix what's wrong with https://github.com/Vita3K/ext-boost I think. Especially since it also doesn't work on windows (tried last night :/).

building.md
ichisadashioko: I think we shouldn't include binary images in the repo as it might be changed and increase the repo size significantly. Yeah, someone might remove the images or the hosting providers will remove our images sometime in the future. But these images are not essential to our project so we can replace them anytime we want.

gen-linux.sh
ichisadashioko: Yeah, I think we should fix `ext-boost` too. I think we should either include the full `boost` (lazy and fast) or write some scripts to read all C header files in `Vita3K` to pull all the required files from `boost` which will take quite a while.

ichisadashioko: Let's update `Vita3K` after `ext-boost` is fixed. I was temping to remove `ext-boost` from `.gitmodule` a while ago.

ichisadashioko: We updated `ext-boost` in this [PR](https://github.com/Vita3K/ext-boost/pull/2). Please check to see if they are all working correctly with the updated `ext-boost` (`ext-boost` reference has been updated).

gen-windows.bat
```diff
@@ -11,8 +11,10 @@ IF "%CI%"=="" IF NOT EXIST external\boost-build (
 	mkdir external\boost-build
 	cd external\boost
 
+	call bootstrap.bat
+
 	REM Build our Boost subset
-	b2 -j5 --build-dir=../boost-build --stagedir=../boost-build toolset=msvc stage
+	call b2 -j5 --build-dir=../boost-build --stagedir=../boost-build toolset=msvc stage
```
ichisadashioko, line 17: Is there a `$(nproc)` equipvalent for Windows?

ichisadashioko:
There is some problem with `ext-boost`. I will fix it before re-opening this PR.

.gitignore
```diff
@@ -7,4 +7,9 @@
...
 .vs
 .vscode
-build
\ No newline at end of file
+build
+# generated on macOS
+*.cmake
+*.log
+CMakeFiles
+.DS_Store
```
1whatleytay: Umm... this stuff should be a build folder, which is included above. Maybe another commit?

ichisadashioko: These files may be generated everywhere. Users may do something like this to store build log for example `./gen-linux.sh > gen-linux.log`. `.DS_Store` may be generated in every directories that are visited by Finder in macOS.

1whatleytay: Ehhh? What if they try to do `./gen-linx.sh > gen-linux.txt`? I wouldn't just exclude files willy nilly like that.

ichisadashioko: I think `.log` is a widely used extension for storing `log`. I am not sure if that is the case for `.txt`. I am not saying that we should include every possible extensions that users might use - just acceptable ones.

pent0: The *.log are personal preferences. I think unless you can't do anything about it (like *.cmake are generated by cmake tools, .DS_Store are generated by system, .obj by compiler), should not include it in .gitignore, just manage it yourself

gen-windows.bat
ichisadashioko: We have `%NUMBER_OF_PROCESSORS%` for Windows.
Everyone can try to `echo %NUMBER_OF_PROCESSORS%` in Command Prompt.

.gitignore
ichisadashioko: Ok. I will remove `*.log` entry.

1whatleytay: Wait wait, we already exclude CMakeFiles and `*.cmake` though by `build-*`, why do we have to exclude them again?

ichisadashioko: They are generated in project directory when building on macOS. Just to make sure `cmake` configured information is stored in `CMakeList.txt`, isn't it?

1whatleytay: Ehh? I work on macos and I have no problem with any cmake files (other than the .DS_Store stuff, that's annoying). You should generate your cmake files in a subdirectory via `cmake ..` so you still have control over them. It's not a good idea to mix build files and source files anyway.

ichisadashioko: I will re-check it. If it is you said, I will update the PR.

ichisadashioko: Yeah. They aren't generated when I re-check the build. I updated the PR.

PR #692, opened 2020-05-05
emu: fix homebrew installation and add game patches

IllusionMan1212:
~`git blame watchful`~
i also added game patches just so they don't fall under the last else statement and overwrite the original game files

### DIFF
--- a/vita3k/interface.cpp
+++ b/vita3k/interface.cpp
@@ -122,15 +122,18 @@ bool install_vpk(HostState &host, GuiState &gui, const fs::path &path) {
     sfo::get_data_by_key(host.game_version, sfo_handle, "APP_VER");
     sfo::get_data_by_key(host.game_category, sfo_handle, "CATEGORY");
     sfo::get_data_by_key(content_id, sfo_handle, "CONTENT_ID");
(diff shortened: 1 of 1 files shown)


KorewaWatchful:
but `ux0:patch` is not supported right?

IllusionMan1212:
game patches don't work in `ux0:patch` afaik, but this is just a precautionary measure if someone decided to install a patch

vita3k/interface.cpp
```diff
@@ -122,15 +122,18 @@ bool install_vpk(HostState &host, GuiState &gui, const fs::path &path) {
     sfo::get_data_by_key(host.game_version, sfo_handle, "APP_VER");
     sfo::get_data_by_key(host.game_category, sfo_handle, "CATEGORY");
     sfo::get_data_by_key(content_id, sfo_handle, "CONTENT_ID");
-    dlc_foldername = content_id.substr(20);
     fs::path output_path;
 
     if (host.game_category == "gd") {
```
Zangetsu38, line 127: delete this if, is not nheed with using else in end ^^ 
```suggestion
```

PR #694, opened 2020-05-07
Replace AppVeyor/Travis by GitHub Actions

scribam:
Example on my fork (with ccache enabled): https://github.com/scribam/Vita3K/actions/runs/99061727

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -1,0 +1,64 @@
+name: C/C++ CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest, ubuntu-latest, windows-latest]
+        config: [Release]
+
+    steps:
+      - name: Set up build environment (macos-latest)
+        run: |
+          brew install ccache boost
+          echo "::add-path::/usr/local/opt/ccache/libexec"
+          echo "::set-env name=CCACHE_DIR::/tmp/ccache"
+        if: matrix.os == 'macos-latest'
+
+      - name: Set up build environment (ubuntu-latest)
+        run: |
+          sudo apt-get update
+          sudo apt-get -y install ccache libboost-filesystem-dev libboost-program-options-dev libboost-system-dev libgtk-3-dev libsdl2-dev
+          echo "::set-env name=CCACHE_DIR::/tmp/ccache"
+        if: matrix.os == 'ubuntu-latest'
+
+      - name: Set up build environment (windows-latest)
+        run: |
+          echo "::set-env name=BOOST_ROOT::$env:BOOST_ROOT_1_72_0"
+        if: matrix.os == 'windows-latest'
+
+      - uses: actions/cache@v1
+        with:
+          path: /tmp/ccache
+          key: ccache-${{ matrix.os }}-${{ matrix.config }}-${{ github.sha }}
+          restore-keys: ccache-${{ matrix.os }}-${{ matrix.config }}-
+        if: matrix.os != 'windows-latest'
+
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          submodules: recursive
+
+      - name: CMake configure
+        run: cmake -B build -DCI=ON -DCMAKE_BUILD_TYPE=${{ matrix.config }}
+
+      - name: CMake build
+        run: cmake --build build --config ${{ matrix.config }} --parallel 2
+
+      - name: CTest
+        working-directory: build
+        run: ctest --build-config ${{ matrix.config }} --output-on-failure --parallel 2
+
+      - name: Compute git short sha
+        id: git_short_sha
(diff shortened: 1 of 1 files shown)


KorewaWatchful:
woah

1whatleytay:
zxsfsdfdsf
Appveyor does not approve XD

I think this is a good idea. Not sure if we're paying appveyor/travis anything right now, but I'm glad github actions is free (I think).

.github/workflows/c-cpp.yml
```diff
@@ -0,0 +1,60 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest, ubuntu-latest, windows-latest]
+        config: [Release]
+
+    steps:
+      - name: Set up build environment (macos-latest)
+        run: |
+          brew install ccache boost
```
ichisadashioko, line 18: Should we use pre-built `boost`? Someone could break local `boost` or use some `boost` features that haven't beed added to `ext-boost` and CI will ignore it. 🙈

.github/workflows/c-cpp.yml
```diff
@@ -0,0 +1,60 @@
...
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          submodules: recursive
+
+      - name: CMake configure
+        run: cmake -B build -DCI=ON -DCMAKE_BUILD_TYPE=${{ matrix.config }}
+
+      - name: CMake build
+        run: cmake --build build --config ${{ matrix.config }} --parallel 2
```
Margen67, line 51: Why not use `j` instead of `parallel`?

ichisadashioko: Does Github Action have cross-platforms `nproc`? This will reduce build time significantly.

ichisadashioko: Now, I know why those builds took forever to complete on CI.

scribam: @Margen67 `parallel` and `j` are the same. Source: https://cmake.org/cmake/help/latest/manual/cmake.1.html

@ichisadashioko As far as I know, GitHub Actions don't have cross-platform `nproc`. I an using the value "2" because it is the number of cpu available. Source: https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources

.github/workflows/c-cpp.yml
scribam: Or should we have builds with pre-built `boost` and others without to cover all cases? Currently, I think only the macOS builds on Travis use `boost` from the submodule.

ichisadashioko: What should we do about pre-built binaries @1whatleytay ?

.github/workflows/c-cpp.yml
Margen67: > 
> 
> @Margen67 `parallel` and `j` are the same. Source: https://cmake.org/cmake/help/latest/manual/cmake.1.html
> 
> @ichisadashioko As far as I know, GitHub Actions don't have cross-platform `nproc`. I am using the value "2" because it is the number of cpu available. Source: https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources

For Windows there's an environment variable: `NUMBER_OF_PROCESSORS`
For macOS you can do use this (or a similar) command: `sysctl hw.ncpu`

ichisadashioko: Yeah, but Github Action uses the same command for all OSs.

1whatleytay:
Will deployment be an issue?

scribam:
All generated artifacts will be visible on GitHub Actions. If you go there (https://github.com/scribam/Vita3K/actions/runs/99061727), you can see 3 artifacts, one for each OS. Let me know if it fits your needs.
This will also mean no more tags or releases like it is currently done at each merge/push to master

.github/workflows/c-cpp.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      - name: CTest
+        working-directory: build
+        run: ctest --build-config ${{ matrix.config }} --output-on-failure --parallel 2
+
+      - uses: actions/upload-artifact@v2
+        with:
+          name: vita3k-${{ github.sha }}-${{ matrix.os }}
```
ichisadashioko, line 59: We should shorten the hash. This is too long and I have no idea how to do it because some nasty OS-specific command.

Margen67: `git rev-parse --short HEAD`?

scribam: Done

ichisadashioko:
We can keep both CI systems together. That will let maintainers decide when to deprecate one and update the release pipeline.

Margen67:
> 
> 
> We can keep both CI systems together. That will let maintainers decide when to deprecate one and update the release pipeline.

Travis is on an outdated version of macOS: #660

scribam:
Fine to me, AppVeyor and Travis restored

ichisadashioko:
LGTM

PR #700, opened 2020-05-12
gui: Add missing PopStyleColor

scribam:
The number of PopStyleColor calls must match the number of PushStyleColor calls. It triggers an assert from ImGui when it is not the case on a debug build.


Zangetsu38:
Are you sur ?  I have not see this problème and i have good set it normally

Zangetsu38:
i see just here lol, why i have forget one 🤦

Zangetsu38:
@scribam you can close it now, sorry for disturb, i have open my proper pr for fix it with fix also text size on app list

PR #721, opened 2020-05-23
kernel: Fix "Load of value 102, which is not a valid value for type 'Code'" from UBSan

scribam:
Triggered when a cast to `Code` is done with a value which does not match the enum values.

Found in Flood It! as the game has an unhandled relocation code:
> |W| [relocate_entry]: Unhandled relocation code 102.


Zangetsu38:
oh gret, i have see it some time

PR #727, opened 2020-05-29
external: Rework unicorn build integration

scribam:
The latest versions of unicorn bring a CMakeLists.txt file which I used to simplify the build process of Vita3k.
Python is no longer required to build unicorn.


scribam:
Ready for review

Issue #726, opened 2020-05-28
Add option to give the path of work.bin instead of zRIF


EXtremeExploit:
self explanatory

Zangetsu38:
Hmm i can probably set this

EXtremeExploit:
i was just saying cuz it's awful to use a converter to convert the work.bin to a zRIF, it would be less of a problem to just do it automatically

PR #729, opened 2020-05-30
pkg: add mode choice for zRIF key.

Zangetsu38:
-  Refactor look design
- Add choice for can inpout zRIF key or select work.bin file for key with fix issue request #726
- Add info content installed. 
- Fix theme list not showed if fw not installed

# Result: 
image

### DIFF
--- a/vita3k/gui/src/pkg_install_dialog.cpp
+++ b/vita3k/gui/src/pkg_install_dialog.cpp
@@ -21,111 +21,132 @@
 #include <host/functions.h>
 #include <host/pkg.h>
 #include <misc/cpp/imgui_stdlib.h>
+#include <rif2zrif.h>
 #include <util/log.h>
 
 #include <nfd.h>
 
 namespace gui {
 
-nfdchar_t *pkg_path = nullptr;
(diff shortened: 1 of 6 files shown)


Zangetsu38:
change pushed with totaly refactor pkg dialog look

vita3k/gui/src/pkg_install_dialog.cpp
```diff
@@ -21,111 +21,132 @@
...
 
 #include <nfd.h>
 
 namespace gui {
 
-nfdchar_t *pkg_path = nullptr;
+static nfdchar_t *pkg_path, *work_path = nullptr;
+static std::string state, title, zRIF;
+static bool draw_file_dialog = true;
+static bool delete_pkg_file, delete_work_file = false;
```
KorewaWatchful, line 34: desgroup asked not to use static variables in the gui, that is why they are in state.h

Zangetsu38: Set this inside state no have sense, is only for pkg dialog, and use static here is absoluly not problème

PR #733, opened 2020-06-02
vita3k: recommend ninja over make for compiling on linux

Johnnynator:
this increases the build time tremendously since ninja actually compiles
more files in parallel, while the Makefile results that both nids.cpp
and module_parent.cpp are compiled alone, which probably doubles intial
build time (if not more)

I can provide exact numbers if wanted tomorrow
And update the ci script probably.

### DIFF
--- a/building.md
+++ b/building.md
@@ -71,7 +71,7 @@ cd Vita3K
 - Install dependencies.
 
 ```sh
-sudo apt install git cmake ninja-build libsdl2-dev pkg-config libgtk-3-dev
+sudo apt install git cmake ninja-build libsdl2-dev pkg-config libgtk-3-dev clang
 ```
 
 - Clone this repo.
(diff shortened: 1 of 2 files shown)


Johnnynator:
Make build time (already partially cached, so full build would be actually slightly slower):
```
Executed in   57.21 mins   fish           external 
   usr time   83.02 mins  261.00 micros   83.02 mins 
   sys time    3.14 mins   44.00 micros    3.14 mins 
```
Build time with ninja (full clean build)
```
Executed in   30.74 mins   fish           external 
   usr time   90.66 mins  267.00 micros   90.66 mins 
   sys time    1.68 mins  122.00 micros    1.68 min
```
So ninja halves the build time at least on my machine, but should help any system that has 2 threads or more..

Johnnynator:
This requires #727 , right now a truly fresh build will fail with
`ninja: error: '../external/unicorn-src/libunicorn.a', needed by 'bin/Vita3K', missing and no known rule to make it`

Johnnynator:
I did rebase my branch

PR #737, opened 2020-06-04
Improve gdb and add code injection to address

sunho:
As the TODO comment in gdp.cpp foreshadowed, the breakpoint doesn't work in thumb mode. Fortunately, IDA provided the instruction set information through gdb protocol. I used it to support thumb breakpoint.

Additionally, I added manual code injection. (I couldn't figure out better name 😢 )

While I'm exploring the assembly of the game, I discovered a lot of log functions without printing to stdout. (maybe they were removed in deployment process) However, it's tiring to use gdb breakpoint in this case. I don't want to stop the game every time it logs something. Thus, I supported something like below.

```c
add_breakpoint(host.mem, false, true, 0x810B431C, [](CPUState& cpu, MemState& mem) -> void {
    auto msg_ptr = Ptr<char>(read_reg(cpu, 0)); // argument 0
    auto variable_ptr = Ptr<void>(read_reg(cpu, 1)); // argument 1
    auto msg = msg_ptr.get(mem);
    auto variable = variable_ptr.get(mem);
    printf(("[printf] " + std::string(msg) + "\n").c_str(), variable);
});
```
By adding this code to interface.cpp, when the instruction at 0x810B431C is launched, the lambda function will be run, and then, it returns to the ordinary routine. (like breakpoint does)

Here's the result:
Screen Shot 2020-06-05 at 1 22 04 AM

This can be the base for modding. We can freely add and remove the code. Especially, we can use the power of libraries outside the emulated environment this way.


vita3k/renderer/src/gl/texture.cpp
```diff
@@ -109,7 +109,7 @@ static size_t decompress_compressed_swizz_texture(SceGxmTextureBaseFormat fmt, v
         return (((width + 3) / 4) * ((height + 3) / 4) * ((ubc_type > 1) ? 16 : 8));
     } else if ((fmt >= SCE_GXM_TEXTURE_BASE_FORMAT_PVRT2BPP) && (fmt <= SCE_GXM_TEXTURE_BASE_FORMAT_PVRTII4BPP)) {
         // TODO, is not perfect for PVRT-II.
-        pvr::PVRTDecompressPVRTC(data, (fmt == SCE_GXM_TEXTURE_BASE_FORMAT_PVRT2BPP) || (fmt == SCE_GXM_TEXTURE_BASE_FORMAT_PVRTII2BPP), width, height, 
+        pvr::PVRTDecompressPVRTC(data, (fmt == SCE_GXM_TEXTURE_BASE_FORMAT_PVRT2BPP) || (fmt == SCE_GXM_TEXTURE_BASE_FORMAT_PVRTII2BPP), width, height,
```
frangarcj, line 112: Remove this from pr

sunho: done

PR #740, opened 2020-06-05
Add hle nid variable exports

sunho:
# About this PR

This PR adds support for HLE NID variable exports and export _pLibPerfCaptureFlagPtr as non-zero value.

# Related bug

The saekano game crashed immediately without even tracing the stack. When the _sceUltUlthreadRuntimeCreate was run, it printed that _sceFiberInitializeWithInternalOptionImpl was not implemented and crashed. This kind of error was reported before in https://github.com/Vita3K/compatibility/issues/331.  

It seemed like _sceFiberInitializeWithInternalOptionImpl was the problem here. But, the actual code that crasehd the game turned out to be

```
ldr r1, [r0]
```

Which dereferenced some pointer. It turned out the code dereferenced the _pLibPerfCaptureFlagPtr which was imported in libult elf module. And, if the value is 0, it breakpoints, otherwise do nothing.

However, as _pLibPerfCaptureFlagPtr variable is not exported anywhere, the previous implementation just stubbed it with 0xDEADBEEF, causing the BAD_MEMORY_ACCESS.

### DIFF
--- a/vita3k/host/src/load_self.cpp
+++ b/vita3k/host/src/load_self.cpp
@@ -82,8 +82,6 @@ static bool load_var_imports(const uint32_t *nids, const Ptr<uint32_t> *entries,
         if (export_address_it != kernel.export_nids.end()) {
             export_address = export_address_it->second;
         } else {
-            // TODO: Proper HLE support for var imports (map with varnid, pointer, size)
-
             const char *const name = import_name(nid);
             constexpr auto STUB_SYMVAL = 0xDEADBEEF;
             LOG_WARN("\tNID NOT FOUND {} ({}) at {}, setting to stub value {}", log_hex(nid), name, log_hex(entry.address()), log_hex(STUB_SYMVAL));
--- a/vita3k/nids/CMakeLists.txt
+++ b/vita3k/nids/CMakeLists.txt
@@ -3,6 +3,7 @@ add_library(
 	STATIC
 	include/nids/functions.h
 	include/nids/nids.h
+	include/nids/nid_vars.h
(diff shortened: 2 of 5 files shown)


vita3k/nids/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ add_library(
 	STATIC
 	include/nids/functions.h
 	include/nids/nids.h
+    include/nids/nid_vars.h
```
VelocityRa, line 6: use 1 space here like the other lines

vita3k/host/src/load_self.cpp
```diff
@@ -97,7 +95,7 @@ static bool load_var_imports(const uint32_t *nids, const Ptr<uint32_t> *entries,
             // Use same stub for other var imports
             kernel.export_nids.emplace(nid, export_address);
         }
-
+
```
VelocityRa, line 98: remove this change

vita3k/nids/CMakeLists.txt
ichisadashioko: a tab?

VelocityRa: tab yeah

frangarcj:
One of our oldest ToDos. Thanks a lot.

sunho:
I updated my code following your reviews.

ichisadashioko:
Oh we need to keep all PR in one commit only. And the commit message needs to be a little more meaningful as you has been suggested on Discord.

sunho:
I squashed them into one commit.

PR #743, opened 2020-06-07
Improve code log and memory log & add toggling them through gui

sunho:
# About PR

- Add pc information in import call and memory log (unicorn hook)
- If the export function is reached, log its name
- Optionally add register values in code log (LOG_REGISTERS)
- Add variable access watch
- Add toggling code, memory log, and import call log through gui
- Add thread id in code and memory log

# Variable access watch

You can trace the access to variables you're interested in by adding the following code in interface.cpp or anywhere that can assess KernelState.

```cpp
add_watch_memory(kernel, 0x5353, 10); 
// 0x5353 is the address of the variable
// 10 is the size of the variable
```
# Reason for toggling

It's very hard to debug the unicorn errors since we don't have something like stack traceback. Code, memory, and import call logs might serve as a great alternative to the traceback. However, they drastically slow down the game to the extent it seems to take forever to reach the unicorn error point. (adding unicorn hooks slows down the emulation a lot) Thus, the ability to toggle them in runtime is really helpful.

# Screenshots

gui:
Screen Shot 2020-06-07 at 9 57 23 AM

improved logs:
Screen Shot 2020-06-07 at 9 58 14 AM

### DIFF
--- a/vita3k/cpu/include/cpu/functions.h
+++ b/vita3k/cpu/include/cpu/functions.h
@@ -18,6 +18,7 @@
 #pragma once
 
 #include <mem/mem.h> // Address.
+#include <util/types.h>
 
 #include <cstdint>
 #include <functional>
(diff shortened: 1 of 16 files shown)


sunho:
Done the requested changes. There was a bug in string_utils::replace. (if the in and out contains the same substring, infinite loop) I changed the code to fix the bug.

PR #747, opened 2020-06-09
SceFiber: implement null addrContext behaviour

sunho:
# About PR

- Restore the fiber with null addrContext back to the initial state in switchContext and returnToThread
- Implement _sceFiberInitializeWithInternalOptionImpl

# Related bug

Libult uses two fibers in each thread to receive a job. The first fiber waits for a job, and the second fiber executes the job. It initially fiberRun the first fiber. When a job arrives in the first fiber, it switches to the second fiber. If the job is completed in the second fiber. It switches back to the first fiber. However, if I just return to where switchFiber is called, it asserted.

Screen Shot 2020-06-10 at 8 32 27 AM

The first fiber was with null addrContext. If I restore first fiber back to initial state in fiberSwitch, this portion of libult works as expected.


vita3k/kernel/include/kernel/state.h
```diff
@@ -294,6 +305,9 @@ struct KernelState {
     SceKernelModuleInfoPtrs loaded_modules;
     LoadedSysmodules loaded_sysmodules;
     ExportNids export_nids;
+
+    InitialFibers initial_fibers;
```
frangarcj, line 309: Fibers are stored on vita at the thread level. is it possible to move this to thread state?

PR #749, opened 2020-06-10
Upload github CI master branch artifcats to a CI Release tag

Johnnynator:
This creates a Release tag that will contain the latest github ci builds so they can easily linked to. Right now it doesn't set it as pre release automatically and doesn't recreate the tag, so the release will show a old commit number (but contain the latest artifacts)

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -14,8 +14,21 @@ jobs:
       - name: Format check
         run: .github/format-check.sh
 
+  move-tag:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Move ci tag
+        uses: s3krit/walking-tag-action@d04f7a5
+        with:
+          TAG_NAME: continous
+          TAG_MESSAGE: |
+            Last commit build by the CI
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


Margen67:
*artifacts

sunho:
@Johnnynator can we merge this?

Johnnynator:
This is at least in a working state now. Not fully as I would like it (ci commit will be moved, even if builds do fail). But it kinda works for having a link to all the latest artifacts. And commit specific artifacts are available under the actions tab on github as always. So right now the tag should have the latest successful builds from each system.

.github/workflows/c-cpp.yml
```diff
@@ -74,3 +87,25 @@ jobs:
...
+
+      - name: Update the CI tag
+        uses: Xotl/cool-github-releases@v1
+        with:
+          mode: update
+          isPrerelease: true
+          tag_name: continous
+          release_name: "Automatic CI builds"
+          body_mrkdwn: |
+            TODO: fill this with something note worthy
```
sunho, line 106: Can you just remove this? Ill just create an issue for this.

Johnnynator: Just forgot about it, in theory just having the last successfully commit for reference in there would be enough. If someone is bored enough, they can add stuff like commit message.

PR #753, opened 2020-06-11
If it's not ci, ask for build

ibrakap:
This PR adds a build option if it's not ci.

### DIFF
--- a/gen-linux.sh
+++ b/gen-linux.sh
@@ -23,7 +23,17 @@ if [ "$CLANG" ]; then
 	cmake_args="-DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++"
 fi
 
+pushd .
+
 # Generate project files
 mkdir -p build-linux
 cd build-linux
 cmake .. -GNinja ${cmake_args}
+
+if [[ -z "${CI}" ]]; then
+	read -p "Do you want to build?(Y/n)" selection
+	if [ "${selection}"=="Y" ]; then
+		ninja
+	fi
+	popd
+fi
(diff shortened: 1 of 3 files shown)


1whatleytay:
I don't understand this PR. The script is called `gen-windows` so I'd presume it would *generate* the project files you need to build the project. I don't think it needs to build it. If we wanted to change it so the gen scripts *also* built the project, there should be changes to the `gen-macos` and `gen-linux` scripts too.

Zangetsu38:
same, no understund here

ibrakap:
Linux script updated

ibrakap:
I'm sorry for late update because installing mac os in virtual machine is headache. It took my days to configure useful setup. It's ready for reviews.

ichisadashioko:
I'm not sure about this. I'm not very font of terminal options. I would prefer it to do what I commanded and terminate then I will give commands to build or do whatever next. The prompt for choosing VS 2017 or VS 2019 has already annoyed me.

- On Windows, we will need to open Visual Studio for building and debugging.
- On macOS, we need Xcode to do those things.
- On Linux, we just need to run `ninja` to build.

ichisadashioko:
If people just want to build `vita3k` but not modify code then they can probably grab them from Github Actions, Travis, or AppVeyor. These scripts are for generating projects.

1whatleytay:
Yeah, if you're going to go through the trouble of building vita3k it will take two whole extra seconds to open up your IDE or run `make`. I am likely not going to approve this, but maybe the other maintainers will?

ibrakap:
I thought a little bit, @ichisadashioko was right. Using prompts not an efficient solution.
Instead of using prompts we can use command line parameters for both build and compiler version selection(I thought for gen-windows.bat but i would apply both gen-macos.sh and gen-linux.sh). Okey, now i will set as draft.

ibrakap:
I wrote something but it didn't satisfy to me. I'm closing this pr for now.

PR #757, opened 2020-06-11
Add fomat check github actions job

sunho:
# About PR

If code is not formatted, stop the whole workflow.

Shout out to @Johnnynator

# Format Check Fail Example

https://github.com/Johnnynator/Vita3K/pull/3

### DIFF
--- a/.github/format-check.sh
+++ b/.github/format-check.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+set -ex
+
+if [ "$GITHUB_EVENT_NAME" == "pull_request" ]; then 
+    for f in $(git diff --diff-filter=AM --name-only origin/$GITHUB_BASE_REF 'vita3k/**.cpp' 'vita3k/**.h' 'tools/**.cpp' 'tools/**.h'); do
+        if [ "$(diff -u <(cat $f) <(clang-format $f))" != "" ]
+        then
+            echo "run format"
+            exit 1
+        fi
+    done
+fi
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -3,7 +3,19 @@ name: C/C++ CI
 on: [push, pull_request]
 
 jobs:
+  format-check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Format check
+        run: .github/format-check.sh
+
   build:
+    needs: format-check
     runs-on: ${{ matrix.os }}
 
     strategy:


Johnnynator:
You should run that only on changed file, not on every file

format-check.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+set -ex
+
+for f in $(find vita3k tools/gen-modules tools/native-tool \( -name *.cpp -o -name *.h \)); do
```
Johnnynator, line 4:
```suggestion
for f in $(git diff --name-only ${GITHUB_BASE_REF:-HEAD~1} HEAD)
```
You could try like this. Not fully sure if it works as intended

Johnnynator: Scratch that, also needs a check if it is `.cpp` or `.h` So add smth like
```bash
if [ "${f%.cpp}" != "${f}" ] || [ "${f%.h}" != "${f}" ]
``` 
around the other if block

.github/workflows/c-cpp.yml
```diff
@@ -3,7 +3,17 @@ name: C/C++ CI
 on: [push, pull_request]
 
 jobs:
+  format-check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
```
Margen67, line 10: Why not use v2?

.github/workflows/c-cpp.yml
```diff
@@ -3,7 +3,17 @@ name: C/C++ CI
...
+
+      - name: Format check
+        run: bash format-check.sh
```
Margen67, line 13:
```suggestion
        run: ./format-check.sh
```

sunho:
For some reason I can't comment on the reviews. I'm writing down here.

Margen67// v2 doesn't fetch old commits

Johnnynator// I updated the PR using your code

Margen67:
> 
> 
> For some reason I can't comment on the reviews. I'm writing down here.
> 
> Margen67// v2 doesn't fetch old commits

```
with:
  fetch-depth: 0
```
makes it fetch all commits

.github/format-check.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+set -ex
+
+if [ "$GITHUB_EVENT_NAME" == "pull_request" ]; then 
+    for f in $(git diff --name-only origin/$GITHUB_BASE_REF | grep -E "(vita3k|tools/gen-modules|tools/native-tool)/.*(.cpp|.h)" ); do
```
Johnnynator, line 6: You could use ```git diff --name-only origin/$GITHUB_BASE_REF HEAD 'vita3k/**.cpp' 'vita3k/**.h' 'tools/**.cpp' 'tools/**.h'``` to make this line like 10+ times faster, but I suppose it doesn't matter much. In theory you could also add `--diff-filter=AM` (only show `added` or `modified` files, could be extended to`R` and `C` (renamed/copied)) so you don't need to check if the file exists.
But otherwise it looks ok.

sunho: ✅

.github/workflows/c-cpp.yml
sunho: ✅

.github/format-check.sh
```diff
@@ -0,0 +1,13 @@
...
+
+if [ "$GITHUB_EVENT_NAME" == "pull_request" ]; then 
+    for f in $(git diff --diff-filter=AM --name-only origin/$GITHUB_BASE_REF 'vita3k/**.cpp' 'vita3k/**.h' 'tools/**.cpp' 'tools/**.h'); do
+        if [ "$(diff -u <(cat $f) <(clang-format $f))" != "" ]
+        then
+            echo "run format"
+            exit 1
+        fi
+    done
+fi
```
1whatleytay, line 13: Maybe a newline here?

PR #764, opened 2020-06-15
PVF initial implementation

sunho:
# About PR

This PR implements basic functionalities of pvf library with HLE approach.

# Why HLE

Some games load every glyph in the loading stage. LEE pvf was slowing down this process.


sunho:
Ready for review

VelocityRa:
> Some games load every glyph in the loading stage. LEE pvf was slowing down this process.

What do you mean? What exactly is slow? If it's I/O bottlenecked how does HLE fix the issue? Did you do any benchmarks to see if it's actually faster and by how much?

I don't see the point of this frankly, it's incomplete and even if it was complete (which would be a lot more code/effort I'm guessing) there's a lot of maintainance burden.
All while we can just load the library as we've been doing and get perfect accuracy/compatibility.

There really needs to be a good reason for this to exist for me to agree with merging.

sunho:
Here's my benchmark result.
| Build | HLE | LLE |
| ----- | ---- | ---- |
| Debug | 8s | 35s |
| Release | 1s | 4s |

Used this code to benchmark.
```cpp
static int time_;
add_breakpoint(host.mem, false, true, 0x810AC932, [](CPUState& cpu, MemState& mem) {
  time_ = time(NULL);
});
add_breakpoint(host.mem, false, true, 0x810AE15E, [](CPUState& cpu, MemState& mem) {
  LOG_INFO("BENCHMARK {}", time(NULL) - time_);
});
```

Here's information about my system.
Screen Shot 2020-06-19 at 6 30 05 AM

The game I tested was PCSG00543.

I'm not specifically sure why it's slow. I don't think IO is the problem here. Because the bottleneck was caused by not openfont but getcharinfo and getimage.

I'm seeing the libpvf is calling a lot of memset.

**Stack traceback of our clibMemset:**
Screen Shot 2020-06-19 at 6 24 52 AM

I guess interrupts are slowing down the library here?

I agree with you that I need to work quite a lot to complete this work... It was more complicated than my thought. For that reason, we should use LLE pvf for now. But, I also think that we shouldn't just discard the code here. It can be the base for perfect HLE pvf that works great.

VelocityRa:
Hm, alright.
Someone else can review.
You should probably test more than 1 game btw.

sunho:
I close this as I thought now is not the right time to do this. I will pr the complete one at some moment.

PR #770, opened 2020-06-19
decrypt more firmware install and module loading change

Zangetsu38:
Split refactor kernel pr
- fw install/pup: decrypt full os0 and vs0 partiton
- interface/run app: load fw pre-load module if app don't have it.

### DIFF
--- a/vita3k/host/src/pup.cpp
+++ b/vita3k/host/src/pup.cpp
@@ -253,25 +253,36 @@ void install_pup(const std::string &pup, const std::string &pref_path) {
 
     if (fs::file_size(pup_dec + "/os0.img") > 0) {
         extract_fat(pup_dec, "os0.img", pref_path);
-    }
-    if (fs::file_size(pup_dec + "/sa0.img") > 0) {
-        extract_fat(pup_dec, "sa0.img", pref_path);
-    }
-    if (fs::file_size(pup_dec + "/vs0.img") > 0) {
-        extract_fat(pup_dec, "vs0.img", pref_path);
-        for (const auto &file : fs::recursive_directory_iterator(pref_path + "/vs0/app")) {
-            if (file.path().filename() == "eboot.bin") {
-                self2elf(file.path().string(), file.path().string() + "elf", SCE_KEYS, 0);
-                fs::rename(file.path().string() + "elf", file.path().string());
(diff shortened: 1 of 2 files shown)


vita3k/host/src/pup.cpp
```diff
@@ -253,25 +253,34 @@ void install_pup(const std::string &pup, const std::string &pref_path) {
...
+        for (const auto &file : fs::recursive_directory_iterator(pref_path + "/vs0")) {
+            if (fs::is_regular_file(file.path())) {
+                const auto extention = file.path().filename().extension();
+                const auto is_self = ((extention == ".suprx") || (extention == ".skprx") || (extention == ".self"));
+                if ((file.path().filename() == "eboot.bin") || is_self) {
+                    self2elf(file.path().string(), file.path().string() + "elf", SCE_KEYS, 0);
+                    fs::rename(file.path().string() + "elf", file.path().string());
+                    make_fself(file.path().string(), file.path().string() + "fself");
+                    fs::rename(file.path().string() + "fself", file.path().string());
+                }
```
1whatleytay, line 282: Extract `file.path().string()` lol, you use it like 17 times here and lines of code above.

I suggest something like the following with better naming lol
```
        for (const auto &file : fs::recursive_directory_iterator(pref_path + "/vs0")) {
            fs::path filePath = file.path();
            std::string fileString = filePath.string();

            if (fs::is_regular_file(filePath) {
                const auto extention = filePath.filename().extension();
                const auto is_self = ((extention == ".suprx") || (extention == ".skprx") || (extention == ".self"));
                if ((file.path().filename() == "eboot.bin") || is_self) {
                    self2elf(fileString, file.path().string() + "elf", SCE_KEYS, 0);
                    fs::rename(fileString + "elf", fileString);
                    make_fself(fileString, fileString + "fself");
                    fs::rename(fileString + "fself", fileString);
                }
            }
        }
```

Zangetsu38: hmm i don't like lokl is more ugly with more create variable not needed ^^

Zangetsu38: code for this is untouched, i have just added for check extension file :)

Zangetsu38: i have change one things like you suggest :)

PR #777, opened 2020-06-22
renderer: consider different uniform sampler order

sunho:
# About PR

Samplers cannot be bound by uniform buffer object. We need to bind them manually.

# Fixes

I think, literally all games that use shader with uniform sampler

I checked that it fixes lbp intro screen little bit.

### DIFF
--- a/vita3k/renderer/include/renderer/gl/types.h
+++ b/vita3k/renderer/include/renderer/gl/types.h
@@ -36,9 +36,6 @@ typedef std::unique_ptr<void, std::function<void(SDL_GLContext)>> GLContextPtr;
 struct SceGxmProgramParameter;
 
 namespace renderer::gl {
-static constexpr GLint COLOR_ATTACHMENT_TEXTURE_SLOT_IMAGE = 0; ///< The slot that has our color attachment (for programmable blending) - image2D.
-static constexpr GLint COLOR_ATTACHMENT_TEXTURE_SLOT_SAMPLER = 12; ///< The slot that has our color attachment (for programmable blending) - sampler2D.
-
 struct ExcludedUniform {
     std::string name;
     GLuint program;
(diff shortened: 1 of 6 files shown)


sunho:
This is maybe workaround. It seems making sceGxmSetFragmentTexture to bind its sampler is more legitimate way.

pent0:
Doing this is right! Sometimes textures are not in order or because of vendor (AMD), their location in generated shader might be different from what's in the GXP (location 1 might not be for texture 1). Though don't do this every frame, and I prefer the location (binding = ) method, but not all OpenGL version supported it. Fine not listening to me and do your own investigation instead though.

sunho:
Of course, I will mind you. :) I looked around based on your comment, and I was able to grasp what's going on. Thank you so much for helpful hints. I'm going to edit my code to fix the problem properly.

sunho:
Complete. I also like (binding = ) method. Is there a reason why we're not using glsl 420 all the time?

pent0:
It's at least OpenGL 4.2, unless there is support for Vulkan so that Mac users can be able to use this emulator, you are stuck with OpenGL 4.1.

sunho:
I see. Then, we can't help it.

VelocityRa:
Well we can, if we drop macOS.
I see no issue with that, as I've said many times.

We can't keep wasting so much effort working around its shitty old drivers.
Plus it has deprecated OpenGL.

If it wasn't for @1whatleytay using it I would have already done so myself a while ago.

When our Vulkan backend is mature enough we can get MoltenVK to work with it.

1whatleytay:
Hi, I'm still working on vulkan stuff.

In the meantime, can you wrap it in `#ifdef APPLE`? I thought @sunho used macOS too... I guess I am alone :(

sunho:
I switched to windows for vita3k! I updated the code to use binding method if opengl4.2 is available. On the day we can use vulkan backend confidently, we can just drop every if (!features.use_binding)

PR #779, opened 2020-06-22
renderer: add texture dumping

sunho:
# About PR

Based on https://github.com/Vita3K/Vita3K/pull/522 by @VelocityRa 

I elaborated logging and supported compressed texture dumping.

### DIFF
--- a/vita3k/renderer/src/gl/sync_state.cpp
+++ b/vita3k/renderer/src/gl/sync_state.cpp
@@ -12,6 +12,8 @@
 
 #include <cmath>
 
+static constexpr bool dump_textures = false;
+
 namespace renderer::gl {
 
 static GLenum translate_depth_func(SceGxmDepthFunc depth_func) {
@@ -248,7 +250,7 @@ void sync_front_depth_bias(const GxmContextState &state) {
 }
 
 void sync_texture(GLContext &context, const GxmContextState &state, const MemState &mem, std::size_t index,
-    bool enable_texture_cache) {
+    bool enable_texture_cache, const std::string &base_path, const std::string &title_id) {
     const SceGxmTexture &texture = state.fragment_textures[index];
 
     if (texture.data_addr == 0) {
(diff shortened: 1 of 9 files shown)


vita3k/renderer/src/gl/texture.cpp
```diff
@@ -283,5 +285,67 @@ void upload_bound_texture(const SceGxmTexture &gxm_texture, const MemState &mem)
     }
 }
 
+// Dumps bound texture to a file
+void dump(const SceGxmTexture &gxm_texture, const MemState &mem, const std::string &parameter_name, const std::string &base_path, const std::string &title_id, Sha256Hash program_hash) {
+    static uint32_t gg_tex_index = 0;
```
VelocityRa, line 290: you meant `g_tex_index`?
`g_` is supposed to be for globals and below you use it for a local instead
not sure what "gg" is

sunho: Absolutely right. I fixed it.

vita3k/renderer/src/state_set.cpp
```diff
@@ -312,12 +312,15 @@ COMMAND_SET_STATE(stencil_ref) {
 COMMAND_SET_STATE(fragment_texture) {
     const std::uint32_t texture_index = helper.pop<std::uint32_t>();
     SceGxmTexture texture = helper.pop<SceGxmTexture>();
-
+    auto frag_program = state->fragment_program.get(mem);
+    auto program = frag_program->program.get(mem);
+    const Sha256Hash hash_bytes = sha256(program, program->size);
```
VelocityRa, line 317: isn't this is hashing every texture even if texture dumping isn't enabled?
there should be no additional hashing if it's disabled

vita3k/renderer/src/gl/sync_state.cpp
```diff
@@ -327,6 +344,8 @@ bool sync_state(GLContext &context, const GxmContextState &state, const MemState
 
     // Textures.
     const SceGxmFragmentProgram &gxm_fragment_program = *state.fragment_program.get(mem);
+    const auto program = gxm_fragment_program.program.get(mem);
+    const Sha256Hash hash_bytes = sha256(program, program->size);
```
VelocityRa, line 348: same here

vita3k/renderer/src/state_set.cpp
sunho: fixed!

vita3k/renderer/src/gl/sync_state.cpp
sunho: fixed!

PR #781, opened 2020-06-22
shader: implement kill instruction

sunho:
# About PR

- Implement kill instruction

# Related bug

It removes some glitches in lbp intro screen.


VelocityRa:
Is this ready for merging?

PR #784, opened 2020-06-27
Add save/restore scroll pos for themes list.

Zangetsu38:
- Add save and restore scroll pos for can resore it.

Very needed if you have very lot diferent theme in list for it not re go on 0 after set back.

### DIFF
--- a/vita3k/gui/src/themes.cpp
+++ b/vita3k/gui/src/themes.cpp
@@ -419,6 +419,8 @@ void draw_start_screen(GuiState &gui, HostState &host) {
 
 static std::string popup, menu, selected, title, start, delete_user_background;
 static ImGuiTextFilter search_bar;
+static float scroll_pos;
+static bool set_scroll_pos;
 
 void draw_themes_selection(GuiState &gui, HostState &host) {
     const auto display_size = ImGui::GetIO().DisplaySize;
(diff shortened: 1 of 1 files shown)


vita3k/gui/src/themes.cpp
```diff
@@ -419,6 +419,8 @@ void draw_start_screen(GuiState &gui, HostState &host) {
 
 static std::string popup, menu, selected, title, start, delete_user_background;
 static ImGuiTextFilter search_bar;
+static float scroll_pos;
+static bool set_scroll_pos;
```
1whatleytay, line 423: sad static

vita3k/gui/src/themes.cpp
```diff
@@ -490,8 +496,16 @@ void draw_themes_selection(GuiState &gui, HostState &host) {
...
     } else if (menu == "theme") {
+        // Theme List
         if (selected.empty()) {
             title = "Theme";
+
+            // Set Scroll Pos
+            if (set_scroll_pos) {
+                ImGui::SetScrollY(scroll_pos);
+                set_scroll_pos = false;
+            }
```
1whatleytay, line 507: Strange to me why you cant just do `ImGui::SetScrollY(scroll_pos);` on line 790.

Zangetsu38: is outside begin_child, you can't, if you set this here, you change position scroll inside begin :D

vita3k/gui/src/themes.cpp
Zangetsu38: static is perfect on general :)

PR #785, opened 2020-06-27
shader: fix spir-v errors

sunho:
# About PR

Related to #629 

Fixed some spir-v errors following through the validation errors.

Optimization in general fails if the validation fails. (internally they use same parsing function with validation) So, this can benefit the mentioned PR.


sunho:
I saw this outputs completely different glsl for many shaders. It might potentially fixes some game, but I think it would cause more regressions. I'm making this draft for this reason.

sunho:
Turns out it was some stupid coding mistake that caused regressions.

sunho:
👌

Issue #751, opened 2020-06-11
I can't open the Vita3k.exe


Boru69:
When i try to open the application, this message always appear
[12:51:04.229] |I| [config::init_config]: backend-renderer: OpenGL
[12:51:04.234] |I| [config::init_config]: log-level: 0
[12:51:04.243] |I| [SDL_main]: Vita3K v0.1 1453-d0a42162
[12:51:05.675] |E| [app::init]: Failed to init audio! Your computer probably does not have an audio device.
My audio devices are Realtek High Definition Audio ( speakers and microphone )
My laptop has windows 10, intel core i7 8750H NVDIA  GeForce GTX 1050 TI

SaribMadni:
This message appears when I try to run application
[20:42:11.325] |I| [config::init_config]: backend-renderer: OpenGL
[20:42:11.646] |I| [config::init_config]: log-level: 0
[20:42:16.222] |I| [SDL_main]: Vita3K v0.1 1447-6ae63c7f
[20:42:19.363] |I| [renderer::gl::create]: Swap interval = 0

PR #787, opened 2020-06-29
app: don't crash if it couldn't find audio device

sunho:
# About PR

Tries to skip no audio device error.

Audio callback will not be called if SDL couldn't open the audio properly. So, we can just warn the users and keep going.

fix https://github.com/Vita3K/Vita3K/issues/751


ichisadashioko:
Why don't we also get the error message from `SDL` to provide for information about why does it failed.

Mine was because I don't have an audio device.

image

sunho:
Wouldn't that say the same message every time.

sunho:
But, the more log the better. Updated the code.

Issue #792, opened 2020-06-30
Improve shader disassembler


sunho:
# About

Our disassembler has some room for improvement. It especially lacks testing.

# TODOs
- [ ] Organize disassembly format
- [ ] Separate disassembler from translator
- [x] Instruction parser update script
- [ ] Modifier/mask on many instructions
- [ ] Common repeat mode parse function
- [ ] Add automatic tests
- [ ] Add missing log_disasm
- [ ] Order the instructions as in the gxp

PR #794, opened 2020-07-01
Move usse-decoder-gen to Vita3k & improve it

sunho:
# About PR

- Move usse-decoder-gen to here
- autogen.py now directly update the actual source code
- Update grammar.yaml to reflect recent changes

# Motivation

It's generally good practice to manage auto-generated code by the tool that generated it. But, another reason is that it's very tiring to edit the bit-field of existing decoder. (after all, the motivation behind usse-decoder-gen was the difficulty of bit-field as I know)

# Words for reviewers

Notice that the function parameters don't change in cpp files (except nop), and the bit fields and the order of matchers don't change in entry.cpp in this PR.

### DIFF
--- a/tools/usse-decoder-gen/autogen.py
+++ b/tools/usse-decoder-gen/autogen.py
@@ -157,7 +157,7 @@ def update_visitor():
 
     # update sources
     def src_pat(name):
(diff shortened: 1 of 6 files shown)


sunho:
Related to https://github.com/Vita3K/Vita3K/issues/792

Issue #508, opened 2019-06-03
Fix SceAvPlayer/libscemp4 LLE


VelocityRa:
Used by _many_ commercial games.

Can use this hb for testing: http://vitadb.rinnegatamante.it/#/info/364

VelocityRa:
#512 improves this.

sunho:
SceAvPlayer uses libscemp4 under the hood.

sunho:
It looks like the games that only call scemp4 functions are statically linked with sceavplayer and using those static sceavplayer functions.

I think we have two options
1. LLE both scemp4 and sceavplayer
2. HLE scemp4 and LLE sceavplayer

We can't only HLE sceavplayer. Although HLE avplayer would work in many games, some games are statically linked with sceavplayer, so we need to make sure it works in LLE mode.

I'm first trying to go with option 1.

sunho:
LLE scemp4 was stuck at initialization. The init function spawns one worker thread and waits for worker thread to do the work. But, the worker thread somehow doesn't do any work. I'm investigating more.

VelocityRa:
The statically linked thing was already known btw. Should have mentioned it here. I guess I figured it out afterwards and forgot to.
You can ask next time, I don't mind getting pinged.

sunho:
More robust implementation of audioDecDecode is needed to move forward. Especially, AAC decoding part. Audio part of avplayer is just broken without this. And, apps that use audio part of avplayer in certain way are not working.

bookmist:
Finally implemented by this https://github.com/Vita3K/Vita3K/pull/1686

Zangetsu38:
Very old issue, 😅 should can close it

PR #802, opened 2020-07-07
Make lle mp4 and avplayer bootable

sunho:
# About Pr
- Fix some issues regarding full mspace
- Stub codecengine
- Stub avcdec memsize functions

# About codecengine

It's used to give work memory for codec library. But, codecengine apparently allocates it at separate hardware unit. So, if lle mp4 and avplayer turned out to use this worker memory directly, it can be quite troublesome. I just stubbed it for the moment, but if some of those lle functions are not working correctly, we might look at codecengine again.

# Further works

Video is now working in many games with this. But, it randomly got stuck or crash in the middle of video. This is maybe related to sync primitives which we need to investigate on. Also, audio is not working correctly because of incomplete audiodec. ngs pr (https://github.com/Vita3K/Vita3K/pull/657) improves the situation, but we also need to implement decoders for more codecs. (e.g. AAC)

### DIFF
--- a/vita3k/kernel/include/kernel/state.h
+++ b/vita3k/kernel/include/kernel/state.h
@@ -39,10 +39,13 @@ struct WatchMemory;
 
 struct InitialFiber;
 
+struct CodecEngineBlock;
+
 typedef std::vector<InitialFiber> InitialFibers;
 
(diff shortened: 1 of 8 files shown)


sunho:
Related to https://github.com/Vita3K/Vita3K/issues/508

Issue #809, opened 2020-07-12
Callback functions


sunho:
Unicorn apparently crashes if we stop(cpu) & emu_start(cpu, until) in interrupt handler and give control back to unicorn. This complicates implementation of callback functions. (for example, npservicestatecallback or netctlcallback)

sunho:
This is related to run_on_current function.

sunho:
We need to fix a bug in unicorn-engine to cleanly solve this issue. More specifically, if we stop the unicorn inside the interrupt handler, unicorn crashes.

sunho:
https://github.com/Vita3K/Vita3K/pull/845 This one attempts to solve this problem by creating another cpu instance if the current instance is running. This is the best idea I can think out. If this turns out to be a bottleneck, we can briefly cache the instances or pull them to another "worker" thread. For thread way, we might parallelize the execution of callback function. (if callback functions are 5, we can execute them over 5 threads)

PR #810, opened 2020-07-12
Fix cpu state corruption when calling guest callback function

sunho:
# About PR

- Fix cpu state corruption when calling guest callback function

# Context saving

Our previous context saving was not enough for unicorn to continue the emulation normally. It turned out unicorn relied on some internal states that could've changed as it runs the code. (which caused the crash after restoring context) Fortunately, unicorn provides its own context saving api that also deal with internal states. I used it in this PR.

# Some words

run_on_current still just continues the main emulation without going back to the caller. Notice that run_thread(thread, false) will not stop until the current thread dies. So, we can't use this function to use random guest function and multiple netctl callbacks are not working. (after processing one callback it immediately returns to the guest)

I'm searching for clean way to solve this problem.

### DIFF
--- a/vita3k/cpu/include/cpu/functions.h
+++ b/vita3k/cpu/include/cpu/functions.h
@@ -26,8 +26,10 @@
 
 struct CPUState;
 struct MemState;
+struct uc_context;
 
 struct CPUContext {
+    uc_context *context;
     uint32_t cpu_registers[16];
     uint32_t sp;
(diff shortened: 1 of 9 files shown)


sunho:
Related to https://github.com/Vita3K/Vita3K/issues/809

sunho:
I'll redo this.

PR #818, opened 2020-07-22
Fix kiil and vtest instruction + v_Position

sunho:
# About PR
- Implement int32 subs in vtest
- Correct register type in vtest (no double regs)
- Match more kill instructions
- Map gl_fragcoord to v_Position in fragment input 

# v_Position

This is equivalent to WPOS. So, we can just give gl_fragcoord to it.

# Related issues

With updated vtest & v_Position, vitagl texture sample is now working correctly.

Updated v_Position fixes a compile error of one fragment shader in LBP.


sunho:
WPOS_centroid. I’ll look into it.

PR #826, opened 2020-07-26
Fix uchar sampler and sop2

sunho:
# About PR
- Fix some errors in sop2 implementation
- Correctly handle uchar sampler
- Stub uint8 pack & unpack

# Related issue

Half-fixes the yellow screen in xeodrifter
Fixes usampler test shader

# Uint8

I tried unpacking 4-bytes into 4 1-byte uint and static cast to float. (so that 255 becomes 255.0) The unpacked value was used as an operand of sop2. However, we assumed operands of sop2 to be in c10. 

I need to investigate more on sop2. In the meantime, I'm just stubbing uint8 packer & unpacker with c10 implementation to prevent shader crash and see more.

### DIFF
--- a/vita3k/shader/src/usse_utilities.cpp
+++ b/vita3k/shader/src/usse_utilities.cpp
@@ -217,6 +217,7 @@ static spv::Function *make_f16_pack_func(spv::Builder &b, const FeatureState &fe
 }
 
 spv::Id shader::usse::utils::unpack_one(spv::Builder &b, SpirvUtilFunctions &utils, const FeatureState &features, spv::Id scalar, const DataType type) {
+    // TODO: doing this for uint8 is probably not right
     switch (type) {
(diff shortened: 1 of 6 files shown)


vita3k/shader/src/usse_utilities.cpp
```diff
@@ -732,13 +736,14 @@ void shader::usse::utils::store(spv::Builder &b, const SpirvShaderParameters &pa
...
-    case DataType::INT8:
-    case DataType::UINT8: {
+    case DataType::UINT8:
+    case DataType::INT8: {
         dest.type = DataType::C10;
         break;
     }
 
     case DataType::F32:
+
```
Zangetsu38, line 746: Space needed here ?

sunho: :O

PR #827, opened 2020-07-27
gxm: Add RET_ERRORS log on nearly every function

EXtremeExploit:


vita3k/modules/SceGxm/SceGxm.cpp
```diff
@@ -208,6 +208,24 @@ EXPORT(int, sceGxmColorSurfaceInit, SceGxmColorSurface *surface, SceGxmColorForm
...
+    if (outputRegisterSize != SCE_GXM_OUTPUT_REGISTER_SIZE_32BIT)
+        RET_ERROR(SCE_GXM_ERROR_INVALID_ALIGNMENT);
+
+    if (width <= 0)
+        RET_ERROR(SCE_GXM_ERROR_INVALID_ALIGNMENT);
+
+    if (height <= 0)
+        RET_ERROR(SCE_GXM_ERROR_INVALID_ALIGNMENT);
+
+    if (strideInPixels <= 0)
```
Zangetsu38, line 223: ugly lol,  using "||"  no create news value for return same value

vita3k/modules/SceGxm/SceGxm.cpp
```diff
@@ -261,6 +287,12 @@ EXPORT(int, sceGxmCreateContext, const SceGxmContextParams *params, Ptr<SceGxmCo
     assert(params != nullptr);
     assert(context != nullptr);
 
+    if (!params)
+         RET_ERROR(SCE_GXM_ERROR_INVALID_POINTER);
+
+    if (!context)
+         RET_ERROR(SCE_GXM_ERROR_INVALID_POINTER);
```
Zangetsu38, line 294: same

sunho:
Thanks! Please run format.bat to pass the ci. And, can you do like ```#define LOG_ERROR_ONCE RET_ERROR``` and use it instead? So that we can only look at "LOG_ERROR_ONCE" to correctly handle them later.

PR #828, opened 2020-07-28
Fix uint8 in shader

sunho:
# About PR
- Fix trivial errors in sop2 implementation
- Implement pack & unpack of uint8
- Handle uint8 (or uchar) sampler
- Handle uint8 output, input and colorattachment
- Update SOP2 to receives and returns uint8 values
- Update VPCK to support uint8

# SOP2

Analyzing shaders using SOP2, I observed four behaviors about sop2.

1. It can use uchar value coming from uniform parameter directly as its sources.
2. The output from one sop2 can be also used directly as sources of another sop2.
3. It can do y = ```(1-x)*color1 + x*color2```. And 1 was immediate value.
4. In 3, x can be 0xff in uint8 bytes.

From 1 and 2, I thought that sop2 actually takes uint8 and returns uint8. Before, we were inputting and outputting fp8 values.

From 3 and 4, I thought that sop2 is actually normalizing the sources internally. I saw so many evidences that uint8 is statically converted to float value (so that 255 -> 255.0 and vice versa) So, if sop2 doesn't normalize sources, in 4 (1-x) would become signed -254, which is really weird if we think that inputs are unsigned values.

# Related issue
- Fixes the yellow screen in xeodrifter with one upcoming PR
- Fixes usampler test shader
- No regression on P4G TV that motivated using fp8 for input and output of sop2 before

### DIFF
--- a/vita3k/modules/SceGxm/SceGxm.cpp
+++ b/vita3k/modules/SceGxm/SceGxm.cpp
@@ -2382,7 +2382,7 @@ EXPORT(int, sceGxmTextureSetStride, SceGxmTexture *texture, uint32_t byteStride)
 
 EXPORT(int, sceGxmTextureSetUAddrMode, SceGxmTexture *texture, SceGxmTextureAddrMode mode) {
     assert(texture);
-    if (texture)
+    if (!texture)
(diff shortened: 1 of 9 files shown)


sunho:
Do not merge yet. Confirming something.

PR #830, opened 2020-07-28
renderer: add ability to log active default uniform buffers

sunho:
It's useful for debugging shaders. Renderdoc does some dumping, but it's in scientific notation. I found hex more handy.

### DIFF
--- a/vita3k/renderer/src/gl/draw.cpp
+++ b/vita3k/renderer/src/gl/draw.cpp
@@ -7,6 +7,8 @@
 #include <renderer/state.h>
 #include <renderer/types.h>
 
+#include <sstream>
+
 #include <gxm/types.h>
 #include <util/log.h>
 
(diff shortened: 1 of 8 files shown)


vita3k/renderer/src/gl/draw.cpp
```diff
@@ -74,6 +76,14 @@ void draw(GLState &renderer, GLContext &context, GxmContextState &state, const F
         const std::string hash_text_v = hex_string(state.vertex_program.get(mem)->renderer_data->hash);
 
         LOG_DEBUG("\nVertex  : {}\nFragment: {}", hash_text_v, hash_text_f);
+
+        std::stringstream vert_ub("");
```
VelocityRa, line 80:
```suggestion
        std::stringstream vert_ub;
```
it's already default-constructed

vita3k/renderer/src/gl/draw.cpp
```diff
@@ -74,6 +76,14 @@ void draw(GLState &renderer, GLContext &context, GxmContextState &state, const F
...
+        dump_hex(context.ubo_data[0], vert_ub);
+        LOG_DEBUG("Vertex default uniform buffer: \n{}", vert_ub.str());
+
+        std::stringstream frag_ub("");
```
VelocityRa, line 84:
```suggestion
        std::stringstream frag_ub;
```
ditto

sunho:
I'll fix it tonight. Thanks for review!

PR #831, opened 2020-07-30
Shader uint register storage

sunho:
I thought this one is needed for integer uniform buffer. Because some encoding of ints are NaN in float that might mess up things. (https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/intBitsToFloat.xhtml) It turned out this is not needed at all for my gpus (nvidia and amd). I'm just putting this here, hoping it might mitigate some issues later.


sunho:
Probably not needed!

PR #835, opened 2020-08-01
Stub broken lle pvf functions

sunho:
# About PR

Some part of LLE pvf is broken and it's causing regressions on some games. This stubs those part of pvf by using working lle pvf function openuserfile.

# Related issue
- Fixes loading error of all atelier games I checked (total 5)

# Hack

This is a hack. One sceft2 function (nid 0x20CBDA2D) called by pvf is hanging, we should look into this to properly solve this issue.


sunho:
I'll look into this more

PR #851, opened 2020-08-14
io/SceAppUtil: fix saving

IllusionMan1212:
- undertale writes its save file in truncate mode, so that's been fixed
- some mages games consider a save file to be corrupt if param-status isn't 0
- SCE_O_WRONLY shouldn't create files, only SCE_O_CREAT can


sunho:
Yes!!! Please run format then perfect.

vita3k/modules/SceAppUtil/SceAppUtil.cpp
```diff
@@ -234,6 +234,12 @@ EXPORT(int, sceAppUtilSaveDataDataSave, SceAppUtilSaveDataFileSlot *slot, SceApp
             create_dir(host.io, file_path.c_str(), 0777, host.pref_path, export_name);
             break;
         case SCE_APPUTIL_SAVEDATA_DATA_SAVE_MODE_FILE_TRUNCATE:
+            if (files[i].buf.get(host.mem) != nullptr) {
```
sunho, line 237: You can just if(files[i].buf) here

IllusionMan1212: done

PR #855, opened 2020-08-15
interace: make game installation work even if game is not in the root zip

EXtremeExploit:


vita3k/interface.cpp
```diff
@@ -94,27 +94,40 @@ bool install_archive(HostState &host, GuiState *gui, const fs::path &path) {
     }
 
     int num_files = mz_zip_reader_get_num_files(zip.get());
-    fs::path sfo_path = "sce_sys/param.sfo";
```
Zangetsu38, line 97: You can keep it

vita3k/interface.cpp
```diff
@@ -96,25 +96,39 @@ bool install_archive(HostState &host, GuiState *gui, const fs::path &path) {
     int num_files = mz_zip_reader_get_num_files(zip.get());
     fs::path sfo_path = "sce_sys/param.sfo";
     bool theme = false;
+    std::string extra_path = "";
```
sunho, line 99: std::string extra_path; will work too :) just fyi

vita3k/interface.cpp
```diff
@@ -96,25 +96,39 @@ bool install_archive(HostState &host, GuiState *gui, const fs::path &path) {
...
 
     for (int i = 0; i < num_files; i++) {
         mz_zip_archive_file_stat file_stat;
-        if (!mz_zip_reader_file_stat(zip.get(), i, &file_stat)) {
+        if (!mz_zip_reader_file_stat(zip.get(), i, &file_stat))
             continue;
-        }
-        if (fs::path(file_stat.m_filename) == "sce_module/steroid.suprx") {
+
+        if (fs::path(file_stat.m_filename).string().find("sce_module/steroid.suprx") != std::string::npos) {
```
sunho, line 106: can you just make fs::path(file_stat.m_filename).string() variable and use it below?

vita3k/interface.cpp
```diff
@@ -96,25 +96,39 @@ bool install_archive(HostState &host, GuiState *gui, const fs::path &path) {
...
             LOG_CRITICAL("A Vitamin dump was detected, aborting installation...");
             fclose(vpk_fp);
             return false;
         }
-        if (fs::path(file_stat.m_filename) == "theme.xml")
+
+        if (fs::path(file_stat.m_filename).string().find("theme.xml") != std::string::npos)
```
sunho, line 112: here would be then if (variable.find("theme.xml") != std::string::npos)

PR #861, opened 2020-08-17
Implement unpack_one & pack_one of int and uint values

sunho:
This implements full packing and unpacking of uint and int values. It can still pack and unpack the types supported before (uint8 and half)

### DIFF
--- a/vita3k/shader/include/shader/usse_utilities.h
+++ b/vita3k/shader/include/shader/usse_utilities.h
@@ -12,12 +12,10 @@ struct FeatureState;
 namespace shader::usse::utils {
 
 struct SpirvUtilFunctions {
-    spv::Function *pack_f16{ nullptr };
-    spv::Function *unpack_f16{ nullptr };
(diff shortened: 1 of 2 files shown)


vita3k/shader/include/shader/usse_utilities.h
```diff
@@ -12,12 +12,10 @@ struct FeatureState;
 namespace shader::usse::utils {
 
 struct SpirvUtilFunctions {
-    spv::Function *pack_f16{ nullptr };
-    spv::Function *unpack_f16{ nullptr };
+    std::map<DataType, spv::Function *> unpack_funcs;
+    std::map<DataType, spv::Function *> pack_funcs;
     spv::Function *pack_fx8{ nullptr };
```
frangarcj, line 17: fx8 is not similar?

sunho: I'm not sure about this type yet. I'd like to leave it there for now.

PR #868, opened 2020-08-19
GUI: Some refactor

Zangetsu38:
- gui/main menubar: fix set value.
- gui/app sellector: Add information bar.
- gui: reworks app selector and live area switch game.

# Result:
- Live area app with arrow left/right for switch app opened
image
 
- App Selector with arrow in left for open it on live area 
image

- App running go live area and left for show app selector
(for can go on system app for now, can launch another app in futur like psvita)
image


sunho (approved):
Good refactor 👍

Issue #880, opened 2020-08-22
Failed to build on Ubuntu


Sunderland93:
Hello. I'm trying to build Vita3K on Ubuntu 20.04, but both Clang and GCC gives an error:

```
/build/vita3k-0.1~git20200821/vita3k/gui/src/pkg_install_dialog.cpp:59:50: error: format string is not a string literal (potentially insecure) [-Werror,-Wformat-security]
        ImGui::TextColored(GUI_COLOR_TEXT_TITLE, title.c_str());
                                                 ^~~~~~~~~~~~~
/build/vita3k-0.1~git20200821/vita3k/gui/src/pkg_install_dialog.cpp:59:50: note: treat the string as an argument to avoid this
        ImGui::TextColored(GUI_COLOR_TEXT_TITLE, title.c_str());
                                                 ^
                                                 "%s", 
1 warning and 1 error generated.
make[3]: *** [vita3k/gui/CMakeFiles/gui.dir/build.make:274: vita3k/gui/CMakeFiles/gui.dir/src/pkg_install_dialog.cpp.o] Error 1
make[3]: Leaving directory '/build/vita3k-0.1~git20200821/obj-x86_64-linux-gnu'
make[2]: *** [CMakeFiles/Makefile2:2269: vita3k/gui/CMakeFiles/gui.dir/all] Error 2
make[2]: Leaving directory '/build/vita3k-0.1~git20200821/obj-x86_64-linux-gnu'
make[1]: *** [Makefile:144: all] Error 2
make[1]: Leaving directory '/build/vita3k-0.1~git20200821/obj-x86_64-linux-gnu'
dh_auto_build: error: cd obj-x86_64-linux-gnu && make -j1 returned exit code 2
make: *** [debian/rules:10: build] Error 2
dpkg-buildpackage: error: debian/rules build subprocess returned exit status 2
```

PR #883, opened 2020-08-23
gui: fix a compilation error and minor misc changes

scribam:
Fix the second compilation error reported in #880


IllusionMan1212 (approved):

sunho:
@scribam Do you why our CI failed to detect these compilation errors?

scribam:
@sunho IIRC these errors are triggered by clang-10 which is not set up on our CI.

sunho:
I see. We have to be more careful about this error.

PR #886, opened 2020-08-24
Refactor GXP input parsing

sunho:
# About PR

- Separate gxp input parsing part from spirv recompiler
- Fix non-ubo mode uniform bug
- Refactor various utility functions
- Refactor create_parameters of spirv_recompiler.cpp

# Motivation

GXP input parsing part can be reused for many purposes (fuzzing/debugging/optimizing/etc) But, we had it inside private recompiler function that also deals with spir-v setting up, hiding useful parsed information.

With this PR, now we can get almost all information about input data of gxp shader program using code like below. 
```cpp
const auto prgoram_input = shader::get_program_input(program);
```

### DIFF
--- a/vita3k/shader/CMakeLists.txt
+++ b/vita3k/shader/CMakeLists.txt
@@ -13,6 +13,7 @@ add_library(
 	include/shader/usse_translator_entry.h
 	include/shader/usse_translator_types.h
 	include/shader/usse_utilities.h
+	include/shader/gxp_parser.h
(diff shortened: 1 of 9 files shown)


Zangetsu38:
can probably squash and remove last commit for set in inside other ^^

sunho:
Alright, I'll do that

frangarcj (approved):

VelocityRa (approved):

PR #942, opened 2020-08-29
Cast non float in sceGxmSetUniformDataF

sunho:
We should cast to non floats in this function. 

Look how VitaGL does: https://github.com/Rinnegatamante/vitaGL/blob/0d2110ac7b001e1e70688c6450a263b361a2b337/source/vitaGL.c#L1366


scribam (approved):
These changes fix Flood-It! which was broken by #886

sunho:
@scribam Can you check if https://github.com/Vita3K/Vita3K/pull/941 fixed the issue in Flood-it? It's surprising to hear this pr fixed regression caused by https://github.com/Vita3K/Vita3K/pull/886.

scribam:
#941 does not fix the issue:

Capture

sunho:
:eyes:

frangarcj (approved):

scribam:
For some strange reasons, Flood It! is still broken after this merge so my first comment must be wrong.

PR #944, opened 2020-08-30
Correct the size and order of vertex outputs & fix some bugs in loading and storing them

sunho:
This improves p4g rendering a lot.

I'm not sure whether this size is actually right. I'm going to confirm tomorrow.


sunho:
Confirmed 2 is the actual size. I also realized that v_Fog should be written after v_Color.

sunho:
Ready for review.

frangarcj (approved):

PR #951, opened 2020-09-04
Rework uniform buffers

sunho:
# About PR
- Don't analyze shader instructions to find out the size of uniform buffers
- Deal with half register half memory uniform buffer

# Related issue

Fixes p4g sunny character rendering.

# Description 

We used to analyze the program to determine the size of memory uniform buffer. However, this method will require tons of works when dealing with ldr with temporary register for base field. (this can happen)

From my observation, uniform parameters of memory uniform buffer are not stripped away. Utilizing this, I tried to determine uniform buffer size from uniform parameters. I also observed half register half memory uniform buffer. (this seems to happen in large default uniform buffer) In this case the size_in_f32 is less than the total size of uniform buffer, and the ldr instruction will be used to access fields not available in register. Also, the base of ldr instruction should start from the end of "register block " of uniform buffer. I updated recompiler to consider this information.

### DIFF
--- a/vita3k/shader/include/shader/gxp_parser.h
+++ b/vita3k/shader/include/shader/gxp_parser.h
@@ -8,6 +8,5 @@ namespace shader {
 
 usse::GenericType translate_generic_type(const gxp::GenericParameterType &type);
 std::tuple<usse::DataType, std::string> get_parameter_type_store_and_name(const SceGxmParameterType &type);
-std::vector<usse::UniformBuffer> get_uniform_buffers(const SceGxmProgram &program);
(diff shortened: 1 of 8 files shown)


sunho:
THEY ARE STRIPPING PARAMETERS.

sunho:
It turned out size info is inside array_size of uniform buffer parameter. Regressions in danganronpa2 and is it wrong to are fixed. This is ready for review.

VelocityRa (approved):
Found the size after all! Nice work.

pent0:
😀😀😀

PR #590, opened 2019-10-12
[WIP] Hide HostState members to improve compilation time

petmac:

PR #952, opened 2020-09-04
Hide HostState members to improve compilation time (rebased)

scribam:
PR #590 rebased-ish


sunho:
Amazing

Zangetsu38:
Mac os 41 min ? 🤔 compare my other pr
image

petmac:
I suspect that's due to ccache.

sunho:
@scribam Can you resolve conflicts? We can see the CI build time again with updated cache.

sunho:
Reading the logs of CI, I can see that ccache is from the master branch (https://github.com/Vita3K/Vita3K/commit/407ea1296ed771202797a10afb42127d42669a50)

sunho:
With ccache, the speed's almost same. (mac build is currently slower even in other pr)
image

sunho (approved):
I think we should merge this soon. This pr's so big, so we'd need to rebase again if we don't merge.

VelocityRa (commented):
No no no.

Why force indirection for every access in all these _extremely_ frequently accessed members?
Cache coherency should be taken into account in decisions that affect so much.

I _very_ much disagree with this, even if it decreases compile times slightly.

sunho:
I was leaning hiding state way because it can achieve encapsulation and improve build time. Hearing velo's perspective about cache, I'm now more in favor of keeping with current way.

scribam:
So let's close this then.

PR #961, opened 2020-09-08
Limit Framerate when selecting applications

Ristellise:
Small fix to prevent high CPU usage when selecting applications. Turns out we are running too fast.

Current limit is 60fps, though feel free to tweak.

After tweak: 
image  
Current Master:  
image

### DIFF
--- a/vita3k/main.cpp
+++ b/vita3k/main.cpp
@@ -136,8 +136,24 @@ int main(int argc, char *argv[]) {
         auto discord_rich_presence_old = host.cfg.discord_rich_presence;
 #endif
 
+        std::chrono::system_clock::time_point present = std::chrono::system_clock::now();
+        std::chrono::system_clock::time_point later = std::chrono::system_clock::now();
+        const double frame_time = 1000.0 / 60.0;
         // Application not provided via argument, show game selector
         while (run_type == app::AppRunType::Unknown) {
+            // get the current time & get the time we worked for
+            present = std::chrono::system_clock::now();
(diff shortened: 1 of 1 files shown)


sunho:
Cool!!

vita3k/main.cpp
```diff
@@ -136,8 +136,24 @@ int main(int argc, char *argv[]) {
         auto discord_rich_presence_old = host.cfg.discord_rich_presence;
 #endif
 
+        std::chrono::system_clock::time_point present = std::chrono::system_clock::now();
+        std::chrono::system_clock::time_point later = std::chrono::system_clock::now();
+        const double frame_time = 16.67;
```
sunho, line 141: const double frame_time = 1000.0 / 60.0;

You make 60.0 a constant if you want.

Ristellise: Fixed

sunho (approved):

frangarcj:
It's not possible to use sdl sync like the renderer does?

Ristellise:
> It's not possible to use sdl sync like the renderer does?

Not sure... I havent used SDL when I was programming opengl last year, but used chrono instead.

frangarcj:
https://github.com/Vita3K/Vita3K/blob/f08ec1e667dd12a5293ccb060fe59f8690ff6a28/vita3k/renderer/src/gl/renderer.cpp#L222-L224

Ristellise:
@frangarcj Doesn't work.
image

Issue #964, opened 2020-09-09
Memory leak in renderer


sunho:
The currently reported increasing ram is surprisingly casued by one or two leaks. 

This can be fixed by using delete[] in https://github.com/Vita3K/Vita3K/blob/master/vita3k/renderer/src/state_set.cpp#L109 and adding delete[] stream_data; in https://github.com/Vita3K/Vita3K/blob/master/vita3k/renderer/src/state_set.cpp#L374.

But, these changes somehow causes crash in release build. (check out p4g)

ichisadashioko:
I will try to look into this bug.

PR #966, opened 2020-09-09
Fix memory leak in renderer 

KorewaWatchful:
Fixes issue #964, thanks to @sunho and @Ristellise for the solution.

### DIFF
--- a/vita3k/renderer/src/state_set.cpp
+++ b/vita3k/renderer/src/state_set.cpp
@@ -106,7 +106,7 @@ COMMAND_SET_STATE(uniform_buffer) {
     switch (renderer.current_backend) {
     case Backend::OpenGL: {
         gl::set_uniform_buffer(*reinterpret_cast<gl::GLContext *>(render_context), is_vertex, block_num, size, data, config.log_active_shaders);
-        delete[size] data;
+        delete[] data;
 
         break;
     }
--- a/vita3k/renderer/src/state_set.cpp
+++ b/vita3k/renderer/src/state_set.cpp
@@ -371,7 +371,7 @@ COMMAND_SET_STATE(vertex_stream) {
     case Backend::OpenGL: {
         gl::upload_vertex_stream(*reinterpret_cast<gl::GLContext *>(render_context), stream_index, stream_data_length,
             stream_data);
-        delete[stream_data_length] stream_data;
+        delete[] stream_data;
         break;
     }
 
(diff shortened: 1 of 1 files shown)


frangarcj (approved):

vita3k/renderer/src/state_set.cpp
```diff
@@ -106,7 +106,7 @@ COMMAND_SET_STATE(uniform_buffer) {
     switch (renderer.current_backend) {
     case Backend::OpenGL: {
         gl::set_uniform_buffer(*reinterpret_cast<gl::GLContext *>(render_context), is_vertex, block_num, size, data, config.log_active_shaders);
-        delete data;
+        delete[size] data;
```
VelocityRa, line 109: This is not valid C++.
MSVC probably accepted it because MSVC is dumb and accepts all sorts of non-compliant code.
```suggestion
        delete[] data;
```

vita3k/renderer/src/state_set.cpp
```diff
@@ -371,7 +371,7 @@ COMMAND_SET_STATE(vertex_stream) {
     case Backend::OpenGL: {
         gl::upload_vertex_stream(*reinterpret_cast<gl::GLContext *>(render_context), stream_index, stream_data_length,
             stream_data);
-
+        delete[stream_data_length] stream_data;
```
VelocityRa, line 374: ditto

VelocityRa (changes requested):

vita3k/renderer/src/state_set.cpp
frangarcj:
```suggestion
        delete[] stream_data;
```

sunho:
This is not ready to merge fyi. standard delete[] somehow causes crashes in release build. (not in debug build) Really weird bug.

frangarcj:
I saw your conversation and I thought the size between [] fixed it.

Then I saw @VelocityRa and noticed that the CI was not building for mac or linux.

KorewaWatchful:
Uhm, so I just downloaded the artifact on windows and we're not getting a crash anymore, went in-game, walked the dream dungeon, etc. Ram usage staying at .5GB after ~10 mins of playing. 
Screenshot 2020-09-11 091439

KorewaWatchful:
I'm gonna test on linux in an hour, but it might've just fixed itself (worst kind of fixed).

sunho:
Solved by https://github.com/Vita3K/Vita3K/pull/989 without erratic crashes in release build.

PR #969, opened 2020-09-13
installation progress bar

IllusionMan1212:
run game and firmware installation on a separate thread to avoid locking up the emulator and keep it responsive

image

### DIFF
--- a/vita3k/gui/src/archive_install_dialog.cpp
+++ b/vita3k/gui/src/archive_install_dialog.cpp
@@ -34,11 +34,16 @@ static nfdchar_t *archive_path;
 void draw_archive_install_dialog(GuiState &gui, HostState &host) {
     nfdresult_t result = NFD_CANCEL;
 
+    static std::mutex install_mutex;
     static bool draw_file_dialog = true;
     static bool content_install_confirm = false;
     static bool finished_installing = false;
-    static float progress;
-    static bool reinstalling;
+    static std::atomic<float> progress = 0;
+    static bool reinstalling = false;
+    static const auto progress_callback = [&](float updated_progress) {
+        progress = updated_progress;
+    };
+    std::lock_guard<std::mutex> lock(install_mutex);
 
     if (draw_file_dialog) {
         result = NFD_OpenDialog("vpk,zip", nullptr, &archive_path);
@@ -47,11 +52,16 @@ void draw_archive_install_dialog(GuiState &gui, HostState &host) {
 
         if (result == NFD_OKAY) {
             std::thread installation([&host, &gui]() {
-                if (install_archive(host, &gui, fs::path(string_utils::utf_to_wide(archive_path)), &progress))
+                if (install_archive(host, &gui, fs::path(string_utils::utf_to_wide(archive_path)), progress_callback)) {
+                    std::lock_guard<std::mutex> lock(install_mutex);
                     content_install_confirm = true;
-                else if (!gui.content_reinstall_confirm)
+                } else if (!gui.content_reinstall_confirm) {
                     ImGui::OpenPopup("Content installation failed");
-                finished_installing = true;
+                }
+                {
+                    std::lock_guard<std::mutex> lock(install_mutex);
+                    finished_installing = true;
+                }
             });
             installation.detach();
         } else {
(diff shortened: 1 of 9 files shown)


Ristellise:
I thinking of doing it a while back lol. Was wondering if there is a standardised way to do threads in vita3k. Anyways, good job :)

Zangetsu38:
Omg finnaly, good works 👀

vita3k/gui/src/pkg_install_dialog.cpp
```diff
@@ -102,12 +105,16 @@ void draw_pkg_install_dialog(GuiState &gui, HostState &host) {
             if (ImGui::Button("Ok", BUTTON_SIZE) && !zRIF.empty())
                 state = "install";
         } else if (state == "install") {
-            if (install_pkg(std::string(pkg_path), host, zRIF))
-                state = "succes";
-            else
-                state = "fail";
-            zRIF.clear();
-        } else if (state == "succes") {
+            std::thread installation(([&host]() {
```
Zangetsu38, line 108: why here have double "("  after installation ? other have only one 🤔

IllusionMan1212: Lol I must have just typed it twice

vita3k/host/src/pkg.cpp
```diff
@@ -252,7 +254,10 @@ bool install_pkg(const std::string &pkg, HostState &host, std::string &p_zRIF) {
             LOG_ERROR("The pkg file size is too small, possibly corrupted");
             return false;
         }
-
+        float file_count = byte_swap(pkg_header.file_count);
+        if (i % static_cast<int>(round(file_count / 100.f)) == 0) {
+            *progress = *progress + 0.6;
```
Zangetsu38, line 259: Can write more short
```suggestion
            *progress += 0.6;
```

sunho:
We probably need to do this more properly if we want to support cancelling installation. (we shouldn't detach) But, for now, it looks fine generally.

vita3k/gui/src/archive_install_dialog.cpp
```diff
@@ -34,25 +36,59 @@ void draw_archive_install_dialog(GuiState &gui, HostState &host) {
...
-            if (install_archive(host, &gui, fs::path(string_utils::utf_to_wide(archive_path))))
-                content_install_confirm = true;
-            else if (!gui.content_reinstall_confirm)
-                ImGui::OpenPopup("Content installation failed");
+            std::thread installation([&host, &gui]() {
+                if (install_archive(host, &gui, fs::path(string_utils::utf_to_wide(archive_path)), &progress))
+                    content_install_confirm = true;
+                else if (!gui.content_reinstall_confirm)
+                    ImGui::OpenPopup("Content installation failed");
+                finished_installing = true;
```
sunho, line 54: It's little dirty to use flag (finished_installing) here (because it has race condition) Try to use condition_variable.

vita3k/gui/src/archive_install_dialog.cpp
```diff
@@ -34,25 +36,59 @@ void draw_archive_install_dialog(GuiState &gui, HostState &host) {
...
         } else {
             gui.file_menu.archive_install_dialog = false;
             draw_file_dialog = true;
         }
     }
 
+    if (reinstalling) {
+        finished_installing = false;
+        std::thread reinstallation([&host, &gui]() {
+            if (install_archive(host, &gui, fs::path(string_utils::utf_to_wide(archive_path)), &progress)) {
```
sunho, line 66: There's a big race condition on progress. You should use mutex.

vita3k/host/include/host/functions.h
```diff
@@ -22,7 +22,7 @@
 
 struct SfoFile;
 
-void install_pup(const std::string &pref_path, const std::string &pup_path);
+void install_pup(const std::string &pref_path, const std::string &pup_path, uint32_t *progress = 0);
```
sunho, line 25: I think it's more clean to have a callback function that reports the progress so that the caller can manage the race condition things, instead of requiring pointer.

vita3k/host/src/pkg.cpp
sunho: Maybe we can just split this to multiple progresses (extract_progress, copy_progress) And do weighted average one it. (progress =  0.2 * extract_progress + 0.8 * copy_progress)

IllusionMan1212:
putting this on hold until i can properly implement it

Zangetsu38:
you need rebase andf fix conflict here already

Zangetsu38:
need rebase and fix conflict

sunho (approved):
Good works

PR #970, opened 2020-09-14
gui: support for more than 1 folder inside vpk root

EXtremeExploit:
Fixes:
* Drag and drop for vpk/zip files
* Some zip/vpk had more than 1 folder depth ("game.zip/foo/bar/eboot.bin") where the game is at, this has now been fixed too

### DIFF
--- a/vita3k/interface.cpp
+++ b/vita3k/interface.cpp
@@ -201,6 +201,9 @@ bool install_archive(HostState &host, GuiState *gui, const fs::path &archive_pat
         }
 
         std::string m_filename = file_stat.m_filename;
+        if (extra_path.find(m_filename) != std::string::npos) {
+            continue;
+        }
         std::string replace_filename = m_filename.substr(extra_path.size());
         const fs::path file_output = { output_path / replace_filename };
         if (mz_zip_reader_is_file_a_directory(zip.get(), i)) {


sunho (approved):

vita3k/interface.cpp
```diff
@@ -267,7 +270,7 @@ static ExitCode load_app_impl(Ptr<const void> &entry_point, HostState &host, con
         return InvalidApplicationPath;
 
     vfs::FileBuffer params;
-    bool params_found = vfs::read_app_file(params, host.pref_path, host.io.title_id, "sce_sys/param.sfo");
+    bool params_found = vfs::read_app_file(params, host.pref_path, host.app_title_id, "sce_sys/param.sfo");
```
Zangetsu38, line 273: Fail lol d'ont touch this broken all

Zangetsu38, line 273: Remove this, get broken load app

Zangetsu38 (changes requested):

vita3k/interface.cpp
sunho: You can see that people are actually confused by this.

sunho: We should really move this to gui state.

Zangetsu38: No relate, is all time writed like that is not news get by my code

sunho:
Does  this work if you use host.io.title_id,?

EXtremeExploit:
nope

sunho:
This issue is also related to broken cli vpk execution and console mode. The issue happens because noone fills host.io.title_id. We need to just refactor this part of code. It was confusing before (a big side effect) and became more confusing by https://github.com/Vita3K/Vita3K/pull/856 (side effect changing the public state that is only used privately).

EXtremeExploit:
Remade fix drag and drop, should uso I/O now

vita3k/main.cpp
```diff
@@ -186,6 +186,7 @@ int main(int argc, char *argv[]) {
     }
 
     Ptr<const void> entry_point;
+    host.io.title_id = host.app_title_id;
```
Zangetsu38, line 189: delete this, broken boot for some case.
look code before, also, io.title_id is already set

Zangetsu38 (changes requested):

vita3k/main.cpp
EXtremeExploit: done

Zangetsu38: thx, sorry for this, i try fix myself this issue

Zangetsu38:
oàuc, you need reébase correctly here

PR #975, opened 2020-09-20
gui: workaround memory leak when opening settings

EXtremeExploit:
Adds comparasion overloads for config, we compare the config to a temporary copy, if they are different then save it that way we don't spam save config

### DIFF
--- a/vita3k/app/src/app_init.cpp
+++ b/vita3k/app/src/app_init.cpp
@@ -105,6 +105,7 @@ bool init(HostState &state, Config &cfg, const Root &root_paths) {
     };
 
     state.cfg = std::move(cfg);
+    state.cfg_temp = std::move(cfg);
     state.kernel.wait_for_debugger = state.cfg.wait_for_debugger;
 
     state.base_path = root_paths.get_base_path_string();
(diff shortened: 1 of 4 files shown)


Zangetsu38:
No need that att all, just need add button for can save config

vita3k/util/include/util/overloaded.h
```diff
@@ -1,4 +1,4 @@
 template <class... Ts>
 struct overloaded : Ts... { using Ts::operator()...; };
 template <class... Ts>
-overloaded(Ts...)->overloaded<Ts...>;
\ No newline at end of file
+overloaded(Ts...)->overloaded<Ts...>;
```
Zangetsu38, line 4: very need found how fix this things, i get same on run format all time lol

Zangetsu38:
after talk with other dev, this issue is no possible fix, is yaml issue itself, and your solution is good, but in this case no can really good works :/, memory continue increment on your all change setting, better way on this case is my workaround pr with button, like that, change all setting and save, for reduce maximum memory leak

Zangetsu38:
commiçt name is wrong ^^ remove profil manager

PR #976, opened 2020-09-20
gui: workaround mem leak for setting dialog

Zangetsu38:
workaround memory leak issue for yaml.
- save setting if press save button

Not real fix, issue is itself yaml submodule


Ristellise:
Doesn't actually fix the root cause which is the yaml error, just makes it much less common.

Zangetsu38:
Yaml error is other thinks

Zangetsu38:
Already done

PR #982, opened 2020-10-11
gui: Add missing PopStyleColor

scribam:
To match with: https://github.com/Vita3K/Vita3K/blob/8369d0287bb0e525b77a7488edad4c9d59390fc3/vita3k/gui/src/common_dialog.cpp#L69


VelocityRa (approved):
👍

PR #984, opened 2020-10-11
io: Fix sceIoRemove and sceIoRmdir return codes on success

scribam:
`sceIoRemove` and `sceIoRmdir` must return 0 on success which is not the case on master as `remove_file` and `remove_dir` return true (aka 1) on success


VelocityRa (approved):
Thanks.

PR #989, opened 2020-10-20
renderer : fix mem leaks in data uploading commands

sunho:
This should remove most of the reported memory leaks when running games for a long time.

Fix https://github.com/Vita3K/Vita3K/issues/964


VelocityRa (approved):
👍

PR #995, opened 2020-10-21
shader: memory load instruction with actual base and float scalar array

sunho:
I observed some shader that actually use some value mutated by alu as base. So current method will fail because it assumes that shader will always use certain registers listed in meta as base.

This pr tries to better emulate uniform buffer part of gpu memory by the following method. 

- Create an array varialbe that is big enough
- Load all uniform buffer data to that variable
- The base will be just index of that array 
- Use float scalar array instead of vector array to ease the implementation difficulty


VelocityRa (approved):
nice

PR #1004, opened 2020-10-29
renderer: fix few bugs

sunho:
# About PR

- Fix the mismatch between vertex attribute name of shader and runtime
- Shrink the size of semantic field of gxm parameter struct

# Why shrink

sceGxmProgramFindParameterBySemantic tried to find parameter with semantic 0x6, but it failed because the desired parameter has semantic 0x0106. 0x01 should be from another field. (I guess it's a flag)

# Related issue

This fixes the black screen render of ingame and title of puyo puyo tetris.

### DIFF
--- a/vita3k/modules/SceCommonDialog/SceCommonDialog.cpp
+++ b/vita3k/modules/SceCommonDialog/SceCommonDialog.cpp
@@ -1009,7 +1009,7 @@ EXPORT(int, sceSaveDataDialogGetResult, Ptr<SceSaveDataDialogResult> result) {
         return RET_ERROR(SCE_COMMON_DIALOG_ERROR_NULL);
     }
 
-    if (host.common_dialog.status != SCE_COMMON_DIALOG_STATUS_FINISHED && host.common_dialog.substatus != SCE_COMMON_DIALOG_STATUS_FINISHED) {
+    if (host.common_dialog.status != SCE_COMMON_DIALOG_STATUS_FINISHED || host.common_dialog.substatus != SCE_COMMON_DIALOG_STATUS_FINISHED) {
         return RET_ERROR(SCE_COMMON_DIALOG_ERROR_NOT_FINISHED);
     }
 
--- a/vita3k/gxm/include/gxm/types.h
+++ b/vita3k/gxm/include/gxm/types.h
@@ -1397,7 +1397,8 @@ struct SceGxmProgramParameter {
         bf_t<uint16_t, 8, 4> component_count; // applicable for constants, not applicable for samplers (select size like float2, float3, float3 ...)
         bf_t<uint16_t, 12, 4> container_index; // applicable for constants, not applicable for samplers (buffer, default, texture)
     };
-    uint16_t semantic; // applicable only for for vertex attributes, for everything else it's 0
+    uint8_t semantic; // applicable only for for vertex attributes, for everything else it's 0
+    uint8_t pad;
     uint32_t array_size;
     int32_t resource_index;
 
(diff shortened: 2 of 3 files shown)


frangarcj (approved):

vita3k/modules/SceCommonDialog/SceCommonDialog.cpp
```diff
@@ -1009,7 +1009,7 @@ EXPORT(int, sceSaveDataDialogGetResult, Ptr<SceSaveDataDialogResult> result) {
         return RET_ERROR(SCE_COMMON_DIALOG_ERROR_NULL);
     }
 
-    if (host.common_dialog.status != SCE_COMMON_DIALOG_STATUS_FINISHED && host.common_dialog.substatus != SCE_COMMON_DIALOG_STATUS_FINISHED) {
+    if (host.common_dialog.status != SCE_COMMON_DIALOG_STATUS_FINISHED || host.common_dialog.substatus != SCE_COMMON_DIALOG_STATUS_FINISHED) {
```
Zangetsu38, line 1012:
```suggestion
    if ((host.common_dialog.status != SCE_COMMON_DIALOG_STATUS_FINISHED) || (host.common_dialog.substatus != SCE_COMMON_DIALOG_STATUS_FINISHED)) {
```
maybe ?

sunho: Nope. https://en.cppreference.com/w/cpp/language/operator_precedence

Zangetsu38: All rpcs3 dev have said me need zdd () if you usine two things like this.
Like if (i == 2 || i == 3) request add () for both for correctly separe it

sunho: It's probably coding convention in rpcs3 to avoid mistake when adding more expressions.

Zangetsu38: Notre really i have see is normal for all dev, vita3k dev using also this

sunho: All dev? https://github.com/google/googletest/blob/master/googletest/src/gtest-port.cc#L322 https://github.com/tensorflow/tensorflow/blob/master/tensorflow/cc/framework/cc_op_gen.cc#L60 https://github.com/tensorflow/tensorflow/blob/master/tensorflow/cc/framework/cc_op_gen.cc#L282 https://github.com/google/googletest/blob/master/googletest/src/gtest-port.cc#L1011 https://github.com/Vita3K/Vita3K/blob/master/vita3k/modules/SceGxm/SceGxm.cpp#L1665

Zangetsu38: no, i want said, i seeé majory case explain need ad () for all if (())
vita3k dev have already request me i add this () because have two == inside if

sunho: I just don't see the need here. If you can give me more proper reason to me, I will add them. Why do you keep bothering with weird idea again and again. I'm tired of this.

Zangetsu38: sorry, i have not wanted tired you :/ i have just explazin exacltyy like other dev have explain me that all :(

Zangetsu38 (approved):

VelocityRa (approved):

PR #1005, opened 2020-10-29
ci: Use Xcode 11.7 until compilation with Xcode 12 is solved

scribam:


sunho (approved):

sunho:
This is related to ffmpeg. (precisely nasm) It's known problem so hopefully solved by ffmpeg team if we wait. If it's not solved for a long time, we might need to get ffmpeg without assembly optimization. (configure --disable-asm)

PR #1008, opened 2020-11-07
np: set state.inited to false on deinit

scribam:
Otherwise, if you call `sceNpInit` =>`sceNpTerm` => `sceNpInit`, you have a `SCE_NP_ERROR_ALREADY_INITIALIZED` error. Same thing for the trophy functions.

### DIFF
--- a/vita3k/np/src/init.cpp
+++ b/vita3k/np/src/init.cpp
@@ -31,6 +31,7 @@ bool init(NpState &state, const np::CommunicationID *comm_id) {
 
 bool deinit(NpState &state) {
     // Reserved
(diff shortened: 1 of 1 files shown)


sunho (approved):

sunho:
How about netctl?

scribam:
netctl is ok to me according to this code:
https://github.com/Vita3K/Vita3K/blob/aa46de69c0d6fd7720b4d66775749603343fc9bd/vita3k/modules/SceNet/SceNet.cpp#L366-L375

sunho:
Got it. :+1: for finding these!

PR #1016, opened 2020-11-12
gxm: Optimize graphics hashing with XXH3

EXtremeExploit:
* XXH3 is faster than accumulate and SHA256
* Also optimize nids find inside resolve_export function
# Testing
Since my PC is not that great i am unable to detect any performance boost with this, but technically it should, if someone could do some benchmarking justo to get some numbers would be great

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -81,3 +81,6 @@
 [submodule "external/psvpfstools"]
 	path = external/psvpfstools
 	url = https://github.com/Vita3K/psvpfstools.git
+[submodule "external/xxHash"]
+	path = external/xxHash
+	url = https://github.com/Cyan4973/xxHash
--- a/vita3k/kernel/include/kernel/state.h
+++ b/vita3k/kernel/include/kernel/state.h
@@ -29,6 +29,7 @@
 #include <map>
 #include <mutex>
 #include <queue>
+#include <unordered_map>
 #include <vector>
 
 struct ThreadState;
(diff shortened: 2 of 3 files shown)


sunho (changes requested):
This is clever optimization but this hash value is used to identify the shader from various games. Using different hashing algorithm will break the database of shaders collected for debugging purpose (e.g. shader 342hlsfd3424 from game A is broken)

sunho:
I thought about having an option to enable this optimization but that would confuse communication sometimes. (e.g. tester gives a shader log to developer but the names are different) So, I'm afraid we shouldn't pull this change if the issues I mentioned are not addressed.

Frain-Breeze:
You could add a prefix to the hash to identify what algorithm was used, like xxh3_HASH.

sunho:
We can take change in texture_cache.cpp confidently, can you separate the pr?

EXtremeExploit:
Should i also remove the changes to [this file](https://github.com/Vita3K/Vita3K/pull/1016/files#diff-5ff5ee152e1a16923b44fa35ba31fc48b696354b521ccadd88e3fdc233113ee7R58) ?

sunho:
That looks great. 👍 Just the part that change the name of shader is problematic.

sunho (approved):
:)

sunho:
I'm sorry for rejecting some of your code.

sunho:
I just noticed that xxHash is not added in cmake file. Is it okay?

EXtremeExploit:
it worked just fine and the compilation didn't fail so i guess its ok

PR #1017, opened 2020-11-14
SceGxm: fix SCE_GXM_MAX_VERTEX_STREAMS value

scribam:


sunho (approved):
Good find. I think some code use hardcoded value for max vertex stream. I'll look around.

PR #1019, opened 2020-11-14
SceGxm: init surface correctly in sceGxmDepthStencilSurfaceInitDisabled

scribam:


sunho:
https://github.com/Vita3K/Vita3K/blob/aebc9fb0fd25dbf399a4e3f97fd2135609fb42fc/vita3k/gxm/include/gxm/types.h#L1053

Shouldn't this set it 1.0?

scribam:
For the first change, yes. For the second one, you have a `memset(surface, 0, sizeof(*surface));` before so I assume it sets the value to 0.0

sunho (changes requested):
Can you change sceGxmDepthStencilSurfaceInitDisabled to copy default initialized 
 SceGxmDepthStencilSurface  struct like sceGxmDepthStencilSurfaceInit do? That way we can just use the struct definition for changing default value.

scribam:
@sunho done, I have used the code from sceGxmDepthStencilSurfaceInit to initialize the structures. Please have a look again

sunho (approved):

PR #1021, opened 2020-11-15
SceGxm: add missing returns before RET_ERROR

scribam:
And remove some invalid checks in `sceGxmTextureGet*` functions (verified by RE)

### DIFF
--- a/vita3k/modules/SceGxm/SceGxm.cpp
+++ b/vita3k/modules/SceGxm/SceGxm.cpp
@@ -208,12 +208,12 @@ EXPORT(int, sceGxmColorSurfaceInit, SceGxmColorSurface *surface, SceGxmColorForm
     assert(data);
 
     if (!surface || !data)
-        RET_ERROR(SCE_GXM_ERROR_INVALID_POINTER);
+        return RET_ERROR(SCE_GXM_ERROR_INVALID_POINTER);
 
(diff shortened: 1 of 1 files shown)


vita3k/modules/SceGxm/SceGxm.cpp
```diff
@@ -755,7 +755,7 @@ EXPORT(void, sceGxmFinish, SceGxmContext *context) {
     assert(context);
 
     if (!context)
-        RET_ERROR(SCE_GXM_ERROR_INVALID_POINTER);
+        return;
```
Zangetsu38, line 758: return nothing normal ?

scribam: The return type of the function is `void` so yes, it is normal

Zangetsu38: oh, i see lol

sunho (approved):
Thanks for doing this

sunho:
IIRC returning made regressions in many games (e.g. new game, p4g) Even if this is from RE, we better keep track of how these changes affect compatibility.

scribam:
I tried a new game in P4G and haven't found any issues so far. Only the changes in `sceGxmTextureGet*` functions are from RE, the other ones are to "fix" the fact `RET_ERROR` only doesn't not return if you don't add a return before.
I plan to push more changes about error checking after this PR.

PR #1023, opened 2020-11-15
add support for various texture formats

Frain-Breeze:
I'm not happy with the way the depalletization is split in two parts, but keeping them together doesn't seem to function correctly. Insight would be appreciated.


Frain-Breeze:
this is a cleaner implementation and makes more sense. I have tried a few games to see if it breaks anything, but I haven't seen any odd behavior resulting from the change.

zkyhzxf (approved):

zkyhzxf (approved):

cdwcgt (approved):

sunho (approved):
🎉

PR #1028, opened 2020-11-18
SceGxm: add function arguments and error checks

scribam:
Also:
- remove assert from the code when there is already an explicit check which returns an error (RET_ERROR)
- add assert to the code when there is no explicit check found in RE
- change return types to more accurate ones


sunho (approved):
👍

PR #1031, opened 2020-11-22
SceGxm: do not check outputRegisterSize value in sceGxmColorSurfaceInit

scribam:
I didn't found a similar check during RE, not sure why it was added.


Zangetsu38:
Goot fix régression on gravity rush boot

frangarcj (approved):

PR #1032, opened 2020-11-24
shader: fix bugs in uniform buffer & improve logging

sunho:
# About PR

- Correct the unit of src1 in ldr (it was in words instead of bytes)
- Correct repeat offset of vbw (it needed /2)
- Adjust null checks 
- Improve logging

# Related issue

Fixes regression in BlazBlue.

### DIFF
--- a/vita3k/shader/src/gxp_parser.cpp
+++ b/vita3k/shader/src/gxp_parser.cpp
@@ -189,19 +189,22 @@ ProgramInput shader::get_program_input(const SceGxmProgram &program) {
 
         const uint32_t offset = base_offset + buffer_info->base_offset;
 
-        const auto buffer = uniform_buffers.at(buffer_info->reside_buffer);
+        const auto buffer = uniform_buffers.find(buffer_info->reside_buffer);
+        // buffer = null seems to happen when there's a leftover uniform buffer (uniform buffer that's not used in shader code)
+        // This case needs more invastigation
+        if (buffer != uniform_buffers.end()) {
+            Input item;
+            item.type = DataType::UINT32;
+            item.offset = offset;
+            item.component_count = 1;
+            item.array_size = 1;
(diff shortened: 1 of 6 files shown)


vita3k/shader/src/gxp_parser.cpp
```diff
@@ -189,19 +189,22 @@ ProgramInput shader::get_program_input(const SceGxmProgram &program) {
...
-        item.array_size = 1;
-        UniformBufferInputSource source;
-        source.base = buffer.reg_block_size;
-        source.index = buffer.index;
-        item.source = source;
-
-        program_input.inputs.push_back(item);
+        const auto buffer = uniform_buffers.find(buffer_info->reside_buffer);
+        // buffer = null seems to happen when there's a leftover uniform buffer (uniform buffer that's not used in shader code)
+        // This case needs more invastigation
```
scribam, line 194:
```suggestion
        // This case needs more investigation
```

vita3k/shader/src/spirv_recompiler.cpp
```diff
@@ -1215,7 +1215,7 @@ static SpirvCode convert_gxp_to_spirv(const SceGxmProgram &program, const std::s
     return spirv;
 }
 
-static std::string convert_spirv_to_glsl(SpirvCode spirv_binary, const FeatureState &features, TranslationState &translation_state) {
+static std::string convert_spirv_to_glsl(std::string shader_name, SpirvCode spirv_binary, const FeatureState &features, TranslationState &translation_state) {
```
scribam, line 1218:
```suggestion
static std::string convert_spirv_to_glsl(const std::string &shader_name, SpirvCode spirv_binary, const FeatureState &features, TranslationState &translation_state) {
```

scribam (approved):

PR #1035, opened 2020-11-24
SceGxm: implement missing sceGxmProgramIs* functions

scribam:


sunho (approved):
Interesting finds. We might be able to make use of these in other part as well.

Issue #392, opened 2019-01-26
macOS: My Vita3K.app instantly closes!


1whatleytay:
**Steps to reproduce:** Double click the .app file on macOS. You can get your .app file from your development environment on the website, worked both ways for me.
**What happens:** It instantly closes! How rude!
**What I think should happen:** It works, starts up, and I can play my pre-installed games without any workarounds.

**Workarounds:** It's not impossible to start Vita3K though, you'll just have to try something different.

- Right click on Vita3K and click "Show Package Contents". Then, navigate Contents -> MacOS and double click on Vita3K. It'll open a terminal and run perfectly!
<img width="378" alt="screen shot 2019-01-26 at 11 05 16 am" src="https://user-images.githubusercontent.com/32211852/51789779-9c072d00-215a-11e9-954d-4679410f19d0.png">
<img width="881" alt="screen shot 2019-01-26 at 11 05 38 am" src="https://user-images.githubusercontent.com/32211852/51789789-ba6d2880-215a-11e9-81e0-67411730638a.png">

- Opening it in your development environment directly seems to work fine as well, wether it be XCode or Clion.
- Dropping your Vita3K executable in the terminal directly also does the trick!

**What the heck is going wrong!?** I was having difficulty properly debugging this bug (or at least, I hope it is) because in every development environment Vita3K worked fine. I tried opening it in Instruments too, but it started and worked normally.
**My best guess:** I'm thinking it has something to do with Vita3K's logging. Development environments and the workarounds above all open terminals which Vita3K uses to print, but directly opening the .app does not. Maybe its hitting some bad access?

If this is expected behaviour, then feel free to close this issue. It might be nice to clear this up in #info-faq though. This same issue was encountered by IOI#6476 in discord, so I thought I would at least say something here.

johnothwolo:
I can confirm that happens on the latest nightly build on Mojave 10.14.2.

VelocityRa:
@1whatleytay You can try adding `raise(SIGTRAP);` line by line to see where it fails.

Also I'm assuming you've tried running it by attaching debuggers to the appveyor builds, correct? As opposed to building + running Debug builds from the environment. If not, try it.

1whatleytay:
I just double checked, the when I try to run the Travis build with Instruments it works fine. Just double clicking it makes it crash... how irritating!

I'll try dropping `raise(SIGTRAP);`s everywhere in a bit. Thanks for the help.

VelocityRa:
[...]

PR #1036, opened 2020-11-24
Fix macOS rpath

johnothwolo:
Fixes issue #392. `discord_game_sdk.bundle` changed to match the dylib dependency `@rpath/discord_game_sdk.dylib`. Also changed the rpath to `@loader_path` according to https://blog.krzyzanowskim.com/2018/12/05/rpath-what. Tested and verified on 2 different machines. However, according to Apple it'd be best to move all libraries to the Frameworks directory.


sunho:
@IllusionMan1212

IllusionMan1212 (approved):
looks good 👍

PR #1038, opened 2020-11-25
SceGxm: more error checks update

scribam:
[Danganronpa Another Episode: Ultra Despair Girls [PCSB00788]](https://github.com/Vita3K/compatibility/issues/294): Nothing => Menu

Capture1

There is some checks where the value `multisampleMode` is compared to others but they exist in a more complex way than the one we have in the code. So, until further investigation is done, I think it is better to remove this inaccurate check.

### DIFF
--- a/vita3k/modules/SceGxm/SceGxm.cpp
+++ b/vita3k/modules/SceGxm/SceGxm.cpp
@@ -225,6 +225,9 @@ EXPORT(int, sceGxmColorSurfaceInit, SceGxmColorSurface *surface, SceGxmColorForm
     if (strideInPixels & 1)
         return RET_ERROR(SCE_GXM_ERROR_INVALID_ALIGNMENT);
 
+    if ((strideInPixels < width) || ((data.address() & 3) != 0))
+        return RET_ERROR(SCE_GXM_ERROR_INVALID_VALUE);
+
     memset(surface, 0, sizeof(*surface));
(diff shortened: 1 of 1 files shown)


vita3k/modules/SceGxm/SceGxm.cpp
```diff
@@ -225,6 +225,9 @@ EXPORT(int, sceGxmColorSurfaceInit, SceGxmColorSurface *surface, SceGxmColorForm
     if (strideInPixels & 1)
         return RET_ERROR(SCE_GXM_ERROR_INVALID_ALIGNMENT);
 
+    if ((strideInPixels < width) || ((data.address() & 3) != 0))
+        return RET_ERROR(SCE_GXM_ERROR_INVALID_VALUE);
```
sunho, line 229: Shouldn't we return SCE_GXM_ERROR_INVALID_ALIGNMENT for data.address() & 3 case?

scribam: Unless I misinterpreted something during RE, this is how it is supposed to be although I agree with you, SCE_GXM_ERROR_INVALID_ALIGNMENT  would be more appropriated for this case.

Zangetsu38:
you can squash last with first ^^ no need other commit for same things :)

sunho (approved):

PR #1043, opened 2020-12-05
gxm: fix is_sampler_cube method

scribam:
Previously, we had:
```
uint16_t semantic;
[...]
bool is_sampler_cube() const {
    return (semantic >> 12) & 1;
}
```
then the `semantic` variable split into two uint8_t:
```
uint8_t semantic;
uint8_t semantic_index;
```
so I believe the `is_sampler_cube` method needs an update. Also, it makes no sense to shift an uint8_t to the right by 12 bits


sunho (approved):
You're totally right!!

Issue #1050, opened 2020-12-20
Regression 3d model color (school idol paradise)


Zangetsu38:
On last master
 ![image](https://cdn.discordapp.com/attachments/442344583293304833/790031089556389898/unknown.png)

on little old master

![image](https://user-images.githubusercontent.com/5261759/102703370-2f293e80-426e-11eb-804c-306179ad90e0.png)

Caused by commit by @sunho  https://github.com/Vita3K/Vita3K/commit/1149e099bde78e699bd6a6373d99288f4396362b 

sb_mode 0 missing inside smp
in log, i have 
>Unsupported sb_mode: 0

if i add case 0 with case 1 get works fine

Zangetsu38:
Fixed by #1051

PR #1051, opened 2020-12-20
shader: add sb_mode 0 in smp

Zangetsu38:
add case 0 request by game, fix regression  #1050

### DIFF
--- a/vita3k/shader/src/translator/texture.cpp
+++ b/vita3k/shader/src/translator/texture.cpp
@@ -181,11 +181,10 @@ bool USSETranslatorVisitor::smp(
     spv::Id result = do_fetch_texture(sampler, { coord, static_cast<int>(DataType::F32) }, DataType::F32);
 
     switch (sb_mode) {
-    case 1: {
+    case 0:
+    case 1:
         store(inst.opr.dest, result, 0b1111);
         break;
-    }
-
     case 3: {
         // TODO: figure out what to fill here
         //store(inst.opr.dest, stub, 0b1111);


vita3k/shader/src/translator/texture.cpp
```diff
@@ -181,11 +181,10 @@ bool USSETranslatorVisitor::smp(
     spv::Id result = do_fetch_texture(sampler, { coord, static_cast<int>(DataType::F32) }, DataType::F32);
 
     switch (sb_mode) {
-    case 1: {
+    case 0:
```
VelocityRa, line 184:
```suggestion
    case 0: // TODO: is this correct?
```

Zangetsu38: Not sur i wating @sunho info, but is user like that before is probably same mode 1

VelocityRa: ok let's see what sunho says

pent0:
Sb_mode is sample bypassing mode. It's to decide which sample values to gather during the smp call. [Docs](http://cdn.imgtec.com/sdk-documentation/PowerVR+Instruction+Set+Reference.pdf) here talks a bit about it.

From @sunho work:
- mode 3 has some extra info (4 components), and mode 1 is just normal sample result.
- Based on the docs there, I assume mode 0 is default mode, same as mode 1.
- Mode 2 is info mode (contains sample info, i dont know what it contains yet, need hardware test. I guess it has mipmap level maybe).
- Mode 3 is both, since 4 latter components are sample data, 4 beginning components should be info mode

Add mode 0 fix rainy texture in menu of Virtua Tennis 4. All results gather here are likely correct.

a

lol i see zang already made pr

vita3k/shader/src/translator/texture.cpp
Zangetsu38: Is approuve

frangarcj (approved):

sunho:
:+1::+1::+1:

PR #1052, opened 2020-12-20
sceAppUtilDrmOpen: fix error code when dirName is NULL

scribam:
On master, the function returns `SCE_APPUTIL_ERROR_NOT_INITIALIZED` instead of `SCE_APPUTIL_ERROR_PARAMETER`  when dirName is NULL.

### DIFF
--- a/vita3k/modules/SceAppUtil/SceAppUtil.cpp
+++ b/vita3k/modules/SceAppUtil/SceAppUtil.cpp
@@ -101,11 +101,11 @@ EXPORT(int, sceAppUtilDrmClose) {
     return UNIMPLEMENTED();
 }
 
-EXPORT(int, sceAppUtilDrmOpen, SceAppUtilDrmAddcontId *dirName, SceAppUtilMountPoint *mountPoint) {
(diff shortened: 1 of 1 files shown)


vita3k/modules/SceAppUtil/SceAppUtil.cpp
```diff
@@ -101,14 +101,14 @@ EXPORT(int, sceAppUtilDrmClose) {
...
+    if (!dirName)
+        return RET_ERROR(SCE_APPUTIL_ERROR_PARAMETER);
 
-    if (dirName == nullptr)
-        return RET_ERROR(SCE_APPUTIL_ERROR_NOT_INITIALIZED);
+    const auto drm_content_id_path{ fs::path(host.pref_path) / (+VitaIoDevice::ux0)._to_string() / host.io.device_paths.addcont0 / reinterpret_cast<const char *>(dirName->data) };
 
     if (!fs::exists(drm_content_id_path) || (fs::is_empty(drm_content_id_path)))
-        return SCE_ERROR_ERRNO_ENOENT;
+        return RET_ERROR(SCE_ERROR_ERRNO_ENOENT);
```
Zangetsu38, line 111: accorded with fran before, is not error, it check dlc so if no exist is normal not error, for that i have not used RET_ERROR ^^

VelocityRa (approved):

PR #1061, opened 2020-12-23
config/gui/renderer: add configuration to dump textures

scribam:


Zangetsu38 (approved):
thx i have wanted set t'his before but no found way ^^

VelocityRa (approved):

PR #1063, opened 2020-12-25
SceCommonDialog: fix savedata dialog logic

IllusionMan1212:
i think this is how it should be, i tested a bunch of games and found no regressions


Zangetsu38 (approved):

1whatleytay (approved):
It would be nice if we moved `result.get(host.mem)` off to a variable or something but that's not the purpose of this PR.

PR #1070, opened 2020-12-29
SceGxm: add error checks and stub some functions

scribam:


vita3k/modules/SceGxm/SceGxm.cpp
```diff
@@ -124,19 +124,21 @@ EXPORT(int, sceGxmBeginCommandList, SceGxmContext *deferredContext) {
...
     if (!context) {
         return RET_ERROR(SCE_GXM_ERROR_INVALID_POINTER);
     }
+
+    if (flags & 0xFFFFFFF0) {
+        return RET_ERROR(SCE_GXM_ERROR_INVALID_VALUE);
+    }
+
+    if (!renderTarget || (vertexSyncObject != nullptr)) {
+        return RET_ERROR(SCE_GXM_ERROR_INVALID_POINTER);
```
Zangetsu38, line 136: value return "SCE_GXM_ERROR_INVALID_POINTER" already writed for context, you an add check inside same ?

scribam: No as this will not respect the order of the checks

Zangetsu38: oh i see :)

IllusionMan1212 (approved):

PR #1076, opened 2020-12-31
shader: Add Opcode ISUB16/ISUBU16.

Zangetsu38:
Request by Blaze Blue, already approved by kd

### DIFF
--- a/vita3k/shader/src/translator/alu.cpp
+++ b/vita3k/shader/src/translator/alu.cpp
@@ -468,11 +468,12 @@ spv::Id USSETranslatorVisitor::do_alu_op(Instruction &inst, const Imm4 dest_mask
         break;
     }
 
+    case Opcode::ISUBU16:
+    case Opcode::ISUB16:
     case Opcode::ISUBU32:
-    case Opcode::ISUB32: {
+    case Opcode::ISUB32:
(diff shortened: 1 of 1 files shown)


vita3k/shader/src/translator/alu.cpp
```diff
@@ -468,11 +468,12 @@ spv::Id USSETranslatorVisitor::do_alu_op(Instruction &inst, const Imm4 dest_mask
         break;
     }
 
+    case Opcode::ISUBU16:
+    case Opcode::ISUB16:
     case Opcode::ISUBU32:
-    case Opcode::ISUB32: {
+    case Opcode::ISUB32:
         result = m_b.createBinOp(spv::OpISub, m_b.makeIntType(32), vsrc1, vsrc2);
```
scribam, line 475: Is `m_b.makeIntType(32)` right for ISUBU16 and ISUB16? Shouldn't be  `m_b.makeIntType(16)`? 🤔

pent0: Integer are all extracted as 32-bit. The storing of them is 16-bit

Zangetsu38: Yes, kd explain after i try 16 and get error you can't using 16-bit, need 32, check value already here

scribam (approved):

PR #1079, opened 2021-01-01
Fix for XCode 12

1whatleytay:
Addresses #985 

Merge https://github.com/Vita3K/ffmpeg-core/pull/2 first. I want to look at CI and see how it does with this new build under an older version of macOS (but it'll probably just have submodule issues).

PR #1080, opened 2021-01-01
Use threadsafe versions of std::localtime and std::gmtime

1whatleytay:
`std::localtime` and `std::gmtime` use static state by default, meaning they aren't especially threadsafe. Platform specific methods fix this by letting you allocate your own `tm` structure that won't be overwritten by a rogue function call.

I don't think this is a priority, I just want something simple to go through right now, and CodeQL has some simple warnings that are easy to resolve. Will probably be going through more until #1079 gets merged.

### DIFF
--- a/vita3k/gui/src/content_manager.cpp
+++ b/vita3k/gui/src/content_manager.cpp
@@ -26,8 +26,9 @@
 #include <io/VitaIoDevice.h>
 #include <io/functions.h>
 
-namespace gui {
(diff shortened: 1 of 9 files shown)


VelocityRa (dismissed):

Zangetsu38:
Fail on format check ^^

Zangetsu38 (approved):

PR #1084, opened 2021-01-04
util: Support f32 to f16 conversion even without AVX2

bookmist:
Check if AVX2 is suported at run-time and use appropriate conversion function.

I check on Intel(R) Core(TM) i5-2500K CPU and MSVC 2019 compiler and it works well, but it needs check on other CPUs and compilers.

### DIFF
--- a/vita3k/util/src/util.cpp
+++ b/vita3k/util/src/util.cpp
@@ -250,7 +250,7 @@ std::int64_t byte_swap(std::int64_t val) {
     return byte_swap(static_cast<std::uint64_t>(val));
 }
 
-void float_to_half(const float *src, std::uint16_t *dest, const int total) {
+void float_to_half_AVX_F16C(const float *src, std::uint16_t *dest, const int total) {
     float toconvert[8] = { 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f };
     int i = 0;
 
@@ -266,6 +266,48 @@ void float_to_half(const float *src, std::uint16_t *dest, const int total) {
         dest += 8;
     }
 }
+
+#if (defined(__AVX__) && defined(__F16C__)) || defined(__AVX2__)
+//forced use AVX+F16C instruction set
+//AVX2 checked intentionally cause MSVC does not have __F16C__ macros
+//and checking AVX is not enough for some CPU architectures (Intel Sandy bridge)
+void float_to_half(const float *src, std::uint16_t *dest, const int total) {
+    float_to_half_AVX_F16C(src, dest, total);
+}
+#else
+//check and use AVX+F16C instruction set if possible
+#include <util/instrset_detect.h>
+
+void float_to_half_basic(const float *src, std::uint16_t *dest, const int total) {
+    int i = 0;
+    while (i < total) {
+        uint32_t x = *((uint32_t *)&src[i]);
+        dest[i] = ((x >> 16) & 0x8000) | ((((x & 0x7f800000) - 0x38000000) >> 13) & 0x7c00) | ((x >> 13) & 0x03ff);
+        i++;
+    }
+}
+
+// use function variable as imitation of self-modifying code.
+// on first use we check processor features and set appropriate realisation, later we immediately use appropriate realisation
+void float_to_half_init(const float *src, std::uint16_t *dest, const int total);
+
+void (*float_to_half_var)(const float *src, std::uint16_t *dest, const int total) = float_to_half_init;
+
+void float_to_half_init(const float *src, std::uint16_t *dest, const int total) {
+    if (util::instrset::hasF16C()) {
+        float_to_half_var = float_to_half_AVX_F16C;
+        LOG_INFO("AVX+F16C instruction set supported. Use fast f32 to f16 conversion");
+    } else {
+        float_to_half_var = float_to_half_basic;
+        LOG_INFO("AVX+F16C instruction set not supported. Use basic f32 to f16 conversion");
+    }
+    (*float_to_half_var)(src, dest, total);
+}
+
+void float_to_half(const float *src, std::uint16_t *dest, const int total) {
+    (*float_to_half_var)(src, dest, total);
+}
+#endif
 
 // pent0 found on stackoverflow
 // https://stackoverflow.com/questions/4398711/round-to-the-nearest-power-of-two/4398799
--- a/vita3k/util/CMakeLists.txt
+++ b/vita3k/util/CMakeLists.txt
@@ -10,6 +10,7 @@ add_library(
     include/util/find.h
     include/util/fs.h
     include/util/function_info.h
+    include/util/instrset_detect.h
     include/util/semaphore.h
     include/util/lock_and_find.h
     include/util/log.h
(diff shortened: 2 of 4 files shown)


scribam:
`_mm256_load_ps` requires AVX and `_mm256_cvtps_ph` requires F16C but you seem to check for AVX2 support in your code :thinking: 
You can find F16C detection here : https://github.com/vectorclass/version2/blob/master/instrset_detect.cpp

Also, wouldn't be better to move the instruction set detection code to a specific file? Or at least out of the `#ifdef __AVX2__ #else #endif` block to it can be reused for other usages?
If you prefer to use a third party library (in the external folder), I am okay with that.

bookmist:
> `_mm256_load_ps` requires AVX and `_mm256_cvtps_ph` requires F16C but you seem to check for AVX2 support in your code 🤔
> You can find F16C detection here : https://github.com/vectorclass/version2/blob/master/instrset_detect.cpp
> 
> Also, wouldn't be better to move the instruction set detection code to a specific file? Or at least out of the `#ifdef __AVX2__ #else #endif` block so it can be reused for other usages?
> If you prefer to use a third party library (in the external folder), I am okay with that.

Changed detection and moved to specific file.

bookmist:
@scribam @frangarcj Please, review.

vita3k/util/src/util.cpp
```diff
@@ -267,6 +267,47 @@ void float_to_half(const float *src, std::uint16_t *dest, const int total) {
     }
 }
 
+#ifdef __AVX2__
```
scribam, line 270: Why AVX2? It should be AVX + F16C

vita3k/util/src/util.cpp
```diff
@@ -267,6 +267,47 @@ void float_to_half(const float *src, std::uint16_t *dest, const int total) {
...
+//forced use F16C instruction set
+//AVX2 checked intentionally cause checking AVX is not enough for some CPU architectures (Sandy bridge)
```
scribam, line 272: Update comment to remove reference to AVX2

vita3k/util/src/util.cpp
```diff
@@ -267,6 +267,47 @@ void float_to_half(const float *src, std::uint16_t *dest, const int total) {
...
+}
+
+// use function variable as imitation of self-modifying code.
+// on first use we check processor features and set appropriate realisation, later we immediately use appropriate realisation
+void float_to_half_init(const float *src, std::uint16_t *dest, const int total);
+
+void (*float_to_half_var)(const float *src, std::uint16_t *dest, const int total) = float_to_half_init;
+
+void float_to_half_init(const float *src, std::uint16_t *dest, const int total) {
+    if (util::instrset::hasF16C()) {
```
scribam, line 296: Requires AVX + F16C

vita3k/util/src/util.cpp
```diff
@@ -267,6 +267,47 @@ void float_to_half(const float *src, std::uint16_t *dest, const int total) {
...
+        float_to_half_var = float_to_half_F16C;
+        LOG_INFO("F16C instruction set supported. Use fast f32 to f16 conversion");
```
scribam, line 298: AVX + F16C

vita3k/util/src/util.cpp
```diff
@@ -267,6 +267,47 @@ void float_to_half(const float *src, std::uint16_t *dest, const int total) {
...
+    } else {
+        float_to_half_var = float_to_half_basic;
+        LOG_INFO("F16C instruction set not supported. Use basic f32 to f16 conversion");
```
scribam, line 301: AVX + F16C

vita3k/util/src/util.cpp
```diff
@@ -250,7 +250,7 @@ std::int64_t byte_swap(std::int64_t val) {
     return byte_swap(static_cast<std::uint64_t>(val));
 }
 
-void float_to_half(const float *src, std::uint16_t *dest, const int total) {
+void float_to_half_F16C(const float *src, std::uint16_t *dest, const int total) {
```
scribam, line 253: Probably rename it to float_to_half_AVX_F16C or similar

scribam (changes requested):
These lines force the requirements to avx + f16c for specific compilers and should be removed I think
https://github.com/Vita3K/Vita3K/blob/24385524586edb63274eaa935e04a9205ac979eb/CMakeLists.txt#L52-L56

vita3k/util/src/util.cpp
bookmist: I don't know how to check exactly AVX + F16C at compile-time. common compiler macros are __AVX__ and __AVX2__  I've found some checks but they were compiler-specific.
As sayed on wikipedia Intel Sandy Bridge has AVX but does not have F16C, so checking __AVX__ is not enough.

vita3k/util/src/util.cpp
bookmist: F16C include AVX (hasF16C function firstly check AVX)

vita3k/util/src/util.cpp
scribam: `#if defined(__AVX__) && defined(__F16C__)` does not work?

vita3k/util/src/util.cpp
scribam: Miss that, all good then 👍

vita3k/util/src/util.cpp
bookmist: I can't check, but as told here https://git.sailfishos.org/mer-core/qtbase/commit/280e321e52fd4e86545f3f0d4bd4e047786a897e and there https://stackoverflow.com/questions/16348909/how-do-i-know-if-i-can-compile-with-fma-instruction-sets "Unfortunately MSVC doesn't define any instruction set specific macros other than __AVX__ and __AVX2__"

bookmist: maybe will work #if (defined(__AVX__) && defined(__F16C__)) || defined(__AVX2__)

CMakeLists.txt
```diff
@@ -49,12 +49,6 @@ if(CMAKE_GENERATOR MATCHES "Visual Studio" AND (CMAKE_C_COMPILER_ID MATCHES "MSV
 	string(APPEND CMAKE_CXX_FLAGS " /MP")
 endif()
 
-if ((CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
-		OR (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
-		OR (CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang"))
-	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -msse2 -mavx -mf16c")
```
scribam, line 55: Actually, only `-mavx -mf16c` has to be removed. SSE2 is still a requirement.

bookmist:
:( disabling  -mavx -mf16c makes f16c intrinsics impossible to use in GCC and Clang. Looks like this two compilers have 2 ways of work with instruction sets: on and it's used everywhere and off and it can't be used at all. Looks like all my code is useful only on MSVC. Common way of runtime detection with GCC or CLang is detect on install or loading different .dll/.so for different instruction sets.

scribam (approved):

PR #1099, opened 2021-01-15
Bankai libkernel module

Zangetsu38:
Split other pr with all already fine, for can finally merge it on master and can works more on it by any other dev

- This PR will address a big refactor of Vita3K.
- We will preload some user modules that are preloaded on real hw.
- That means that some vita3k user implementations will be moved to kernel ones.
- If we are lucky the number of functions needed to be implemented will be reduced.

Some already done tasks:

- interface/run app: load os0:kd/us/ modules like driver_us.suprx and libkernel.suprx 
- modules/SceAppMgrUser, SceDispayUser and SceRtcUser move all function inside kernel modules version
- implement some other functions

### DIFF
--- a/vita3k/modules/SceKernelModulemgr/SceModulemgr.cpp
+++ b/vita3k/modules/SceKernelModulemgr/SceModulemgr.cpp
@@ -151,8 +151,20 @@ EXPORT(int, sceKernelGetLibraryInfoByNID) {
     return UNIMPLEMENTED();
 }
 
-EXPORT(int, sceKernelGetModuleIdByAddr) {
-    return UNIMPLEMENTED();
+EXPORT(int, sceKernelGetModuleIdByAddr, Ptr<void> addr) {
+    KernelState *const state = &host.kernel;
+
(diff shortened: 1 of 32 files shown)


vita3k/modules/SceKernelModulemgr/SceModulemgr.cpp
```diff
@@ -16,18 +16,111 @@
...
+            error_val = RET_ERROR(SCE_ERROR_ERRNO_EINVAL);
+            return false;
+        }
+
+        if (seek_file(file, 0, SCE_SEEK_SET, host.io, export_name) < 0) {
+            error_val = RET_ERROR(static_cast<int>(size));
+            return false;
+        }
+
+        auto *data = new char[static_cast<int>(size) + 1]; // null-terminated char array
```
1whatleytay, line 52: It would be nice if this was `std::vector<char>` or something, RAII would cleanup everything nicely.

1whatleytay (approved):
Just a quick look.

1whatleytay (approved):

PR #1105, opened 2021-01-23
Fix failed vpk crash and made error popup make more sense

EXtremeExploit:


vita3k/gui/src/archive_install_dialog.cpp
```diff
@@ -166,7 +170,7 @@ void draw_archive_install_dialog(GuiState &gui, HostState &host) {
         ImGui::Spacing();
         ImGui::Separator();
         ImGui::Spacing();
-        ImGui::Checkbox("Delete the archive for the content that failed to install?", &delete_archive_file);
+        ImGui::Checkbox("Delete vpk file of the content that failed to install?", &delete_archive_file);
```
Zangetsu38, line 173: remove this, i have changed vpk to archive because is zip and vpk, so is archive file

Zangetsu38 (changes requested):

Zangetsu38 (approved):

PR #1110, opened 2021-01-24
ci: Attempt to fix move-tag job

scribam:


Zangetsu38 (approved):
if i good see is fixed fine

PR #1128, opened 2021-02-06
modules/SceGxm: Initialise background texture of color surface

pent0:
image

The game (motorstorm) uses precomputed, and use background texture to calculate the surface size and submit to display. But anyway since we don't fill it all controls words are 0 and size return 1, making it display nothing

texx

fine now


poudink:
What game?

pent0:
Motorstorm

Zangetsu38 (approved):

PR #1146, opened 2021-02-13
gxm: Implement sceGxmProgramIsEquivalent

EXtremeExploit:
* Implement sceGxmProgramIsEquivalent

### DIFF
--- a/vita3k/modules/SceGxm/SceGxm.cpp
+++ b/vita3k/modules/SceGxm/SceGxm.cpp
@@ -1597,9 +1597,12 @@ EXPORT(bool, sceGxmProgramIsDiscardUsed, const SceGxmProgram *program) {
 }
 
 EXPORT(bool, sceGxmProgramIsEquivalent, const SceGxmProgram *programA, const SceGxmProgram *programB) {
-    assert(programA);
-    assert(programB);
-    return UNIMPLEMENTED();
+    if (!programA || !programB) {
+        LOG_ERROR("SCE_GXM_ERROR_INVALID_POINTER");
+        return false;
+    }
+
+    return (programA->size == programB->size) && (memcmp(programA, programB, programA->size) == 0);
 }
 
 EXPORT(bool, sceGxmProgramIsFragColorUsed, const SceGxmProgram *program) {


vita3k/modules/SceGxm/SceGxm.cpp
```diff
@@ -1612,7 +1612,17 @@ EXPORT(bool, sceGxmProgramIsDiscardUsed, const SceGxmProgram *program) {
...
+    if (!programA)
+        LOG_ERROR("ProgramA is invalid");
+    if (!programB)
+        LOG_ERROR("ProgramB is invalid");
+
+    STUBBED("Compare programs, not output");
+    if (programA == programB)
+        return true;
+    else
+        return false;
```
Zangetsu38, line 1625:
```suggestion
    if (!programA || !programB)
        return RET_ERROR(SCE_GXM_ERROR_INVALID_POINTER);

    return programA == programB ? true : false;
```

Zangetsu38: remove also both assert

Zangetsu38 (changes requested):

vita3k/modules/SceGxm/SceGxm.cpp
EXtremeExploit: function returns bool, RET_ERROR would return nonsense there, isn't better to just log the error and return false?

vita3k/modules/SceGxm/SceGxm.cpp
```diff
@@ -1612,7 +1612,12 @@ EXPORT(bool, sceGxmProgramIsDiscardUsed, const SceGxmProgram *program) {
 EXPORT(bool, sceGxmProgramIsEquivalent, const SceGxmProgram *programA, const SceGxmProgram *programB) {
     assert(programA);
     assert(programB);
-    return UNIMPLEMENTED();
+
+    if (!programA || !programB)
+        LOG_ERROR(SCE_GXM_ERROR_INVALID_POINTER);
+
+    STUBBED("Compare programs, not output");
+    return programA == programB;
```
Zangetsu38, line 1620:
```suggestion
    return (programA->size == programB->size) && (memcmp(programA, programB, programA->size) == 0);
```

Zangetsu38: confirmed, is good implement like that, and no stub

Zangetsu38 (changes requested):

vita3k/modules/SceGxm/SceGxm.cpp
```diff
@@ -1612,7 +1612,12 @@ EXPORT(bool, sceGxmProgramIsDiscardUsed, const SceGxmProgram *program) {
 EXPORT(bool, sceGxmProgramIsEquivalent, const SceGxmProgram *programA, const SceGxmProgram *programB) {
     assert(programA);
     assert(programB);
-    return UNIMPLEMENTED();
+
+    if (!programA || !programB)
+        LOG_ERROR(SCE_GXM_ERROR_INVALID_POINTER);
```
Zangetsu38, line 1617:
```suggestion
        return RET_ERROR(SCE_GXM_ERROR_INVALID_POINTER);
```small error here

Comment (Zangetsu38) [reply vita3k/modules/SceGxm/SceGxm.cpp:1617]: it return error on log + send false

Comment (EXtremeExploit) [reply vita3k/modules/SceGxm/SceGxm.cpp:1617]: doesn't return false, only way that an int casted to bool is false is when the int is 0, so this would return true even if it is false

File: vita3k/modules/SceGxm/SceGxm.cpp (hunk 2 above)
Comment (EXtremeExploit) [reply vita3k/modules/SceGxm/SceGxm.cpp:1620]: both programs can be different primitively and have same output iirc

File: vita3k/modules/SceGxm/SceGxm.cpp (hunk 3 above)
Comment (Zangetsu38) [reply vita3k/modules/SceGxm/SceGxm.cpp:1617]: No, error value is négative, so it return false, if value is > 0 get true if value is < 1 get false

File: vita3k/modules/SceGxm/SceGxm.cpp [hunk 4]
```diff
@@ -1612,7 +1612,12 @@ EXPORT(bool, sceGxmProgramIsDiscardUsed, const SceGxmProgram *program) {
...
+
+    STUBBED("Compare programs, not output");
```
Zangetsu38, line 1619: Remove stubbed, is full implémented now

vita3k/modules/SceGxm/SceGxm.cpp
Zangetsu38: No, hete with yhis code you compare size and shader in memory directy

vita3k/modules/SceGxm/SceGxm.cpp
EXtremeExploit: http://cpp.sh/8pwas
This should prove it, as i said, booleans are only false if int is 0, if it isn't 0 it is true

Zangetsu38: Yes sorry, so using return false and after get erreur

Zangetsu38 (approved):

PR #1148, opened 2021-02-14
implement LodBias and LodMin

Zangetsu38:
thx @kd-11 for help, i hop is good
- Implement both on GXM  and set glTexParameterf in gl/texture .


vita3k/renderer/src/gl/texture.cpp
```diff
@@ -272,11 +272,19 @@ void upload_bound_texture(const SceGxmTexture &gxm_texture, const MemState &mem)
 
         glPixelStorei(GL_UNPACK_ROW_LENGTH, static_cast<GLint>(pixels_per_stride));
 
+        GLuint sampler_state = 0;
```
kd-11, line 275: This complicates things. Only image unit 0 will have the settings you set up, the others will ignore it. Since textures are constantly uploaded, just set the LODs in the texture itself. Also the object is removed before it is used for drawing making it pointless as these changes are forgotten immediately after creation.
Just use glTexParameterf for now until a proper persistent texture is in place.

kd-11 (approved):

PR #1150, opened 2021-02-15
Random GXM game fixes for the chads

pent0:
- Enable viewport by default (I have noticed a bunch of games don't set viewport mode at all)
- Only set viewport mode to enable/disable when in function sceGxmSetViewportEnable. We overwritten the viewport setting in sceGxmBeginScene, if this is set outside. Fix Haiyore rendering
- Allocate each input attribute in shader a location
    * Previously we manually queried them. I don't know if this break homebrew hand-written shader.
- Calculate attribute location using linear allocation, not by dividing register index.
    * There are some cases when the input variable in shader are not aligned by 4 floats, so for example half4 a, half4 b will have the same attribute location (0/4 and 2/4 = 0), if we use old method
    * Old method results in some attributes being disabled and overwritten.
    * => Should fix Berserk rendering


pent0:
image

Berserk video

image

Haiyore

Zangetsu38 (dismissed):

Zangetsu38 (approved):

PR #1151, opened 2021-02-16
Remove legacy compabilities with hand-written shader

pent0:
- It's a big block before, but now they are removed so I removed them.
- Sync some states again when viewport flip changes. It should help some game with some part being black (ff10), fix flood it hardware flip
- Remove sync_state function, no one uses it


frangarcj (approved):
LGTM, thanks

sunho:
Nice. I've been always wondering why we were keeping these.

PR #1152, opened 2021-02-16
gl/sync_state: fix sync_cull.

Zangetsu38:
- fix or improve render on some game

# Result:
-  Improve body and fix some transparante on Ff10 
image

- fix yellow screen
image

- fix character texture
image

### DIFF
--- a/vita3k/renderer/src/gl/sync_state.cpp
+++ b/vita3k/renderer/src/gl/sync_state.cpp
@@ -194,11 +194,11 @@ void sync_cull(GLContext &context, const GxmContextState &state) {
     switch (state.cull_mode) {
     case SCE_GXM_CULL_CCW:
         glEnable(GL_CULL_FACE);
-        glCullFace(context.viewport_flip[1] == 1.0f ? GL_FRONT : GL_BACK);
+        glCullFace(GL_BACK);
         break;
     case SCE_GXM_CULL_CW:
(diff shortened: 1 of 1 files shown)


pent0 (approved):

pent0:
Zang told me he has tested all and things work fine.

frangarcj:
@pent0 are you sure the condition needs to be removed? Maybe we just had inverted culls. GL_FRONT <---> GL_BACK

pent0:
GXM coordinates vs OpenGL coordinates are inverted. Face that is culled by default is back face, front by default is CCW. So we invert it for GXM.

For the condition yes, it seems when viewport is flipped, the vertex order in GXM is also flipped, and game handle that by itself (by default GXM games dont flip viewport, but 2 games I have seen flip viewport and also flip cull mode by itself, akiba and ff10). Homebrews always not flip viewport so nothing changed.

frangarcj:
understood...

but for clip we have to invert? any idea for the viewport on ffx?

Zangetsu38:
ffx10 is not viewport, is other thingsbefore get menu, this things take full screen, and this things can also get fullscreen depending camera move etc

pent0:
Yes for clip we have to invert the vertex and clipping only, thr cull is handled by game

frangarcj:
thanks to both of us

PR #1138, opened 2021-02-08
AvPlayer event callback

bookmist:
This help to break throught dead pause on video player. It's very rough code and need to be restructured and reformatted. But it can help with big problem in many games.
I'm asking for help for good realisation of running callbacks. Now I can't pass more when 3 parameters and put garbage in threads list. Any advices how to realyse it right.

PR #1155, opened 2021-02-17
kernel: reimplement run_callback

sunho:
This should be without erratic bug when calling lle callback function (after running one callback function it will go outside the interrupt handler) 

I used cpu instance pool to solve the problem. It grabs one cpu instance from the pool, execute the lle code on it, and return instance to the pool. If service call is made inside lle callback code, it should work fine, since it's brand new cpu instance separate from the original cpu instance of the thread (the one called something like sceNpCheckCallback) The cursed nested loop like service call -> callback -> service call -> callback -> service call also work fine. But, if the depth is higher than the volume of cpu pool, it would break. I set it to 10 for now. I think it should work fine as long as there's no infinite recursion or something.

The reason for pooling is because of unicorn instance tlb cache. The instance holds its own jit cache, and the overhead for translation is quite high. Pooling can help this. However, since we almost randomly draw an instance from the pool, the instance might not have helpful jit cache. Solving this problem is interesting further work. We can try using thread_id to draw more appropriate instance. (e.g. this instance was used by thread 3 before, so it should have jit cache helpful for lle callback execed from thread 3)

### DIFF
--- a/vita3k/app/include/app/functions.h
+++ b/vita3k/app/include/app/functions.h
@@ -23,6 +23,7 @@ struct Config;
 struct HostState;
 struct SDL_Window;
 struct ImGui_State;
+struct CPUDepInject;
 template <class T>
 class Ptr;
 class Root;
@@ -39,7 +40,7 @@ enum class AppRunType {
     Vpk,
 };
 
-bool init(HostState &state, Config &cfg, const Root &root_paths);
+bool init(HostState &state, Config &cfg, const Root &root_paths, CPUDepInject inject);
 void destroy(HostState &host, ImGui_State *imgui);
 void update_viewport(HostState &state);
 void error_dialog(const std::string &message, SDL_Window *window = nullptr);
--- a/vita3k/app/src/app_init.cpp
+++ b/vita3k/app/src/app_init.cpp
@@ -94,7 +94,7 @@ void update_viewport(HostState &state) {
     }
 }
 
-bool init(HostState &state, Config &cfg, const Root &root_paths) {
+bool init(HostState &state, Config &cfg, const Root &root_paths, CPUDepInject inject) {
     const ResumeAudioThread resume_thread = [&state](SceUID thread_id) {
         const auto thread = lock_and_find(thread_id, state.kernel.threads, state.kernel.mutex);
         const std::lock_guard<std::mutex> lock(thread->mutex);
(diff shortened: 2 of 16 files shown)


sunho:
LBP and downfall (network checking game in general) are affected by this pr. I'd appreciate if someone can test them. 

I also want to confirm that the issue related to https://github.com/Vita3K/Vita3K/pull/1138 is okay with this. @bookmist

Zangetsu38:
same for killzone, npcheckcallback, and some other stop boot after this get, i try

sunho:
downwell boots fine for me. But the graphics has regressed in master build.

Zangetsu38:
yes, i have same, render is very darkness, diffucult to see

Zangetsu38:
i see regression on video playing for Hatsune mikudiva F, i now crash on start play video

frangarcj:
I still don't get how the arguments are passed. Can you explain it?

bookmist:
> I also want to confirm that the issue related to #1138 is okay with this. @bookmist

Looks ok. Many thanks, I hoped someone rewrite this part of my code properly. Some of my old code was not cleaned, so there is the patches

` vita3k/modules/SceAvPlayer/SceAvPlayer.cpp | 1 -
 1 file changed, 1 deletion(-)

diff --git a/vita3k/modules/SceAvPlayer/SceAvPlayer.cpp b/vita3k/modules/SceAvPlayer/SceAvPlayer.cpp
index 725e472c..370b3fba 100644
--- a/vita3k/modules/SceAvPlayer/SceAvPlayer.cpp
+++ b/vita3k/modules/SceAvPlayer/SceAvPlayer.cpp
@@ -186,7 +186,6 @@ EXPORT(int, sceAvPlayerAddSource, SceUID player_handle, const char *path) {
 EXPORT(int, sceAvPlayerClose, SceUID player_handle) {
     const PlayerPtr &player_info = lock_and_find(player_handle, host.kernel.players, host.kernel.mutex);
     run_event_callback(host, thread_id, player_info, SCE_AVPLAYER_STATE_STOP, 0, Ptr<void>(0));
-    assert(player_info->callback_depth_counter == 0);
     host.kernel.players.erase(player_handle);
     return 0;
 }`

` vita3k/kernel/include/kernel/state.h | 2 --
 1 file changed, 2 deletions(-)

diff --git a/vita3k/kernel/include/kernel/state.h b/vita3k/kernel/include/kernel/state.h
index 08d136b0..fb12a879 100644
--- a/vita3k/kernel/include/kernel/state.h
+++ b/vita3k/kernel/include/kernel/state.h
@@ -234,8 +234,6 @@ struct PlayerInfoState {
     SceAvPlayerMemoryAllocator memory_allocator;
     SceAvPlayerFileManager file_manager;
     SceAvPlayerEventManager event_manager;
-    SceUID player_callback_thread_id = SCE_KERNEL_ERROR_ILLEGAL_THREAD_ID;
-    uint32_t callback_depth_counter = 0;
 };
 
 typedef std::shared_ptr<PlayerInfoState> PlayerPtr;
`

bookmist:
> 
> 
> I still don't get how the arguments are passed. Can you explain it?

Arguments passed in first 4 registers of CPU. Fifth and more arguments should be puched into stack, but i don't know how to do it.

Zangetsu38 (approved):

PR #1168, opened 2021-02-21
gxm: Implement untype attribute format

EXtremeExploit:
untyped attribute formats are simply 32 bit untyped data, GL_INT is 32 bit so it is ok to use it here


pent0 (changes requested):
Upload data with GL_INT type requires to use glVertexAttribIPointer, you should add a case [here](https://github.com/Vita3K/Vita3K/blob/master/vita3k/renderer/src/gl/sync_state.cpp#L368) too

pent0 (approved):

PR #1170, opened 2021-02-21
vita3k: Some refactor on compiler warnings

EXtremeExploit:
Gets rid of some compiler warnings and corrects type of sceKernelGetTimerBaseWide

### DIFF
--- a/vita3k/modules/SceGxm/SceGxm.cpp
+++ b/vita3k/modules/SceGxm/SceGxm.cpp
@@ -189,7 +189,7 @@ static int init_texture_base(const char *export_name, SceGxmTexture *texture, Pt
     if (texture_type == SCE_GXM_TEXTURE_SWIZZLED) {
         // Find highest set bit of width and height. It's also the 2^? for width and height
         static auto highest_set_bit = [](const int num) -> std::uint32_t {
-            for (std::uint32_t i = 12; i >= 0; i--) {
+            for (std::uint32_t i = 12; i != 0; i--) {
                 if (num & (1 << i)) {
                     return i;
                 }
--- a/vita3k/gui/src/imgui_impl_sdl.cpp
+++ b/vita3k/gui/src/imgui_impl_sdl.cpp
@@ -176,6 +176,7 @@ IMGUI_API ImTextureID ImGui_ImplSdl_CreateTexture(ImGui_State *state, void *data
 #endif
     default:
         LOG_ERROR("Missing ImGui init for backend {}.", static_cast<int>(state->renderer->current_backend));
+        return (void *)0;
     }
 }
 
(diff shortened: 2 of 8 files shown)


vita3k/modules/SceGxm/SceGxm.cpp
```diff
@@ -189,7 +189,7 @@ static int init_texture_base(const char *export_name, SceGxmTexture *texture, Pt
     if (texture_type == SCE_GXM_TEXTURE_SWIZZLED) {
         // Find highest set bit of width and height. It's also the 2^? for width and height
         static auto highest_set_bit = [](const int num) -> std::uint32_t {
-            for (std::uint32_t i = 12; i >= 0; i--) {
+            for (std::uint32_t i = 12; true; i--) {
```
Zangetsu38, line 192: True ?

EXtremeExploit: i is an unsigned int so it is impossible to have "i >= 0" be false

Zangetsu38: Yes but true is totaly wrong, is loop, do need condition relate i, is more i == 0

EXtremeExploit: the break of the loop is at line [194](https://github.com/Vita3K/Vita3K/pull/1170/files#diff-f2f7f9a543fa0fce25491627413b0513bff2432dd8deeaf900cd968997881ce7R194), the return breaks the loop

Zangetsu38: I see but true like that no have sense for look using like is already is fine

EXtremeExploit: its the goal the PR, getting rid of some compiler warnings and make code easier to understand

Zangetsu38: Oki so change too == 0 is more understund true here is wrong

EXtremeExploit: true its literally the same behaviour, easier to understand how the loop works and also gets rid of the compiler warning at that line, true isn't wrong here, its literally the best option you can put there

Zangetsu38: No is wrong true no have sense in loop, hat is infinité loop, and sorry i have mistake is more need set "i /= 0"

EXtremeExploit: it already was an infinite loop and thats the goal of that loop, it is called "highest_set_bit"

Zangetsu38, line 192:
```suggestion
            for (std::uint32_t i = 12; i != 0; i--) {
```

Zangetsu38: change like that confirmed good on my test

Zangetsu38 (changes requested):

Zangetsu38 (approved):

PR #1198, opened 2021-03-19
modules/SceMotion: stub GetBasicOrientation.

Zangetsu38:
fix orientation change alone on pinball aracade.


vita3k/modules/SceMotion/SceMotion.cpp
```diff
@@ -21,8 +21,16 @@ EXPORT(int, sceMotionGetAngleThreshold) {
...
+EXPORT(int, sceMotionGetBasicOrientation, SceFVector3 *basicOrientation) {
+    if (!basicOrientation)
+        return RET_ERROR(SCE_MOTION_ERROR_NULL_PARAMETER);
+
+    STUBBED("set default value");
+    basicOrientation->x = 0;
+    basicOrientation->y = 0;
+    basicOrientation->z = 1;
+
+    return SCE_OK;
```
scribam, line 33:
```suggestion
    return 0;
```

vita3k/modules/SceMotion/SceMotion.h
```diff
@@ -19,6 +19,21 @@
 
 #include <module/module.h>
 
+enum SceMotionErrorCode {
+    SCE_OK = 0,
```
scribam, line 23: Remove this line

scribam (changes requested):
The constant `SCE_OK` is never used, I think `0` is preferred.

vita3k/modules/SceMotion/SceMotion.cpp
Zangetsu38: is done for both

scribam (approved):

PR #1209, opened 2021-04-01
emu: add can switch/exit game.

Zangetsu38:
using execv previously implemented for can using this.
- Add ability to can switch/exit  game,  for moment, and like for LoadExec, it restart emu.


frangarcj (approved):

Croden1999:
sir, i want request about this issue, the Vita3K emulator just only switch/exit once game, if switch/exit game on second times, and then crash.

PR #1215, opened 2021-04-05
audio: Programmatic audio control

tictakk:
Implementing sceAudioOutSetVolume to allow software to adjust volume

This turns out to be a little more complicated than expected. Since audio relies on the **SDL_MixAudio** function, that function takes a volume param and applies the volume when mixing the samples. But the sceAudioOutSetVolume function can specify which channel it wants to lower volume on, which you cannot do with SDL_MixAudio. What I did in this PR is average the left and right channel volumes together and applied that to the **SDL_MixAudio** function, which I'm assuming we're not going to want to do. 

Possible solutions:
Apply the volume to the samples in the streams before mixing, but then we're running essentially the same function twice over the same set of sample data, also we'd have to account for every possible format the sample data can be received in.

Mix our own audio, which wouldn't be too hard, just mimicking what SDL_MixAudio does, but accounting for individual channels. 

I may be making this way more complicated than it is an there's an obvious solution that I am missing or don't know about.

### DIFF
--- a/vita3k/audio/include/audio/state.h
+++ b/vita3k/audio/include/audio/state.h
@@ -27,6 +27,9 @@
 #include <mutex>
 #include <vector>
 
+#define SCE_AUDIO_OUT_MAX_VOL 32768 //!< Maximum output port volume
+#define SCE_AUDIO_VOLUME_0DB SCE_AUDIO_OUT_MAX_VOL //!< Maximum output port volume
+
 typedef std::shared_ptr<SDL_AudioStream> AudioStreamPtr;
 typedef std::function<void(SceUID)> ResumeAudioThread;
 
@@ -43,6 +46,11 @@ struct SharedAudioOutPortState {
 struct AudioOutPort {
     ReadOnlyAudioOutPortState ro;
     SharedAudioOutPortState shared;
+    // Channel range from 0 - 32768
+    int left_channel_volume = SCE_AUDIO_VOLUME_0DB;
+    int right_channel_volume = SCE_AUDIO_VOLUME_0DB;
+    // Volume range from 1 - 128
+    float volume = SDL_MIX_MAXVOLUME;
 };
 
 struct AudioInPort {
--- a/vita3k/audio/src/audio.cpp
+++ b/vita3k/audio/src/audio.cpp
@@ -52,7 +52,7 @@ static void mix_out_port(uint8_t *stream, uint8_t *temp_buffer, int len, AudioOu
     const int bytes_got = SDL_AudioStreamGet(port.shared.stream.get(), temp_buffer, bytes_to_get);
     lock.unlock();
     if (bytes_got > 0) {
-        SDL_MixAudio(stream, temp_buffer, bytes_got, SDL_MIX_MAXVOLUME);
+        SDL_MixAudio(stream, temp_buffer, bytes_got, port.volume);
     }
 }
 
--- a/vita3k/modules/SceAudio/SceAudio.cpp
+++ b/vita3k/modules/SceAudio/SceAudio.cpp
@@ -49,6 +49,11 @@ enum SceAudioOutErrorCode {
     SCE_AUDIO_OUT_ERROR_OUT_OF_MEMORY = 0x8026000D
 };
 
+enum SceAudioOutChannelFlag {
+    SCE_AUDIO_VOLUME_FLAG_L_CH = 1, //!< Left Channel
+    SCE_AUDIO_VOLUME_FLAG_R_CH = 2 //!< Right Channel
+};
+
 EXPORT(int, sceAudioOutGetAdopt) {
     return UNIMPLEMENTED();
 }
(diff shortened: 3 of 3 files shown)


vita3k/audio/include/audio/state.h
```diff
@@ -43,6 +43,7 @@ struct SharedAudioOutPortState {
 struct AudioOutPort {
     ReadOnlyAudioOutPortState ro;
     SharedAudioOutPortState shared;
+    float volume = 100.0;
```
pent0, line 46: If this is intended to br from 0.0f to 1.0f make it 1.0f by default

vita3k/audio/src/audio.cpp
```diff
@@ -50,9 +50,10 @@ static void mix_out_port(uint8_t *stream, uint8_t *temp_buffer, int len, AudioOu
     // Mix as much as we need.
     const int bytes_to_get = std::min(len, bytes_available);
     const int bytes_got = SDL_AudioStreamGet(port.shared.stream.get(), temp_buffer, bytes_to_get);
+    const int vol = SDL_MIX_MAXVOLUME * port.volume;
```
pent0, line 53:
```suggestion
    const int vol = static_cast<int>(SDL_MIX_MAXVOLUME * port.volume);
```

pent0: Warning if no cast

vita3k/modules/SceAudio/SceAudio.cpp
```diff
@@ -164,8 +164,16 @@ EXPORT(int, sceAudioOutSetPortVolume_forUser) {
     return UNIMPLEMENTED();
 }
 
-EXPORT(int, sceAudioOutSetVolume) {
-    return UNIMPLEMENTED();
+EXPORT(int, sceAudioOutSetVolume, int port, int channelFlags, int *vol) {
+    // 32768 is max volume
+    int left = *vol;
```
pent0, line 169: Also check channelFlags, there might be instances where only one channel is to be set

pent0:
```suggestion
    const int left = vol[0];
```

vita3k/modules/SceAudio/SceAudio.cpp
```diff
@@ -164,8 +164,16 @@ EXPORT(int, sceAudioOutSetPortVolume_forUser) {
...
+    int right = *vol++;
```
pent0, line 170: Same

vita3k/modules/SceAudio/SceAudio.cpp
```diff
@@ -164,8 +164,16 @@ EXPORT(int, sceAudioOutSetPortVolume_forUser) {
...
+    float volume = ((float)((left + right) / (float)32768));
```
pent0, line 171: Use static cast

pent0: You can import 32768 constant and the channel flags by including vitasdk header

vita3k/audio/include/audio/state.h
pent0: Make note also of its range

pent0 (changes requested):

tictakk:
@pent0  added the notes and checks for flags. I also decided to keep the state of each channel so if the program only adjusts one channel, we still have data on what the other channel is (since they will be combined and used). I also made it do all the calculations in the **sceAudioOutSetVolume** function so the volume calculation isn't done each time mix_audio is called.

vita3k/modules/SceAudio/SceAudio.cpp
```diff
@@ -49,6 +49,11 @@ enum SceAudioOutErrorCode {
     SCE_AUDIO_OUT_ERROR_OUT_OF_MEMORY = 0x8026000D
 };
 
+typedef enum SceAudioOutChannelFlag {
```
Zangetsu38, line 52: Remove typedef

vita3k/modules/SceAudio/SceAudio.cpp
```diff
@@ -49,6 +49,11 @@ enum SceAudioOutErrorCode {
...
+    SCE_AUDIO_VOLUME_FLAG_L_CH = 1, //!< Left Channel
+    SCE_AUDIO_VOLUME_FLAG_R_CH = 2 //!< Right Channel
+} SceAudioOutChannelFlag;
```
Zangetsu38, line 55: Remove this just keep };

Zangetsu38 (changes requested):
For push change just édit last commit and force push you can also squash your commit here

vita3k/modules/SceAudio/SceAudio.cpp
```diff
@@ -164,8 +169,23 @@ EXPORT(int, sceAudioOutSetPortVolume_forUser) {
...
-EXPORT(int, sceAudioOutSetVolume) {
-    return UNIMPLEMENTED();
+EXPORT(int, sceAudioOutSetVolume, int port, SceAudioOutChannelFlag ch, int *vol) {
+    if (!ch) //no channel selected, no changes
+        return 0;
+
+    const AudioOutPortPtr prt = lock_and_find(port, host.audio.shared.out_ports, host.audio.shared.mutex);
+
+    const int left = (ch & SCE_AUDIO_VOLUME_FLAG_L_CH) ? vol[0] : prt->left_channel_volume;
+    const int right = (ch & SCE_AUDIO_VOLUME_FLAG_R_CH) ? vol[1] : prt->right_channel_volume;
```
pent0, line 179: I'm not sure about this line, maybe make a note in case we encounter regression

pent0: Reason is maybe if only R_CH is available, volumes array size is only 1? Im not sure, but we have not seen this yet

pent0 (approved):

Zangetsu38 (approved):

pent0:
Zang's requests are complimented

PR #1235, opened 2021-04-16
lang: Fix English weekday capitalisation, add English (United Kingdom) language file

Yunotchi:
In English, weekdays are considered proper nouns and therefore its first letter is capitalised. This is the case on the Vita itself too:

I also added an English (United Kingdom) lang file (`en-gb.xml`). Currently, this is identical to English (United States) (`en.xml`) -- I couldn't see any changes to make here.


Zangetsu38:
you have write, rco and psvita confirm me that, no idea why, i have see in french is like that so, but for other languague i have take day in clear so is fine, i have only changed apital for english by thinks is like french

Zangetsu38 (approved):

PR #1241, opened 2021-04-19
util: Use std::optional for C++17 builds

totlmstr:
Subset of #660. Done for easier merging.

This adds in the ability to compile Vita3K as a C++17 program. C++17 compilation will be used by default, but we still need Boost for the rest of the program (such as `psvpsfstools`), so `external/boost` won't be modified in this PR for safety.

~~Note that the minimum version requirement of Boost version 1.72 is for `boost::optional::has_value()` (which is effectively the same as `boost::optional::is_initialized()`).~~

Edit: I originally had the Boost requirement as 1.72, but that will only be built by the user, the CI will never need to use that, and APT is *for some absurd reason* still stuck on 1.71. Since we already have a local Boost at 1.72 and C++17, this requirements seems to be fully satisfied.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,7 @@ macro(pre_configure_boost)
 	set (Boost_USE_STATIC_LIBS ON)
 	find_package(Boost COMPONENTS filesystem system program_options QUIET)
 
-	if (Boost_FOUND)
+	if (${Boost_FOUND})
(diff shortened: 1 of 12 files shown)


pent0 (approved):

pent0:
This is the way, Boost is complicated the build process pretty much

PR #1244, opened 2021-04-23
modules/SceGxm: Implement SetVertexTexture.

Zangetsu38:


Zangetsu38:
Oups...  arg need more things

sunho:
You shouldn't set fragment_texture there. It will collide with fragment shader.

Zangetsu38:
yes i know lol need create value etc for vertex_texture i ask bent for that

sunho:
Isn't this going to be problematic if vertex texture index and fragment texture index are identical? e.g. set vertex texture 1 and set fragment texture 1 at the same time.

Zangetsu38:
normally  no, it using totaly diferent variable for store texture and for execute it

sunho:
It will overlap even though we use different variable. Because index used in "glActiveTexture" will be used to store & bind texture.

sunho:
I also realized that we need to tweak the texture index inside glsl product in that case.

Zangetsu38:
hmm for boçth is GL_TEXTURE0 yes, but normally, and i have check, it never run few texture in same time so probably fine

pent0:
Need to find where vertex shader uses a texture, then I can verify this PR

bookmist:
This textures are set in "ar nosurge: ode to unborn star", but i dont know are them used in shaders. Is recompiling texture fetch in vertex shader implemented? (Are vertex textures used in vertex shaders? right?)

Zangetsu38:
wipeout HD using it, and extract texture works, i get some texture for one animation

pent0:
Yes, but you see, there is nothing like that in GLSL

vita3k/renderer/src/gl/sync_state.cpp
```diff
@@ -312,7 +312,51 @@ void sync_texture(GLContext &context, const GxmContextState &state, const MemSta
...
+        LOG_WARN("Texture has null data.");
+        return;
+    }
+
+    const SceGxmTextureFormat format = gxm::get_format(&texture);
+    if (gxm::is_paletted_format(format) && texture.palette_addr == 0) {
+        LOG_WARN("Ignoring null palette texture");
+        return;
+    }
+    glActiveTexture(static_cast<GLenum>(static_cast<std::size_t>(GL_TEXTURE0) + index));
```
kd-11, line 335: Won't work, you have to rebind your vertex texture indices so that they do not overlap fragment texture indices. e.g vertex texture index 0 cannot have a different texture than fragment texture index 0 in GL because the texture indices are the hw resource being shared. You should remap to something like GL_TEXTURE16 + index (assuming 16 fragment texture slots in vita hw) This is what I meant by extra considerations on shader validation in our previous chat.

Zangetsu38: i see 🤔

kd-11: FYI this also applies to shader bindings, image units cannot overlap or you get invalid shader on compliant drivers. The shader may compile and link but mysteriously fail to render textures on some hw due to TIU aliasing rules. See https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glValidateProgram.xhtml

sunho: I read that the maximum texture unit 16 in opengl. Is GL_TEXTURE16 + index going to be fine even in binding process?

pent0 (approved):

PR #1246, opened 2021-04-23
gui: improve IME keyboard.

Zangetsu38:
On my first pr i have fovus my works on virtual keyboard working, here is more complete and visual improvement design.
- refactor code for can adding numeric pad.
- refactor visuel for get more close on psvita aspect
- remove space in margin side with remove child.
- add change relate ponctuation if is shift or not
- reworks move carret for improve preedit.
- Add select keyboard with choice,

# Result: 
image

- French Keyboard
image
- choice of select language
image

- current keyèboard support
image


Zangetsu38:
if any dev can help me here
i get crash with thy ussing all char u8
in moment it convert string to str16 i crash
image
image
relate this all char u8
image

Zangetsu38:
Comment unicode text for moment, any dev can try fix itself later

vita3k/config/include/config/config.h
```diff
@@ -86,14 +86,16 @@
     code(bool, "show-welcome", true, show_welcome)                                                      \
     code(bool, "asia-font-support", false, asia_font_support)                                           \
     code(bool, "video-playing", true, video_playing)                                                    \
-    code(bool, "spirv-shader", false, spirv_shader)
+    code(bool, "spirv-shader", false, spirv_shader)                                                     \
+    code(uint64_t, "current-ime-lang", 4, current_ime_lang)
```
pent0, line 90: Is uint64_t really necessary, maybe uint32_t

vita3k/gui/src/ime.cpp
```diff
@@ -42,129 +74,391 @@ enum CapsLevel {
...
+    update_str(ime, key);
+    reset_preedit(ime);
+}
+
+static std::map<int, float> lang_keyboard_pos;
+static std::map<int, std::vector<std::u16string>> lang_key, shift_lang_key, ponct;
+static std::string space_str;
+
+void init_ime_lang(Ime &ime, const SceImeLanguage &lang) {
+    // Set Ponctuation
```
pent0, line 112: Punctuation

vita3k/gui/src/ime.cpp
```diff
@@ -42,129 +74,391 @@ enum CapsLevel {
...
-        if (host.ime.edit_text.preeditIndex > host.ime.edit_text.caretIndex)
-            host.ime.edit_text.preeditIndex = host.ime.caretIndex = host.ime.edit_text.caretIndex;
-        host.ime.edit_text.editLengthChange = -1;
-        if (host.ime.edit_text.preeditLength > 0)
-            --host.ime.edit_text.preeditLength;
-        host.ime.event_id = SCE_IME_EVENT_UPDATE_TEXT;
-        if (host.ime.str.empty() && (host.ime.caps_level == NO))
-            host.ime.caps_level = YES;
+    // Set Space String
+    switch (lang) {
```
pent0, line 135: Maybe a map is better for this

pent0 (changes requested):

vita3k/config/include/config/config.h
Zangetsu38: number can be get very big i have used value set for supportedlanguage inside param, is uint64_t, and is this value can give it of lang

vita3k/gui/src/ime.cpp
Zangetsu38: no understund map ? i ussing diferent switch because for some thinks is same other and for some is diferent

sunho:
By the way, I think we should just properly support ime seriously. We can't implement virtual keyboard for every language. How are you going to deal with Chinese, Korean, or Japanese for instance?

Zangetsu38:
> 
> 
> By the way, I think we should just properly support ime seriously. We can't implement virtual keyboard for every language. How are you going to deal with Chinese, Korean, or Japanese for instance?

i already thinks for other languague, herte is just improvement, and no complete, complete for japan and, chinese is hardjap have 2 keyboard, i can set only one for now and chinese have 3 same ^^

sunho:
image
You gotta implement something like this. Combinating characters. (ㄱ+ㅏ+ㄱ+ㅅ=갃, ㅇ+ㅏ=아) for korean. For chinese, we must ship a database for roman chracter to chinese chracter.

Zangetsu38:
hmm no sur, but for now, iu just want merge this ^^ i continue later

sunho:
Japanese is even more complicated. 
image
If you type in some chracters, "candidate outputs" show up. Sometimes a mix of hiragana and kanji.

Zangetsu38:
i want refer and write maximum like on real hw i set all like that

sunho:
Korean keyboard on real hw btw is the same as what I mentioned. We can't deal with this complications. You took some efforts so we can keep what you have done till now. But, we can't keep working on virtual keyboard things. It's just so much works.

Zangetsu38:
last- works adding full russian keyboar dand switch Russian/Qwerty keyboard
image
image

pent0 (approved):

PR #1247, opened 2021-04-23
Some very minor refactorings, improvements and compiler warnings fixes

bookmist:
Not sure in "app/screen_render: add stensil clear before draw"
and "TLS init fix" but tls init fix is better then before (but may be still buggy)

### DIFF
--- a/vita3k/kernel/src/thread/thread.cpp
+++ b/vita3k/kernel/src/thread/thread.cpp
@@ -116,14 +116,15 @@ SceUID create_thread(Ptr<const void> entry_point, KernelState &kernel, MemState 
 
     alloc_name = fmt::format("TLS for thread {} (#{})", name, thid);
 
-    auto tls_address = alloc(mem, 0x800, alloc_name.c_str()) + 0x800;
+    constexpr auto tls_size = 0x800;
+    auto tls_address_ptr = Ptr<void>(alloc(mem, tls_size, alloc_name.c_str()));
+    auto tls_address = tls_address_ptr.address() + tls_size;
 
     write_tpidruro(*thread->cpu, tls_address);
 
-    Ptr<void> tls_address_ptr(tls_address);
-    memset(tls_address_ptr.get(mem), 0, 0x800);
+    memset(tls_address_ptr.get(mem), 0, tls_size);
     if (kernel.tls_address) {
-        assert(kernel.tls_psize <= 0x800 / 4);
+        assert(kernel.tls_psize <= tls_size / 4);
         memcpy(tls_address_ptr.get(mem), kernel.tls_address.get(mem), 4 * kernel.tls_psize);
     }
     WaitingThreadState waiting{ name };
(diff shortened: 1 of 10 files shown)


vita3k/kernel/src/thread/thread.cpp
```diff
@@ -116,11 +116,11 @@ SceUID create_thread(Ptr<const void> entry_point, KernelState &kernel, MemState
 
     alloc_name = fmt::format("TLS for thread {} (#{})", name, thid);
 
-    auto tls_address = alloc(mem, 0x800, alloc_name.c_str()) + 0x800;
+    Ptr<void> tls_address_ptr = Ptr<void>(alloc(mem, 0x800, alloc_name.c_str()));
```
Zangetsu38, line 119: you can also using auto here, you using Ptr for intial value so is good on auto too :)

Zangetsu38:
no take your head, jun run format :D and set format change inside commit relate :)

sunho:
Can you tell me little more about tls init fix?

vita3k/app/src/screen_render.cpp
```diff
@@ -102,7 +102,7 @@ void gl_screen_renderer::render(const HostState &host) {
     glViewport(static_cast<GLint>(host.viewport_pos.x), static_cast<GLint>(host.viewport_pos.y), static_cast<GLsizei>(host.viewport_size.x),
         static_cast<GLsizei>(host.viewport_size.y));
 
-    glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
+    glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);
```
pent0, line 105: We don't use stencil in main framebuffer, so it's no need. What is the reason?

bookmist: I had yellow and blue screen on loading ar nosurge game. At some time this change and maybe some more half-random changes replace yellow to black (that, I think, is more like original behavior). Now i rechecked this and see that this change only has no significant effect.
Thank you very much for explaining, I'll revert this commit.

bookmist:
> Can you tell me little more about tls init fix?

About tls_fix. Before my commit emu did the following:
alloc 0x800 bytes of memory
set pointer to the end of allocated memory
memset from this pointer (END of allocated mem) 0x800 bytes to zero

I changed this to memset from BEGIN of allocated mem.

I'm not sure should I init this memory with kernel.tls_address* and should I allocate 0x800 bytes or kernel.tls_msize bytes.
Also maybe function get_thread_tls_addr should be simplified and kernel.tls field can be deleted.

sunho:
You are totally right. Nice find!

If kernel.tls_address is not empty, we should memcopy from the BEGINNING of allocated memory. And, we should allocate 4* kernel.tls_msize in stead of 0x800 in this case. 

I agree that we can just simplify get_tls_address. kernel.tls field is merely serving as cache. No need for caching this kind of thing.

sunho:
As I didn't write that code on tls, I'll ask frangarj who wrote that code just in case to confirm. But, it's fairly clear we should do that from beginning of that area looking at old code.

sunho (approved):
We should merge this soon. Faulty tls is a huge deal.

sunho:
@bookmist tell me if I can merge this.

bookmist:
@sunho Yes, of course.

PR #1250, opened 2021-04-25
Fix eventflag wait

sunho:
Fixes to love ru visual novel game booting.


pent0 (approved):
From logic point totally cool

PR #1253, opened 2021-04-28
Add Texture format handle support.

Zangetsu38:
- Add support for X8D24 depth-stencil format read.
- Add support for F32M.

# Result:
- fix crash on boot for Urban trial and get menu with 3d font render (yellow issue)
image
image


pent0 (approved):
Lezit

PR #1255, opened 2021-04-30
Shader fix

Zangetsu38:
- fix shader crash on god of war 1 & 2 for can get ingame on both
- fix other shader crash with fix generate shader all on atelier game, MS RC, Accel >World vs Sao and more

# Result: 
- God of war 1
image
image
image
image

- God of war 2
image
image
image


sunho (approved):
Some room for cleaner code but looks fine. Do FP16 and UINT16 use a same set of constants?

sunho:
Wait how come op.type be uint there. Does int instructions specify fpconstant register?

pent0:
Yes. Its a bitwise instruction specified this. VBW

PR #1257, opened 2021-04-30
kernel: Fix bug and memory leak in thread local storages. Need help.

bookmist:
It works, but hard to understand and not much c++ style. If someone can help me with code style and self explaining code, it would be good.
Fix Call of Duty crash


vita3k/kernel/src/thread/thread.cpp
```diff
@@ -117,16 +117,18 @@ SceUID create_thread(Ptr<const void> entry_point, KernelState &kernel, MemState
 
     alloc_name = fmt::format("TLS for thread {} (#{})", name, thid);
 
-    constexpr auto tls_size = 0x800;
-    auto tls_address_ptr = Ptr<void>(alloc(mem, tls_size, alloc_name.c_str()));
-    auto tls_address = tls_address_ptr.address() + tls_size;
+    auto tls_size = kernel_tls_size + kernel.tls_msize;
```
sunho, line 120: What happens if you just use tls_msize when its not zero?

bookmist: Maybe it's the same tls, but I think it's unlikely.

sunho: Oh cool that sounds compelling. I want to check out more about this but I have a feeling youve gotten there.

vita3k/kernel/src/kernel.cpp
```diff
@@ -27,17 +27,12 @@
...
 Ptr<Ptr<void>> get_thread_tls_addr(KernelState &kernel, MemState &mem, SceUID thread_id, int key) {
-    SlotToAddress &slot_to_address = kernel.tls[thread_id];
-
-    const SlotToAddress::const_iterator existing = slot_to_address.find(key);
-    if (existing != slot_to_address.end()) {
-        return existing->second;
+    Ptr<Ptr<void>> address(0);
+    //magic numbers taken from decompiled source. There is 0x400 unused bytes of unknown usage
+    if (key <= 0x100) {
+        const ThreadStatePtr thread = util::find(thread_id, kernel.threads);
```
sunho, line 33: better to log when key is more than 0x100. That means things are off.

vita3k/kernel/src/thread/thread.cpp
```diff
@@ -117,16 +117,18 @@ SceUID create_thread(Ptr<const void> entry_point, KernelState &kernel, MemState
...
+    auto base_tls_address_ptr = Ptr<uint8_t>(alloc(mem, tls_size, alloc_name.c_str()));
+    memset(base_tls_address_ptr.get(mem), 0, tls_size);
+    thread->tls = std::make_shared<Resource<Address>>(base_tls_address_ptr.address(), stack_deleter);
```
sunho, line 123: Don't think making this shared is good idea. Just using normal Resource should work just fine. Even if it doesn't work we better make this unique_ptr, because the owner is apparent.

sunho: If Resource is not used by many places, we can just get rid of its copy constructor. That way we can avoid dealing with hassles of return value optimizations etc.

sunho: I'll dig into this Resource thing and pr maybe tommorow. I have some ideas.

sunho:
btw code is quite self explantory for me. We don't really have some style guideline yet. So as long as code is readable and not out of place, it's fine...

sunho:
You can try std::copy instead of memcpy if you want.

vita3k/kernel/src/thread/thread.cpp
bookmist: Yes, but I don't know how to write it working. changing to 
struct ThreadState {
    ....
    Resource<Address> tls;
};
////
    thread->tls = Resource<Address>(base_tls_address_ptr.address(), stack_deleter);
does not compile

bookmist:
Two tls was hard to found and understand. And very different. Maybe we should use two different pointers and memory areas for them?

sunho:
I'm fairly convied about two types of tls. They match with what i have observed. If they are indeed differnt areas we can just alloc two areas as you said.

bookmist:
Now it's like allocating one memory area and use different parts of it for different purposes. More optimised, but less comprehensive. 
And also this tls field is used only for memory free on ThreadState destructor. register tpidruro is used for access

vita3k/kernel/src/thread/thread.cpp
sunho: https://github.com/Vita3K/Vita3K/pull/1258

Here we go

bookmist:
Thanks sunho for Block type. I use it and now PR is ready. Thanks for help.

vita3k/kernel/src/thread/thread.cpp
```diff
@@ -102,16 +102,18 @@ SceUID create_thread(Ptr<const void> entry_point, KernelState &kernel, MemState
...
-    write_tpidruro(*thread->cpu, tls_address);
+    auto user_tls_address_ptr = base_tls_address_ptr + kernel_tls_size;
+
+    write_tpidruro(*thread->cpu, user_tls_address_ptr.address());
 
-    memset(tls_address_ptr.get(mem), 0, tls_size);
     if (kernel.tls_address) {
-        assert(kernel.tls_psize <= tls_size / 4);
-        memcpy(tls_address_ptr.get(mem), kernel.tls_address.get(mem), 4 * kernel.tls_psize);
+        assert((kernel.tls_psize = 0xcccccccc ? 0 : kernel.tls_psize) <= (kernel.tls_msize = 0xcccccccc ? 0 : kernel.tls_msize));
```
sunho, line 115: ==

bookmist: Oh. My bad. Tnx.

bookmist: Oh. My bad. Tnx.

sunho (approved):
Looks fine but we need to keep track of this issue further.

Zangetsu38:
Issue regression of call of is keep no fixed, but just better and no unicorn crash

bookmist:
@sunho PR can be merged now. No more changes, just recommit to rerun failed check.
About call of: reverting my changes does not helps much. It still freese or crash but a bit later, so i can see menu.

sunho (approved):
Alright but can you keep trying to fix call of duty?

PR #1260, opened 2021-05-02
gui: fix and refactor dpi scale/font scale.

Zangetsu38:
# About 
After have see some strange things, like font is set for scale on dpi but ,after divide on this dpi get so no scaled, i have so reworks it completely.
- settings dialog: fix lle kernel for custom config.

# Result: 
Arrow added in app selector/live area + texte centered on column
image


sunho (approved):
This looks renaming variables to me. Well, I guess that can be considered as refactor.

Zangetsu38:
> This looks renaming variables to me. Well, I guess that can be considered as refactor.

Yes i have renaming one value but is not only this i have totaly refactor code for font scale and also d'or scale is totaly diferent before

pent0 (approved):

PR #1271, opened 2021-05-07
ci/workflows: Add note about release date

pent0:
For outsiders not confusing about it


sunho (approved):
I like the fact you preserved laziness.

PR #1273, opened 2021-05-08
gui/settings dialog: Add cpu backend choice.

Zangetsu38:
# About
- Add cpu backend choice in setting dialog.
- Add cpu backend used in log.
- reafctor cpu backend in app init

# Result:
image

### DIFF
--- a/vita3k/app/src/app_init.cpp
+++ b/vita3k/app/src/app_init.cpp
@@ -117,6 +117,8 @@ bool init(HostState &state, Config &cfg, const Root &root_paths) {
         state.pref_path = string_utils::utf_to_wide(state.cfg.pref_path);
     }
 
+    state.kernel.cpu_backend = state.cfg.cpu_backend == "Dynarmic" ? CPUBackend::Dynarmic : CPUBackend::Unicorn;
+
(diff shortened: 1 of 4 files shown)


vita3k/host/include/host/state.h
```diff
@@ -79,6 +79,7 @@ struct HostState {
     uint32_t fps = 0;
     uint32_t ms_per_frame = 0;
     WindowPtr window = WindowPtr(nullptr, nullptr);
+    CPUBackend cpu_backend;
```
sunho, line 82: No please don't do this

Zangetsu38: why ? i need this

Zangetsu38: > 
> 
> Just use host.kernel.cpu_backend

oh hmm i check sorry

Zangetsu38: fixed sorry, i have don't see this value exist

sunho (approved):
Thanks for doing this. Love your ui.

There's a little bug that the backend change for callback pool apply after restarting emu. I'll deal with this in another pr.

PR #1286, opened 2021-05-10
gui: Add dialog warning for app close.

Zangetsu38:
# About:
- Adding dialog warning for can close app if one running.
- Improve update app list openend more psvitya close, app already open is move in first place, and for one app run if is last app in list, it erase just before it.

# Result:
image

### DIFF
--- a/lang/en-gb.xml
+++ b/lang/en-gb.xml
@@ -131,6 +131,10 @@ Do not power off the system or close the application.</warning_saving>
 	</save_data>
 </dialog>
 
+<game_data>
+	<app_close>The following application will close.</app_close>
+</game_data>
+
 <indicator>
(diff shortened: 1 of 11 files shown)


Zangetsu38:
Sorry but you hazve no understund, i add only warning dialog for can launch othyer game, nothing more, nothing low code is same before

Zangetsu38:
> 
> 
> Ha.. alright. I'll remove the hacks when we get proper host clean up. At least, UI's looking good so it's not the end of the world to keep this.

i hope lol, i have said to fran needc correct, but have said, and right, for now using this hack for LoadExec, after i have used this code for can just switch app, iis very usefull lol
i want too can find how pause in open live area like on psvita with all thread pause i thinks, and back on game resume all like before

PR #1298, opened 2021-05-11
modules/SceLibKernel: Better handling of thread exit status

bookmist:
Not sure is it used anywhere.


bookmist:
Sorry, I was stupid. Thank you very much for this explaining.

PR #1302, opened 2021-05-12
renderer: Implement sceGxmNotificationWait

sunho:


Zangetsu38 (approved):

sunho:
It's done by pent's pr. I'll pr the linux/mac atomic patch if it turns out to be required. 

We might not need atomic here, since game will not likely submit another scene while waiting, so any order of load and store work and buffered store is not problem...

Also, we better just use OS sync primitive here not spin loop.

PR #1304, opened 2021-05-12
Fix ubc with width or height not divisible by 4 again

sunho:
The ubc was swizzled outside 4x4 block so my recent pr that assumed linear layout was wrong. 

This pr reverts that recent pr and tris to solve the original problem again. It does by padding destination buffer to multiples of 4. It will not run out of source because source is destined to have enough amount of blocks. (since bc format doesn't define "half block") Texture will be read correctly by opengl since it's using appropriate stride that accounts for padding and supply "orig_width," and "orig_height" as texture size.


sunho:
Perheps, we must pad it to nearest power of two because of nature of swizzling. This pr will certainly solve heap corruption problem, but it might not solve partial textures. I'm not able to see the actual rendering of 1x80 texture used in gravity rush yet. I will visit this issue again after I've got some scene to test this out more thoroughly.

Zangetsu38 (approved):

PR #1308, opened 2021-05-13
settings_dialog: Add support cpu backend for custom config.

Zangetsu38:
# About:
- Simply add possibility change cpu backend for custom config.
- remove set value in kernel, value is already set in app init, so no need re set again.

### DIFF
--- a/vita3k/config/include/config/state.h
+++ b/vita3k/config/include/config/state.h
@@ -101,6 +101,7 @@ public:
 
     // Current config
     struct CurrentConfig {
+        std::string cpu_backend;
         bool lle_kernel = false;
         bool auto_lle = false;
         std::vector<std::string> lle_modules = {};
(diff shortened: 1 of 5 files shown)


sunho:
I'll pr a fix to make backend change apply without restarting emu.

sunho (approved):

Zangetsu38:
> I'll pr a fix to make backend change apply without restarting emu.

Is already possible for me ^^ i can change cpu backend and just run app and ut using it

sunho:
Callback pool is not affected at all. It's going to be quite massy.

Zangetsu38:
> Callback pool is not affected at all. It's going to be quite massy.

Ah yes is what is init on boot but that works without for that

sunho:
So some part is emulated using unicorn while other is emulated usimg dynarmic. It's mixing both of them. I'll deal with it.

Zangetsu38:
> So some part is emulated using unicorn while other is emulated usimg dynarmic. It's mixing both of them. I'll deal with it.

Yes i have get surprised with switch on dynamic and run app with it using it ^^

PR #1314, opened 2021-05-14
kernel, load_self: Fix for empty tls

bookmist:


pent0 (approved):
Your solution is more thorough, thanks

vita3k/kernel/src/thread/thread.cpp
```diff
@@ -107,18 +107,18 @@ SceUID create_thread(Ptr<const void> entry_point, KernelState &kernel, MemState
...
-    write_tpidruro(*thread->cpu, user_tls_address_ptr.address());
-
     if (kernel.tls_address) {
-        assert((kernel.tls_psize == 0xcccccccc ? 0 : kernel.tls_psize) <= (kernel.tls_msize == 0xcccccccc ? 0 : kernel.tls_msize));
+        auto user_tls_address_ptr = base_tls_address_ptr + kernel_tls_size;
+        write_tpidruro(*thread->cpu, user_tls_address_ptr.address());
+        assert(kernel.tls_psize <= kernel.tls_msize);
         memcpy(user_tls_address_ptr.get(mem), kernel.tls_address.get(mem), kernel.tls_psize);
+    } else {
+        write_tpidruro(*thread->cpu, 0);
```
sunho, line 121: We should set kernel tls address here.

sunho:
```
auto user_tls_address_ptr = base_tls_address_ptr + kernel_tls_size;
write_tpidruro(*thread->cpu, user_tls_address_ptr.address());
```
Just move this out and delete else. Then, looks fine!

bookmist: No, we should not, We have zero size tls, so I set null to tpidruro to crash on unexpected tls using.

sunho: If we lle some library that access kernel tls through tpidruro, we can know immediately in your way.

sunho: LGTM

PR #1319, opened 2021-05-15
arm_dynarmic: Minor changes

MerryMage:
Ensure all raised exceptions are handed.
Also reduce tick count (it's actually internally interpreted as a signed integer).


VelocityRa (approved):
lgtm, thanks ✨

PR #1324, opened 2021-05-17
Redesign guest function calling

sunho:
You can call guest function through three ways.

1. request_callback(current_thread, address, args) inside the interrupt handler

This queue a callback to current thread. After hle function returns, the thread will run the callback function, and return to the previous point. Any number of callbacks work, callback inside callback also works.

2. run_guest_function(kernel, address, args) from anywhere

This will queue a quest function call to single global worker thread. The worker thread will run requested quest functions one by one. This function is blocking and returned value is what guest function returned.

3. run_guest_function(kernel, worker_thread, address, args) from anywhere

If you don't want to use global worker in concern of bottleneck, you can create additional thread and specify here. GXM display callback is using a custom worker thread.

DON'T USE RUN_GUEST_FUNCTION WITH CURRENT THREAD INSIDE INTERRUPT HANDLER. It will be broken after callback finish. If you want to use current thread to run callback, use request_callback function.

### DIFF
--- a/vita3k/modules/SceAvPlayer/SceAvPlayer.cpp
+++ b/vita3k/modules/SceAvPlayer/SceAvPlayer.cpp
@@ -313,7 +313,9 @@ EXPORT(int32_t, sceAvPlayerEnableStream, SceUID player_handle, uint32_t stream_n
 EXPORT(bool, sceAvPlayerGetAudioData, SceUID player_handle, SceAvPlayerFrameInfo *frame_info) {
     const auto state = host.kernel.obj_store.get<AvPlayerState>();
     const PlayerPtr &player_info = lock_and_find(player_handle, state->players, state->mutex);
-
+    if (!player_info) {
+        return SCE_AVPLAYER_ERROR_ILLEGAL_ADDR;
+    }
     Ptr<uint8_t> buffer;
 
     if (player_info->paused) {
--- a/vita3k/modules/SceAvPlayer/SceAvPlayer.cpp
+++ b/vita3k/modules/SceAvPlayer/SceAvPlayer.cpp
@@ -19,6 +19,7 @@
 
 #include <codec/state.h>
 #include <io/functions.h>
+#include <kernel/functions.h>
 #include <kernel/thread/thread_functions.h>
 #include <util/lock_and_find.h>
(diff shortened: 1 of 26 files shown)


vita3k/modules/SceIme/SceIme.cpp
```diff
@@ -27,7 +27,7 @@ EXPORT(void, SceImeEventHandler, Ptr<void> arg, const SceImeEvent *e) {
     Ptr<SceImeEvent> e1 = Ptr<SceImeEvent>(alloc(host.mem, sizeof(SceImeEvent), "ime2"));
     memcpy(e1.get(host.mem), e, sizeof(SceImeEvent));
     auto thread = lock_and_find(thread_id, host.kernel.threads, host.kernel.mutex);
-    run_callback(host.kernel, *thread, thread_id, host.ime.param.handler.address(), { arg.address(), e1.address() });
+    request_callback(*thread, host.ime.param.handler.address(), { arg.address(), e1.address() });
```
Zangetsu38, line 30: i have try, keyboard is now totaly broken

vita3k/modules/SceAvPlayer/SceAvPlayer.cpp
```diff
@@ -223,7 +224,7 @@ uint run_event_callback(HostState &host, SceUID thread_id, const PlayerPtr playe
     constexpr char *export_name = "SceAvPlayerEventCallback";
     if (player_info->event_manager.event_callback) {
         auto thread = lock_and_find(thread_id, host.kernel.threads, host.kernel.mutex);
-        return run_callback(host.kernel, *thread, thread_id, player_info->event_manager.event_callback.address(), { player_info->event_manager.user_data, event_id, source_id, event_data.address() });
+        return run_guest_function(host.kernel, player_info->event_manager.event_callback.address(), { player_info->event_manager.user_data, event_id, source_id, event_data.address() });
```
Zangetsu38, line 227: all game using this function is now broken, a virus named tome etc

sunho:
With this, we don't need callback pool. Which reduced my memory usage on saenai from 8gb to 4gb.

Zangetsu38 (approved):

PR #1333, opened 2021-05-18
config: add disable at9 audio decoder option in settings dialog

Zangetsu38:
# About: 
After request change, i so add option in setting dialog for can disable At9 audio decoder in setting dialog.
Obligatory request for moment in gravity rush.
- Hack for moment external for lle MP4

# Result:
image

### DIFF
--- a/vita3k/gui/src/settings_dialog.cpp
+++ b/vita3k/gui/src/settings_dialog.cpp
@@ -144,6 +144,7 @@ static bool get_custom_config(GuiState &gui, HostState &host, const std::string 
             // Load Emulator Config
             if (!custom_config_xml.child("emulator").empty()) {
                 const auto emulator_child = custom_config_xml.child("emulator");
+                config.disable_at9_decoder = emulator_child.attribute("disable-at9-decoder").as_bool();
                 config.disable_ngs = emulator_child.attribute("disable-ngs").as_bool();
                 config.video_playing = emulator_child.attribute("video-playing").as_bool();
             }
@@ -166,6 +167,7 @@ void init_config(GuiState &gui, HostState &host, const std::string &app_path) {
         config.lle_kernel = host.cfg.lle_kernel;
         config.auto_lle = host.cfg.auto_lle;
         config.lle_modules = host.cfg.lle_modules;
+        config.disable_at9_decoder = host.cfg.disable_at9_decoder;
         config.disable_ngs = host.cfg.disable_ngs;
         config.video_playing = host.cfg.video_playing;
     }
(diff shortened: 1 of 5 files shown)


hobyst:
Wouldn't it be better to apply game-specific hacks on separate branches for each game rather than applying them on the master branch which is intended for general use. That way it will be much efficient to track any applied hacks by running git diff against both master and the game-specific branch?

Zangetsu38:
> Wouldn't it be better to apply game-specific hacks on separate branches for each game rather than applying them on the master branch which is intended for general use? That way it will be much efficient to track any applied hacks by running git diff against both master and the game-specific branch.

No that is very bad idea, is particuly because us general can works for all user, and here is détailed for why is here, is fine

Zangetsu38:
> Devs can use this branch for development purpose. But, I strongly disagree to merge this into master branch.

Big problèm, dev no obligatory know this issue and also no know this branch 🤷‍♂️ 
So more easy for dev and user for can get it

sunho:
You can make it togglable through gui so that it's easier to keep track of hacks. I'm fine if it's in that way.

hobyst:
> > Devs can use this branch for development purpose. But, I strongly disagree to merge this into master branch.
> 
> Big problèm, dev no obligatory know this issue and also no know this branch 🤷‍♂️
> So more easy for dev and user for can get it

But we don't know yet if this change is going to negatively impact other games and cause a regression.

sunho:
I know what's going on and how to fix this. It just takes a lot of time, so I decided to dig in later. Having hard coded game detection is like admitting that we have no intention to solve this, which is not the case.

Zangetsu38:
> > > Devs can use this branch for development purpose. But, I strongly disagree to merge this into master branch.
> > 
> > 
> > Big problèm, dev no obligatory know this issue and also no know this branch 🤷‍♂️
> > So more easy for dev and user for can get it
> 
> But we don't know yet if this change is going to negatively impact other games and cause a regression.

That is litteraly impossible, that affect Only gravity rush...

Zangetsu38:
> I know what's going on and how to fix this. It just takes a lot of time, so I decided to dig in later. Having hard coded game detection is like admitting that we have no intention to solve this, which is not the case.

Yes i know, but if you read my comment you understood is not this

sunho (approved):
We compromised to make it as a config instead of detecting game.

vita3k/gui/src/settings_dialog.cpp
```diff
@@ -385,13 +391,18 @@ void draw_settings_dialog(GuiState &gui, HostState &host) {
     if (ImGui::BeginTabItem("Emulator")) {
         ImGui::PopStyleColor();
         ImGui::Spacing();
+        ImGui::Checkbox("Disable At9 audio decoder", &config.disable_at9_decoder);
+        if (ImGui::IsItemHovered())
+            ImGui::SetTooltip("Disable At9 audio decoder can help some game to more progress.\nThis option is at the moment obligatory request for Gravity Rush/Daze");
```
sunho, line 396: Enable the hack that disables AT9 audio decoder.\nIt is required to prevent the crash in certain games such as Gravity Rush or Daze for the time being.

sunho (approved):

PR #1341, opened 2021-05-19
interface: load original eboot if detect mai-dump.

Zangetsu38:
# About
User and myself have see mai dump boot is roken, and i have thinks too, using original eboot is better for compat, mai dump is strange and ussless for emu.
- fix boot mai dump app and is better for compat.

PS: @sunho i know you fwill fix it later, like need fix lib kernel loading,  but for mai-dump is beter using origin eboot.


sunho (approved):
Yea I'll  revert this after I fixed.

PR #1344, opened 2021-05-21
shader/vpck: Correct no cycling

sunho:
Fix atlier color


Zangetsu38 (approved):
I funny like somme time problème is fixed with very small change 😅

PR #1345, opened 2021-05-22
gui/app selector: Add button on app icon if custom config exist.

Zangetsu38:
# About: 
- Add fake button for can view directly what game have custom config, i have lost time with all time check what game have custom config or no.
Why fake button ? already, button is no possible click because have selectable, and is good because, using text only alone is bad, for font text, can be hidden by depending color of icon.
Why CC ? because no can using icon, or need create it, so i have just founded CC for Custom Config.
-  disable count of start screen if any settings or control dialog is open.

# Result:
- In list
image
- In grid
image


sunho (approved):
cc

PR #1356, opened 2021-05-24
config: Use dynarmic as default

sunho:
All games using libult are essentially broken in unicorn. (which is quite a lot of games) Even if games use unimplemented/unpredictable instructions, we've got unicorn fallback. Why not making dynarmic default.


Zangetsu38 (approved):
I have thinks same recently

PR #1357, opened 2021-05-24
debugger: Fix crash when update_watches is called without valid parent

MerryMage:
Happens when Debug settings are configured prior to initialization of kernel


sunho (approved):

sunho:
We can initialize kernel earlier now, as we removed the callback pool. Lazy init was not necessary in the first place after all. I'll clean this part up after a while.

PR #1360, opened 2021-05-25
gui/Notice: fix load new notice selected.

Zangetsu38:
# About:
I have see after get one or some new notice, it is no select in blue if emu reboot, but only if get new notice during current running, i have so see in code, value for set notice info new is no set on init notice info.
- reworks order saved in xml.
- enable seletable for now all notification, with new code, game can switch in other game, so old code is no necessary anymore.
- fix position for cc button in app selector.

### DIFF
--- a/vita3k/gui/src/information_bar.cpp
+++ b/vita3k/gui/src/information_bar.cpp
@@ -186,7 +186,8 @@ void init_notice_info(GuiState &gui, HostState &host) {
                         });
                         notice_list[user.first].erase(notice_index);
                         save_notice_list(host);
-                    }
+                    } else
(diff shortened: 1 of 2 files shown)


vita3k/gui/src/information_bar.cpp
```diff
@@ -186,7 +186,8 @@ void init_notice_info(GuiState &gui, HostState &host) {
                         });
                         notice_list[user.first].erase(notice_index);
                         save_notice_list(host);
-                    }
+                    } else
+                        notice_info_new[notice.date] = notice_list_new[user.first][notice.date];
                 }
```
Zangetsu38, line 191: is that i have forget set before on rewroks my preview code, is now fixed so

sunho (approved):

PR #1371, opened 2021-05-27
gui/archive install dialog: refactor dialog.

Zangetsu38:
# About:
Refactor and reworks archive install dialog for can install exemple 100 app/dlc/theme in one time with select directory.
- allow can using folder for install some archive in same time.

# Result:
- new select choice
image
- Exemple on install directory
image
- after finish
image


Zangetsu38:
ready for review.
reviewer can just check too finished part for check if english text is fine :)

sunho (approved):

PR #1377, opened 2021-05-29
renderer: Allow writing depth when both side fragment program is disabled

sunho:
If both side is disabled, fragment shader discards every pixel, not writing to depth. But, we want it to still write to depth as long as depthwrite is enabled. This pr just mask color when both side is disabled to mitigate the issue. This fixes the regression in asphalt and progresses it by fixing the skybox.

I personally don't like the way we're implementing disabled side. Discarding pixel according to facing can be done by culling, so we don't need to have some sort of flag in fragbufferblock or have slow conditional discarding in fragment shader. But, I'm not sure what's going to happen when there is culling at the same time disabled side. So, I'm just keeping it for now.

### DIFF
--- a/vita3k/renderer/src/gl/draw.cpp
+++ b/vita3k/renderer/src/gl/draw.cpp
@@ -108,14 +108,6 @@ void draw(GLState &renderer, GLContext &context, const FeatureState &features, S
 
     glUseProgram(program_id);
 
-    if (context.record.is_maskupdate) {
-        // Tests bypassed in maskupdate
-        glDisable(GL_DEPTH_TEST);
-        glDisable(GL_STENCIL_TEST);
-
-        glBindFramebuffer(GL_FRAMEBUFFER, context.render_target->maskbuffer[0]);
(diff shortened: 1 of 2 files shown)


vita3k/renderer/src/gl/draw.cpp
```diff
@@ -185,6 +190,10 @@ void draw(GLState &renderer, GLContext &context, const FeatureState &features, S
         glBindFramebuffer(GL_FRAMEBUFFER, context.render_target->framebuffer[0]);
     }
 
+    if (both_side_fragment_program_disabled) {
+        sync_blending(context.record, mem);
```
pent0, line 194: Hmm why need sync again?

sunho: Restore color mask. I doubt there exist such crazy game but it can technically do setFragmentProgram -> disable side -> draw -> enable side -> draw.

Zangetsu38 (approved):

PR #1380, opened 2021-05-30
install content folder and some refactor

Zangetsu38:
# About
With this, for moment, you can just with open emu, drag & drop any one or few content, but that works also if you drag & drop one folder with inside 200 content like dlc/app


Zangetsu38:
Is now ready to review, for moment, is only working in drag and drop, open or closed

sunho (approved):

PR #1382, opened 2021-05-31
Partial mac opengl fix

sunho:
Mac opengl support is currently broken. This is a partial fix that works in general cases. I will not finish or clean up this pr in near future.


Zangetsu38:
is keep alive or ?

PR #1386, opened 2021-06-01
modules/SceGxm: Stub BeginSceneEx.

Zangetsu38:
# About
after see issue i get with this game is caused by shader issue like some toher game,  and no by this function directly, i thinks can stub it for now and fine
- allow some game move to ingame, ex: valkyrie drive.
# Result:
image
- hack shader get fine render
image

this game and one other move to bootable to ingame


sunho (approved):
Should be better than nothing.

PR #1393, opened 2021-06-03
gl/texture_formats: fix value for U2F10F10F10.

Zangetsu38:
# About:
- fix gl error and visible render on Valkyrie drive.

# Result:
- tv crypted effect and black and white is caused by one frag shader broken.
image
- using hack on this broken shader
image

Render keep have glitch on color, skin is full white etc


kd-11 (approved):
This is correct. See https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glTexSubImage2D.xhtml for the list of accepted types.

PR #1398, opened 2021-06-08
[WIP] renderer: add some color format.

Zangetsu38:
# About:
- add U10U10U10U2_RGBA for Chronovolt.
- add F10F10F10U2_RGBA for F1 2011 & silent hill.
- add F16F16F16F16_RGBA for Wipeout 2048 (Photos mode).
- add U8U8_GR for Grim fandango.
- add F32_R for Gundam Extreme.
- add U8U8U8U8_BGRA for To Love-Ru: Battle Ecstasy.
- add U1U5U5U5_ABGR for Hyperdimension Neptunia PP

- fix sizzle for for both U2_RGBA
- fix swizzle for U8U8U8X8_BGR1

# Result: 
- with some shader hack for improve broken texture
image
- with idsable process for pcm decoder
image

### DIFF
--- a/vita3k/renderer/src/gl/renderer.cpp
+++ b/vita3k/renderer/src/gl/renderer.cpp
@@ -442,9 +442,8 @@ void set_context(GLContext &context, const MemState &mem, const GLRenderTarget *
 void get_surface_data(GLContext &context, size_t width, size_t height, size_t stride_in_pixels, uint32_t *pixels, SceGxmColorFormat format) {
     R_PROFILE(__func__);
 
-    if (pixels == nullptr) {
+    if (!pixels)
         return;
-    }
 
     glPixelStorei(GL_PACK_ROW_LENGTH, static_cast<GLint>(stride_in_pixels));
 
@@ -454,15 +453,14 @@ void get_surface_data(GLContext &context, size_t width, size_t height, size_t st
         glReadPixels(0, 0, static_cast<GLsizei>(width), static_cast<GLsizei>(height), GL_RGBA, GL_UNSIGNED_INT_8_8_8_8_REV, pixels);
         break;
     case SCE_GXM_COLOR_FORMAT_U8U8U8U8_ARGB:
+    case SCE_GXM_COLOR_FORMAT_U8U8U8U8_BGRA:
         glReadPixels(0, 0, static_cast<GLsizei>(width), static_cast<GLsizei>(height), GL_BGRA, GL_UNSIGNED_INT_8_8_8_8_REV, pixels);
         break;
     case SCE_GXM_COLOR_FORMAT_U8U8U8U8_RGBA:
-        glReadPixels(0, 0, static_cast<GLsizei>(width), static_cast<GLsizei>(height), GL_RGBA, GL_UNSIGNED_BYTE, pixels);
-        for (int i = 0; i < width * height; ++i) {
-            uint8_t *pixel = reinterpret_cast<uint8_t *>(&pixels[i]);
-            std::swap(pixel[0], pixel[3]);
-            std::swap(pixel[1], pixel[2]);
-        }
+        glReadPixels(0, 0, static_cast<GLsizei>(width), static_cast<GLsizei>(height), GL_RGBA, GL_UNSIGNED_INT_8_8_8_8, pixels);
+        break;
+    case SCE_GXM_COLOR_FORMAT_U1U5U5U5_ABGR:
+        glReadPixels(0, 0, static_cast<GLsizei>(width), static_cast<GLsizei>(height), GL_RGBA, GL_UNSIGNED_SHORT_1_5_5_5_REV, pixels);
         break;
     case SCE_GXM_COLOR_FORMAT_U4U4U4U4_ARGB:
         glReadPixels(0, 0, static_cast<GLsizei>(width), static_cast<GLsizei>(height), GL_BGRA, GL_UNSIGNED_SHORT_4_4_4_4_REV, pixels);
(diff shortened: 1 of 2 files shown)


vita3k/renderer/src/gl/renderer.cpp
```diff
@@ -486,7 +486,9 @@ void get_surface_data(GLContext &context, size_t width, size_t height, size_t st
         glReadPixels(0, 0, static_cast<GLsizei>(width), static_cast<GLsizei>(height), GL_RED, GL_UNSIGNED_BYTE, pixels);
         break;
     case SCE_GXM_COLOR_FORMAT_U2F10F10F10_ABGR:
+    case SCE_GXM_COLOR_FORMAT_F10F10F10U2_RGBA:
     case SCE_GXM_COLOR_FORMAT_U2U10U10U10_ABGR:
+    case SCE_GXM_COLOR_FORMAT_U10U10U10U2_RGBA:
         glReadPixels(0, 0, static_cast<GLsizei>(width), static_cast<GLsizei>(height), GL_RGBA, GL_UNSIGNED_INT_2_10_10_10_REV, pixels);
```
kd-11, line 492: If 2-10-10-10 is using GL_UNSIGNED_INT_2_10_10_10_REV then the new ones which are reversed should use GL_UNSIGNED_INT_10_10_10_2.

Zangetsu38: Oh’ i have just used your commit’ is untouched

Zangetsu38: you can re see it here
https://github.com/Zangetsu38/Vita3K/commit/1ca94c0cfcbd3a562b8dc0b4e95bb1370df9b661

Zangetsu38: and cceck it for my version on other value, i have used what you said before and on my current build
https://github.com/Zangetsu38/Vita3K/commit/64eb69a94f377c9d08fb459c3a6c9219bdd03913

Zangetsu38: i am just sur for current value here for U2F10F10F10_ABGR
image
using other value get olor full broken

Zangetsu38:
@kd-11 if you can review again please

vita3k/renderer/src/gl/renderer.cpp
```diff
@@ -486,9 +488,15 @@ void get_surface_data(GLContext &context, size_t width, size_t height, size_t st
...
         break;
+    case SCE_GXM_COLOR_FORMAT_F10F11F11_BGR:
+    case SCE_GXM_COLOR_FORMAT_F11F11F10_RGB:
+        glReadPixels(0, 0, static_cast<GLsizei>(width), static_cast<GLsizei>(height), GL_BGR, GL_UNSIGNED_INT_10F_11F_11F_REV, pixels);
```
kd-11, line 498: One is BGR the other is RGB but you're using the same readback format. Now way for both to be right.

vita3k/renderer/src/gl/renderer.cpp
```diff
@@ -486,9 +488,15 @@ void get_surface_data(GLContext &context, size_t width, size_t height, size_t st
         glReadPixels(0, 0, static_cast<GLsizei>(width), static_cast<GLsizei>(height), GL_RED, GL_UNSIGNED_BYTE, pixels);
         break;
     case SCE_GXM_COLOR_FORMAT_U2F10F10F10_ABGR:
+    case SCE_GXM_COLOR_FORMAT_F10F10F10U2_RGBA:
     case SCE_GXM_COLOR_FORMAT_U2U10U10U10_ABGR:
+    case SCE_GXM_COLOR_FORMAT_U10U10U10U2_RGBA:
```
kd-11, line 493: Same problem here, you have different order all using the same readback settings making it impossible for all of them to be right.

vita3k/renderer/src/gl/renderer.cpp
```diff
@@ -454,6 +454,7 @@ void get_surface_data(GLContext &context, size_t width, size_t height, size_t st
         glReadPixels(0, 0, static_cast<GLsizei>(width), static_cast<GLsizei>(height), GL_RGBA, GL_UNSIGNED_INT_8_8_8_8_REV, pixels);
         break;
     case SCE_GXM_COLOR_FORMAT_U8U8U8U8_ARGB:
+    case SCE_GXM_COLOR_FORMAT_U8U8U8U8_BGRA:
```
kd-11, line 457: One is BGRA, the other RGBA, using the same readback format. One of them is going to be incorrect.

vita3k/renderer/src/gl/renderer.cpp
frangarcj: @Zangetsu38 

example
SCE_GXM_COLOR_FORMAT_F11F11F10_RGB -> GL_UNSIGNED_INT_10F_11F_11F_REV
SCE_GXM_COLOR_FORMAT_F10F11F11_BGR -> GL_UNSIGNED_INT_10F_11F_11F

PS: I don't know which one is REV

Zangetsu38: Only this value existe, so no have choice

vita3k/renderer/src/gl/renderer.cpp
Zangetsu38: Yes normal, like other already set bgra for both

vita3k/renderer/src/gl/renderer.cpp
Zangetsu38: I have explain all in dm obligatory need use same value for works

vita3k/renderer/src/gl/renderer.cpp
kd-11: ReadPixels spec only defines the REV version, but this means post-transformation is required on the CPU client. The data is wrong either way.

Zangetsu38: OK so how set it correctly ? I have no idea here 😢

vita3k/renderer/src/gl/renderer.cpp
kd-11: It is impossible for both to be right, so if adding this fixes something there are 2 possibilities:
1. The existing format _ABGR is wrong.
2. There is a bug somewhere else that is making this work.
I get the need for some results, but this doesn't fix the real problem. Someone will have to redo all of these properly later.

vita3k/renderer/src/gl/renderer.cpp
kd-11: Same as previous. If both are working you have serious problems somewhere else.

Zangetsu38:
this swizzle change fix this two game, but broken dustforce, and probably any other using it, me here 💀

pent0:
Waiting for kd results from tests.

Zangetsu38:
outdated, no request anymore

PR #1405, opened 2021-06-14
gui/app context menu: Add last time played in information.

Zangetsu38:


vita3k/gui/src/app_context_menu.cpp
```diff
@@ -97,6 +98,57 @@ static const char OS_PREFIX[] = "open ";
 static const char OS_PREFIX[] = "xdg-open ";
 #endif
 
+static std::map<std::string, std::map<std::string, time_t>> last_time_apps_used;
```
pent0, line 101: Should move it to UI State, then access it there

vita3k/gui/src/app_context_menu.cpp
```diff
@@ -97,6 +98,57 @@ static const char OS_PREFIX[] = "open ";
...
+
+void get_last_time_apps_used(HostState &host) {
```
pent0, line 103: If the plan is to move the upper variable to GuiState, you can pass another argument

```suggestion
void get_last_time_apps_used(GuiState &gui, HostState &host) {
```

vita3k/gui/src/app_context_menu.cpp
Zangetsu38: I want keep it here for moment

vita3k/gui/src/app_context_menu.cpp
Zangetsu38: Yes i know’ but for moment i keep like that

vita3k/gui/src/app_context_menu.cpp
Zangetsu38: done, moved in gui state.

pent0 (approved):

PR #1406, opened 2021-06-14
Integrate Boost build and introduce CMake toolchain files

hobyst:
This pull request updates the build process of Vita3K to completely remove the need to use command line scripts each time a project has to be generated due to the Boost building step, which now will automatically be done by CMake at configure time.

It also introduces the use of CMake toolchain files to set platform-specific variables and settings that used to be set by the terminal scripts and updates the GitHub workflows to utilize the toolchain files as well, so any new settings common to both developer and CI setups but unique to a certain target platform can now be added directly into the corresponding toolchain file for the target platform instead of having to set them in multiple places. Toolchain files are also used by CMake to allow cross-compilation, which will get used at some point with target platforms like Android.

The new build instructions can be found at `./building.md`.


Zangetsu38:
I don't really like the removal of the gen-windows.bat, which is just really handy
try keep it for no need open any cmd command or any things, just run bat and go :)

sunho:
Whoa.

sunho:
@Zangetsu38 You can just make gen-windows.bat file with content "cmake .." and that's it.

Zangetsu38:
> 
> 
> @Zangetsu38 You can just make gen-windows.bat file with content "cmake .." and that's it.

Is full no sense, and i just lost time, file is already here and ready, just change command line inside that all, but no delete, is totaly no sense

gen-windows.bat
```diff
@@ -1,26 +0,0 @@
-@echo off
```
Zangetsu38, line 1: Don't delete it, back this file and just update command with news version please

Zangetsu38 (changes requested):

sunho:
Let's leave gen-x scripts for zang. Official build guide should instruct cmake --build method though.

hobyst:
> > @Zangetsu38 You can just make gen-windows.bat file with content "cmake .." and that's it.
> 
> Is just full no sense, and is just lost time, file is already here and ready, just change command line inside that all, but no delete, is totaly no sense

Didn't even thought about that option 😅. Should the script run both project generation and building or just project generation?

Zangetsu38:
> 
> 
> > > @Zangetsu38 You can just make gen-windows.bat file with content "cmake .." and that's it.
> > 
> > 
> > Is just full no sense, and is just lost time, file is already here and ready, just change command line inside that all, but no delete, is totaly no sense
> 
> Didn't even thought about that option 😅. Should the script run both project generation and building or just project generation?

no idea, no sur how works yoiur new build ^^ 

just jkeep file here and update command inside :)

sunho:
@hobyst Just make gen-x generate project files like before and forget about it.

hobyst:
> > > > @Zangetsu38 You can just make gen-windows.bat file with content "cmake .." and that's it.
> > > 
> > > 
> > > Is just full no sense, and is just lost time, file is already here and ready, just change command line inside that all, but no delete, is totaly no sense
> > 
> > 
> > Didn't even thought about that option sweat_smile. Should the script run both project generation and building or just project generation?
> 
> no idea, no sur how works yoiur new build ^^
> 
> just jkeep file here and update command inside :)


The new system pretty much makes running `bootstrap.sh` / `bootstrap.bat` and `b2` unnecessary as now CMake will do it. Boost now also gets built inside the CMake binary directory (the same one that gets specified at project generation with `-B`), which for example would mean that with a binary directory named `build-windows`, Boost would get built at `./build-windows/external/boost` instead of `./external/boost-build`. This allows building the emulator for multiple platforms using the same host machine without having to constantly delete and rebuild Boost each time the target platform is changed.
The new command line scripts would essentially be just the project generation command found in the `building.md`. Example for Windows:
```cmd
cmake -S . -B build-windows -DCMAKE_TOOLCHAIN_FILE=./cmake/toolchain/windows-x64.cmake
```
And what I'm asking is whether the script should also build/compile the project as well or just generate the it (like they did before).

Zangetsu38:
ah no, i speek just like before, soi just generate project, that all ^^, just like before
and for windows you said is 
>cmake -S . -B build-windows -DCMAKE_TOOLCHAIN_FILE=./cmake/toolchain/windows-x64.cmake

so just set inside inside gen-windosws

sunho:
I'm fully aware that adding gen-x scripts would mean just copy&paste from building.md. But, zang just doesn't want to open command line in the first place, and it's very hard to change his mind once he comes to wanting something. Just copy & paste cmake command to gen-x and forget about it. (building guide shouldn't instruct gen-x however) Anyways, you did a totally awesome job. 👍 I love your pr.

Zangetsu38:
> 
> 
> I'm fully aware that adding gen-x scripts would mean just copy&paste from building.md. But, zang just doesn't want to open command line in the first place, and it's very hard to change his mind once he comes to wanting something. Just copy & paste cmake command to gen-x and forget about it. (building guide shouldn't instruct gen-x however) Anyways, you did a totally awesome job. 👍 I love your pr.

hey 😭 , i said this for all, batch is just very usseful :)

hobyst:
Even tho I have now recovered the command line scripts, wait until my next push to review the PR. There's something I want to change from the code I wrote in CMakeLists.txt I want to improve.

gen-windows.bat
```diff
@@ -1,26 +1,4 @@
...
-
-REM Generate project files
-IF "%CI%"=="" (
-	call cmake -G "Visual Studio 16 2019" ..
-) ELSE (
-	call cmake -G "Visual Studio 16 2019" -DCI:BOOL=ON -DCMAKE_CONFIGURATION_TYPES=%CONFIGURATION% ..
-)
-popd
+REM Generate project files for Visual Studio 2019
+call cmake -S . -B build-windows -G "Visual Studio 2019 16" -DCMAKE_TOOLCHAIN_FILE=./cmake/toolchain/windows-x64.cmake
```
Zangetsu38, line 4: thx, no forget add new line here ^^

hobyst:
Okay, now everything should be good to go and ready to review.

hobyst:
I forgot to add new line on `gen-windows.bat` in the last push 😅. Now everything should be fine.

scribam (commented):
These toolchains look redundant to me with the default values initialized by cmake.  Maybe, Linux is the exception as you can have GCC and Clang as default depending of your system configuration. Some versions of GCC doesn't work well with Vita3K (IIRC GCC 8+) but wouldn't be better to blacklist these versions (e.g. using CMAKE_CXX_COMPILER_ID / CMAKE_CXX_COMPILER_VERSION) with a message saying the compiler is incompatible?

I understand one of the goal of this PR is to help users to build Vita3k and I think the boost macros (although I have a system boost setup on my Windows environment) and improving the documentation are enough to do so 👍

hobyst:
> These toolchain files look redundant to me with the default values initialized by cmake. Maybe, Linux is the exception as you can have GCC and Clang as default depending of your system configuration. Some versions of GCC doesn't work well with Vita3K (IIRC GCC 8+) but wouldn't be better to blacklist these versions (e.g. using CMAKE_CXX_COMPILER_ID / CMAKE_CXX_COMPILER_VERSION) with a message saying the compiler is incompatible?
> 
> I understand one of the goal of this PR is to help users to build Vita3k and I think the boost macros (although I have a system boost setup on my Windows environment) and improving the documentation are enough to do so 👍

The idea behind the toolchain files is to get them incorporated so if at any time certain development environment settings related to a specific target platform need to be either or modified or added, editing the toolchain files would be enough for that change to be applied to every development/building environment (including CI) without having to edit documentation or the GitHub workflow scripts. They also are used by CMake to enable cross-compilation (at the moment the toolkit for desktop targets files are scripted to block cross-compilation), which will be useful once porting to mobile platforms is investigated.

Zangetsu38 (approved):

sunho:
Tell if there's problem.

PR #1407, opened 2021-06-15
gui/settings: add arrow for scroll in theme list.

Zangetsu38:
# About:
is same like for app list
Is very useful for can scroll without wheel, without mouse, iwht just keyoard and click on arrow

# Result:
image

### DIFF
--- a/vita3k/gui/src/settings.cpp
+++ b/vita3k/gui/src/settings.cpp
@@ -567,8 +567,6 @@ void draw_start_screen(GuiState &gui, HostState &host) {
 
 static std::string popup, settings_menu, menu, sub_menu, selected, title, delete_user_background, delete_theme;
 static ImGuiTextFilter search_bar;
-static float scroll_pos;
-static bool set_scroll_pos;
 
 static std::string get_date_format_sting(GuiState &gui, DateFormat &date_format) {
     std::string date_format_str;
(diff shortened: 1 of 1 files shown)


vita3k/gui/src/settings.cpp
```diff
@@ -590,6 +588,9 @@ static const std::vector<std::string> LIST_SYS_LANG = {
     "Dansk", "Norsk", "Polskis", u8"Português (Brasil)", "English (United Kingdom)", u8"Türkçe"
 };
 
+static const ImU32 ARROW_COLOR = 4294967295; // White
```
pent0, line 591: Better to be in hexcimal

vita3k/gui/src/settings.cpp
```diff
@@ -590,6 +588,9 @@ static const std::vector<std::string> LIST_SYS_LANG = {
...
+static float set_scroll_pos, current_scroll_pos, max_scroll_pos;
```
pent0, line 592: These variables are better, to be moved in GuiState

Zangetsu38: No, is variable for settings only keep here

vita3k/gui/src/settings.cpp
Zangetsu38: i already using this in oher place,this function request ImU32 value only

pent0: 0xFFFFFFFF

Zangetsu38: No i said only using this value no hexa

vita3k/gui/src/settings.cpp
pent0: current_scroll_pos is definitely used again from previous state

Zangetsu38: No understund is full fine here

vita3k/gui/src/settings.cpp
pent0: Why no hexdacimal? This is for people to recognise this RGB code immediately without doubt.

Zangetsu38: done

pent0 (approved):

PR #1408, opened 2021-06-16
project: fix typo on gen-windows.

Zangetsu38:
# About: 
command is wrong, number is inversed for cmake, fix generate project.
- add pause in end script for can view result before close alone.


pent0 (approved):
Valid, see [here](https://cmake.org/cmake/help/git-stage/generator/Visual%20Studio%2016%202019.html)

PR #1410, opened 2021-06-18
gui: add time app used & refactor.

Zangetsu38:
# About:
Add time app used inside app context menu in information.
Refactor code for can using new value for no get double value exist for same things.
Using so now app list by user.

- move cfg to new value.

# Result
- is fake time used here off course, is just for test etc :D 
image

### DIFF
--- a/vita3k/kernel/include/kernel/object_store.h
+++ b/vita3k/kernel/include/kernel/object_store.h
@@ -49,7 +49,7 @@ public:
     }
 
     template <typename T, typename... Args>
-    bool create(Args &&... args) {
+    bool create(Args &&...args) {
         std::lock_guard<std::mutex> lock(mutex);
         auto ptr = std::make_shared<T>(std::forward<Args>(args)...);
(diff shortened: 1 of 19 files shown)


vita3k/kernel/include/kernel/object_store.h
```diff
@@ -49,7 +49,7 @@ class ObjectStore {
     }
 
     template <typename T, typename... Args>
-    bool create(Args &&... args) {
+    bool create(Args &&...args) {
```
pent0, line 52: hm

pent0: Unrelated format changes not to be here.

Zangetsu38: is done by format bat ^^

pent0: bbb vv zz

pent0 (approved):

PR #1432, opened 2021-07-31
settings: improve dialog selet lang/time format.

Zangetsu38:
# About:
Remove space contourn popup select, get better result.

### DIFF
--- a/vita3k/gui/src/settings.cpp
+++ b/vita3k/gui/src/settings.cpp
@@ -176,10 +176,19 @@ static void get_themes_list(GuiState &gui, HostState &host) {
     }
 }
 
-static std::string popup, settings_menu, menu, sub_menu, selected, title, delete_user_background, delete_theme;
+static std::string popup, menu, sub_menu, selected, title, delete_user_background, delete_theme;
 static ImGuiTextFilter search_bar;
 
 static float set_scroll_pos, current_scroll_pos, max_scroll_pos;
+
+enum SettingsMenu {
+    SELECT,
+    THEME_BACKGROUND,
+    DATE_TIME,
+    LANGUAGE
+};
+
+static SettingsMenu settings_menu = SELECT;
 
 void draw_settings(GuiState &gui, HostState &host) {
     const auto display_size = ImGui::GetIO().DisplaySize;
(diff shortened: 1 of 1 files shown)


marcoluc97:
Hi @Zangetsu38 I do not know if this is useful but if you are excluding NPXS10015 because is the Settings app maybe is better to add also NPXS10016, or i'm just wrong so ignore this.

[PSVita Dev Wiki Title ID](https://playstationdev.wiki/psvitadevwiki/index.php?title=Title_ID)

Zangetsu38:
> 
> 
> Hi @Zangetsu38 I do not know if this is useful but if you are excluding NPXS10015 because is the Settings app maybe is better to add also NPXS10016, or i'm just wrong so ignore this.
> 
> [PSVita Dev Wiki Title ID](https://playstationdev.wiki/psvitadevwiki/index.php?title=Title_ID)

yes, i using only one setting, no other

vita3k/gui/src/app_selector.cpp
```diff
@@ -31,7 +31,13 @@ using namespace std::string_literals;
 namespace gui {
 
 void init_user_apps(GuiState &gui, HostState &host) {
-    gui.apps_background.clear();
+    if (gui.live_area.settings) {
+        for (const auto &app : gui.apps_list_opened) {
+            if (app != "NPXS10015")
+                gui.apps_background.erase(app);
+        }
+    } else
```
IllusionMan1212, line 39: please add curly braces here, it makes it more readable especially since the if statement has braces

IllusionMan1212 (changes requested):

vita3k/gui/src/app_selector.cpp
Zangetsu38: Is not needed with one return value

IllusionMan1212: it works yes, but its a lot more readable to have braces

Zangetsu38: no really more readable :) i no want touch it is fine here

Zangetsu38: request approve, i no want change it here, is good like that

Zangetsu38: review again, this code change is totaly removed, no need anymore ^^

IllusionMan1212 (approved):

PR #1437, opened 2021-08-09
Settings: improve system lang

Croden1999:
improve settings application translation of system languages.


Zangetsu38:
Sorry, for this things i have set this for good reason, using this string request other font and take more memory, so i have keep english text

Croden1999:
> Sorry, for this things i have set this for good reason, using this string request other font and take more memory, so i have keep english text

i understand. thank you so much.

PR #1438, opened 2021-08-09
change chinese lang strings

Croden1999:
The template.xml file for the settings application is shown as the ch and zh  lang strings, so i change them.

# Result


Zangetsu38:
Is wrong, XML file for lang using zh-t/zh-s, this is inside firmware and inside all rco file
That is totaly diferent compare live area template

Zangetsu38:
But, @Croden1999   if i trust your tapante, is interesting for live area part i need check it, thx for this found

Croden1999:
> But, @Croden1999 if i trust your tamplate, is interesting for live area part i need check it, thx for this found

thx a lot, Mr. Zangetsu.

PR #1483, opened 2021-10-30
Warning fixes

bookmist:
I'm not sure if I fix this warnings properly, but original code looks strange for me.
@sunho please, check this. The most strange moment for me is float registers in Dynarmic and Unicorn.

### DIFF
--- a/vita3k/modules/SceAvPlayer/SceAvPlayer.cpp
+++ b/vita3k/modules/SceAvPlayer/SceAvPlayer.cpp
@@ -51,7 +51,7 @@ struct AvPlayerState {
 };
 
 struct SceAvPlayerMemoryAllocator {
-    uint32_t user_data;
+    uint32_t user_data = 0;
 
     // All of these should be cast to SceAvPlayerAllocator or SceAvPlayerDeallocator types.
     Ptr<void> general_allocator;
@@ -61,7 +61,7 @@ struct SceAvPlayerMemoryAllocator {
 };
 
 struct SceAvPlayerFileManager {
-    uint32_t user_data;
+    uint32_t user_data = 0;
 
     // Cast to SceAvPlayerOpenFile, SceAvPlayerCloseFile, SceAvPlayerReadFile and SceAvPlayerGetFileSize.
     Ptr<void> open_file;
@@ -71,7 +71,7 @@ struct SceAvPlayerFileManager {
 };
 
 struct SceAvPlayerEventManager {
-    uint32_t user_data;
+    uint32_t user_data = 0;
 
     // Cast to SceAvPlayerEventCallback.
     Ptr<void> event_callback;
@@ -310,7 +310,7 @@ EXPORT(bool, sceAvPlayerGetAudioData, SceUID player_handle, SceAvPlayerFrameInfo
     const auto state = host.kernel.obj_store.get<AvPlayerState>();
     const PlayerPtr &player_info = lock_and_find(player_handle, state->players, state->mutex);
     if (!player_info) {
-        return SCE_AVPLAYER_ERROR_ILLEGAL_ADDR;
+        return false;
     }
     Ptr<uint8_t> buffer;
 
@@ -376,6 +376,9 @@ EXPORT(uint32_t, sceAvPlayerGetStreamInfo, SceUID player_handle, uint stream_no,
 EXPORT(bool, sceAvPlayerGetVideoData, SceUID player_handle, SceAvPlayerFrameInfo *frame_info) {
     const auto state = host.kernel.obj_store.get<AvPlayerState>();
     const PlayerPtr &player_info = lock_and_find(player_handle, state->players, state->mutex);
+    if (!player_info) {
+        return false;
+    }
 
     Ptr<uint8_t> buffer;
 
--- a/vita3k/renderer/src/gl/draw.cpp
+++ b/vita3k/renderer/src/gl/draw.cpp
@@ -128,7 +128,8 @@ void draw(GLState &renderer, GLContext &context, const FeatureState &features, S
     set_uniform_buffer(context, true, SCE_GXM_REAL_MAX_UNIFORM_BUFFER, sizeof(GXMRenderVertUniformBlock), &vert_ublock, false);
 
     GXMRenderFragUniformBlock frag_ublock;
-    const bool both_side_fragment_program_disabled = context.record.front_side_fragment_program_mode && (context.record.front_side_fragment_program_mode || context.record.back_side_fragment_program_mode);
+    const bool both_side_fragment_program_disabled = (context.record.front_side_fragment_program_mode == SCE_GXM_FRAGMENT_PROGRAM_DISABLED)
+        && ((context.record.back_side_fragment_program_mode == SCE_GXM_FRAGMENT_PROGRAM_DISABLED) || (context.record.two_sided == SCE_GXM_TWO_SIDED_DISABLED));
     if (both_side_fragment_program_disabled) {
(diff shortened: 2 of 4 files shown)


vita3k/modules/SceAvPlayer/SceAvPlayer.cpp
```diff
@@ -310,7 +310,7 @@ EXPORT(bool, sceAvPlayerGetAudioData, SceUID player_handle, SceAvPlayerFrameInfo
     const auto state = host.kernel.obj_store.get<AvPlayerState>();
     const PlayerPtr &player_info = lock_and_find(player_handle, state->players, state->mutex);
     if (!player_info) {
-        return SCE_AVPLAYER_ERROR_ILLEGAL_ADDR;
+        return false;
```
Zangetsu38, line 313: tjhx, i have see it some time but no see hors fix it correctly ^^

vita3k/cpu/src/unicorn_cpu.cpp
```diff
@@ -364,7 +374,7 @@ CPUContext UnicornCPU::save_context() {
     ctx.set_pc(is_thumb_mode() ? get_pc() | 1 : get_pc());
 
     for (size_t i = 0; i < ctx.fpu_registers.size(); i++) {
-        ctx.fpu_registers[i] = get_float_reg(i);
+        ctx.fpu_registers[i] = float_to_binary(get_float_reg(i));
```
sunho, line 377: Nice catch. You can just do reinterpret_cast<uint32_t&>(float) and reinterpret_cast<float&>(uint32_t)

vita3k/renderer/src/gl/draw.cpp
```diff
@@ -128,7 +128,7 @@ void draw(GLState &renderer, GLContext &context, const FeatureState &features, S
     set_uniform_buffer(context, true, SCE_GXM_REAL_MAX_UNIFORM_BUFFER, sizeof(GXMRenderVertUniformBlock), &vert_ublock, false);
 
     GXMRenderFragUniformBlock frag_ublock;
-    const bool both_side_fragment_program_disabled = context.record.front_side_fragment_program_mode && (context.record.front_side_fragment_program_mode || context.record.back_side_fragment_program_mode);
+    const bool both_side_fragment_program_disabled = context.record.front_side_fragment_program_mode && context.record.back_side_fragment_program_mode;
```
sunho, line 131: Nice one. Can you change the name of variable to both_side_fragment_program_enabled?

bookmist: No. 
`enum SceGxmFragmentProgramMode {
    SCE_GXM_FRAGMENT_PROGRAM_DISABLED = 0x00200000u,
    SCE_GXM_FRAGMENT_PROGRAM_ENABLED = 0x00000000u
};`
so True mean disabled.
Or is there also some logic bug?

vita3k/cpu/src/unicorn_cpu.cpp
bookmist: I can't convert function result with reinterpret_cast. :(

vita3k/renderer/src/gl/draw.cpp
sunho: https://github.com/Vita3K/Vita3K/blob/9cb2063d50ae5fe0603594ea6122069cf8d67510/vita3k/shader/src/spirv_recompiler.cpp#L1271

We disable it when disabled != 0.0f
 
should be context.record.front_side_fragment_program_mode == SCE_GXM_FRAGMENT_PROGRAM_DISABLED && context.record.back_side_fragment_program_mode == SCE_GXM_FRAGMENT_PROGRAM_DISABLED ;

vita3k/cpu/src/unicorn_cpu.cpp
sunho: actually better way is just change type of fpu_registers to array of float

sunho: it's our type

sunho: https://github.com/Vita3K/Vita3K/blob/168dfb19caa42c1216152e11bbc209ec7d2b8cb0/vita3k/cpu/src/dynarmic_cpu.cpp#L411
here you can just use memcpy to copy it directly. Bit pattern should be exactly same.

sunho: memcpy(&ctx.fpu_registers, dctx.ExtRegs().data(), sizeof(ctx.fpu_registers));

vita3k/renderer/src/gl/draw.cpp
bookmist: What about this block?
https://github.com/Vita3K/Vita3K/blob/399fe1766f0f985d440813def0ab2aef53ea12ea/vita3k/renderer/src/gl/draw.cpp#L194-L196

bookmist: should it be  if (both_side_fragment_program_enabled) { 
or  if (!both_side_fragment_program_enabled) {

sunho: No that line should be fine. When fragment program is disabled, we should not write the color that was given by fragment program but should still write depth. But, we're using discard to implement this behavior so depth is also thrown away. We're temporarily masking color write when both side is disabled, but doesn't discard pixels so that depth buffer is written. It's syncing blending state to restore the tweaked color mask.

sunho: A cleaner way to approach this thing is don't use discard but write the previous color. Then, we don't need this color write trickery. This fix needs a bit more code tho. Let's only fix the logic error of both_side_fragment_program_disabled in this pr.

sunho: both_side_fragment_program_disabled was the right name btw. Sorry for confusion.

bookmist:
I hope now it works as intended. @sunho recheck it, please.

vita3k/renderer/src/gl/draw.cpp
```diff
@@ -128,10 +128,11 @@ void draw(GLState &renderer, GLContext &context, const FeatureState &features, S
     set_uniform_buffer(context, true, SCE_GXM_REAL_MAX_UNIFORM_BUFFER, sizeof(GXMRenderVertUniformBlock), &vert_ublock, false);
 
     GXMRenderFragUniformBlock frag_ublock;
-    const bool both_side_fragment_program_disabled = context.record.front_side_fragment_program_mode && (context.record.front_side_fragment_program_mode || context.record.back_side_fragment_program_mode);
+    const bool both_side_fragment_program_disabled = (context.record.front_side_fragment_program_mode == SCE_GXM_FRAGMENT_PROGRAM_DISABLED)
+        && ((context.record.back_side_fragment_program_mode == SCE_GXM_FRAGMENT_PROGRAM_DISABLED) || (context.record.two_sided == SCE_GXM_TWO_SIDED_DISABLED));
     if (both_side_fragment_program_disabled) {
-        frag_ublock.front_disabled = 0.0f;
-        frag_ublock.back_disabled = 0.0f;
+        frag_ublock.front_disabled = 1.0f;
```
sunho, line 134: needs to be 0.0f; cause we don't know it to discard.

sunho: otherwise fine!

bookmist: Thank you for your help!

sunho (approved):

PR #1487, opened 2021-11-05
modules/SceThreadmgr: Imlement sceKernelGetSemaInfo.

Zangetsu38:
# About:
- Implement GetSemaInfo.
- fix some find callbackId, is currently inversed with show error if found it, correct is get error is not found.
- fix return error on this callback function too, for show it in log.
 
# Result: 
value get in log
image


Zangetsu38:
Oups is Implement, i need glass 🤦‍♂️🤣 i change later

frangarcj (approved):

PR #1485, opened 2021-11-03
Add partial Callback implementation

CreepNT:

PR #1488, opened 2021-11-05
modules/SceKernelThreadMgr: fix mutex deadlock and clean code of callback implementation

bookmist:
Without deadlock fix almost all games stops almost immediately.
Sorry for lack of review in #1485 PR


Zangetsu38:
Good catch i have missing it in my last pr

Zangetsu38 (approved):

PR #1495, opened 2021-11-09
kernel: Fix callback execution

sunho:
Try to fix callback execution

run_guest_function of thread shouldn't be called in the hle handler of that thread.


sunho:
need some more care. it's more complicated.

Zangetsu38 (approved):
LGTM i have rebase and fixed conflict, very important for other stuff and code fine works, @sunho if you want fix any things wrong on it later you are free, but here is request for all lego game thx

Issue #1489, opened 2021-11-05
Regression after deferred PR


Zangetsu38:
# About: 
I open issue for list all game have get regression after this PR #1195 relate all time implement notification wait commit https://github.com/Vita3K/Vita3K/commit/b6e4400b971fe213c24d1d2733bc25d667085eee

- Knyt Undergrund: Stop boot on first screen loading, no load file, infinite wait inside notification wait
- f1 2011: crash inside sceGxmEndScene on vertexNotification set.
- final fantasy X/X-2: infinite wait inside notification wait.
- OlliOlli2: Welcome to Olliwood infinite wait inside notification wait.

# Log:
-  log for ffx and Knyt undergrund, value is diferent but issue is same, value and notif value have - 1 
![image](https://user-images.githubusercontent.com/5261759/140583698-0b28db6e-ac09-401c-a8b6-3d3b0bc25569.png)

Zangetsu38:
OUPSY, is fixed now

PR #1500, opened 2021-11-11
Try fix signal Notification issue

Zangetsu38:
# About:
- Inspired by GxmMidFlush improvement, i have wanted see if using same method can fix infinite wait caused bad bad value set on some game, with all time it same + 2 on value so get all time example on notifwait 12 vs 11.

- try fix issue #1489

I have see no using add command fix it and set so exactly good value, and for ratchet no need any if (val).

And see for ffx too, issue is no get by only VertexNotif but also on FragmentNotif.

No want said need merge pr like that, is just here for search and expertise and try fix this issue.
I waiting so expertise of @sunho or @pent0 

F1 keep broken, but it no infinite wait anymore and no crash, just freeze 🤷‍♂️ 

# Result:
- fix ffx(-2) for this screen is vertexnotif, but for can boot game, is fragment notif if waiting infinite before load self
image
image

- fix knyt undergrund too
image

### DIFF
--- a/vita3k/modules/SceGxm/SceGxm.cpp
+++ b/vita3k/modules/SceGxm/SceGxm.cpp
@@ -1379,13 +1379,13 @@ EXPORT(int, sceGxmEndScene, SceGxmContext *context, Ptr<SceGxmNotification> vert
         ++host.renderer->last_scene_id);
 
     if (vertexNotification) {
-        renderer::add_command(context->renderer.get(), renderer::CommandOpcode::SignalNotification,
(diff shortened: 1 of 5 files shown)


sunho:
no this is wrong

Zangetsu38:
i know is wrong, i have said is for can check how fixing it, because this works

sunho:
yea we can look into this. But, this kind of fix is just wrong completely. No reason to keep this thing. Make an issue.

Zangetsu38:
you have no read description of pr here, issue exist

PR #1517, opened 2021-11-24
external: update dynarmic submodule.

Zangetsu38:


sunho:
we should use our fork. it contains vita3k specific fixes.

Zangetsu38:
> we should use our fork. it contains vita3k specific fixes.

Oh, i see, sorry, i have no see special commit for that

PR #1526, opened 2021-11-30
modules/SceGxm: fix regression in loop of both StateSetAllUniformBuffers

Zangetsu38:
# About:
- fix loop set value too much, max size is 14, so have only 14 array available (0-13), fix crash with dynarmic error on load ingame

image


pent0 (approved):
Note to self: These functions never allowed to set BUFFER14, need separate functions

PR #1523, opened 2021-11-28
Implement all sceDisplayWaitVblank function

Zangetsu38:
# About:
-  all function of sceDisplayWaitVblank get correctly implemented.

PR #1539, opened 2021-12-07
renderer: Correct front and back disabled value.

Zangetsu38:


sunho:
This is going to discard the depth. Does this not regress asphalt?

sunho:
It's messy tho, proper fix is just write previous color in glsl.

pent0:
Yes as I say before, why did you forget so fast :(

sunho:
Ah didn't saw it was your commit.

hobyst:
I've tested the commit and it doesn't fix the issue, at least not in PlayStation Vita Pets.

Zangetsu38:
> I've tested the commit and it doesn't fix the background blinking issue, at least not in PlayStation Vita Pets.

normal, request also my pr of #1523 , dbz have same, it very flashing without it, with get litterally black screen

Zangetsu38:
hmmm, to myserious for this

PR #1547, opened 2021-12-11
build: add suport of VS 2022.

Zangetsu38:
# About:
- add support of Visual Studio 2022 (VC143), with update boost on 1.78, and with update gen window

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -44,7 +44,7 @@ jobs:
             extra_cmake_args: -DCMAKE_TOOLCHAIN_FILE=./cmake/toolchain/linux-x64.cmake -DUSE_DISCORD_RICH_PRESENCE=OFF
           - os: windows-latest
             cache_path: ~\AppData\Local\Mozilla\sccache
-            extra_cmake_args: -DBOOST_ROOT=C:\hostedtoolcache\windows\Boost\1.72.0\x86_64 -DCMAKE_TOOLCHAIN_FILE=./cmake/toolchain/windows-x64.cmake -DCMAKE_C_COMPILER_LAUNCHER=sccache -DCMAKE_CXX_COMPILER_LAUNCHER=sccache
+            extra_cmake_args: -DBOOST_ROOT=C:\hostedtoolcache\windows\Boost\1.78.0\x86_64 -DCMAKE_TOOLCHAIN_FILE=./cmake/toolchain/windows-x64.cmake -DCMAKE_C_COMPILER_LAUNCHER=sccache -DCMAKE_CXX_COMPILER_LAUNCHER=sccache
 
     steps:
       - name: Set up build environment (macos-latest)
@@ -61,9 +61,9 @@ jobs:
 
       - name: Set up build environment (windows-latest)
         run: |
-          $Url = "https://sourceforge.net/projects/boost/files/boost-binaries/1.72.0/boost_1_72_0-msvc-14.2-64.exe"
+          $Url = "https://sourceforge.net/projects/boost/files/boost-binaries/1.78.0/boost_1_78_0-msvc-14.2-64.exe"
           (New-Object System.Net.WebClient).DownloadFile($Url, "$env:TEMP\boost.exe")
-          Start-Process -Wait -FilePath "$env:TEMP\boost.exe" "/SILENT","/SP-","/SUPPRESSMSGBOXES","/DIR=C:\hostedtoolcache\windows\Boost\1.72.0\x86_64"
+          Start-Process -Wait -FilePath "$env:TEMP\boost.exe" "/SILENT","/SP-","/SUPPRESSMSGBOXES","/DIR=C:\hostedtoolcache\windows\Boost\1.78.0\x86_64"
           Invoke-Expression (New-Object System.Net.WebClient).DownloadString('https://get.scoop.sh')
           scoop install ninja sccache --global
           echo "${env:PATH}" >> ${env:GITHUB_PATH}
--- a/building.md
+++ b/building.md
@@ -4,7 +4,7 @@ Vita3K uses CMake for its project configuration and generation and should in the
 
 Target OS | Host OS | [Project generator](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html) (`-G`) | C/C++ Compiler (`-DCMAKE_C_COMPILER` and `-DCMAKE_CXX_COMPILER`)
 --- | --- | --- | ---
-Windows | Windows | Visual Studio / MSBuild | `cl` a.k.a Microsoft Visual C/C++ Compiler (included as a part of the [Build Tools for Visual Studio 2019](https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2019) and the ` Dekstop development with C++` Visual Studio workload)
+Windows | Windows | Visual Studio / MSBuild | `cl` a.k.a Microsoft Visual C/C++ Compiler (included as a part of the [Build Tools for Visual Studio 2022](https://aka.ms/vs/17/release/vs_BuildTools.exe) and the ` Desktop development with C++` Visual Studio workload)
 macOS | macOS | Xcode | Clang (`clang` and `clang++`)
 Linux | Linux | Ninja | Clang (`clang` and `clang++`)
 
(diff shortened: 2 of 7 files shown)


Zangetsu38:
Help request, no succes myself, and it no found path of "VsDevCmd.bat", for me is good inside it
image

hobyst:
@Zangetsu38 For the issues you are having in Visual Studio, try [downloading](https://cmake.org/download/) the latest release of CMake. Visual Studio 2022 compatibility was officially brought to CMake in CMake 3.21. As far as I know, VS comes bundled with CMake 3.20, so you might be having issues due to that. After the update, delete your `build` directory entirely, enter a Visual Studio 2022 x64 Native Tools Command Prompt, run `cmake --version` to be sure that the new CMake version is picked up instead of the one bundled with Visual Studio and run the `gen-windows.bat` script again.

As for the Boost upgrade, I'll try to compile it by myself on Linux and reply back with whether it works or not.

Zangetsu38:
@hobyst i have already cmake 3.2.1 and i no have problème for compile boost is fine, problème is for compile emu project look screen

Zangetsu38:
@hobyst mast screen you see is for compile inside github action, i have set new path of visual studio 2022 but check log on it, file no found

hobyst:
@Zangetsu38 Try removing the toolchain file argument from the CMake configure command. I think Microsoft changed the name of the MSVC compiler binary and they haven't really said anything about it. The GitHub Action is failing because is trying to find the `cl` compiler but it isn't there anymore, and it might be due to the VS 2022 update.

hobyst (dismissed):

gen-windows.bat
```diff
@@ -1,5 +1,5 @@
 @echo off
 
-REM Generate project files for Visual Studio 2019
-call cmake -S . -B build-windows -G "Visual Studio 16 2019" -DCMAKE_TOOLCHAIN_FILE=./cmake/toolchain/windows-x64.cmake
+REM Generate project files for Visual Studio 2022
+call cmake -S . -B build-windows -G "Visual Studio 17 2022" -DCMAKE_TOOLCHAIN_FILE=./cmake/toolchain/windows-x64.cmake
```
djtuBIG-MaliceX, line 4: Recommend remove `-G "Visual Studio 17 2022"` and let whatever visual studio version compiler installed be used.

.github/workflows/c-cpp.yml
```diff
@@ -33,7 +33,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-latest, ubuntu-latest, windows-latest]
+        os: [macos-latest, ubuntu-latest, windows-2022]
```
djtuBIG-MaliceX, line 36: Something tells me this should not be changed.

.github/workflows/c-cpp.yml
```diff
@@ -59,15 +59,15 @@ jobs:
           sudo apt-get -y install ccache libboost-filesystem-dev libboost-program-options-dev libboost-system-dev libgtk-3-dev libsdl2-dev ninja-build
         if: matrix.os == 'ubuntu-latest'
 
-      - name: Set up build environment (windows-latest)
+      - name: Set up build environment (windows-2022)
```
djtuBIG-MaliceX, line 62: Should this really be changed?

building.md
```diff
@@ -4,7 +4,7 @@ Vita3K uses CMake for its project configuration and generation and should in the
 
 Target OS | Host OS | [Project generator](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html) (`-G`) | C/C++ Compiler (`-DCMAKE_C_COMPILER` and `-DCMAKE_CXX_COMPILER`)
 --- | --- | --- | ---
-Windows | Windows | Visual Studio / MSBuild | `cl` a.k.a Microsoft Visual C/C++ Compiler (included as a part of the [Build Tools for Visual Studio 2019](https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2019) and the ` Dekstop development with C++` Visual Studio workload)
+Windows | Windows | Visual Studio / MSBuild | `cl` a.k.a Microsoft Visual C/C++ Compiler (included as a part of the [Build Tools for Visual Studio 2022](https://aka.ms/vs/17/release/vs_BuildTools.exe) and the ` Dekstop development with C++` Visual Studio workload)
```
djtuBIG-MaliceX, line 7: s/Dekstop/Desktop

building.md
```diff
@@ -56,9 +56,9 @@ From there, the project will be ready to build right from the Visual Studio UI.
 
 - Generate the project:
   ```cmd
-  cmake -S . -B build-windows -G "Visual Studio 16 2019" -DCMAKE_TOOLCHAIN_FILE=./cmake/toolchain/windows-x64.cmake
+  cmake -S . -B build-windows -G "Visual Studio 17 2022" -DCMAKE_TOOLCHAIN_FILE=./cmake/toolchain/windows-x64.cmake
```
djtuBIG-MaliceX, line 59: explicit generator not necessary. just use whatever VS is installed on the machine.

djtuBIG-MaliceX (changes requested):

gen-windows.bat
Zangetsu38: Eu what do you said ? is obligatory for cake

.github/workflows/c-cpp.yml
Zangetsu38: Is obligatory too

.github/workflows/c-cpp.yml
Zangetsu38: Same

building.md
Zangetsu38: ?

building.md
Zangetsu38: Explicit is request, cmake no works without this

gen-windows.bat
djtuBIG-MaliceX: cmake can autodetect latest MSVC in the developer tools command prompt/powershell. there's no need to explicitly specify the generator.

building.md
djtuBIG-MaliceX: Spelling.

building.md
djtuBIG-MaliceX: Yes it does.

gen-windows.bat
djtuBIG-MaliceX: image

hobyst: > Recommend remove `-G "Visual Studio 17 2022"` and let whatever visual studio version compiler installed be used. Don't need to specify a specific VS version in my opinion.

`-G` argument is not for compilers but for [project generators](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html), and they aren't the same thing. Even if CMake would normally default to use MSBuild as the project generator on a VS Command Prompt, it also depends on the compilation environment. Since the environment is not even controlled by the devs but rather by GitHub, leaving an explicit declaration to the project generator is safer than not including it.

Zangetsu38: i see, i have all time see dev said explicit version, and cmake gui also request so 🤔

djtuBIG-MaliceX: OK I wasn't aware this was being used as an artifact for GitHub Actions CI.

.github/workflows/c-cpp.yml
hobyst: > Something tells me this should not be changed.

Look at the specification of included packages for the GitHub Actions [`windows-latest` // `windows-2019`](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md#visual-studio-enterprise-2019) and [`windows-2022`](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2022-Readme.md#visual-studio-enterprise-2022) virtual environments. `windows-2022` it's the only one set up with Visual Studio 2022.

djtuBIG-MaliceX: Yeah cool. I guess that's more a GitHub naming issue. Just saw the detail in https://github.com/actions/virtual-environments/issues/4488 so all good.

gen-windows.bat
Zangetsu38: small see, totaly forget, but after moving on github action, it no using anymore this gen-window, is only for Windows user/dev, for compile it only using indicztion inside workflow

building.md
Zangetsu38: oki, thx fixed, i have no see this typo error

1whatleytay (approved):
CI likes it so I will assume its perfect. I don't remember the problem that required that `choice` prompt...

Zangetsu38:
> 

it using workfllow now, cmd bad now is only for user :D

PR #1549, opened 2021-12-11
Profiler: Replace Microprofile with Tracy

hobyst:
The purpose of this PR is to replace the now broken Microprofile profiler implementation with a completely revamped one using the Tracy profiler. As of today, this PR only offers a peak at what would be possible and my goal is to be able to implement Tracy on the entire emulator to know exactly how each game interacts with the emulator and its different modules (audio, graphics renderer, HLE kernel modules...).

# How it works

Tracy uses a server-client model to operate, where the client is the application that reports profiling information (in this case, Vita3K) and a server, which receives and displays the profiling information, both needing to use the same version of the Tracy protocol to work. The client application compiles and links against `tracy/TracyClient.cpp` while the server is built as a project-agnostic standalone application.

Thanks to the `TRACY_ENABLE` compile definition, even if Tracy is linked as a logical target inside CMake, Tracy will only get compiled if this compile definition is defined. The CMake build system has already been setup to only compile Tracy in `Debug` and `RelWithDebInfo` builds.

# How to use it

Due to the constant updates that Tracy receives on its `master` branch and after discussion with @sunho, it was decided to go with the Tracy stable releases.
1. Download the Tracy server application that matches the version that Vita3K implements (which is [v0.8.1](https://github.com/wolfpld/tracy/releases/tag/v0.8.1)) and compile Vita3K in one of the builds that include debugging information (`RelWithDebInfo` is preferred, as it will give results closer to `Release` builds).
2. Run the Tracy server and Vita3K, the order doesn't matter.
3. After a few seconds, the Tracy server should automatically detect the Vita3K instance running and allow you to hook the server to it. As soon as the connection is established, the Tracy server will start to collect and display profiling information coming from Vita3K.

As a quick tip, the bar on the top of the Tracy server UI there's a pause button that is used to "stop" or "resume" the UI of the server from following and showing the information stream in real-time so that you can start to navigate the profiling report independently of at which point the execution of the client is. Please notice this is just a visual thing and profiling recollection only stops when the connection between server and client breaks, either by the client ending execution or the user manually cutting the connection on the server. For more information on how to use it, check the [manual](https://github.com/wolfpld/tracy/releases/download/v0.8.1/tracy.pdf).

Vita3K also implements the `TRACY_ON_DEMAND` compiler definition which prevents unnecessary memory usage when running a debugging build without really intending to use Tracy by only starting to collect profiling data after a connection with a Tracy server has been established. To collect profiling information right from the start of the execution of Vita3K:
1. Run the server before running Vita3K.
 2. Force the server to connect to `127.0.0.1` 
 3. After the server has started to wait for a connection, run Vita3K.
 
 # Current status of the implementation

Right now, Vita3K utilizes the ability of Tracy to track multiple frame streams to track three of them: the actual host frame stream, the host audio stream and the emulated frame buffer. These will be displayed on top of the trace graph.

In terms of scopes, it tracks the main function, the game rendering and UI rendering loops as well as every call to HLE module functions. Please note that the way tracing of HLE functions is implemented is "fake", as it doesn't track every function as a scope but rather it "tricks" the visual representation in the Tracy graph to change the name of the scope as HLE functions are invoked. This results on an accurate visual representation of the trace without having to do explicit scoping for each function, but it renders Tracy's statistic analysis features useless for HLE functionality. The intention is to migrate to per-function scoping as time goes by.

"Advanced" profiling is also possible by logging arguments on every call by serializing them and sending them to Tracy as additional data. To check how this can be implemented, check commit `SceAudio: Implement advanced Tracy profiling` for reference, but in short you need to:
- Register the module as an available module for advanced profiling in `vita3k/config/include/config/state.h`.
- Abstract all module-specific data structures and other type definitions in the header file of the module. That way, the Tracy serialization functions you create will also be able to use it without having to copy them into the file.
- Set up a `.cpp` and `.h` file with the functions that will be called every single time the HLE module function is called by the emulator to serialize its arguments and send the info to Tracy.
- Add the `.cpp` containing the serialization functions to the build system. The header doesn't need to be included as is only intended to be used by the source file of the module itself, and therefore both files will be on the same folder. Header files that are in the same folder as the source code files that include them are automatically detected and compiled without any more indications. This behavior is a standard for C and C++ compilers.

The final result will look something like this, a timed call with arguments of the function you are trying to profile:
image

On the Vita3K settings menu it is possible to control which modules are logged into the trace (except for the `renderer` and `shader` components, since the `host` status isn't shared with any of their functions so it has to be enabled in code in `vita3k/renderer/include/renderer/profile.h` and `vita3k/shader/include/shader/profile.h`).

image

# Future plans

~~My intention is to keep improving this PR over time to be able to trace as much as possible of the emulator, as well as use Tracy's OpenGL and Vulkan (after the render engine of the emulator is reworked to use it) integration. Tracy also supports zone filtering on runtime, so certain zones/scopes can be included or excluded from the profiling reports. The idea would be to use this to group zones into logical groups like "rendering", "audio", "kernel", "file loading"... so that developers could use the emulator's UI to enable and disable and focus the profiling reports on the aspect(s) of the emulator they want to profile. But before that, I would like to know which profiling groups would make sense to create and which functions or scopes should be included in them (ex. the "rendering" group could include native and SceGxm rendering functions).~~

Now that the foundation for all the integration with the profiler is built, any developer can casually log just one specific function of their choice or implement advanced profiling for certain parts of the emulator as needed.

For the OpenGL and Vulkan integrations, that can be worked on in the future or by another developer.

Please feel free to provide any kind of feedback about this should be handled and/or implemented!

# Screenshots of Tracy reports

image

image


sunho:
Is this still not ready?

Zangetsu38:
> Is this still not ready?

he have said is ready for  review

sunho:
To my knowledge noone is using microprofile, so it generally looks good to me.

external/CMakeLists.txt
```diff
@@ -304,4 +304,42 @@ set_property(TARGET xxhash PROPERTY FOLDER externals)
 # Soundtouch audio processing library for timing, pitch and playback rate manipulation
 add_subdirectory(soundtouch)
 set_property(TARGET SoundTouch PROPERTY FOLDER externals)
-set_property(TARGET soundstretch PROPERTY FOLDER externals)
\ No newline at end of file
+set_property(TARGET soundstretch PROPERTY FOLDER externals)
+
+# Tracy
+option(TRACY_ENABLE "Enable profiling of the emulator using Tracy" ON)
```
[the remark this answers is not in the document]
hobyst: CI builds will never have Tracy as CMake already disables Tracy on `Release` and `MinSizeRel` builds by not defining the `TRACY_ENABLE` pre-processor macro. Also I might rename that CMake option since it's just to control whether the `module` (`vita3k/module`) CMake target links against Tracy or not so that HLE modules can be compiled without requiring the parent project to have Tracy support in their build system, like in vita2hos.

hobyst:
> To my knowledge noone is using microprofile, so it generally looks good to me.

I've revised everything just now and the shader recompiler is still requiring Microprofile to compile. I'll have to fix that.

external/CMakeLists.txt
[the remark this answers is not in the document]
sunho: gotcha

sunho:
Tell me when it's ready

vita3k/config/src/config.cpp
```diff
@@ -276,4 +277,27 @@ ExitCode init_config(Config &cfg, int argc, char **argv, const Root &root_paths)
     return Success;
 }
 
+bool is_tracy_advanced_profiling_active_for_module(std::vector<std::string> &active_modules, std::string module, int *index) {
```
hobyst, line 280: Should I warp this one in a `#ifdef TRACY_ENABLE ... #endif` check as well so it won't get compiled on Release builds or should I leave it as it is right now? The only code that executes it is in `vita3k/gui/src/settings_dialog.cpp` and that code is already set not to be compiled on Release builds.

vita3k/config/src/config.cpp
```diff
@@ -276,4 +277,27 @@ ExitCode init_config(Config &cfg, int argc, char **argv, const Root &root_paths)
...
+
+    // If index is being requested and the module name was found
+    if ((index != nullptr) && (result)) {
+        // Calculate the distance between the first element in the module and
+        *index = std::distance(active_modules.begin(), iterator);
+    }
+
+    return result;
+}
+
```
hobyst, line 302: Should I warp this one in a `#ifdef TRACY_ENABLE ... #endif` check as well so it won't get compiled on Release builds or should I leave it as it is right now? The only code that executes it is in `vita3k/gui/src/settings_dialog.cpp` and that code is already set not to be compiled on Release builds.

bookmist: I think yes. ifdef is a visible sign, that this is needed for another build configuration and not a dead code

hobyst:
@sunho Ready to merge.

bookmist:
Great work and great documentation. I hope description of this PR would not lost. So we definitely need this text or link to it in wiki or readme.
(review in process)

vita3k/renderer/src/gl/renderer.cpp
```diff
@@ -128,7 +128,7 @@ static void before_callback(const char *name, void *funcptr, int len_args, ...)
 #endif // MICROPROFILE_ENABLED
```
bookmist, line 128: microprofile remainings

vita3k/audio/src/audio.cpp
```diff
@@ -15,21 +15,19 @@
 // with this program; if not, write to the Free Software Foundation, Inc.,
 // 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
+#include "Tracy.hpp"
```
bookmist, line 18: Is it needed to add #ifdef TRACY_ENABLE here too?

vita3k/renderer/src/gl/renderer.cpp
hobyst: Yes, Microprofile references were left in some cases as a reminder to help keeping track of parts of the emulator that were once profiled with Microprofile but that haven't been implemented yet with Tracy, specially in the renderer and the shader recompiler of the emulator due to complexity. Since Tracy has advanced integration with OpenGL and Vulkan being capable of displaying frame buffers and other features, using regular profiling in the renderer wouldn't make much sense so I left the Microprofile code in purpose to keep which functions and parts of the renderer were profiled before so that they can be re-implemented at a later date either by me or by any other dev using the integration features that Tracy has for the graphics API.

vita3k/audio/src/audio.cpp
hobyst: The compiler define check I implemented in `module`, `modules`, `renderer` and `shader` are so that in case a third-party application wants to re-use that code from the emulator, the compiler will not complain about not being able to access the Tracy headers. This was heavily inspired by the fact that vita2hos reuses parts of the emulator, but mainly the renderer and the shader recompiler. The Microprofile implementation didn't guard the header includes in any of the emulator code, so I decided to guard the headers in the core components of the emulator so that projects like vita2hos can compile them without having Tracy in their source tree. Since the audio host implementation is specific to Vita3K, I don't think it will be a problem to leave it like that.

bookmist (approved):

sunho (approved):
nice work

PR #1550, opened 2021-12-12
live area: using app icon when gate icon no exist.

Zangetsu38:
# About:
- using like psvita when live area content is missing

# Result:
- reference
image
- result
image


1whatleytay (approved):
lgtm

PR #1553, opened 2021-12-14
gui: Prefer standard library algorithms

tcoyvwac:
Cleans up and abstracts some ranged-for loops for their standard library equivalent.

Happy to condense / squash into one commit. :smile_cat: 
Currently PR is in this commit form to show changeset history (before squash / merge).

### DIFF
--- a/vita3k/gui/src/content_manager.cpp
+++ b/vita3k/gui/src/content_manager.cpp
@@ -17,7 +17,9 @@
 
 #include "private.h"
 
+#include <algorithm>
 #include <boost/algorithm/string/trim.hpp>
+#include <boost/range/numeric.hpp>
 
 #include <gui/functions.h>
 
@@ -29,6 +31,19 @@
 #include <util/safe_time.h>
 
 namespace gui {
+namespace {
+template <typename T>
+auto get_recursive_directory_size(const T &path) {
+    const auto &path_list = fs::recursive_directory_iterator(path);
+    const auto pred = [](const auto acc, const auto &app) {
+        if (fs::is_regular_file(app.path()))
+            return acc + fs::file_size(app.path());
+        return acc;
+    };
+    return boost::accumulate(path_list, boost::uintmax_t{}, pred);
+}
+} // namespace
+
 void get_app_info(GuiState &gui, HostState &host, const std::string &app_path) {
     const auto APP_PATH{ fs::path(host.pref_path) / "ux0/app" / app_path };
     gui.app_selector.app_info = {};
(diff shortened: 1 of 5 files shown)


vita3k/gui/src/content_manager.cpp
```diff
@@ -29,6 +31,19 @@
...
+namespace {
+template <typename T>
+auto get_recursive_directory_size(const T &path) {
+    const auto &path_list = fs::recursive_directory_iterator(path);
+    const auto pred = [](const auto acc, const auto &app) {
+        if (fs::is_regular_file(app.path()))
+            return acc + fs::file_size(app.path());
+        return acc;
+    };
+    return boost::accumulate(path_list, 0ul, pred);
```
hobyst, line 43: Increasing dependency on Boost when `std::accumulate` already exists is redundant. Are there any advantages over using the `std` implementation?

vita3k/gui/src/content_manager.cpp
```diff
@@ -590,9 +590,7 @@ void draw_content_manager(GuiState &gui, HostState &host) {
                     content.second = false;
             }
         }
-        const auto is_enable = std::find_if(contents_selected.begin(), contents_selected.end(), [&](const auto &cs) {
-            return cs.second;
-        }) != contents_selected.end();
+        const auto is_enable = boost::algorithm::any_of(contents_selected, [&](const auto &cs) { return cs.second; });
```
hobyst, line 593: Increasing dependency on Boost when `std::algorithm::any_of` already exists is redundant. Are there any advantages over using the `std` implementation?

tcoyvwac:
I am happy to remove `boost` algorithms marked and use the `std` version.
Currently, the only reason for using the `boost` version is for code clarity / minimalism.

For future PRs, does this project accept using `boost::range`? This is quite a useful library which can do things that the `std` library cannot do (easily or at all).

hobyst:
> I am happy to remove `boost` algorithms marked and use the `std` version. Currently, the only reason for using the `boost` version is for code clarity / minimalism.
> 
> For future PRs, does this project accept using `boost::range`? This is quite a useful library which can do things that the `std` library cannot do (easily or at all).

As far as I know, there's no explicit policy on what implementations to use. But regarding to Boost, keep in mind that while CI builds use standard and already-compiled Boost distributions, developer builds and environments use [`Vita3K/ext-boost`](https://github.com/Vita3K/ext-boost). Check if the Boost library you want to use it's there and if it doesn't then PR to `ext-boost` adding the library before using it in Vita3K.

vita3k/gui/src/content_manager.cpp
tcoyvwac: For this function & case, there is one advantage / difference:
* `fs::recursive_directory_iterator` is accepted as a "first-class citizen" datatype and the special `fs` iterators are accepted "as is", just like `std` iterators.
* In the `std` version, you must use special `fs::{begin, end}()` free functions for `fs::recursive_directory_iterator` to work.

In this case, the `boost` version is very streamlined & minimalistic to use, making it less error-prone and understandable.

vita3k/gui/src/content_manager.cpp
tcoyvwac: There is no difference compared to the `std` version except its function signature / parameters.

tcoyvwac:
Humm, any reason for the failing build on Windows? It suggests something wrong with the `CMake` part of the CI, but I cannot see further detail & information.

Zangetsu38:
> Humm, any reason for the failing build on Windows? It suggests something wrong with the `CMake` part of the CI, but I cannot see further detail & information.

You need just look end log in compile, it said error caused it

tcoyvwac:
Ahh, was logged in, looking at the logs, but it was blank when checking for more details! ...But it seems I was having problems with GitHub's JS! :sweat_smile: Oh GitHub. Thanks for the assist @Zangetsu38 .

Zangetsu38:
for error
image

pent0 (approved):
Good place to use these functions, thanks! If you are good to merge please tell.

tcoyvwac:
Thanks for the encouragement and review! :smile_cat: I am happy for the PR to be (squashed? and) merged.

Zangetsu38:
> Thanks for the encouragement and review! 😸 I am happy for the PR to be (squashed? and) merged.

Sorry is done totaly forget ^^

Zangetsu38:
@tcoyvwac master is now broken after this pr
master branch no have include <boost/range/numeric.hpp>

PR #1555, opened 2021-12-15
Update version of SPIRV-Cross and glslang

bookmist:
It looks like we update dependencies, so why not to update glslang and SPIRV-Cross.

It's not the last version of glslang, but it's last version I can make working here.

Changes are precision, discard and spirv-version (if version is newer, then spirv-cross does not emit global variables)

### DIFF
--- a/vita3k/shader/include/shader/usse_utilities.h
+++ b/vita3k/shader/include/shader/usse_utilities.h
@@ -90,7 +90,7 @@ void make_for_loop(spv::Builder &b, spv::Id iterator, spv::Id initial_value_ite,
 
     b.setBuildPoint(&blocks.head);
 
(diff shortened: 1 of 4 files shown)


sunho:
I feel hesitatant to update spir-v cross, as we're using it in a non standard way. It's easy to regress as spriv cross author works on it to improve standard use case. (disassemble glsl compiler output) We're not really akeen to the spir-v cross internals, so regression on spir-v level can be frustrating.

scribam (dismissed):
Looks fine to me, the branch has a conflicting file though.
As sunho said, regression can happen but Vita3K is still an experimental emulator🙂

Zangetsu38:
need fix conflict here

bookmist:
conflict fixed, rebased to master
@scribam thank you very much for your attention to this PR

Zangetsu38 (approved):

PR #1557, opened 2021-12-17
gui: display draw shader compiled.

Zangetsu38:
# About:
- sugested and based on idea by @1whatleytay  after see is not possible display in real time compiling shaders message
- display in down left corner message for said have compiled one/few shaders.
- delay for display dialog is 2 sec.

# Result:
image
image


zh82 (approved):

1whatleytay (approved):
lgtm?

frangarcj (approved):

PR #1558, opened 2021-12-19
Implement pre-compile shaders cache.

Zangetsu38:
# About: 
- implement pre-compile shaders and program in boot game.
- Add progress bar during pre compile shader.
- remove useless shaderlog in io and app context
- move shader path in good path

# Result:
- Draw pre compile shaders progress bar
image

video demonstration
https://www.youtube.com/watch?v=bYAXP3wmZq4

### DIFF
--- a/vita3k/main.cpp
+++ b/vita3k/main.cpp
@@ -28,6 +28,7 @@
 #include <host/state.h>
 #include <modules/module_parent.h>
 #include <renderer/functions.h>
+#include <renderer/gl/functions.h>
 #include <shader/spirv_recompiler.h>
 #include <util/log.h>
 #include <util/string_utils.h>
@@ -71,7 +72,7 @@ int main(int argc, char *argv[]) {
                 fs::remove_all(fs::path(root_paths.get_pref_path()) / "ux0/app" / *cfg.delete_title_id);
                 fs::remove_all(fs::path(root_paths.get_pref_path()) / "ux0/addcont" / *cfg.delete_title_id);
                 fs::remove_all(fs::path(root_paths.get_pref_path()) / "ux0/user/00/savedata" / *cfg.delete_title_id);
-                fs::remove_all(fs::path(root_paths.get_pref_path()) / "shaderlog" / *cfg.delete_title_id);
+                fs::remove_all(fs::path(root_paths.get_base_path()) / "cache/shaders" / *cfg.delete_title_id);
             }
             if (cfg.pkg_path.has_value() && cfg.pkg_zrif.has_value()) {
                 LOG_INFO("Installing pkg from {} ", *cfg.pkg_path);
@@ -213,12 +214,6 @@ int main(int argc, char *argv[]) {
     if (host.io.title_id.find("PCS") != std::string::npos)
         host.app_sku_flag = get_license_sku_flag(host, host.app_content_id);
 
-    Ptr<const void> entry_point;
-    if (const auto err = load_app(entry_point, host, string_utils::utf_to_wide(host.io.app_path)) != Success)
-        return err;
-    if (const auto err = run_app(host, entry_point) != Success)
-        return err;
-
     if (cfg.console) {
         auto main_thread = host.kernel.threads.at(host.main_thread_id);
         auto lock = std::unique_lock<std::mutex>(main_thread->mutex);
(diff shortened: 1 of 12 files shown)


vita3k/main.cpp
```diff
@@ -213,6 +214,10 @@ int main(int argc, char *argv[]) {
     if (host.io.title_id.find("PCS") != std::string::npos)
         host.app_sku_flag = get_license_sku_flag(host, host.app_content_id);
 
+    // Pre-Compile Shader only for glsl, spriv is broken
+    if (!host.cfg.spirv_shader)
+        renderer::gl::pre_compile_programs(static_cast<renderer::gl::GLState &>(*host.renderer), host.base_path.c_str(), host.io.title_id.c_str(), host.renderer->shaders_count_compiled);
```
pent0, line 219: Rather then making a shader count compiled variable very unrelated to host, why not just stored it in renderer?

vita3k/renderer/include/renderer/types.h
```diff
@@ -148,6 +148,11 @@ struct FragmentProgram : ShaderProgram {
 struct VertexProgram : ShaderProgram {
 };
 
+struct ShaderHash {
```
pent0, line 151: std::pair is better. Maybe be better to name ProgramHash

vita3k/renderer/src/gl/compile_program.cpp
```diff
@@ -105,8 +105,210 @@ static SharedGLObject compile_spirv(GLenum type, const std::vector<std::uint32_t
...
+        size_t size;
+        shaders_hashs.read((char *)&size, sizeof(size));
+
+        // Check version of cache
+        size_t version_size;
+        shaders_hashs.read((char *)&version_size, sizeof(version_size));
+        std::vector<char> buffer(version_size);
+        shaders_hashs.read(buffer.data(), version_size);
+        const std::string versionInFile(buffer.begin(), buffer.end());
+        if (versionInFile != "1.00") {
```
pent0, line 123: You should make version being int variable. Read an integer then check version.

Ifstream fi;
Int version;
Fi.read(&version, sizeof(int32_t));
if (version!=1)
 .....

vita3k/renderer/src/gl/compile_program.cpp
```diff
@@ -105,8 +105,210 @@ static SharedGLObject compile_spirv(GLenum type, const std::vector<std::uint32_t
...
+            LOG_WARN("Current version of cache: {}, is outdated, recreate it.", versionInFile);
+            return false;
+        }
+
+        // Read Hashe info value
+        for (size_t a = 0; a < size; a++) {
+            auto read = [&shaders_hashs]() {
+                size_t size;
+
+                shaders_hashs.read((char *)&size, sizeof(size));
```
pent0, line 133: Should store it in binary instead of readable hex strimg, for smaller size!

pent0 (dismissed):

vita3k/main.cpp
Zangetsu38: i no understund yoiur ask here, is already inside renderer state here

vita3k/renderer/include/renderer/types.h
Zangetsu38: i have thinks some diferent, but here is better, pair is good, but only struct can said me what is frag/vert, 
and no, is wrong, here is ShaderHash, no Programhach, and ProgramHash already exist for program hashes so

vita3k/renderer/src/gl/compile_program.cpp
Zangetsu38: no idea how done this, this code is full based on app cache code writed by @1whatleytay

vita3k/renderer/src/gl/compile_program.cpp
Zangetsu38: i have used it for app cache,i prefer, i can set 1.01 like that, int is to limited

vita3k/renderer/include/renderer/types.h
pent0: There are hashes of two shader with two types, so better named something else then ShaderHash. I can accept the first one

vita3k/renderer/src/gl/compile_program.cpp
pent0: Just dont convert it to hex string in write

pent0: You can also ask taylor for more.

Zangetsu38: > Just dont convert it to hex string in write

no understand your message here, is not hex string here, is original string give by game in fragment_program.hash, i convert it hex only for create/read shader file name

pent0: Then cool. Why don't understand? Just yes or no.

Zangetsu38: sorry,  i just said i no understrund you ask, you speek me no register hex string, but is no hex string, for that i no understand your message

vita3k/renderer/src/gl/compile_program.cpp
Zangetsu38: done, i have move it to int

IllusionMan1212 (approved):

PR #1559, opened 2021-12-21
renderer: Fix PVRTII decompression

hobyst:
This PR fixes PVRTII texture decompression issue caused by uninitialized memory in the modulation parameter arrays.

It also includes a fix for a Clang-specific issue uncovered by the PVRTII decompression fix and caused due to compiler optimizations that makes texture caching fail. The issue is likely caused by some optimizations done by Clang conflicting with the way pointers are implemented in the texture caching system, so the whole cache system should be revised at some point to detect more possible points of failure related to memory protection but for now this should do the job.


hobyst:
Don't merge yet.

hobyst:
Ready to merge.

Zangetsu38 (approved):

PR #1572, opened 2021-12-28
vita3k: Fix compilation errors on the Arch AUR

EXtremeExploit:
#### What does this PR do?
This is the second part of fixing compilation problems in the Arch AUR, turns out that there were more errors than the string literal one, they look to be the same error as before but in different places, they do make sense tho, hopefully this is the last one. If possible i will try to get ownership of it if its a PKGBUILD error, but in the meantime, we should fix these inconsistent strings of the gui

### DIFF
--- a/vita3k/gui/src/settings.cpp
+++ b/vita3k/gui/src/settings.cpp
@@ -461,19 +461,19 @@ void draw_settings(GuiState &gui, HostState &host) {
                         const auto INFO_POS = ImVec2(280.f * SCALE.x, 30.f * SCALE.y);
                         ImGui::SetWindowFontScale(0.94f);
                         ImGui::SetCursorPosY(ImGui::GetCursorPosY() + INFO_POS.y);
-                        ImGui::TextColored(GUI_COLOR_TEXT, is_lang ? lang["name"].c_str() : "Name");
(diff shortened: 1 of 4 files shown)


vita3k/gui/src/settings.cpp
```diff
@@ -422,7 +422,7 @@ void draw_settings(GuiState &gui, HostState &host) {
                         const auto delete_str = is_lang ? lang["delete"] : "This theme will be deleted.";
                         const auto CALC_TEXT = ImGui::CalcTextSize(delete_str.c_str());
                         ImGui::SetCursorPos(ImVec2(POPUP_SIZE.x / 2 - (CALC_TEXT.x / 2.f), POPUP_SIZE.y / 2.f));
-                        ImGui::TextColored(GUI_COLOR_TEXT, delete_str.c_str());
+                        ImGui::TextColored(GUI_COLOR_TEXT, "%s", delete_str.c_str());
```
Zangetsu38, line 425: thx, some is already normaly fixed but waiting insidde my setting pr :/

Zangetsu38:
fix format

Zangetsu38 (approved):
thx sorry for that

EXtremeExploit:
np

PR #1580, opened 2022-01-03
Fixes for the crash issue when the time is uninitialized

iWumboUWumbo2:
I don't know if this is the best way to resolve the problem because as @Zangetsu38 mentioned, the GUI is very complicated but this at least mitigates the crashing.

### DIFF
--- a/vita3k/gui/src/gui.cpp
+++ b/vita3k/gui/src/gui.cpp
@@ -435,6 +435,7 @@ void init_home(GuiState &gui, HostState &host) {
     }
 
     get_time_apps(gui, host);
+    init_last_time_apps(gui, host);
 
     if (!gui.users.empty() && (gui.users.find(host.cfg.user_id) != gui.users.end()) && (is_cmd || host.cfg.auto_user_login)) {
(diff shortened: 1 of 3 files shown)


vita3k/gui/src/gui.cpp
```diff
@@ -435,6 +436,7 @@ void init_home(GuiState &gui, HostState &host) {
     }
 
     get_time_apps(gui, host);
+    init_last_time_apps(gui, host);
```
Zangetsu38, line 439: Some is fine but here is wrong, like i explain is complicated

sunho:
Is the edition to app_selector.cpp still needed after the init fix?

sunho:
by init fix I'm referring to last_time=0

Zangetsu38:
> by init fix I'm referring to last_time=0

Yes both no request, in normal mode value is obligatory 0 if not

Zangetsu38:
Code here no fix issue att all is more Skipped,  you obligiyary need run init last time after init user,i have this and founded how fix it quickely, thx for your works 👌

PR #1585, opened 2022-01-09
gui: add avg and max value with statistic of fps.

Zangetsu38:
# About:
- add avg, min abd max fps and line for statistic of fps

# Result:
image


pent0 (approved):
Sick

PR #1598, opened 2022-02-02
gui/theme: allow install custom theme and more..

Zangetsu38:
# About:
- request by user,  for can allow install custom theme with using drag and drop, for install archive is more complicated
- fix no create empty texture when image is not found with little refactor.
- fix invisible menu bar and indicator when alpha is missing in value inside theme.
- add browser icon, open vita3k website.
- add support of notice font color.
- refactor code of convert hex color to ImU32 value.
- refactor get manual page.
- add support of unicode of theme content id.

thx to @kd-11 for help me on code for convert color


KorewaWatchful (approved):
👍

PR #1599, opened 2022-02-07
shader/translator: refactor using standard algorithms

tcoyvwac:
Refactored to prefer `std::any_of` over raw range-for loop.


Zangetsu38 (approved):

pka4916:
image

PR #1603, opened 2022-02-11
gui/initial setup: fix some string.

Croden1999:
- Fix small issue.


Croden1999:
ahh, some undisplayed text translations will be restored to English. sad.

Zangetsu38:
thx, but missing lang here is normal

Croden1999:
> thx, but missing lang here is normal

Mr. Zangetsu. i just translate a part string. thx a lot.

PR #1618, opened 2022-02-22
codec/pcm: fix intermediate buffer size calculation for adpcm decoder

bookmist:
fix ngs error in RIDGE RACER and other games, which use adpcm codec


Zangetsu38 (approved):
LGTM

PR #1621, opened 2022-02-24
gui: allow erase app when is deleted outside emu.

Zangetsu38:
# About:
- allow erase app when is deleted outside emu with using update button.

### DIFF
--- a/vita3k/gui/src/app_context_menu.cpp
+++ b/vita3k/gui/src/app_context_menu.cpp
@@ -73,8 +73,8 @@ static bool get_update_history(GuiState &gui, HostState &host, const std::string
     return !update_history_infos.empty();
 }
 
-std::vector<TimeApp>::iterator get_time_app_index(GuiState &gui, HostState &host, const std::string app) {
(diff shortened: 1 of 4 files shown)


vita3k/gui/src/app_context_menu.cpp
```diff
@@ -342,7 +368,9 @@ void draw_app_context_menu(GuiState &gui, HostState &host, const std::string &ap
                     open_path(SHADER_LOG_PATH.string());
                 ImGui::EndMenu();
             }
-            if (!host.cfg.show_live_area_screen && ImGui::BeginMenu("Live Area")) {
+            if (ImGui::BeginMenu("Live Area")) {
+                if (ImGui::MenuItem(gui.lang.live_area[START].c_str()))
```
EXtremeExploit, line 372: use string literal here
```suggestion
                if (ImGui::MenuItem("%s", gui.lang.live_area[START].c_str()))
```

Zangetsu38: oh i have forget this pr, i have hurt for corect solution lol

PR #1639, opened 2022-03-07
gxm: split public SceGxm types and internal+opaque SceGxm types

xerpi:
With this change, we are separating the public SceGxm enums/structs from Vita3K's internal implementation of SceGxm opaque types, and also internal Vita3K structs/contexts.

This allows dropping the dependency of `gxm` from `mem` so that `gxm` becomes more generic with very few dependencies to Vita3K.

Vita3K's internal implementation of SceGxm opaque types are moved to `vita3k/renderer/include/renderer/gxm_types.h`.

### DIFF
--- a/vita3k/gxm/CMakeLists.txt
+++ b/vita3k/gxm/CMakeLists.txt
@@ -13,5 +13,5 @@ add_library(
 )
 
 target_include_directories(gxm PUBLIC include)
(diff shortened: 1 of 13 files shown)


Zangetsu38:
fail on format, check log on it for see where fix it

1whatleytay (approved):

PR #1646, opened 2022-03-16
shader: minor fixes

xerpi:
Minor fixes to `shader/`. See individual commits.

### DIFF
--- a/vita3k/shader/src/spirv_recompiler.cpp
+++ b/vita3k/shader/src/spirv_recompiler.cpp
@@ -1528,7 +1528,7 @@ std::string convert_gxp_to_glsl(const SceGxmProgram &program, const std::string 
 
 void convert_gxp_to_glsl_from_filepath(const std::string &shader_filepath) {
     const fs::path shader_filepath_str{ shader_filepath };
-    fs::ifstream gxp_stream(shader_filepath_str, fs::ifstream::binary);
+    std::ifstream gxp_stream(shader_filepath, std::ifstream::binary);
 
     if (!gxp_stream.is_open())
         return;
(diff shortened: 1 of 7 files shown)


Zangetsu38:
check log for more detail
image

vita3k/shader/src/spirv_recompiler.cpp
```diff
@@ -1528,7 +1528,7 @@ std::string convert_gxp_to_glsl(const SceGxmProgram &program, const std::string
 
 void convert_gxp_to_glsl_from_filepath(const std::string &shader_filepath) {
     const fs::path shader_filepath_str{ shader_filepath };
-    fs::ifstream gxp_stream(shader_filepath_str, fs::ifstream::binary);
+    std::ifstream gxp_stream(shader_filepath_str.filename().string(), std::ifstream::binary);
```
Zangetsu38, line 1531: no sur have sense here, can directly using shader_filepath if you no using fs::path att all 🤷‍♂️

xerpi: Yes, I also noticed that but for some reason, I thought you might like this way better!
I'll change it to directly pass `shader_filepath`.
`fs` might be `boost::filesystem`, but since we already have `ifstream` in `std::`I think it's better to use it so one place less that uses `boost` :)

vita3k/shader/src/spirv_recompiler.cpp
```diff
@@ -1528,7 +1528,7 @@ std::string convert_gxp_to_glsl(const SceGxmProgram &program, const std::string
 
 void convert_gxp_to_glsl_from_filepath(const std::string &shader_filepath) {
     const fs::path shader_filepath_str{ shader_filepath };
```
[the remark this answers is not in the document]
Zangetsu38: Ah yes is later ok

pent0 (approved):

PR #1657, opened 2022-03-31
kernel\thread, modules: fix TLS in callback calls

bookmist:
Fix crashes on some games, using thread local storage in SceGxmShaderPatcher callbacks


Zangetsu38 (approved):
LGTM for me in my all test, fix crash oin some game, fix regression, if any other dev see probleme can improve code later

PR #1665, opened 2022-04-11
gui: Fix some elemements not using string literals

EXtremeExploit:
We should make these warnings/messages in windows be treated as errors so that it doesn't happen again and linux can compile


Zangetsu38 (approved):
Oupsy

Zangetsu38:
ah, no my fault here,, is user have coded it, but i have no see in review, my bad xD

Croden1999:
is my mistake, sorry.

PR #1670, opened 2022-04-18
modules/SceAudiodec: Handle multiple init and term library calls with different codecs

Macdu:
Fix a crash that happens when multiple calls to sceAudiodecInitLibrary and sceAudiodecTermLibrary are done.
sceAudiodecInitLibrary is called once for each codec used. However the object AudioCodec is discarded from the object store after a call to sceAudiodecTermLibrary, even though only decoders associated with the codec terminated should be discarded.

As an example the following code is valid but crashes in Vita3K:

sceAudiodecInitLibrary(codec1)
sceAudiodecInitLibrary(codec2)
sceAudiodecTermLibrary(codec2)
Use codec1 -> crash : AudioCodec could not be found

How I fixed it is by only removing decoders associated with the codec passed as a parameter to sceAudiodecTermLibrary when it is called. 

This fix allows Tales of Hearts R to get to the startup menu (and even ingame with AT9 audio decoder disabled)

### DIFF
--- a/vita3k/modules/SceAudiodec/SceAudiodecUser.cpp
+++ b/vita3k/modules/SceAudiodec/SceAudiodecUser.cpp
@@ -33,10 +33,13 @@ enum {
 
 typedef std::shared_ptr<DecoderState> DecoderPtr;
 typedef std::map<SceUID, DecoderPtr> DecoderStates;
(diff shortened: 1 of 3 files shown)


Zangetsu38 (approved):
LGTM

PR #1673, opened 2022-04-19
display: Fix display waits not processing all waits.

Zangetsu38:


EXtremeExploit:
imo vblank frequency should be a slider so you aren't stuck with 60 or 120, other values can work better for other games and by doing this one is able to have the same vblank as the screen Hz, slider can be from 1 to 480 in increments of 30 like RPCS3 does

Zangetsu38:
> imo vblank frequency should be a slider so you aren't stuck with 60 or 120, other values can work better for other games and by doing this one is able to have the same vblank as the screen Hz, slider can be from 1 to 480 in increments of 30 like RPCS3 does

Is too early for that

EXtremeExploit:
literally change from menu item to a simple slider and the value of the slider be the vblank freq, not really complicated

Zangetsu38:
> literally change from menu item to a simple slider and the value of the slider be the vblank freq, not really complicated

You have taly no understund me here 😅

bookmist (approved):

PR #1678, opened 2022-04-20
gdbstub: Fix gdbstub implementation

Macdu:
This pull request fixes gdbstub.
Right now it works on my computer with mingw64 gdb-multiarch.

I think that more testing is required before merging.


lometsj:
test on msys2 mingw gdb-mutliarch
image
after command “target remote :2159”
it seems like cannot break after attach

Alriightyman:
So, I added these changes to the current code base.  For me, the threads lock up waiting on.. something.   One gets to `run_guest_function` and waits on it.  Another thread, goes into wait mode thorugh `to_do` and sits there forever. 

As for debugging using ghidra,  vita3k gets the connection but ghidra seems to be forever 'connecting..' to it.   I'm not sure if that's a ghidra related issue or not.  

Has there been any progress on this?  Looks like nobody has touched it in over a year. 

For reference, I was running this on Windows 10.  I'd be willing to test this more on Windows, if need be.  I'm afraid I'm not that great with threading (or do I know this code base well enough) to really try and fix it myself.

bookmist:
@Alriightyman I was tryed to use ghidra debugger with Vita3k and have no success. I can connect to vita3k from ghidra and run stopped program, but no more. It seems that support of gdb stubs (for different emulators) is not a priority for ghidra developers.

Alriightyman:
So, I've come back to this code and added the changes to the latest version for Windows.  Using arm-none-eabi-gdb.exe version, I can remote connect successfully.   I add a breakpoint using something like `b *0x810b4bde`.   It gdb adds the break point.  I type `c` for continue and visual studio debugger throws an exception in function `Debugger::add_breakpoint()`.  "read access violation."  The line in question it this: 
`std::memcpy(&bk.data, &mem.memory[addr], sizeof(THUMB_BREAKPOINT));`
I know the issue is reading `mem.memory[addr]`.  In fact, any value I put in for addr throws this exception.  Because the actual debug version is broken - can't seem to get that to run after building, I  can't really look into this more.  @Macdu, did you see any issues related to this when you were testing it?  Or anyone else for that matter?
Here is the info from gdb:
```(gdb) target remote localhost:2159
Remote debugging using localhost:2159
warning: No executable has been specified and target does not support
determining executable automatically.  Try using the "file" command.
warning: multi-threaded target stopped without sending a thread-id, using first non-exited thread
0x00000000 in ?? ()
(gdb) b *0x810b4bde
Breakpoint 1 at 0x810b4bde
(gdb) c
Continuing.```

Some extra info to point out:
I never really use gdb, so if I am missing some bits when starting it, that could be helpful.
I'm a bit new to Vita software and in turn, Vita3k's codebase.  So, any help/guidance would be greatly appreciated.

**Comment by Macdu** [comment]
Well, to be honest, the gdb implementation kind of needs to be reimplemented from scratch. Substituting an instruction with a breakpoint causes many issues.

**Comment by Alriightyman** [comment]
> Well, to be honest, the gdb implementation kind of needs to be reimplemented from scratch. 

Well... That's lame. 

> Substituting an instruction with a breakpoint causes many issues.

Oh, so that's what the is doing..... why do it that way?  How should that be done?

**Comment by VelocityRa** [comment]
@Macdu Can at least some of the changes here be merged? I'm willing to PR them myself but not sure which ones are problematic.

**Comment by Macdu** [comment]
I mean, these changes allow gdb to actually detect Vita3K's gdb server. However, the current implementation is quite broken, any program will get corrupted after resuming from any breakpoint. So right now you can basically break once and that's it, which I don't think is really worth it.

Now that we have thread safe cache invalidation, a good way to implement instruction break would be to add the break call in the dynarmic IR generation, but that would require rewriting this whole part. We could also implement data read/write break by disabling the fastmem ptr, but that would also require many additional changes.

**Comment by VelocityRa** [comment]
@Macdu Oh, but I saw [`cmd_remove_breakpoint`](https://github.com/Vita3K/Vita3K/blob/master/vita3k/gdbstub/src/gdb.cpp#L601-L614) was implemented (seems to invalidate the jit cache and everything) so I assumed that worked. Did it work at some point and is now broken then?

**Comment by VelocityRa** [comment]
@Macdu Also, which GDB client am I supposed to use with this?
I saw there's `gdb-multiarch`, but trying wth Vita SDK's `arm-vita-eabi-gdb`, with this command: `arm-vita-eabi-gdb game.elf -ex 'target remote 127.0.0.1:2159'` and getting this:
```
[01:15:36.656] |I| [server_listen]: GDB Server Received Connection
[01:15:36.657] |I| [server_next]: GDB Server Recognized Command as qSupported. qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContSupported+;QThreadEvents+;no-resumed+
[01:15:36.658] |I| [server_next]: GDB Server Recognized Command as vMustReplyEmpty. vMustReplyEmpty
[01:15:36.698] |I| [server_next]: GDB Server Recognized Command as H. Hg0
[01:15:36.740] |I| [server_next]: GDB Server Recognized Command as qTStatus. qTStatus
[01:15:36.780] |I| [server_next]: GDB Server Recognized Command as q. qTfV
[01:15:36.781] |I| [cmd_unimplemented]: GDB Server: Unimplemented Packet. qTfV
[01:15:36.822] |I| [server_next]: GDB Server Recognized Command as ?. ?
[01:15:36.863] |I| [server_next]: GDB Server Recognized Command as qfThreadInfo. qfThreadInfo
[01:15:36.904] |I| [server_next]: GDB Server Recognized Command as qsThreadInfo. qsThreadInfo
[01:15:36.945] |I| [server_next]: GDB Server Recognized Command as qsThreadInfo. qsThreadInfo
[01:15:36.986] |I| [server_next]: GDB Server Recognized Command as qAttached. qAttached
[01:15:37.027] |I| [server_next]: GDB Server Recognized Command as H. Hc-1
[01:15:37.027] |I| [cmd_set_current_thread]: GDB Server Deprecated Continue Option 'c'
[01:15:37.068] |I| [server_next]: GDB Server Recognized Command as qC. qC
[01:15:37.109] |I| [server_next]: GDB Server Recognized Command as q. qOffsets
[01:15:37.109] |I| [cmd_unimplemented]: GDB Server: Unimplemented Packet. qOffsets
[01:15:37.149] |I| [server_next]: GDB Server Recognized Command as g. g
[01:15:37.192] |I| [server_next]: GDB Server Recognized Command as qfThreadInfo. qfThreadInfo
[01:15:37.232] |I| [server_next]: GDB Server Recognized Command as qsThreadInfo. qsThreadInfo
[01:15:37.274] |I| [server_next]: GDB Server Recognized Command as qsThreadInfo. qsThreadInfo
[01:15:37.315] |I| [server_next]: GDB Server Recognized Command as m. m0,4
[01:15:37.357] |I| [server_next]: GDB Server Recognized Command as q. qSymbol::
[01:15:37.357] |I| [cmd_unimplemented]: GDB Server: Unimplemented Packet. qSymbol::
[01:15:37.398] |I| [server_next]: GDB Server Recognized Command as qTStatus. qTStatus
[01:15:37.438] |I| [server_next]: GDB Server Recognized Command as q. qTfP
[01:15:37.439] |I| [cmd_unimplemented]: GDB Server: Unimplemented Packet. qTfP
[01:15:39.848] |I| [server_next]: GDB Server Recognized Command as m. m0,4
[01:15:41.701] |I| [server_next]: GDB Server Recognized Command as p. p19
```
So it's a bit iffy but it does establish a connection because [this PR](https://github.com/Vita3K/Vita3K/pull/3497) fixed it recently.
My plan is to add support for the Ghidra client.
I'm on the Discord server if you want to help/collaborate and prefer chatting there.

VelocityRa:
Oh, I think [this](https://github.com/Vita3K/Vita3K/pull/3497/files#diff-2c888699cb142f1d3daeb678eb5031de69f308dde5b68d3cadbc383eb103115dR302-R303) was maybe wrong compared to your `to_do = ThreadToDo::wait;` threading fix. Not familiar with that code anymore so not sure.

VelocityRa:
Now hitting a stack overflow in dynarmic when a breakpoint triggers.

VelocityRa:
Breakpoints don't seem to work on Unicorn either, getting `UC_ERR_INSN_INVALID` on instructions other than the breakpointed one.
But for Unicorn I'm guessing we could use its hooks system (`UC_HOOK_CODE` etc) anyway.

Macdu:
What is not implemented right now is that, when you add a breakpoint, break on it and continue, the bytes overwritten by the breakpoint never get executed which is really bad. [cmd_remove_breakpoint](https://github.com/Vita3K/Vita3K/blob/master/vita3k/gdbstub/src/gdb.cpp#L601-L614) properly removes the breakpoint and restores the original bytes. But this is only called when you ask gdb to remove the breakpoint, not when resuming after a break.

In general, I think overwritting the memory with a breakpoint is a bad idea. A much nicer way would be just to inject the break when recompiling using the translation hook ([here](https://github.com/Vita3K/Vita3K/blob/a2cfd2ca588c52f635c627415bdb1abfa532bf10/vita3k/cpu/src/dynarmic_cpu.cpp#L120) to be more precise). I've used the translation hook (and also checking the memory read/write handler) a lot for debugging on Vita3K. 

I don't know about a stack overflow using gdb and breakpoints. But this PR is 3 years old, dynarmic has changed a lot since then.

VelocityRa:
Ah ok, I assumed GDB would handle reapplying the original code bytes itself since they're software breakpoints (the stub only advertises `swbreak`) and afaik it technically could.

Either way, the bug I'm facing is before that - the actual breaking on the bp doesn't work, it crashes. So I suppose have to debug that.
I just wanted to make sure I'm not doing something obviously dumb like using the wrong GDB version/settings or whatever. It'd be helpful if you could test GDB on master and tell me whether it works for you, or just _when_ you last tested it.
It seems like there's not much in the way of a guest debugger so I'm curious how you guys manage without one 😅 checks in code/memory hooks and printf-debugging only get you so far, I would think.

Thanks for the pointer regarding the code hook, I was using unicorn for some unrelated tests and didn't even think to check the dynarmic backend, I didn't know it offers such finely grained hooks too.

PR #1681, opened 2022-04-27
kernel: Make ThreadState resume thread safe

Macdu:
When a thread is resumed immediately after being paused, a race condition can occur between the thread transitioning to the wait state and resuming at the same time.

To prevent that, the thread mutex must be locked in the resume function, to make sure the thread is not going into the wait state at the same time.

This race condition can occur in the vblank thread. This fixes the random freeze in Tales of Innocence R.

### DIFF
--- a/vita3k/gdbstub/src/gdb.cpp
+++ b/vita3k/gdbstub/src/gdb.cpp
@@ -454,13 +454,13 @@ static std::string cmd_continue(HostState &state, PacketCommand &command) {
                     auto lock = std::unique_lock(state.kernel.mutex);
                     for (const auto pair : state.kernel.threads) {
                         auto &thread = pair.second;
-                        auto thread_lock = std::unique_lock(thread->mutex);
-                        lock.unlock();
                         if (thread->status == ThreadStatus::suspend) {
+                            lock.unlock();
                             thread->resume();
-                            thread->status_cond.wait(thread_lock, [&]() { return thread->status != ThreadStatus::suspend; });
+                            lock.lock();
+
+                            thread->status_cond.wait(lock, [&]() { return thread->status != ThreadStatus::suspend; });
                         }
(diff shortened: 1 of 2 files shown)


Zangetsu38 (approved):

sunho:
This is supposed to be threaad unsafe function btw but it's fine to make it thread safe. Just need to be careful about deadlocks casue some places assume that this is no thread safe. e.g. https://github.com/Vita3K/Vita3K/blob/931bbe410367991304370796bf8aa406e0457464/vita3k/gdbstub/src/gdb.cpp#L443

sunho:
cleaner way would just add lock at vblank

Zangetsu38:
> cleaner way would just add lock at vblank

If i good remember is bent have sugest this inside kernel directly

Macdu:
Oh I didn't realize it was supposed to be thread unsafe.
Indeed bent suggested that I made the fix directly inside kernel and I think it is cleaner.
If you prefer I can add the thread lock in the vblank thread.
But i can also update gdbstub to assume this fonction is thread safe (I'm the one doing the pull request to fix it).

sunho:
Yeah maybe making it thread safe in the first place is better to prevent mistake like this. Just be sure to edit gdbstub too in order to not deadlock!

sunho:
btw now I think of it, we seriously should consider using recursive_mutex for thread mutex. That way we can just spam mutex lock to pretty much every function without caring about deadlock. It's almost zero overhead compared to plain mutex fyi.

Macdu:
Ok! 
Btw this function is also used by sceKernelSuspendThreadForVM which did not assume that it was thread unsafe.
I made the changes to prevent the deadlock in gdbstub (even though it can't be tested in master because it is broken for other reasons).

Zangetsu38 (approved):

PR #1682, opened 2022-04-28
kernel: Fix fifo event thread queue treated as a priority queue

Macdu:
`SCE_KERNEL_ATTR_TH_FIFO` is defined as `0x00000000U`, therefore `is_fifo` is always set to false, even when the bit corresponding to `SCE_KERNEL_ATTR_TH_PRIO` (`0x00002000U`) is set.
Moreover the way the thread queue for events is considered as a priority or fifo queue is different compared to what is done for semaphores and mutexes. This pull request changes the code so that it is done the same way.

I don't know if this fix improves the state of any game.


sunho (approved):
You're correct. Setting priority zero don't guarantee the fifo behavior at all as we sort the queued thread only by priority and order of same priority values are undefinded. (std multiset seems to not gurantee same elements to be stable sorted by insertion order)

PR #1683, opened 2022-04-29
kernel: Fix deadlocks and handle_timeout

Macdu:
This pull request contains fixes for the following 3 bugs in sync_primitives.cpp:

- When calling `erase` on a `PriorityThreadDataQueue`, all waiting threads with the same priority will be erased because a multiset tests equalities using `<` and not `==`. 
- All mutex locks in sync_primitives are acquired the following way: acquire the primitive (event/mutex/semaphore...) mutex first then the thread mutex second. The same must be done for handle_timeout in order to prevent deadlocks. Unlocking one mutex before locking the other kind of works but it allows race conditions to happen.
- Fix race condition in `eventflag_set`: if the thread mutex cannot be acquired immediatly, the thread will sleep forever. The thread mutex is locked to prevent this issue. There should not be any deadlock with the previous fix.

The last part fixes the random freeze in Resident Evil : Revelations 2.

This pull request also contains performance improvements for ThreadDataQueue.

### DIFF
--- a/vita3k/kernel/src/sync_primitives.cpp
+++ b/vita3k/kernel/src/sync_primitives.cpp
@@ -795,11 +795,7 @@ SceInt32 eventflag_set(KernelState &kernel, const char *export_name, SceUID thre
                 event->flags &= ~waiting_flags;
             }
 
-            const std::unique_lock<std::mutex> waiting_thread_lock(waiting_thread->mutex, std::try_to_lock);
-            if (!waiting_thread_lock) {
-                ++it;
-                continue;
-            }
+            const std::lock_guard<std::mutex> waiting_thread_lock(waiting_thread->mutex);
 
             waiting_thread->update_status(ThreadStatus::run, ThreadStatus::wait);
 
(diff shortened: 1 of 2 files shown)


Zangetsu38 (approved):

vita3k/kernel/src/sync_primitives.cpp
```diff
@@ -797,11 +795,7 @@ SceInt32 eventflag_set(KernelState &kernel, const char *export_name, SceUID thre
                 event->flags &= ~waiting_flags;
             }
 
-            const std::unique_lock<std::mutex> waiting_thread_lock(waiting_thread->mutex, std::try_to_lock);
```
sunho, line 800: nice catch. I messed up by doing ++it here.

sunho:
and yes it does have race condition that can alllow other thread to sneak in in-between state of sync primitive. nice work

Zangetsu38:
> and yes it does have race condition that can alllow other thread to sneak in in-between state of sync primitive. nice work

is ready to merge for you ?

sunho:
erase function can be improved a bit.

Zangetsu38:
> erase function can be improved a bit.

ok, i waiting :)

Macdu:
Ok, so how I did it was I kept an iterator pointing towards the WaitingThreadData when it is inserted in the queue (list and multiset iterators are not invalidated when other items are inserted or removed).

This way, if necessary we can erase the threadData using the pointer instead of its value which is much faster: it takes constant time for the FIFO thread and logarithmic time for the priority queue (instead of linear time right now for both).

Zangetsu38 (approved):

Zangetsu38:
@sunho check last commit, is erase commit but totaly reworked, confirme if is fine now

sunho (approved):
LGTM

PR #1689, opened 2022-05-01
kernel: add mutex lock guard on mspace func.

Zangetsu38:
# About:
- Add lock guard inside all mspace function, fix crash on mp4 lle.


bookmist (approved):
Let's suppose it does not made all slower.

sunho:
I was trying to add this too but I just couldn't believe mspace was the culprit.

PR #1694, opened 2022-05-03
modules/SceGxm: fix precomputed state always having max size

Macdu:
When generating a `SceGxmPrecomputedDraw`, `SceGxmPrecomputedFragmentState` or `SceGxmPrecomputedVertexState`, we always generate one with the maximum possible size and return its size accordingly when asked.

However, this makes these objects weigh much more than expected and can result in a crash if a buffer too small was allocated. This is the case in uncharted where a 2M buffer is used to allocate all 3.5k+ `SceGxmPrecomputedFragmentState`and `SceGxmPrecomputedVertexState`. It is not enough if they all have the maximum size.

Because arrays now don't have a fixed size anymore and we cannot add any overhead, I used pointers to replace existing `std::array`.

This allows Uncharted to get to the boot menu.

### DIFF
--- a/vita3k/modules/SceGxm/SceGxm.cpp
+++ b/vita3k/modules/SceGxm/SceGxm.cpp
@@ -343,6 +343,20 @@ static int init_texture_base(const char *export_name, SceGxmTexture *texture, Pt
     return 0;
 }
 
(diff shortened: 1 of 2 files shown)


Zangetsu38 (approved):
LGTM

PR #1700, opened 2022-05-06
modules/SceVideodecUser: Add support to back res value of AvcdecDecode.

Zangetsu38:
# About:
- Fix wrong resolution of video in few game (Street fighter, omega labyrinth, etc...)
thx special for @bookmist for disasm function.

# Result:
image
image


bookmist (approved):
I still think it needs more refactoring and further work

PR #1708, opened 2022-05-08
modules/SceGxm: Handle consecutive calls to sceGxmDisplayQueueAddEntry

Macdu:
When there are two close calls to `sceGxmDisplayQueueAddEntry` (in particular when the second one is done before the first one has finished). There is a time window where between the time the first call and the second call is done, where a scene can be drawn and mess the buffer to be displayed.

One simple fix is to add wishlist calls to `sceGxmDisplayQueueAddEntry`. However it makes this function blocking (functions calling it can easily wait 2 frames). In order to prevent performance deteriorations, I had to implement specific sync functions in order to allow `sceGxmDisplayQueueAddEntry` calls to be non-blocking between them but to be blocking for other calls.

This also fixes a really tight softlock that can happen in `wishlist`.

This fixes Oreshika's intro and partially the intro of Little King's story.


Zangetsu38:
no ooonly fix intro of king story, but also ingame, before ingame flashing, HUD flashing

Zangetsu38 (approved):

PR #1715, opened 2022-05-10
codec, modules/SceAudiodec: Fix Atrac9 and implement sceAudiodecClearContext

Macdu:
Atrac9 decoding was broken in its current state: `sceAudiodecDecode` should only decode one frame and then stop but for At9 it decoded at superframe which could be 1 frame (ok) or 4 frames (bad) which lead to a buffer overflow of the pcm buffer. This pull request fixes this behavior.
However, because there can be some padding bytes between two consecutive superframes, the Atrac9decoder must remember the decoding state in the superframe and add those padding bytes when needed.

This should fix games that were having random issues (caused by the buffer overflow) using atrac9. In particular Tales of games are now working with the sound enabled.

This pull request also implements sceAudiodecClearContext.  This fixes the crash happening in Tales of games when the background music loops back to the beginning.

### DIFF
--- a/vita3k/codec/src/mp3.cpp
+++ b/vita3k/codec/src/mp3.cpp
@@ -1,5 +1,5 @@
 // Vita3K emulator project
-// Copyright (C) 2021 Vita3K team
+// Copyright (C) 2022 Vita3K team
 //
 // This program is free software; you can redistribute it and/or modify
 // it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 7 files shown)


vita3k/codec/src/mp3.cpp
```diff
@@ -123,6 +123,13 @@ uint32_t Mp3DecoderState::get_es_size() {
     return es_size_used;
 }
 
+void Mp3DecoderState::clear_context() {
+    int err = avcodec_close(context);
+    assert(err == 0);
+    err = avcodec_open2(context, codec, nullptr);
+    assert(err = 0);
+}
+
```
bookmist, line 132: in manual to avcodec_close in avcodec.h there is a warning, that "Opening and closing a codec context multiple times is not supported"

Macdu: Indeed, you're right... Anyway I found a much better way to clear the context for both ffmpeg and libatrac9.

Zangetsu38 (approved):

PR #1717, opened 2022-05-11
Update Turkish

Tarik366:


Tarik366:
I delete before pull request because i have a error but somehow is fixed

PR #1723, opened 2022-05-12
[WIP] Build: Improve CMake configuration process

hobyst:
- [x] Refactor Boost detection and compilation (if needed) at CMake project configuration time. Now it will first go for system Boost if available, if not it will automatically detect and compile the custom Boost distribution in `external/boost`. Either of them can be forced using `VITA3K_FORCE_CUSTOM_BOOST` and `VITA3K_FORCE_SYSTEM_BOOST`.
- [ ] Replace CMake toolchain files with [CMake presets](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html) for proper IDE-agnostic configuration files to replace command-line configuration scripts.


CMakeLists.txt
```diff
@@ -24,119 +24,126 @@ enable_testing()
...
+        if(NOT Boost_FOUND)
+            message(STATUS "No existing custom distribution of Boost has been found in ${BOOST_INSTALLDIR}. Proceeding to compile Boost...")
+
+            # Set compilation toolset for b2 and Boost builds
+            if (WIN32)
+                if(CMAKE_GENERATOR_TOOLSET STREQUAL "clangcl")
+                    set(BOOST_TOOLSET "clang-win")
+                else()
+                    set(BOOST_TOOLSET msvc)
+                endif()
```
bookmist, line 105: I was not good in cmake, so CMAKE_CXX_COMPILER_ID may be checked here too.

hobyst: For MSBuild (Visual Studio) is fine to check against `CMAKE_GENERATOR_TOOLSET`, but yeah it might be better to check against `CMAKE_CXX_COMPILER_ID` so that it also works for builds that don't use MSBuild.

PR #1729, opened 2022-05-14
modules/SceVideodec: Update properly numOfOutput

Macdu:
`numOfOutput` was not updated properly when calling `sceAvcdecDecode` and `sceAvcdecDecodeStop`.

This fixes the softlock occuring when trying to skip fmvs in Tales of games.


Macdu:
Even though this pull request fixes a softlock. This fix does not match ps vita's behavior so it is more of a hack. I'm closing this pull request until I find a proper fix.

PR #1751, opened 2022-05-21
ngs: Fix parse_params and add missing checks

Macdu:
Return `ModuleParameterHeader` to its previous structure to match the sdk, and adapt `parse_params` (which was broken anyway).
Also add missing checks to ngs calls.


Zangetsu38 (approved):
Lgtm

PR #1758, opened 2022-05-25
kernel: better handling of thread priority and mask affinity

Macdu:
Implement kernel functions related to thread affinity and priority.

This allows Dragon Quest builder to boot.


pent0:
My three years old dream🥰🥰

Zangetsu38 (approved):

PR #1769, opened 2022-05-28
renderer: Fix renderer for mesa drivers

Macdu:
This fixes the yellow screen when using mesa drivers (mainly used by linux users).
Textures used for image load/store must be immutable (use glTexStorage2D) (otherwise mesa drivers will silently fail when executing shaders...). This was not the case for the mask texture. I haven't been able to find if this is required by the doc or only specific to mesa drivers.

Also fix sceGxmDestroyContext which was broken for many reasons:
- opengl buffers' deletion is not done on the same thread as other opengl calls
- it frees the memory used by the GxmContext, but it is not up to this function to do so (double free).

Also fix debug builds, opengl debug context and increase the opengl version requirement: we use glBufferStorage which requires openGL 4.4 anyway.

### DIFF
--- a/vita3k/glutil/include/glutil/object_array.h
+++ b/vita3k/glutil/include/glutil/object_array.h
@@ -54,7 +54,6 @@ public:
     const GLuint operator[](size_t i) const {
         assert(i >= 0);
         assert(i < names.size());
-        assert(names[i] != 0);
         return names[i];
     }
 
--- a/vita3k/modules/SceGxm/SceGxm.cpp
+++ b/vita3k/modules/SceGxm/SceGxm.cpp
@@ -988,7 +988,7 @@ EXPORT(int, sceGxmDestroyContext, Ptr<SceGxmContext> context) {
     if (!context)
         return RET_ERROR(SCE_GXM_ERROR_INVALID_POINTER);
 
-    free(host.mem, context);
+    renderer::destroy_context(*host.renderer, context.get(host.mem)->renderer);
 
     return 0;
 }
--- a/vita3k/renderer/include/renderer/commands.h
+++ b/vita3k/renderer/include/renderer/commands.h
@@ -66,7 +66,8 @@ enum class CommandOpcode : std::uint8_t {
 
     SignalNotification = 10,
 
-    DestroyRenderTarget = 11
+    DestroyRenderTarget = 11,
+    DestroyContext = 12
 };
 
 enum CommandErrorCode {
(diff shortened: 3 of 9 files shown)


sunho:
We already dropped opengl 4.1 long ago as we temporarily ditched apple. So, it's totally fine.

Zangetsu38 (dismissed):

hobyst:
Running the first build using an integrated Intel GPU (Mesa) and executing the red rectangle app from the vitasdk examples, which uses a pretty simple shader, makes the emulator crash even before the first frame is rendered:
```bash
[19:38:44.195] |T| [open_file]: sceIoOpen: Opening terminal tty0:
[19:38:44.210] |W| [unimplemented_impl]: Unimplemented sceTouchEnableTouchForce import called.
[19:38:44.212] |W| [unimplemented_impl]: Unimplemented sceHidKeyboardEnumerate import called.
[19:38:44.212] |W| [unimplemented_impl]: Unimplemented sceHidMouseEnumerate import called.
[19:38:44.248] |W| [stubbed_impl]: Stubbed sceGxmMapFragmentUsseMemory import called. (always return success)
[19:38:44.286] |E| [after_callback]: OpenGL: glTexImage2D set error 1282.
[19:38:44.327] |W| [stubbed_impl]: Stubbed sceGxmMapVertexUsseMemory import called. (always return success)
[19:38:44.348] |W| [unimplemented_impl]: Unimplemented sceCommonDialogUpdate import called.
[19:38:48.382] |C| [free]: Freeing unallocated page
```

Zangetsu38 (dismissed):

Macdu:
Yes, my last update to the PR (seconds before you did this comment) should have fixed this, can you confirm?

hobyst:
Same thing as before.

Macdu:
Can you send the log? That's weird because I'm not using glTexImage2D anymore.

hobyst:
[vita3k.log](https://github.com/Vita3K/Vita3K/files/8791637/vita3k.log)
The last thing that happens is the emulator crashing due to a SIGTRAP signal from the OS.

Macdu:
@hobyst Is the emulator crashing without this pr using mesa intel drivers?
If you can run vita3K with the debugger, can you tell me which function call causes the emulator to crash?

hobyst:
So far it seems the emulator is trying to access a memory address not really belonging to Vita3K:
```log
Stop reason: signal SIGSEGV: address access protected (fault address: <address_number>)
```

hobyst:
The debugger doesn't seem to give that much information about which function is causing this, but Tracy does and it seems the crash is produced when the game calls `sceGxmDestroyRenderTarget`. Also something worth noticing is that the crash not only happens with Intel/Mesa but also with NVIDIA and their proprietary drivers, so we might be looking at an actual segmentation due to an illegal memory access.
It wouldn't be the first time I see new code uncovering arrays that haven't been properly initialized. I'll later try it with a Debug build to check if anything changes, as I've been testing on RelWithDebInfo.

Zangetsu38 (approved):

Zangetsu38 (approved):

PR #1770, opened 2022-05-28
 renderer: sync surface data explicit

Zangetsu38:
- should fix spam of ogl erorr

# Todo
- [ ] fix boot of uncharted and acces of resogun ingame
- [ ] fix violet color of uncharted piece


Macdu:
What is causing issues in this pr:
- there is a race condition, if a texture sync data is processed in the main thread while GxmEndScene is called, the texture is unprotected by the main thread, protected back by GxmEndScene and then a read error happens in the main thread. We need to use mutex to handle this.
- Moreover if a texture uses the surface, we need to unprotect the memory if this happens...
- Even worse (what is happening previously does not cause any issue other else than performance loss because of the thread id check): when an opengl function is called, it may do some of this processing on another thread: read the buffer on another thread, causing a missed read, sending a command to the main thread and softlocking the game... A way to solve this would be to handle read/write error only if they come from one of the ps vita game threads.

Zangetsu38:
tooo old and too obsolet

PR #1778, opened 2022-05-30
shader/spirv recompiler: handle default sampler value when no exist.

Zangetsu38:
# About:
- add part of commit missing from #1761


Macdu (approved):

Zangetsu38:
AH 🤦‍♂️  Baka me, i have forget change author safter create commit, author is @pent0 of course 🙏

PR #1803, opened 2022-06-04
kernel: Implement simple events

Macdu:
Implement simple event functions.

This should help games that rely on them.


frangarcj (dismissed):

Zangetsu38 (approved):

vita3k/kernel/include/kernel/sync_primitives.h
```diff
@@ -72,7 +77,6 @@ typedef std::unique_ptr<ThreadDataQueue<WaitingThreadData>> WaitingThreadQueuePt
 // NOTE: uid is copied to sync primitives here for debugging,
 //       not really needed since they are put in std::map's
 struct SyncPrimitive {
-    // FIXME turn this into SimpleEvent!
```
CreepNT, line 75: The meaning of this was not just "we don't have an implementation of SimpleEvent" but "there are many objects currently inheriting from `SyncPrimitive` which actually inherit from `SimpleEvent` on the real kernel". To be more precise: the common class is `SceUIDEventClass` from which both `SceUIDSimpleEventClass`, `SceUIDMsgPipe` and `SceUIDCpuTimerClass` inherit.

And to be complete, `SceUIDCpuTimerClass` is base class for:
* `SceUIDAlarmClass` - for internal usage by kernel only?
* `SceUIDDelayClass` - used by kernel to implement e.g., `sceKernelDelayThread()`
* `SceUIDTimerClass` - accessible by user `sceKernelXxxTimer()` API

vita3k/modules/SceKernelThreadMgr/SceThreadmgr.cpp
```diff
@@ -1052,8 +1053,8 @@ EXPORT(int, sceKernelPollSema, SceUID semaid, int32_t needCount) {
     return SCE_KERNEL_OK;
 }
 
-EXPORT(int, sceKernelPulseEvent) {
-    return UNIMPLEMENTED();
+EXPORT(SceInt32, sceKernelPulseEvent, SceUID event_id, SceUInt32 set_pattern, SceUInt64 user_data) {
+    return simple_event_setorpulse(host.kernel, export_name, thread_id, event_id, set_pattern, user_data, false);
```
CreepNT, line 1057: e.g. this function should work if `eventId` is a MsgPipe

PR #1810, opened 2022-06-05
shader: full delete old broken spirv shader.

Zangetsu38:
# About:
- prepare of next with move on real glsl shader, delete this broken shader full no works


Zangetsu38:
Keep it for later

PR #1817, opened 2022-06-07
gui: Mark Unicorn and Spir-V shader as deprecated

Macdu:
Right now there is no real reason to use either Unicorn or Spir-V, mark them as deprecated to discourage users from using them.

### DIFF
--- a/vita3k/cpu/include/cpu/common.h
+++ b/vita3k/cpu/include/cpu/common.h
@@ -107,8 +107,8 @@ struct CPUContext {
 };
 
 enum class CPUBackend {
+    Dynarmic,
     Unicorn,
-    Dynarmic
 };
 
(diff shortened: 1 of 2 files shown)


Zangetsu38 (approved):

sunho:
You can just drop unicorn if you want. Dynarmic verifies itself very rigorously using unicorn engine and improve on some cases where unicorn is just wrong. So, it's just equivalent to unicorn but improved on some part in terms of functionality.

sunho:
There are some edge case tho, in which we use unicorn as fallback. There was vzip.16 which for some reason ps vita compiler generate out of spec instruction which dynarmic returns unpredictable error, and unicorn fallbacks on it. You might want to look into it if you're going to remove unicorn.

PR #1818, opened 2022-06-07
docs: Add official AUR links

EXtremeExploit:
Now that i have write perms of the vita3k AUR packages, i fixed the -git one and added a -bin one, can say this packages are "official" as they are maintained by a dev
Someday in the future maybe the CI will export linux packages, could be an AppImage, deb. But right now its just the executable which is fine


sunho (approved):
why not

PR #1821, opened 2022-06-08
modules/SceLibKernel: add process callbacks on cb function.

Zangetsu38:
# About
- Add execute callback on all cb function.


Macdu:
Ok, but the way we handle it right now is more like a hack, callbacks must be processed after checking a first time for the wait condition, or it could lead to softlocks.

Macdu (approved):

PR #1847, opened 2022-06-12
modules: Don't download discord gamesdk when its turned off & compiler warnings

EXtremeExploit:
* Don't download discord gamesdk if its turned off, it doesn't make sense to download it while it's turned off, its never used it just takes time
* Fix some compiler warnings, if the fixes for the warnings on texture formats are wrong then i'd happily revert those, i didn't see any regressions of any kind tho


Zangetsu38:
Wrong name of first commit, is external/

Zangetsu38 (approved):

vita3k/gui/src/home_screen.cpp
```diff
@@ -257,6 +257,8 @@ static bool app_filter(const std::string &app) {
         if (!filter_app_region({ "PCS" }))
             return true;
         break;
+    case ALL:
+        return true;
```
Zangetsu38, line 261: remove it, is not needed, and is wrong, here return true want said skip game, one this case you skepp so all app
image

Zangetsu38 (changes requested):

vita3k/gui/src/home_screen.cpp
Zangetsu38, line 261:
```suggestion
    case ALL:
        break;
```
if you adding it for any compile warning because maybe all enum is not set ? using this, for directly go outside switch and return false in end

Zangetsu38: and set this case in first, for respect order of enum 👌

Zangetsu38: also, if you adding all here, delete: if (app_region != ALL) in 554 line, no needed anymore

vita3k/gui/src/home_screen.cpp
```diff
@@ -233,6 +233,8 @@ static bool app_filter(const std::string &app) {
     };
 
     switch (app_region) {
+    case ALL:
```
Zangetsu38, line 236: you have forget add "default:" before

EXtremeExploit: There is no need for default because there is not a missing case there, all the cases are present so its not needed

Zangetsu38: i see

Zangetsu38 (approved):

PR #1869, opened 2022-06-18
modules/SceGxm, renderer: Process everything in the main thread and make gxm functions not blocking

Macdu:
`SceGxmBeginScene` (and other functions) are not supposed to be blocking. Rewrite the sync objects to better match the PS vita.
Also process the transfer functions in the main thread because that's where they should be.

This can fix the graphics of some games and should improve the speed of many games.


vita3k/renderer/src/scene.cpp
```diff
@@ -168,4 +168,121 @@ COMMAND(handle_draw) {
...
+    const SceGxmTransferType dst_type = helper.pop<SceGxmTransferType>();
+
+    const auto src_type_is_linear = src_type == SCE_GXM_TRANSFER_LINEAR;
+    const auto src_type_is_tiled = src_type == SCE_GXM_TRANSFER_TILED;
+    const auto src_type_is_swizzled = src_type == SCE_GXM_TRANSFER_SWIZZLED;
+    const auto dest_type_is_linear = dst_type == SCE_GXM_TRANSFER_LINEAR;
+    const auto dest_type_is_tiled = dst_type == SCE_GXM_TRANSFER_TILED;
+    const auto dest_type_is_swizzled = dst_type == SCE_GXM_TRANSFER_SWIZZLED;
+
+    if (src_type_is_linear && dest_type_is_linear) {
```
Zangetsu38, line 189:
```suggestion
    if (src_type && dst_type ) {
```
remove all const auto before, some gamee using two type same type is compatible

Zangetsu38 (approved):

Zangetsu38:
Note current regression, broken Resogun boot

PR #1894, opened 2022-06-27
vita3K/script: Improve script of update windows build.

Zangetsu38:


Macdu:
I don't think admin rights should be needed in order to download a file from the web.

Zangetsu38:
> I don't think admin rights should be needed in order to download a file from the web.

Is should right but one user no have admin and no can using any commende 🤷‍♂️

hobyst:
It might have to do with [this](https://stackoverflow.com/a/50274056). If that's the case using the argument should be fine since we aren't accessing HTML DOM.

EXtremeExploit:
the script instead of using powershell to download it could do it with curl, windows 10 includes curl by default (https://techcommunity.microsoft.com/t5/containers/tar-and-curl-come-to-windows/ba-p/382409). Same thing goes with tar to unzip the file

hobyst:
> the script instead of using powershell to download it could do it with curl, windows 10 includes curl by default (https://techcommunity.microsoft.com/t5/containers/tar-and-curl-come-to-windows/ba-p/382409). Same thing goes with tar to unzip the file
The issue seems to be that some users might have command line execution and other features of the OS completely blocked due to Group Policies in Windows Pro.

Foxlum:
> > the script instead of using powershell to download it could do it with curl, windows 10 includes curl by default (https://techcommunity.microsoft.com/t5/containers/tar-and-curl-come-to-windows/ba-p/382409). Same thing goes with tar to unzip the file
> 
> The issue seems to be that some users might have command line execution and other features of the OS completely blocked due to Group Policies in Windows Pro.

Well thats a extremally unlikely case usually for security or organization reasons. Both I would assume you'd manually get a new build anyway, or not run Vita3k at all on a enrolled machine.

EXtremeExploit (approved):

PR #1898, opened 2022-06-28
Repurpose host library for host OS abstraction library

hobyst:
Most of the code from `vita3k/host` doesn't even have to do with the emulated PlayStation Vita environment and the name "host" can be mistaken with a host OS abstraction layer, so this PR:
- [x] Moves the code in charge of handling PS Vita package formats handling from `vita3k/host` to `vita3k/package`.
- [x] Renames the current host library (`vita3k/host`) to "emuenv" (emulated environment) (`vita3k/emuenv`) and related
types and functions (ex. `HostState` --> `EmuEnvState`) for more meaningful naming and avoid confusion with future host OS abstraction library


vita3k/package/include/package/sfo.h
```diff
@@ -50,3 +59,35 @@ struct SfoFile {
...
+/**
+ * @brief Get the param info object
+ *
+ * @param app_info App information struct to store the information retrieved from the `.sfo` file
+ specified in `param`
+ * @param param File buffer pointing to the `param.sfo` file to parse
+ * @param sys_lang System language. It is used to get translated strings from `param.sfo`
+ */
+void get_param_info(sfo::SfoAppInfo &app_info, const vfs::FileBuffer param, int sys_lang);
+} // namespace sfo
```
Zangetsu38, line 93: oupsy endline :D

vita3k/modules/CMakeLists.txt
```diff
@@ -210,6 +210,6 @@ set(SOURCE_LIST
 
 add_library(modules STATIC ${SOURCE_LIST})
 target_include_directories(modules PUBLIC include)
-target_link_libraries(modules PRIVATE xxHash::xxhash)
+target_link_libraries(modules PRIVATE xxHash::xxhash package)
```
Zangetsu38, line 213: i more see it inside public juse after, private for me is for external submodule

hobyst: I'll leave it as private then.

vita3k/emuenv/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
...
+	emuenv
+	INTERFACE
+	include/emuenv/app_util.h
+	include/emuenv/state.h
+	include/emuenv/window.h
+)
+
+target_include_directories(emuenv INTERFACE include ${PSVPFSPARSER_INCLUDE_DIR})
+target_link_libraries(emuenv INTERFACE psvpfsparser app audio config ctrl dialog display ime io kernel lang miniz net ngs nids np renderer sdl2 touch gdbstub codec package)
+target_link_libraries(emuenv INTERFACE elfio::elfio FAT16 vita-toolchain)
```
Macdu, line 11: You should be able to remove some of these librairies which were used by the code in package now (for sure elfio, fat16 and vita-toolchain, and psvpfsparser,sdl2 too I think).

hobyst: Just confirmed it could be done so I've applied the changes to the CMake file.

vita3k/emuenv/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+add_library(
+	emuenv
+	INTERFACE
+	include/emuenv/app_util.h
+	include/emuenv/state.h
+	include/emuenv/window.h
+)
+
+target_include_directories(emuenv INTERFACE include ${PSVPFSPARSER_INCLUDE_DIR})
```
Macdu, line 9: Btw (that was the cas before this pr, I'm just thinking about it now), why do the psvpsfparser headers must be included with emuenv? It does not use them and even if it were the case they should be linked, not included.

hobyst: I've checked it and they can be removed so I've done so.

Macdu (approved):

PR #1905, opened 2022-07-01
modules/SceCommonDialog: Fix utf16 ending bytes and improve api errors

Macdu:
utf16 strings must end with 2 null bytes and not 1. This allows digimon next order japanese version to go past the character name screen every time without the string length error.

Also improve the return errors given by the api calls (many of them were incorrect).


vita3k/gui/src/common_dialog.cpp
```diff
@@ -49,7 +49,7 @@ static void draw_ime_dialog(DialogState &common_dialog, float FONT_SCALE) {
         common_dialog.status = SCE_COMMON_DIALOG_STATUS_FINISHED;
         common_dialog.result = SCE_COMMON_DIALOG_RESULT_OK;
         const std::u16string result16 = string_utils::utf8_to_utf16(common_dialog.ime.text);
-        memcpy(common_dialog.ime.result, result16.c_str(), result16.length() * sizeof(uint16_t) + 1);
+        memcpy(common_dialog.ime.result, result16.c_str(), (result16.length() + 1) * sizeof(uint16_t));
```
Zangetsu38, line 52: Oh my bad, you fix my fix xd

PR #1912, opened 2022-07-02
[WIP] fs: Switch to c++ filesystem

Macdu:
Switch from Boost filesystem to C++17 filesystem library by default.

I made it so it is still possible to use Boost filesystem by defining the macro USE_BOOST_FS.
Need a bit more testing to make sure everything is working properly on both windows and linux.


scribam:
@Macdu do you still plan to do this?

PR #1917, opened 2022-07-03
shader: GLSL compiler; renderer: Various changes 

korenkonder:


Macdu:
Will be separated into different PRs.

PR #1918, opened 2022-07-03
renderer: reworked decode of BCn texture types

korenkonder:
- Added direct upload of BCn textures without decoding
- Reworked BCn decoder. Even though we don't need it, it's nice to have it for some cases

### DIFF
--- a/vita3k/renderer/src/gl/texture.cpp
+++ b/vita3k/renderer/src/gl/texture.cpp
@@ -112,6 +112,8 @@ void configure_bound_texture(const renderer::TextureCacheState &state, const Sce
     size_t compressed_size = 0;
     uint32_t mip_index = 0;
 
+    bool block_compressed = renderer::texture::is_compressed_format(base_fmt);
+
     // GXM's cube map index is same as OpenGL: right, left, top, bottom, front, back
     GLenum upload_type = GL_TEXTURE_2D;
 
@@ -124,7 +126,8 @@ void configure_bound_texture(const renderer::TextureCacheState &state, const Sce
     }
 
     while (face_iterated < face_total_count && width && height) {
-        if (!is_swizzled && renderer::texture::is_compressed_format(base_fmt, width, height, compressed_size)) {
+        if (block_compressed) {
+            size_t compressed_size = renderer::texture::get_compressed_size(base_fmt, width, height);
             glCompressedTexImage2D(upload_type, mip_index, internal_format, width, height, 0, static_cast<GLsizei>(compressed_size), nullptr);
         } else if (!is_swizzled || (is_swizzled && can_texture_be_unswizzled_without_decode(base_fmt))) {
             glTexImage2D(upload_type, mip_index, internal_format, width, height, 0, format, type, nullptr);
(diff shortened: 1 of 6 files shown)


vita3k/renderer/src/gl/texture.cpp
```diff
@@ -103,7 +103,7 @@ void configure_bound_texture(const renderer::TextureCacheState &state, const Sce
     const GLenum format = translate_format(base_format);
     const GLenum type = translate_type(base_format);
     const auto texture_type = gxm_texture.texture_type();
-    const bool is_swizzled = (texture_type == SCE_GXM_TEXTURE_SWIZZLED) || (texture_type == SCE_GXM_TEXTURE_CUBE) || (texture_type == SCE_GXM_TEXTURE_SWIZZLED_ARBITRARY) || (texture_type == SCE_GXM_TEXTURE_CUBE_ARBITRARY);
+    bool is_swizzled = (texture_type == SCE_GXM_TEXTURE_SWIZZLED) || (texture_type == SCE_GXM_TEXTURE_CUBE) || (texture_type == SCE_GXM_TEXTURE_SWIZZLED_ARBITRARY) || (texture_type == SCE_GXM_TEXTURE_CUBE_ARBITRARY);
```
Macdu, line 106: why isn't it const anymore?

vita3k/renderer/src/gl/texture.cpp
```diff
@@ -467,25 +615,27 @@ void upload_bound_texture(const SceGxmTexture &gxm_texture, const MemState &mem)
...
-
-        glPixelStorei(GL_UNPACK_ROW_LENGTH, static_cast<GLint>(pixels_per_stride));
+        if (block_compressed) {
+            glPixelStorei(GL_UNPACK_ROW_LENGTH, 0);
 
-        if (need_decompress_and_unswizzle_on_cpu)
-            glTexSubImage2D(upload_type, mip_index, 0, 0, width, height, GL_RGBA, GL_UNSIGNED_BYTE, pixels);
+            const GLenum format = translate_format(base_format);
+            size_t compressed_size = renderer::texture::get_compressed_size(base_format, width, height);
+            glCompressedTexSubImage2D(upload_type, mip_index, 0, 0, width, height, format, static_cast<GLsizei>(compressed_size), pixels);
```
Macdu, line 623: Can you use glCompressedTexImage2D instead, Vita3K is still trying to be compatible with openGL 4.4 (the fact that we were using this function is a mistake).

vita3k/renderer/src/texture_format.cpp
```diff
@@ -211,260 +214,322 @@ bool is_compressed_format(SceGxmTextureBaseFormat base_format, std::uint32_t wid
...
+    case SCE_GXM_TEXTURE_BASE_FORMAT_UBC4:
+    case SCE_GXM_TEXTURE_BASE_FORMAT_SBC4:
+        return ((width + 3) / 4) * ((height + 3) / 4) * 8;
+    case SCE_GXM_TEXTURE_BASE_FORMAT_UBC2:
+    case SCE_GXM_TEXTURE_BASE_FORMAT_UBC3:
+    case SCE_GXM_TEXTURE_BASE_FORMAT_UBC5:
+    case SCE_GXM_TEXTURE_BASE_FORMAT_SBC5:
+        return ((width + 3) / 4) * ((height + 3) / 4) * 16;
+    default:
+        return 0;
```
Macdu, line 229: Can you add a LOG_ERROR here ?

vita3k/renderer/src/gl/texture.cpp
```diff
@@ -151,6 +154,53 @@ void configure_bound_texture(const renderer::TextureCacheState &state, const Sce
...
+ * \param width  Texture width.
+ * \param height Texture height.
+ *
+ * \return Void.
+ */
+static void remove_compressed_arbitrary_blocks(SceGxmTextureBaseFormat fmt, void *dest, const void *data, const std::uint32_t width, const std::uint32_t height) {
+    std::size_t w = (width + 3) / 4;
+    std::size_t h = (height + 3) / 4;
+
+    std::size_t a_w = nearest_power_of_two(width);
```
Macdu, line 172: Can you rename this function to `next_power_of_two` (nearest is a bit misleading)?

vita3k/renderer/src/gl/texture.cpp
korenkonder: https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glCompressedTexSubImage2D.xhtml
`glCompressedTexSubImage2D` is compatible even with OpenGL 2.0 while `glCompressedTextureSubImage2D` isn't

vita3k/renderer/src/gl/texture.cpp
korenkonder: My bad

vita3k/renderer/src/gl/texture.cpp
Macdu: Oops my bad, misread the doc

vita3k/renderer/src/gl/texture.cpp
korenkonder: Done

PR #1923, opened 2022-07-05
vita3k/interface: fix race and crash on application close

bookmist:
add waiting until display thread is finished


vita3k/interface.cpp
```diff
@@ -574,6 +574,7 @@ bool handle_events(EmuEnvState &emuenv, GuiState &gui) {
             emuenv.kernel.exit_delete_all_threads();
             emuenv.gxm.display_queue.abort();
             emuenv.display.abort.exchange(true);
```
Macdu, line 576: btw could you change this line to 
`emuenv.display.abort = true;`
It would look cleaner

bookmist: yes, of course

PR #1924, opened 2022-07-05
renderer/gl/texture formats: add BCs texture format missing.

Zangetsu38:
# About
after #1918 get closed by my fault :(, i have just thinks can already got this to master, i have done same on my pr of color format but with wrong value.


Macdu:
Closing at it is half-complete and a full implementation was already merged in #1929.

PR #1926, opened 2022-07-07
vita3k/interface: fix race and crash on application close again

bookmist:
I forget the case when thread finished too fast (or not yet created)


Zangetsu38 (approved):
fix close direct when you try close without have start any game

PR #1927, opened 2022-07-08
vita3k update: add some lang string and fix.

Zangetsu38:
# About
- Add spanish/en-gb lang
-  Add screen if install update fail for allow cancel and back in emu home screen
- add text if version of vit'a3k is upper compare current
- fix warn compile on line find for windows
- check if powershell version is upper to 3 for can using rest-method
- add timeout of 4 sec on check versionn, if github down  etc...

### DIFF
--- a/lang/en-gb.xml
+++ b/lang/en-gb.xml
@@ -313,4 +313,19 @@ Are you sure you want to continue?</user_delete_warn>
 		<no>No</no>
 		<automatic_user_login>Automatic User Login</automatic_user_login>
 	</user_management>
+
+	<vita3k_update name="Vita3K Update">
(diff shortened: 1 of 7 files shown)


Zangetsu38:
> 



> I found that if the system is lower than Windows 8, Vita3K will be stuck and there is no display Vita3K update. It will take some time to wait, and then the log will display errors.

ah,, thx for info 🤔

Zangetsu38:
@Croden1999  you can enter this command on cmd on your old wintows os like you said for check
>powershell ((Invoke-RestMethod https://api.github.com/repos/Vita3K/Vita3K/releases/latest -timeout 2).body.split([Environment]::NewLine) ^| Select-String -Pattern 'Vita3K Build: ')

PR #1948, opened 2022-07-14
vita3K Update: Init it only in CLI master build.

Zangetsu38:
# About
- request by @EXtremeExploit  add definition of use vita3k update only when is cli master build


EXtremeExploit (approved):

EXtremeExploit (approved):
ah better now, ifdef is cleaner. very nice

PR #1953, opened 2022-07-18
kernel/relocation: Fix ARM Call, Jump24 and Abs8 relocations

lephilousophe:
Previously, some data was set to 0 instead of being preserved


lephilousophe:
Yes, you can take the ScummVM package at https://buildbot.scummvm.org/dailybuilds/master/vita-master-latest.zip

The `SDL_AtomicLock` function contains the instruction:
`812eb198:       ebffffda        bl      812eb108 <SDL_AtomicTryLock>`
with a relocation of type R_ARM_CALL and with the current code, the eb is erased to 00 changing the meaning of the instruction.

frangarcj (approved):

PR #1954, opened 2022-07-18
Fix sceGxmTerminate

lephilousophe:
These patches make sure that sceGxmTerminate doesn't trigger a segmentation fault in SDL thread by emptying the work queue first and by aborting it.
Then, the thread is waited for before killing the SceGxm thread.

This PR also includes a fix on a warning raised by clang about memset being called with an improper size.
Finally, the reset() implementation of Queue is fixed to properly build.


Macdu:
About your `reset` change, it's because your compiler complains about `_queue.clear()`, is that right?

lephilousophe:
Yes, `std::queue` hasn't any `reset` function and compiler only complains at template function instantiation.
I suppose this function was never used.

Macdu:
Ah indeed! Thank you for your contribution.

Macdu (approved):

Zangetsu38 (approved):

PR #1955, opened 2022-07-18
vita3k: Check for admin privileges at startup

EXtremeExploit:
works on windows :+1: . tested and it works on linux


Zangetsu38 (approved):

Ristellise:
Should we just warn the user instead of force closing the application?

EXtremeExploit:
:thinking: maybe allow admin perms if the --admin argument is passed? allowing admin perms can break permissions for some files and would have to reset the emulator to run it as a normal user again. also been thinking of telling the user of another way to not run with admin perms on windows, since the cmd appears for a split second.

PR #1962, opened 2022-07-21
net: Implement linux version of sceNetGetMacAddress

EXtremeExploit:
Implements the linxu version of the get macc adress function, there are some quirks i couldn't get across like getting the correct adapter, so it you use wifi it will probably use the ethernet one since its the first one detected, also added a magic number un case the user doesn't have any adapters connected, `EE:EE:EE:BA:DA:55`
image


Zangetsu38:
ouch is old, you can rebase please ? @EXtremeExploit

PR #1964, opened 2022-07-21
kernel: Partially implement command for sceIoDevctl

EXtremeExploit:
wish that there was a list or a clue of what commands exist, i could only find the one the VITAident uses which uses to get the device capacity status. Implementing cluster size doesn't look very important and it requieres os specific code so i left it out stubbed to 4096 which is the default for an NTFS FS, which is the most commonly used

showing correctly
image
image

### DIFF
--- a/vita3k/io/include/io/types.h
+++ b/vita3k/io/include/io/types.h
@@ -103,3 +103,9 @@ struct SceIoDirent {
     Ptr<void> d_private;
     int dummy;
 };
+
+struct SceIoDevInfo {
+    SceInt64 max_size;
+    SceInt64 free_size;
+    SceSize cluster_size;
+};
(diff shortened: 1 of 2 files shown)


sunho:
I don't think we will ever be able to fully implement ioctl family. It will require very thorough reverse engineered kernel to the death. IIRC, ordinary ioctl function is called by lle fios libs and it turned out that fios doesn't really use that info much in meaningful way. Stll, it should be fine to merge this if this helps homebrew users or smth as we have "STUBBED" to have some clue.

sunho (approved):

EXtremeExploit:
> I don't think we will ever be able to fully implement ioctl family. It will require very thorough reverse engineered kernel to the death. IIRC, ordinary ioctl function is called by lle fios libs and it turned out that fios doesn't really use that info much in meaningful way. Stll, it should be fine to merge this if this helps homebrew users or smth as we have "STUBBED" to have some clue.

indeed, the goal is to add them as we discover more commands

PR #1971, opened 2022-07-25
vita3k: Use correct KB/KiB functions

### LINKED ISSUE #1819: Minor misunderstanding about KB and KiB
Currently in the emulator you can describe 1024 bytes as 1KB, however 1KB is 1000 and not 1024, 1KiB IS 1024, there should be 2 KB/KiB definitions for each value, same goes for other sizes like Mega and Giga
image
https://en.wikipedia.org/wiki/Byte#Units_based_on_powers_of_10 

So following the above, KB, MB and GB should be replaced to KiB, MiB and GiB and add new types for the units based on powers of 2.
Windows is the only OS that does not follow the spec, it uses KB/MB/GB for powers of 2 and KiB/MiB/GiB for powers of 10. So its normal to make this mistake
Not really that important but it should be addressed at some point in the future to use the correct terminology to understand better what is what

EXtremeExploit:
please check code in case i made a mistake, i made sure to use KiB/MiB/GiB on every place it used KB/MB/GB, KB/MB/GB now don't get called anywhere so it should still be the same, but just in case :+1:  
closes #1819


vita3k/gui/src/content_manager.cpp
```diff
@@ -72,12 +72,12 @@ size_t get_app_size(GuiState &gui, EmuEnvState &emuenv, const std::string &app_p
...
-    else if (size >= MB(10))
-        size_str = std::to_string(size / MB(1)) + " MB";
-    else if (size >= KB(1))
-        size_str = std::to_string(size / KB(1)) + " KB";
+    if (size >= GiB(10))
+        size_str = std::to_string(size / GiB(1)) + " GiB";
+    else if (size >= MiB(10))
+        size_str = std::to_string(size / MiB(1)) + " MiB";
+    else if (size >= KiB(1))
+        size_str = std::to_string(size / KiB(1)) + " KiB";
```
Zangetsu38, line 80: revert, i using just same text of psvita, vita no write i

Zangetsu38: I wany keep string if vita

EXtremeExploit: the vita says "(size in MiB) MB", which is wrong you don't want to tell the user the game is using "1000 MB" when in reality its using "1000 MiB", it's like having a math test and the question is 2+2, you know the answer is 4, but because your friend puts 5 you put 5 instead of 4. :brain:

Zangetsu38: no really care, just using MB like psvita, i no speek other things, just keep official text, that ,all

sunho (approved):

vita3k/gui/src/content_manager.cpp
EXtremeExploit: i won :sunglasses:

Zangetsu38: 🤷‍♂️

PR #1973, opened 2022-07-25
renderer: Add Vulkan renderer

Macdu:
This PR implements a Vulkan renderer.
The Vulkan renderer is not completely optimized (there are still some `waitIdle` that I need to remove and the rest will be done in separate PRs later), but it should be on par with the OpenGL renderer. This is a WIP, there are still many games having graphical glitches or more when compared to OpenGL.
This was tested for now only tested on Windows on an Nvidia and Intel iGPU. By default if validation layers are installed they will be used (this has a performance impact but the validation error are really really useful).
With this PR we should also be able to bring Mac support back using MoltenVK.


CMakeLists.txt
```diff
@@ -12,6 +12,7 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 option(USE_DISCORD_RICH_PRESENCE "Build Vita3K with Discord Rich Presence" ON)
 option(USE_VULKAN "Build Vita3K with Vulkan backend." OFF)
 option(USE_VITA3K_UPDATE "Build Vita3K with updater." ON)
+set(USE_VULKAN ON)
```
Zangetsu38, line 15: i thinks, is better just remove if for it :D and so remove this variable

bookmist:
I check a little on ati gpu (AMD Radeon HD 7800 Series). It works good on debug build, but crash almost immediately on RelWithDebugInfo build
Log:
`[21:38:04.758] |W| [unimplemented_impl]: Unimplemented sceScreenShotSetOverlayImage import called.
[21:38:04.759] |W| [unimplemented_impl]: Unimplemented sceScreenShotEnable import called.
[21:38:04.759] |W| [unimplemented_impl]: Unimplemented sceNpActivityInit import called.
[21:38:04.774] |E| [debug_callback]: Validation layer: Validation Error: [ VUID_Undefined ] Object 0: handle = 0x1ca4c8f0c60, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x79de34d4 | vkCreateImage(): Format VK_FORMAT_D24_UNORM_S8_UINT is not supported for this combination of parameters.
[21:38:04.774] |E| [debug_callback]: Validation layer: Validation Error: [ VUID-VkImageViewCreateInfo-None-02273 ] Object 0: handle = 0x998b0f00000005ed, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x3f948fd1 | vkCreateImageView(): pCreateInfo->format VK_FORMAT_D24_UNORM_S8_UINT with tiling VK_IMAGE_TILING_OPTIMAL has no supported format features on this physical device. The Vulkan spec states: The format features of the resultant image view must contain at least one bit. (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkImageViewCreateInfo-None-02273)
[21:38:04.791] |E| [debug_callback]: Validation layer: Validation Error: [ VUID_Undefined ] Object 0: handle = 0x1ca4c8f0c60, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x79de34d4 | vkCreateImage(): Format VK_FORMAT_D24_UNORM_S8_UINT is not supported for this combination of parameters.
[21:38:04.792] |E| [debug_callback]: Validation layer: Validation Error: [ VUID-VkImageViewCreateInfo-None-02273 ] Object 0: handle = 0x29e9e500000005fb, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x3f948fd1 | vkCreateImageView(): pCreateInfo->format VK_FORMAT_D24_UNORM_S8_UINT with tiling VK_IMAGE_TILING_OPTIMAL has no supported format features on this physical device. The Vulkan spec states: The format features of the resultant image view must contain at least one bit. (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkImageViewCreateInfo-None-02273)`
Call stack
`>	Vita3K.exe!renderer::vulkan::set_uniform_buffer(renderer::vulkan::VKContext & context, const renderer::ShaderProgram * program, const bool vertex_shader, const int block_num, const int size) Line 88	C++
 	Vita3K.exe!renderer::cmd_set_state_uniform_buffer(renderer::State & renderer, MemState & mem, Config & config, renderer::CommandHelper & helper, renderer::Context * render_context, const char * base_path, const char * title_id) Line 124	C++
 	[External Code]	
 	[Inline Frame] Vita3K.exe!std::_Func_class<void,renderer::State &,MemState &,Config &,renderer::CommandHelper &,renderer::Context *,const char *,const char *>::operator()(renderer::State & _Args, MemState & _Args, Config & _Args, renderer::CommandHelper & _Args, renderer::Context * _Args, const char * _Args, const char * _Args) Line 880	C++
 	Vita3K.exe!renderer::cmd_handle_set_state(renderer::State & renderer, MemState & mem, Config & config, renderer::CommandHelper & helper, const FeatureState & features, renderer::Context * render_context, const char * base_path, const char * title_id, const char * self_name) Line 493	C++
 	[External Code]	
 	[Inline Frame] Vita3K.exe!std::_Func_class<void,renderer::State &,MemState &,Config &,renderer::CommandHelper &,const FeatureState &,renderer::Context *,const char *,const char *,const char *>::operator()(renderer::State & _Args, MemState & _Args, Config & _Args, renderer::CommandHelper & _Args, const FeatureState & _Args, renderer::Context * _Args, const char * _Args, const char * _Args, const char * _Args) Line 880	C++
 	Vita3K.exe!renderer::process_batch(renderer::State & state, const FeatureState & features, MemState & mem, Config & config, renderer::CommandList & command_list, const char * base_path, const char * title_id, const char * self_name) Line 95	C++
 	Vita3K.exe!renderer::process_batches(renderer::State & state, const FeatureState & features, MemState & mem, Config & config, const char * base_path, const char * title_id, const char * self_name) Line 122	C++
 	Vita3K.exe!SDL_main(int argc, char * * argv) Line 372	C++
 	Vita3K.exe!main_getcmdline() Line 82	C
`

Macdu:
Yes, thank you for the validation error, that's the helpful part. It looks like AMD GPUs do not support the D24U8 depth stencil format. In my latest update I switched to F32U8 depth stencils, it should fix the issue.

bookmist:
Thank you. No more validation error but still crash on RelWithDebugInfo build (msvc-clang compiler)

Macdu:
Do you have more infos on the crash?

bookmist:
Not much. Only call stack (same as it was) and debugger message. I'm going to try other compiler (msvc).

vita3k/features/include/features/state.h
```diff
@@ -22,8 +22,8 @@ struct FeatureState {
     bool support_texture_barrier = false; ///< Second option for blending. Slower but work on 3 vendors.
     bool direct_fragcolor = false;
     bool spirv_shader = false;
-    bool preserve_f16_nan_as_u16 = false; ///< Emit store of 4xU16 to draw buffer 1. This buffer is expected to be U16U16U16U16, which can be casted to F16F16F16F16. This is to preserve some drivers's behaviour of casting NaN to default value when store in framebuffer, not keeping its original value.
     bool support_get_texture_sub_image = false;
+    bool preserve_f16_nan_as_u16 = false; ///< Emit store of 4xU16 to draw buffer 1. This buffer is expected to be U16U16U16U16, which can be casted to F16F16F16F16. This is to preserve some drivers's behaviour of casting NaN to default value when store in framebuffer, not keeping its original value.
```
Zangetsu38, line 26: revert this, no idea why have just changed line ^^

Macdu: Because I want to put features that are only supported by one renderer at the end

bookmist:
I recheck, msvc compiler works fine enough, no crash. Some errors
`[01:20:50.825] |E| [debug_callback]: Validation layer: Validation Error: [ UNASSIGNED-CoreValidation-Shader-FeatureNotEnabled ] Object 0: handle = 0x1f4e39a83d0, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x8c548bfd | Shader requires fragmentStoresAndAtomics but is not enabled on the device
[01:20:50.833] |E| [renderer::cmd_set_state_viewport]: Unimplemented graphics API handler with backend 1`

vita3k/features/include/features/state.h
Zangetsu38: ah eu ok, i have thinks is mistake so oki

vita3k/shaders-builtin/vulkan/compile.bat
```diff
@@ -0,0 +1,2 @@
+@echo off
+for %%i in (*.vert *.frag) do "C:\VulkanSDK\Bin\glslangValidator.exe" -V "%%~i" -o "%%~i.spv"
```
Zangetsu38, line 2: new line missing

xerpi: I think this can be turned into CMake using `add_custom_command`, etc.
Also check `Vulkan::glslangValidator` https://cmake.org/cmake/help/latest/module/FindVulkan.html

Macdu: This was just a personal helper to help me compile glsl files into spv. The shaders are really simple and should rarely (if ever) be changed so I will remove this file.

Moreover because Vita3K is now linking against Vulkan dynamically (and not statically), the Vulkan SDK is no longer required to compiler Vita3K, even with Vulkan enabled (there is no `find_package(Vulkan REQUIRED)` in the cmake anymore). It would be annoying to make it a requirement only to compile 2 static shaders.

EXtremeExploit:
```
[17:52:50.400] |E| [debug_callback]: Validation layer: Validation Error: [ VUID-VkVertexInputAttributeDescription-format-00623 ] Object 0: handle = 0x555559337940, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0xa8c1a564 | vkCreateGraphicsPipelines: pCreateInfo[0].pVertexInputState->vertexAttributeDescriptions[2].format (VK_FORMAT_R16G16B16_SFLOAT) is not a supported vertex buffer format. The Vulkan spec states: format must be allowed as a vertex buffer format, as specified by the VK_FORMAT_FEATURE_VERTEX_BUFFER_BIT flag in VkFormatProperties::bufferFeatures returned by vkGetPhysicalDeviceFormatProperties (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkVertexInputAttributeDescription-format-00623)
[17:52:50.400] |E| [debug_callback]: Validation layer: Validation Error: [ VUID-VkVertexInputAttributeDescription-format-00623 ] Object 0: handle = 0x555559337940, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0xa8c1a564 | vkCreateGraphicsPipelines: pCreateInfo[0].pVertexInputState->vertexAttributeDescriptions[3].format (VK_FORMAT_R16G16B16_SFLOAT) is not a supported vertex buffer format. The Vulkan spec states: format must be allowed as a vertex buffer format, as specified by the VK_FORMAT_FEATURE_VERTEX_BUFFER_BIT flag in VkFormatProperties::bufferFeatures returned by vkGetPhysicalDeviceFormatProperties (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkVertexInputAttributeDescription-format-00623)
```
on Touhou Kobuto V: Burst Battle [PCSE01104]

vita3k/shader/src/spirv_recompiler.cpp
```diff
@@ -1374,14 +1414,47 @@ static spv::Function *make_vert_finalize_function(spv::Builder &b, const SpirvSh
 
                 cond_builder.makeBeginElse();
 
+                // Apply the viewport flip
                 if (translation_state.render_info_id != spv::NoResult) {
```
sunho, line 1418: I'm hazy on memory but the all intentions of this flipping machinary was because opengl didn't have enough tools to do this without hardcoding. With vulkan, it should be possible to use negative viewport size to emulate this cleanly.

vita3k/shader/src/spirv_recompiler.cpp
```diff
@@ -1618,7 +1701,7 @@ static std::string convert_spirv_to_glsl(const std::string &shader_name, SpirvCo
     }
     // Compile to GLSL, ready to give to GL driver.
     std::string source = glsl.compile();
-    return source;
+    return std::move(source);
```
sunho, line 1704: std::move isn't needed as we have copy elision by default since c++17.

vita3k/renderer/src/vulkan/texture.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+    const vk::ComponentMapping swizzle = translate_swizzle(format);
+
+    const bool is_cube = (gxm_texture.texture_type() == SCE_GXM_TEXTURE_CUBE || gxm_texture.texture_type() == SCE_GXM_TEXTURE_CUBE_ARBITRARY);
+
+    uint32_t width = static_cast<uint32_t>(gxm::get_width(&gxm_texture));
+    uint32_t height = static_cast<uint32_t>(gxm::get_height(&gxm_texture));
+    if (gxm::is_block_compressed_format(base_format)) {
+        // align width and height to block size
+        width = (width + 3) & ~3;
```
sunho, line 156: we have align function in util module if you didn't know about this before.

vita3k/renderer/src/vulkan/texture.cpp
```diff
@@ -0,0 +1,351 @@
...
+        vk::BufferCreateInfo buffer_info{
+            .size = upload_size,
+            .usage = vk::BufferUsageFlagBits::eTransferSrc,
+            .sharingMode = vk::SharingMode::eExclusive
+        };
+        std::tie(cache.staging_buffer, cache.alloc) = cache.state.allocator.createBuffer(buffer_info, vkutil::vma_mapped_alloc, cache.alloc_info);
+    }
+
+    memcpy(cache.alloc_info.pMappedData, text_data, upload_size);
+    vk::CommandBuffer cmd_buffer = vkutil::create_single_time_command(cache.state.device, cache.state.transfer_command_pool);
```
sunho, line 312: This is totally fine, but we can do some fancy optimization later by using tansfer queue -- we might have to batch another stream of commands with semaphore syncing between render command list.

vita3k/renderer/src/vulkan/pipeline_cache.cpp
```diff
@@ -0,0 +1,550 @@
...
+        .compareOp = translate_stencil_func(state.func)
+    };
+}
+
+vk::Pipeline PipelineCache::retrieve_pipeline(VKContext &context, SceGxmPrimitiveType &type, MemState &mem) {
+    current_context = &context;
+    const GxmRecordState &record = context.record;
+    // get the hash of the current context
+    constexpr size_t record_pipeline_len = offsetof(GxmRecordState, vertex_streams);
+    uint64_t key = XXH_INLINE_XXH3_64bits(&record, record_pipeline_len);
```
sunho, line 422: What a key to use. I love it. I think this can get interesting when compiler inserts padding though. (e.g. there is no standard way to clear the padding bytes in c++)

sunho (commented):
This is really awesome again. It already looks quite complete and I can already see the bright future we're heading at. Just few things caught my eyes while I'm looking around.

vita3k/shader/src/spirv_recompiler.cpp
Macdu: Oh indeed, I didn't know you could specify a negative width and height in a viewport, this will be much cleaner.

vita3k/shader/src/spirv_recompiler.cpp
Macdu: Ah yes, I added some std::move, realized they were not necessary and removed them. I forgot to remove this one.

vita3k/renderer/src/vulkan/texture.cpp
Macdu: Yes, the code for uploading texture is really poorly optimized (`create_single_time_command` should not be called for anything that happens more than once a frame). I plan on uploading textures on a parallel thread using a transfer queue, but I will do this in a separate PR after this is merged.

vita3k/renderer/src/vulkan/pipeline_cache.cpp
Macdu: You're right, but the record is always the same (context.record), so if there are paddings they should keep the same value while the emulator is running.

vita3k/shader/src/spirv_recompiler.cpp
Macdu: With the latest update, a negative viewport height is used instead of the viewport flip in the shader.

IrisHeartCPU:
the image says all

i just installed vulkan build and this happens to me, dunno why, dunno how to fix it. but its really strange

Raezroth:
I tried Freedom Wars on an AMD RX 580 on Vulkan and it renders upside down XD
On Windows BTW. 

unknown_2022 08 19-14 12_1

Raezroth:
The US version of Freedom Wars renders properly though

unknown_2022 08 19-14 18

Macdu:
Ok, I think this PR is ready to be reviewed. Tthe Vulkan renderer has at least the same (or a better) quality on 80-90% of the games. 

For the rest what happens is that Vulkan is a lot less leniant compared to OpenGL so some things that cause only graphical errors in the OpenGL renderer would cause the Vulkan renderer to crash. I don't really see this as an issue because usually the Vulkan renderer crashes with a nice validation error making it easier to debug and fixing the crash in Vulkan would solve the graphical issue in OpenGL so it's not really a Vulkan issue.

What is missing compared to the OpenGL renderer:

- Real programmable blending with shader interlock. Not really an issue, the only game I'm aware of which uses it is One piece unlimited (for the text).
- The mask bit. As far as I'm aware it is only used by VitaGL homebrews and Minecraft
- Surface sync: only used by a few games, and kills the performance

The cleanup is also not done when the game is closed but that's not really an issue as it is not possible to boot a new game without rebooting Vita3K for now.

Zangetsu38 (approved):

Zangetsu38 (approved):

bokssssss (commented):
Macdu:vulkan

PR #1978, opened 2022-07-27
vita3k: Remove unnecessary dependencies between librairies and files

Macdu:
Vita3K has a really annoying design flaw in that almost all files include `emuenv/state.h` and this file include more than half of all the headers of vita3k (and many external librairies). This leads for exemple ngs headers to include `spirv_recompiler.h`. So when I modify one line in `spirv_recompiler.h` (or in a lot of headers), I have to wait 15 minutes for the whole project to recompile.

This PR forwards declare all the dependencies in `emuenv/state.h` so each header should only include headers it needs (or at least less than before), so I should not have to wait 15 minutes every time I change a line in a header. This has also the side effect to divide by 2 the time to build Vita3K from scratch.

If anyone has a suggestion to make the content of `emuenv/state.h` and `emuenv/emuenv.cpp` cleaner, I will take it.

### DIFF
--- a/vita3k/CMakeLists.txt
+++ b/vita3k/CMakeLists.txt
@@ -127,7 +127,7 @@ add_subdirectory(packages)
 
 add_executable(vita3k MACOSX_BUNDLE main.cpp interface.cpp interface.h performance.cpp)
 
-target_link_libraries(vita3k PRIVATE app gui modules packages)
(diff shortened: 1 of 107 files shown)


EXtremeExploit (approved):
i dont understand the changes at all, they are overwhelming but i approve anyways because why not

Macdu:
I'll ask @hobyst to approve also, just in case.

hobyst (approved):

PR #1979, opened 2022-07-28
AppUtil: Fix read safemem pointer

EXtremeExploit:
Earlier it was writing the buffer pointer pointer, now its the actual buffer pointer where data is located

### DIFF
--- a/vita3k/modules/SceAppUtil/SceAppUtil.cpp
+++ b/vita3k/modules/SceAppUtil/SceAppUtil.cpp
@@ -343,7 +343,7 @@ EXPORT(int, sceAppUtilSaveDataUmount) {
     return UNIMPLEMENTED();
 }
 
-static SceInt32 SafeMemory(EmuEnvState &emuenv, const void *buf, SceSize bufSize, SceOff offset, const char *export_name, bool save) {
+static SceInt32 SafeMemory(EmuEnvState &emuenv, void *buf, SceSize bufSize, SceOff offset, const char *export_name, bool save) {
     std::vector<char> safe_mem(SCE_APPUTIL_SAFEMEMORY_MEMORY_SIZE);
     const auto safe_mem_path = construct_savedata0_path("sce_sys/safemem", "dat");
     SceInt32 res = 0;
@@ -363,7 +363,7 @@ static SceInt32 SafeMemory(EmuEnvState &emuenv, const void *buf, SceSize bufSize
         write_file(fd, safe_mem.data(), SCE_APPUTIL_SAFEMEMORY_MEMORY_SIZE, emuenv.io, export_name);
         close_file(emuenv.io, fd, export_name);
     } else
-        memcpy(&buf, &safe_mem[offset], bufSize);
+        memcpy(buf, &safe_mem[offset], bufSize);
 
     return res;
 }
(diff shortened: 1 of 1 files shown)


vita3k/modules/SceAppUtil/SceAppUtil.cpp
```diff
@@ -363,7 +363,7 @@ static SceInt32 SafeMemory(EmuEnvState &emuenv, const void *buf, SceSize bufSize
         write_file(fd, safe_mem.data(), SCE_APPUTIL_SAFEMEMORY_MEMORY_SIZE, emuenv.io, export_name);
         close_file(emuenv.io, fd, export_name);
     } else
-        memcpy(&buf, &safe_mem[offset], bufSize);
+        memcpy(const_cast<void *>(buf), &safe_mem[offset], bufSize);
```
Macdu, line 366: Can you remove the `const_cast` and instead remove the `const ` in the function declaration?

Zangetsu38: no is const, keep it

Macdu: If you need to const cast, then it's not const, as you can see this function is writing to the location this pointer points to.

Zangetsu38: save is const, load, i hgave code like that for keep const in save

Macdu: Yes, save is const but load is not and this function does both so do not make it const. `const_cast` should only be used really exceptionally.

sunho: it should be changed on function declaration as macdu said. Gotta remote const_cast.

Zangetsu38: > it should be changed on function declaration as macdu said. Gotta remote const_cast.

what function declared ? don't touch on header of official function please 💀

Macdu: We are not talking about the header of the official functions which are correct, we are talking about the static function `SafeMemory`...

Zangetsu38: noworks for me, remove it get crash on load save in escape plain

Zangetsu38: Escape plain using Save mode for load save, and after in everyu chapter finished, no works without const

EXtremeExploit: sceAppUtilSaveSafeMemory has a const void* so i have to use const_cast there or not make the buffer not const on that function, which one is prefered?

Zangetsu38: Const cast is good for me, is not possible remove const for save

Macdu: Yes, use the `const_cast` in `sceAppUtilSaveSafeMemory`, it is better.

EXtremeExploit: should be good now

Macdu (approved):

PR #1980, opened 2022-07-28
Intergrate Conan into build pipeline

hobyst:
## Description
This PR creates a Conan recipe (`conanfile.py`) to allow developers to satisfy the pre-built dependency requirements without having to use pre-built binaries provided by the project. It also creates a new CMake script `CopySharedLib.cmake` which can be run at build time by the underlying build system to copy shared libraries and other artifacts needed in the binary folder that also follows and replicates symbolic link chains.

## Requires
- [x] Vita3K/psvpfstools#9

## To-do
- [x] Test on Windows
- [x] Need help to write a macOS Conan profile and test if everything works properly (thanks to @KhoraLee for the help)
- [ ] ~~Set up [Vita3K/ext-deps](https://github.com/Vita3K/ext-deps) so that people can easily obtain pre-built packages of all the dependencies needed without needing to use Conan~~


sunho:
Nice, why not ready for review?

hobyst:
> Nice, why not ready for review?

Because I'm making some changes locally so that it can be enforced to all code related to the emulator more easily, including Boost and FFmpeg.

cmake/conan/profiles/linux-clang_libc++
```diff
@@ -0,0 +1,16 @@
+[settings]
+os=Linux
+os_build=Linux
+arch=x86_64
+arch_build=x86_64
+compiler=clang
+# Using Clang 11 so that no issues arise in GitHub CI, but any version newer than that can be used just fine
```
sunho, line 7: Do you mean clang lower than 11 fails in CI?

EXtremeExploit:
i think pr name sould be changed, it turned into making linux use libc++ into making vita3k use dependencies from conan and improving their integration

cmake/conan/profiles/linux-clang_libc++
hobyst: I mean that since the default Clang version in CI is Clang 11, setting it to a higher value would probably cause issues with CI and therefore fail the build due to the Conan preset requiring a higher Clang version than what CI actually uses.

sunho:
Maybe worth splitting off pr if you want. I like introducing conan btw -- much better than our fragile DIY boost building.

cmake/conan/profiles/linux-clang_libc++
sunho: Yup that make sense.

hobyst:
> Maybe worth splitting off pr if you want. I like introducing conan btw -- much better than our fragile DIY boost building.

I initially just wanted to enable libc++ (Clang) for Linux builds, in which I succeeded to do so without Conan (@EXtremeExploit has been trying it as well using the first commit I made to this PR and we can confirm substantial performance improvements just by using it instead of libstdc++). But since Boost and FFmpeg were still built with libstdc++ (GNU) I needed to rework the way those two are built so that I could enforce the C++ library I wanted. I could split it into separate PRs but since I re-adapted the libc++ settings to have Conan in mind, it's more straightforward to keep both on the same PR.

sunho:
> > Maybe worth splitting off pr if you want. I like introducing conan btw -- much better than our fragile DIY boost building.
> 
> I initially just wanted to enable libc++ (Clang) for Linux builds, in which I succeeded to do so without Conan (@EXtremeExploit has been trying it as well using the first commit I made to this PR and we can confirm substantial performance improvements just by using it instead of libstdc++). But since Boost and FFmpeg were still built with libstdc++ (GNU) I needed to rework the way those two are built so that I could enforce the C++ library I wanted. I could split it into separate PRs but since I re-adapted the libc++ settings to have Conan in mind, it's more straightforward to keep both on the same PR.

No worries. Just do in the way you prefer to. I'm totally fine with either. I just said that cause I thought libc++ needed more works to make this WIP.

hobyst:
> If libc++ and lld packages are now required would be nice to add them to the building instructions, same goes for conan if necessary

Already did it for Conan but not for libc++ an lld. Thanks for reminding me.

sunho:
> If libc++ and lld packages are now required would be nice to add them to the building instructions, same goes for conan if necessary

I don't think we should default to clang/lld/libc++ family yet. It's nice to have as an option but default should be gcc/ld/libstdc++.

hobyst:
> > If libc++ and lld packages are now required would be nice to add them to the building instructions, same goes for conan if necessary
> 
> I don't think we should default to clang/lld/libc++ family yet. It's nice to have as an option but default should be gcc/ld/libstdc++.

We are already using clang/ld/libstdc++ as default for CI builds. Using the full LLVM C++ stack (clang/lld/libc++) does not only make compilation and linking faster but also makes Vita3K run faster and solves several SIGSEGVs that occur when using clang with libstd++.

hobyst:
Since many of the issues I've faced while doing this PR came from Conan being in a transition state towards Conan 2.0 and even thought by using explicit package revisions and explicit Conan versions no issues should arise, I think I will leave it as a WIP for the time being and only make it ready to review once Conan 2.0 is released as stable and I've migrated everything to work as it should with that version.

Macdu:
From what I understand, Conan is only used for boost right now, right?

hobyst:
> From what I understand, Conan is only used for boost right now, right?

@Macdu Yes. The FFmpeg recipe hasn't been ported to Conan 2.0 yet.

Macdu:
Is openssl for windows supported ? It would be far better than using prebuilts from some 3rd party website. GitHub action is using vcpkg for openssl on Windows right now.

hobyst:
@Macdu @EXtremeExploit The recipe of OpenSSL 3.x seems to now work with Conan 2.0. These are the options exposed by the package recipe configure the build:
<details>
<summary>List of options</summary>

```python
options = {
    "shared": [True, False],
    "fPIC": [True, False],
    "enable_weak_ssl_ciphers": [True, False],
    "386": [True, False],
    "capieng_dialog": [True, False],
    "enable_capieng": [True, False],
    "no_aria": [True, False],
    "no_asm": [True, False],
    "no_async": [True, False],
    "no_blake2": [True, False],
    "no_bf": [True, False],
    "no_camellia": [True, False],
    "no_chacha": [True, False],
    "no_cms": [True, False],
    "no_comp": [True, False],
    "no_ct": [True, False],
    "no_cast": [True, False],
    "no_deprecated": [True, False],
    "no_des": [True, False],
    "no_dgram": [True, False],
    "no_dh": [True, False],
    "no_dsa": [True, False],
    "no_dso": [True, False],
    "no_ec": [True, False],
    "no_ecdh": [True, False],
    "no_ecdsa": [True, False],
    "no_engine": [True, False],
    "no_filenames": [True, False],
    "no_fips": [True, False],
    "no_gost": [True, False],
    "no_idea": [True, False],
    "no_legacy": [True, False],
    "no_md2": [True, False],
    "no_md4": [True, False],
    "no_mdc2": [True, False],
    "no_module": [True, False],
    "no_ocsp": [True, False],
    "no_pinshared": [True, False],
    "no_rc2": [True, False],
    "no_rc4": [True, False],
    "no_rc5": [True, False],
    "no_rfc3779": [True, False],
    "no_rmd160": [True, False],
    "no_sm2": [True, False],
    "no_sm3": [True, False],
    "no_sm4": [True, False],
    "no_srp": [True, False],
    "no_srtp": [True, False],
    "no_sse2": [True, False],
    "no_ssl": [True, False],
    "no_stdio": [True, False],
    "no_seed": [True, False],
    "no_sock": [True, False],
    "no_ssl3": [True, False],
    "no_threads": [True, False],
    "no_tls1": [True, False],
    "no_ts": [True, False],
    "no_whirlpool": [True, False],
    "no_zlib": [True, False],
    "openssldir": [None, "ANY"],
}
```

</details>

The default configuration is to compile OpenSSL as a static library with position-independent code enabled, MD2 support disabled.
```python
default_options = {key: False for key in options.keys()}
default_options["fPIC"] = True
default_options["no_md2"] = True
```
Which options should be overwritten (if any) for Vita3K aside from building it as a shared library?

EXtremeExploit:
ig that having mostly all enabled for now is fine, we can experiment on exactly what we need from openssl as we go on imo

hobyst:
@Macdu @EXtremeExploit After some more testing, it seems the Conan recipe for OpenSSL 3.x is still broken because the header files aren't included in the include path by CMake.

Even though this means, that the PR should be ready to merge, I still have to make some changes before merging to re-implement download of pre-built packages because Zangetsu asked so in order to be able to build the project without using Conan and without any user interaction.

imWatchful:
Any plans to pick this back up?

PR #1982, opened 2022-07-29
vita3k: Fix typo on total ram

EXtremeExploit:


Zangetsu38:
you are sure is MiB for it ? 🤔

EXtremeExploit:
ill have to look further into it, if its MB or MiB

EXtremeExploit:
alright i can confirm it is using MiB, i have 50.465.370.112 bytes of ram and dividing that by 2014 two times brings 48127, which is how much memory vita3k says i have

sunho (approved):

PR #1983, opened 2022-07-30
kernel: Simplify handle mutex timeout function

EXtremeExploit:


sunho:
I actually liked the past way better as it has "if this, do this."

EXtremeExploit:
> I actually liked the past way better as it has "if this, do this."

was thinking of simplifying it in terms of nested if statements, if you prefer the old way i can just close

sunho:
Well, it's really of personal tatse, so I feel like it's better to keep as it is. If someone else think this way is better, we can merge this.

EXtremeExploit:
will close then, if something comes up i might reopen

PR #1988, opened 2022-07-31
modules: Add tracy functions to SceRtc

EXtremeExploit:


vita3k/modules/SceRtc/SceRtc_tracy.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+    std::stringstream tickss;
+    tickss << tick;
+    tracy_settings.args.tick += tickss.str();
+
+    // Send args
+    ___tracy_scoped_zone->Text(tracy_settings.args.tick.c_str(), tracy_settings.args.tick.size());
+}
+
+#endif // TRACY_ENABLE
```
Macdu, line 186: Missing newline here and in SceRtc_tracy.h.

Macdu:
What annoys a bit with this PR (and SceCtrl) is that the content it adds is not really complete and often not that useful for debugging. Like for example when _sceRtcGetCurrentClockLocalTime is called I'd like to get the input timezone and the output of the function (the year, month, minute...) structure, otherwise I don't see it as really useful.

EXtremeExploit:
i wish there was a somewhat better way to handle structs in tracy, i will probably wait for #2007 to be finished and have tracy functions available on the whole project, that way it reduces code size and my sanity, maybe a way to handle structs smartly comes with it, in the meantime i will be drafting

PR #1990, opened 2022-07-31
modules: Add tracy support to SceCtrl

EXtremeExploit:


vita3k/modules/SceCtrl/SceCtrl_tracy.cpp
```diff
@@ -0,0 +1,578 @@
...
+    case SCE_CTRL_MODE_ANALOG_WIDE:
+        tracy_settings.args.mode += "SCE_CTRL_MODE_ANALOG_WIDE";
+        break;
+    }
+
+    // Send args
+    ___tracy_scoped_zone->Text(tracy_settings.args.mode.c_str(), tracy_settings.args.mode.size());
+}
+
+#endif // TRACY_ENABLE
```
Macdu, line 578: Missing newline at the end of the file (also in SceCtrl_tracy.h)

bookmist:
I'm not sure, but maybe `#ifdef TRACY_ENABLE ... #endif`` could be converted into macros. It makes code more compact.

bookmist:
Sorry, i make a bad explanation of my idea. I suggest something like this 
`#ifdef TRACY_ENABLE 
#define TRACY_FUNC(name, ...) \
    bool _tracy_activation_state = config::is_tracy_advanced_profiling_active_for_module(emuenv.cfg.tracy_advanced_profiling_modules, tracy_module_name); \
    ZoneNamedN(___tracy_scoped_zone, #name, _tracy_activation_state); \
    tracy_##name(&___tracy_scoped_zone, _tracy_activation_state, ##__VA_ARGS__); 
#else 
#define TRACY_FUNC(name, ...)
#endif


EXPORT(int, sceCtrlGetControllerPortInfo, SceCtrlPortInfo *info) {
    TRACY_FUNC(sceCtrlGetControllerPortInfo, info)`

EXtremeExploit:
> Sorry, i make a bad explanation of my idea. I suggest something like this
> 
> ```
> #ifdef TRACY_ENABLE 
> #define TRACY_FUNC(name, ...) \
>     bool _tracy_activation_state = config::is_tracy_advanced_profiling_active_for_module(emuenv.cfg.tracy_advanced_profiling_modules, tracy_module_name); \
>     ZoneNamedN(___tracy_scoped_zone, #name, _tracy_activation_state); \
>     tracy_##name(&___tracy_scoped_zone, _tracy_activation_state, ##__VA_ARGS__); 
> #else 
> #define TRACY_FUNC(name, ...)
> #endif
> 
> 
> EXPORT(int, sceCtrlGetControllerPortInfo, SceCtrlPortInfo *info) {
>     TRACY_FUNC(sceCtrlGetControllerPortInfo, info)
> ```

i think i told hobyst about this and he said that it may not work so i didn't even bother, i'll have the current tracy prs merged and then implement multiple modules in a single PR with the macro if it works, thanks for suggestion tho, better than if i would do it on my own lol

vita3k/modules/SceCtrl/SceCtrl_tracy.cpp
```diff
@@ -0,0 +1,578 @@
...
+    } tracy_settings;
+
+    // Log enumerated args
+
+    std::stringstream infoss;
+    infoss << info;
+    tracy_settings.args.info += infoss.str();
+
+    // Send args
+    ___tracy_scoped_zone->Text(tracy_settings.args.info.c_str(), tracy_settings.args.info.size());
```
Macdu, line 49: I'm not sure, but can't all of this be simplified into:
```
const std::string infos = fmt::format("&info : {}", log_hex(static_cast<uin64_t>(info)));
___tracy_scoped_zone->Text(infos.c_str(), infos.size()); 
```
Moreoever, right now on Tracy it shows something like `&info: 000000040154FA6C`. Printing the host pointer is not really useful, the least that can be done would be to print the guest pointer (or better the content of the struct even though it would be a bit harder to code).

EXtremeExploit: the goal of printing the host pointer was to check in case its an invalid pointer, like 0 or 4, which are somewhat common, i would mainly print pointers in case they are out pointers. also yeah you are right, it can be simpler, i was just following the way it was first implemented

vita3k/modules/SceCtrl/SceCtrl.cpp
```diff
@@ -20,12 +20,28 @@
...
+template <>
+std::string to_debug_str<SceCtrlPadInputMode>(const MemState &mem, SceCtrlPadInputMode mode) {
+    switch (mode) {
+    case SCE_CTRL_MODE_DIGITAL:
+        return "SCE_CTRL_MODE_DIGITAL";
+    case SCE_CTRL_MODE_ANALOG:
+        return "SCE_CTRL_MODE_ANALOG";
+    case SCE_CTRL_MODE_ANALOG_WIDE:
+        return "SCE_CTRL_MODE_ANALOG_WIDE";
+    }
```
bookmist, line 42: Maybe add something for other values. I'm going to use 
`default: return to_debug_str(mem, static_cast<typename std::underlying_type<SceCtrlPadInputMode>::type>(mode));`
or something like

bookmist (dismissed):

bookmist (approved):

PR #1995, opened 2022-08-02
Fixed gramatical errors, fixed linux build directory not matching the .gitignore

xero-lib:
Updated the `lang` files with proper English, and modified the `gen-linux.sh` script to build to the proper directory name, `build-linux`, to match the Windows equivalent, `build-windows`.


gen-linux.sh
```diff
@@ -2,4 +2,4 @@
 set -ex
 
 # Generate project files for Ninja
-cmake --preset linux-ninja-clang
+cmake --preset linux-ninja-clang -B build-linux
```
EXtremeExploit, line 5: Actually build-linux is not used anymore, just forgot to remove the os specific build folders out of the gitignore, now it's all in a single build folder, revert this one, the rest i think @Zangetsu38 should handle it

EXtremeExploit: ah look like windows still has its own folder, just linux uses `build`, while windows uses `build-windows`

xero-lib: On it.

PR #2000, opened 2022-08-02
gui/lang: correct lang typo and translation.

Croden1999:
correct lang typo and translation, and also add the "Full Screen" tip in control dialog (someone need this). if someone need me change some things, i'll check it.

plz Mr. @Zangetsu38 check this code.

### DIFF
--- a/lang/de.xml
+++ b/lang/de.xml
@@ -3,7 +3,7 @@
 <lang>
 	<main_menubar>
 		<file name="Datei">
-			<open_pref_path>Öffne Präferenzen-Pfad...</open_pref_path>
+			<open_pref_path>Öffne Präferenzen-Pfad</open_pref_path>
 			<install_firmware>Installiere Firmware</install_firmware>
 			<install_pkg>Installiere .pkg</install_pkg>
 			<install_zip>Installiere .zip, .vpk</install_zip>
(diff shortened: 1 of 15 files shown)


lang/de.xml
```diff
@@ -25,10 +25,10 @@
 	</main_menubar>
 
 	<about name="Über">
-		<vita3k>Vita3K: a PS Vita/PSTV Emulator. The world's first functional PS Vita/PSTV emulator.</vita3k>
+		<vita3k>Vita3K: a PS Vita/PS TV Emulator. The world's first functional PS Vita/PS TV emulator.</vita3k>
```
Zangetsu38, line 28: no sur for this ? 🤔  i have all time thinks is PSTV like PSVita

vita3k/gui/src/controls_dialog.cpp
```diff
@@ -101,6 +101,7 @@ void draw_controls_dialog(GuiState &gui, EmuEnvState &emuenv) {
     ImGui::Separator();
     ImGui::Spacing();
     ImGui::TextColored(GUI_COLOR_TEXT_MENUBAR, "%-16s", lang["gui"].c_str());
+    ImGui::Text("%-16s    %-16s", lang["full_screen"].c_str(), "F11");
```
Zangetsu38, line 104: ah yes good idea lol

lang/de.xml
Croden1999: but in Asia it called "PS Vita TV", USA and EU are called "PS TV", so i think it is name different but they same is one thing.

Zangetsu38: oki i see

Zangetsu38 (approved):

PR #2007, opened 2022-08-04
modules: Universal tracy macros (SceAudiodecUser as example)

bookmist:
I try to make universal macro, which can be used to log function parameters with tracy. Now this PR is only a demo, with one module as example.

### DIFF
--- a/vita3k/modules/include/modules/tracy.h
+++ b/vita3k/modules/include/modules/tracy.h
@@ -1,0 +1,80 @@
+// Vita3K emulator project
+// Copyright (C) 2022 Vita3K team
+//
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation; either version 2 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+#pragma once
+
+#include <mem/state.h>
+#include <sstream>
+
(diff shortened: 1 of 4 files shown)


EXtremeExploit:
hmm having the tracy macros defined in the module is not that great, you would have to copy pase them anywhere, maybe it would he nice to have them in utils or something like that, they will be used around the project

bookmist:
I was going to do so, but had no idea where to put this universal macro. So if utils is a good place - it's great.

vita3k/modules/include/modules/tracy.h
```diff
@@ -0,0 +1,79 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+#pragma once
+
+#include <mem/state.h>
```
EXtremeExploit, line 20: also `#include <sstream>` for stringstream on linux end so compiler doesn't give an error on some systemd

vita3k/modules/include/modules/tracy.h
```diff
@@ -0,0 +1,79 @@
...
+// I'm not sure it's needed only in case of tracy enabled
+template <typename T>
+std::string to_debug_str(const MemState &mem, T data) {
+    std::stringstream datass;
+    datass << data;
+    return datass.str();
+}
+
+template <>
+std::string to_debug_str<std::string>(const MemState &mem, std::string data) {
```
EXtremeExploit, line 31: make this inline

vita3k/modules/include/modules/tracy.h
```diff
@@ -0,0 +1,79 @@
...
+    return data;
+}
+
+std::string to_debug_str(const MemState &mem) {
```
EXtremeExploit, line 35: this inline too

EXtremeExploit (changes requested):

vita3k/modules/include/modules/tracy.h
bookmist: done. Thanks for help

EXtremeExploit (approved):

PR #2013, opened 2022-08-06
readme: Add Scoop in download section

sitiom:
[`vita3k`](https://scoop.sh/#/apps?q=vita3k&s=0&d=1&o=true) is available in the games bucket in Scoop.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,10 @@ Vita3K is licensed under the **GPLv2** license. This is largely dictated by exte
 
 ## Downloads
 * [Windows](https://github.com/Vita3K/Vita3K/actions?query=event%3Apush+is%3Asuccess+branch%3Amaster)
-  * [Visual C++ Redistributables](https://aka.ms/vs/16/release/vc_redist.x64.exe) (Install if not working.) 
+  * Scoop
+    * [vita3k](https://scoop.sh/#/apps?q=vita3k&s=0&d=1&o=true)
+  * Requirements:
+    * [Microsoft Visual C++ 2015-2022 Redistributable](https://aka.ms/vs/17/release/vc_redist.x64.exe) (Install if not working.) 
 * Linux
   * Arch based:
     * [vita3k-bin](https://aur.archlinux.org/packages/vita3k-bin)<sup><small>AUR</small></sup>
--- a/README.md
+++ b/README.md
@@ -36,16 +36,16 @@ The emulator currently runs some homebrew programs. It is also able to load some
 Vita3K is licensed under the **GPLv2** license. This is largely dictated by external dependencies, most notably Unicorn.
 
 ## Downloads
-* [Windows](https://github.com/Vita3K/Vita3K/actions?query=event%3Apush+is%3Asuccess+branch%3Amaster)
-  * Scoop
-    * [vita3k](https://scoop.sh/#/apps?q=vita3k&s=0&d=1&o=true)
+* Windows
+  * [Scoop](https://scoop.sh/#/apps?q=vita3k&s=0&d=1&o=true)
   * Requirements:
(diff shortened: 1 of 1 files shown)


Croden1999:
but we hope no change it, there is based the Vita3K team.

sitiom:
> but we hope no change it, there is based the Vita3K team.

?

Macdu:
The issue that I see with this scoop repository is that you download from [https://github.com/Vita3K/Vita3K/releases/download/continuous/windows-latest.zip](https://github.com/Vita3K/Vita3K/releases/download/continuous/windows-latest.zip) and check the hash. However this url gives the latest build and the hash will change with every new commit so it cannot be hardcoded.

Moreover, if you intend not to give the rights of the repository to a Vita3K dev, could you specify in the readme that the scoop download is unofficial ?

sitiom:
> The issue that I see with this scoop repository is that you download from [`continuous` windows-latest.zip (download)](https://github.com/Vita3K/Vita3K/releases/download/continuous/windows-latest.zip) and check the hash. However this url gives the latest build and the hash will change with every new commit so it cannot be hardcoded.

The bucket checks for manifest updates every 4 hours and updates the version and hashes based on the [`checkver` and `autoupdate` fields](https://github.com/Calinou/scoop-games/blob/3fe019f9b03041fabee8e7b004f6a48ff6578590/bucket/vita3k.json#L33-L57).

> Moreover, if you intend not to give the rights of the repository to a Vita3K dev

That's not how scoop works. Manifests reside in the bucket themselves. However, you are free to "maintain" it by sending out PRs to the [games](https://github.com/Calinou/scoop-games/) repository.

Macdu:
Oh alright, my bad. The only issue I have then is that in the scoop file the lang and shader-builtin folders are marked as persisting between updates whereas they should not be. And the shader folder should persist between update (it has its own mechanism to delete outdated shaders when necessary).

sitiom:
> Oh alright, my bad. The only issue I have then is that in the scoop file the lang and shader-builtin folders are marked as persisting between updates whereas they should not be. And the shader folder should persist between updates (it has its own mechanism to delete outdated shaders when necessary).

@Macdu Thanks. Please check out this PR (https://github.com/Calinou/scoop-games/pull/678).

Macdu:
Thank you, can you modify the scoop pr: the data folder should also not be persisted and it's the cache folder and not shader folder that should be persisted (the latter one does not exist, my mistake)

sitiom:
Just to clarify: `cache` and `config.yml` should only persist, correct?

Macdu:
Right

Macdu:
Right

sitiom:
> Thank you, can you modify the scoop pr: the data folder should also not be persisted and it's the cache folder and not shader folder that should be persisted (the latter one does not exist, my mistake)

Done.

Macdu:
Thanks, about the PR itself, I think your layout is kind of confusing, something like this would make more sense:

Windows
- [Download Artifact]()
- [Scoop]()
- Requirements

What do you think about it?

sitiom:
Thanks, changed the layout to be like this:

* Windows
  * [Scoop](https://scoop.sh/#/apps?q=vita3k&s=0&d=1&o=true)
  * Requirements:
    * [Microsoft Visual C++ 2015-2022 Redistributable](https://aka.ms/vs/17/release/vc_redist.x64.exe)
* Linux
  * Arch based:
    * [vita3k-bin](https://aur.archlinux.org/packages/vita3k-bin)<sup><small>AUR</small></sup>
    * [vita3k-git](https://aur.archlinux.org/packages/vita3k-git)<sup><small>AUR</small></sup>
* Others
  * [Download Artifact](https://github.com/Vita3K/Vita3K/actions?query=event%3Apush+is%3Asuccess+branch%3Amaster)

Macdu (approved):

PR #2020, opened 2022-08-15
Fixed Russian

Danik2343:
Fixed Russian


EXtremeExploit:
Why create a second pr for the same branch? you can just push a second commit to the same branch and it will be in the pr, no need to make another PR, fix indentation and squash the commits and it should be good

Croden1999:
u broken the format of this file, plz revert and change again.

PR #2024, opened 2022-08-19
threadmgr: Write SB register to actual cpu info in vm thread context info

EXtremeExploit:
Needs testing in unity games, i can confirm that HTR+ slot simulation has no change, maybe some other games do but i don't know


Zangetsu38:
Need more stuf and fix, with this current code majority unity game is broken rip digimon lol

Zangetsu38:
Why close ? You have no found how fix problem ? Maybe need also register other missing value 🤔

EXtremeExploit:
honestly i don't know how to fix it, idk how to get the values of st and teehbr, maybe adding those can fix the unity issue? but idk, its a small todo and i don't think it really makes any important progress or anything like that, if it gets done imo it should do the other todos aswell

PR #2027, opened 2022-08-22
cpu: Generalize CPU Error logging

EXtremeExploit:
This puts the logging the status of the CPU in a single place, Now it only displays the time on the unicorn/dynarmic error line and then the cpu status without the time at the start of each line
image

### DIFF
--- a/vita3k/cpu/src/dynarmic_cpu.cpp
+++ b/vita3k/cpu/src/dynarmic_cpu.cpp
@@ -15,6 +15,7 @@
 // with this program; if not, write to the Free Software Foundation, Inc.,
 // 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
+#include "cpu/common.h"
 #include <cpu/disasm/functions.h>
 #include <cpu/impl/dynarmic_cpu.h>
 #include <cpu/impl/interface.h>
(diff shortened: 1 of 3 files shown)


vita3k/cpu/src/dynarmic_cpu.cpp
```diff
@@ -127,6 +128,13 @@ class ArmDynarmicCallback : public Dynarmic::A32::UserCallbacks {
         Ptr<T> ptr{ addr };
         if (!ptr || !ptr.valid(*parent->mem)) {
             LOG_WARN("Invalid read of uint{}_t at address: 0x{:x}", sizeof(T) * 8, addr);
+            LOG_WARN("{}", this->cpu->save_context().description());
+
+            auto pc = this->cpu->get_pc();
+            if (pc == 0)
+                LOG_CRITICAL("PC is 0");
```
bookmist, line 135: null was just an example. In emu first memory page is always bad ( this->cpu->get_pc() < parent->mem->page_size), and for any other value I don't know, and I don't know how to check is pointer bad or not.

bookmist:
Thank you very much for this PR. It makes debugging with dynarmic much easier. I was trying to make something like this but did not finish my ideas.

vita3k/cpu/src/dynarmic_cpu.cpp
EXtremeExploit: ic, afaik the only way to check if the pointer is actually valid when the address looks right is to actually check its value, which would cause a segfault, however it is still better to have some checks than nothing like it was before

bookmist: First page is always bad. See https://github.com/Vita3K/Vita3K/blob/054a8119096596c65007e54bf3db49dc69d450b3/vita3k/mem/src/mem.cpp#L103-L109
My implementation of dynarmic error logging. maybe it explain my idea a bit better https://github.com/Vita3K/Vita3K/commit/5d86be362ffcd9b308a580066408a738343a7465

EXtremeExploit:
@bookmist should be good, now both ends log red text and have the same output except for the first line which says what went wrong, that is dependant on the backend and is good enough imo, i had an idea to also display to what module the PC belongs to on the executing line but emuenv is not available in dynarmic so i gave up ¯\_(ツ)_/¯

bookmist (approved):

PR #2033, opened 2022-08-25
CMake: Use LLD linker with Clang

hobyst:
Makes linking with Clang much faster.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -87,6 +87,9 @@
       "cacheVariables": {
         "CMAKE_C_COMPILER": "clang",
         "CMAKE_CXX_COMPILER": "clang++"
+      },
+      "environment": {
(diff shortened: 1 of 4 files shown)


CMakePresets.json
```diff
@@ -253,4 +256,4 @@
       "configurePreset": "linux-ninja-gnu"
     }
   ]
-}
+}
```
Zangetsu38, line 259: new line missing

hobyst: It's due to JSON formatting rules, which optimize the file to weight as less as possible.

hobyst: I've added the new line.

Zangetsu38 (approved):

Macdu (approved):

PR #2034, opened 2022-08-25
external: Update SDL to 2.24.0

EXtremeExploit:


Zangetsu38:
Curreent sdl is old 🤔 uptate it in same pr no have sense here

EXtremeExploit:
Delted the old code, i can confirm that mouse fully works on this version of code in linux, can you check on windows?

Zangetsu38:
Your code jeté is keep old sdl for build on action, update it

Zangetsu38:
you need update windows too

EXtremeExploit:
> you need update windows too

im confused, where is it specified in the workflow file?

Zangetsu38:
Is submodule

Macdu:
@EXtremeExploit It would be useful to explain (maybe not in the code but at least written in the PR description) that this part of the code can be removed because it is already handled this way starting with SDL 2.22.0 (as explain in the last paragraph of [https://wiki.libsdl.org/SDL_CaptureMouse](https://wiki.libsdl.org/SDL_CaptureMouse)). 
Removing this code would lead to a different behavior when using a sdl version between 2.16.0 and 2.22.0 (so we indeed need to make sure we are using at least sdl 2.22.0).

Zangetsu38:
Update submodule, and squash, onjly need one commit here
and juste named it
> external: Update sdl submodule to v2.24.0

any thinks like that

Macdu (approved):

PR #2035, opened 2022-08-26
vulkan: Add missing include on cache pipeline

EXtremeExploit:


vita3k/renderer/include/renderer/vulkan/pipeline_cache.h
```diff
@@ -19,6 +19,7 @@
 
 #include <array>
 #include <map>
+#include <numeric>
```
Macdu, line 22: I believe it should be 
```suggestion
#include <limits>
```
instead

EXtremeExploit: what about the TODO below it?

Macdu (approved):

PR #2038, opened 2022-08-26
vita3k: Bring back the macOS build

Macdu:
With the Vulkan renderer, we should be able to bring back the macOS build (with OpenGL) disabled.
Because the vulkan functions are fetched at runtime, if molten-vk is installed (`brew install molten-vk`), everything should work out of the box.

I just need someone with a Mac to try it before making this pr ready for review.


Zangetsu38:
this code no can works, you need change config for can set vulkan by default in mac, and you can remove ifndef create of ogl, need change code inside set value of backend render

ArbestRi:
While I have a Mac I don't know how to build.

ArbestRi:
Also, whenever I try to open the application (downloaded from GitHub Actions) with a M1 MacBook, macOS still says "This application cannot be opened".

Zangetsu38:
> 



> Also, whenever I try to open the application (downloaded from GitHub Actions) with a M1 MacBook, macOS still says "This application cannot be opened".

is normal, you need real cpu, m1 is not cpu compatible

merryhime:
Build runs when I `chmod +x` the executable and run it directly but not when executing the App bundle like a normal user would. May look into this when I have time.

Pedro-Beirao:
`xattr -d com.apple.quarantine ./Vita3K.app` solves the app not launching on double click.

Also tried 2 games, Motorstorm RC and Persona 4 golden (both work on windows), neither worked on mac.

Maybe moltenvk does not support some vulkan features needed?

___

> is normal, you need real cpu, m1 is not cpu compatible

It has nothing to do with the cpu. App works on m1

Macdu:
The Vulkan implementation of Vita3K should be able to work on MoltenVK, I made it so its requirements are very low (only Vulkan 1.0 and some standard extension that are supported by MoltenVk). From what I understand there is an error while compiling the MSL shaders generated from the SpirV recompiled shaders. 
However, the spirV generated shaders are all valid and I don't have a Mac, making it really hard for me to debug. If anyone with a Mac wants to try solving these issues, feel free to do so.

merryhime:
> It has nothing to do with the cpu. App works on m1

Vita3k's dynarmic fork would want to cherry-pick https://github.com/merryhime/dynarmic/commit/c60fd3f0ac8165c743ff9c11c300a1a99089fc7c, which works around a Rosetta 2 bug (Rosetta doesn't emulate `sahf` correctly).

Macdu:
Thanks a lot for KhoraLee for finding out what was causing the issue with metal. I've also added a workaround for the fact that metal does not support triangle fans.

vita3k/renderer/src/vulkan/scene.cpp
```diff
@@ -251,8 +251,51 @@ static void bind_vertex_streams(VKContext &context, MemState &mem) {
...
+        for (uint32_t triangle = 0; triangle < nb_triangle; triangle++) {
+            curr_indices[0] = old_indices[0];
+            curr_indices[1] = old_indices[triangle + 1];
+            curr_indices[2] = old_indices[triangle + 2];
+            curr_indices += 3;
+        }
+        old_indices += nb_triangle + 2;
+    }
+
+    delete[] reinterpret_cast<uint8_t *>(old_indices);
```
KhoraLee, line 280: Fails to free old_indices
```
Vita3K(37840,0x7ff853f38740) malloc: *** error for object 0x600002d78b7c: pointer being freed was not allocated
```
Maybe need to point first element before freeing memory?
`old_indices -= (nb_triangle + 2) * instance_count;`

shinra-electric:
The artefact in Github Actions fails to launch because it is missing `libssl.3.dylib` from the Frameworks folder. 

```
Termination Reason:    Namespace DYLD, Code 1 Library missing
Library not loaded: /usr/local/opt/openssl@3/lib/libssl.3.dylib
Referenced from: <36F74C9D-E96F-3793-AE13-FB220C48D8E7> /private/var/folders/*/Vita3K.app/Contents/MacOS/Vita3K
Reason: tried: '/usr/local/opt/openssl@3/lib/libssl.3.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/usr/local/opt/openssl@3/lib/libssl.3.dylib' (no such file), '/usr/local/opt/openssl@3/lib/libssl.3.dylib' (no such file), '/usr/local/lib/libssl.3.dylib' (no such file), '/usr/lib/libssl.3.dylib' (no such file, not in dyld cache)
(terminated at launch; ignore backtrace)
```

Macdu:
Have you installed openssh ? (`brew install openssh`)?

shinra-electric:
Yes, but I'm using an M1, so homebrew installs to `opt/homebrew` rather than `usr/local`. 

If you wanted it to locate it on both x86 and Arm, you should probably use `$(brew --prefix)` instead. 
But since Arm is not supported now, I assume that the x86 version would be better to use.  

But regardless, the the app bundle should have all the required frameworks included, and not require additional installations.

Macdu:
Ah I see, thank you for testing. I'll fix it and include the openssl dylib with Vita3K.

KhoraLee:
@shinra-electric 
Installing x86_64 version of homebrew and openssl will work.
`arch -x86_64 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"`
then
`arch -x86_64 /usr/local/bin/brew install openssl`

Zangetsu38 (dismissed):

KhoraLee:
@Macdu Could you add creating DMG process for CI? https://github.com/KhoraLee/Vita3K/commit/0157861e381ab3dfa126be3e23ec881f7aa5d184
This will preserve executable permission for Vita3K.app. Still require users to go setting and press 'Open Anyway' but they don't have to deal with terminal.
Screenshot 2022-10-30 at 7 52 34 PM
Screenshot 2022-10-30 at 7 52 42 PM
Screenshot 2022-10-30 at 7 53 01 PM

KhoraLee:
> @Macdu Could you add creating DMG process for CI? [KhoraLee@0157861](https://github.com/KhoraLee/Vita3K/commit/0157861e381ab3dfa126be3e23ec881f7aa5d184)
> This will preserve executable permission for Vita3K.app. Still require users to go setting and press 'Open Anyway' but they don't have to deal with terminal.

By the way this quarantine can be removed if we sign and notarize the app.

Macdu:
Thank you, I cherry-picked your commit. About signing the app, I really don't know how to do it on Mac. Feel free to do a PR for it after this PR is merged (it should happen really soon).

hobyst:
> Thank you, I cherry-picked your commit. About signing the app, I really don't know how to do it on Mac. Feel free to do a PR for it after this PR is merged (it should happen really soon).

@Macdu [App notarization for macOS](https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution) isn't something CI can do, at least as it is right now. Not only it requires sending the binary to Apple (and waiting for about an hour for the response) but also an Apple Developer ID which will be associated to the resulting notarization ticket (the data that validates a binary executable as "safe to run").

It could certainly be done in CI, but someone would have to either:
- Give off their Apple ID credentials to whoever manages the Vita3K repository so that they can be added as secrets in the GitHub CI settings and later used in CI builds for every merge on the `master` branch.
- Make the same CI setup above but for their own fork repo, and have their CI builds linked in the README.md as a convenient but unofficial way to get Vita3K for macOS
- Build and notarize Vita3K on their own macOS device and send it to devs. This would likely limit the notarization packages to only be available on major version changes and it should only be done if said 3rd party is some one trustworthy.

That said, if any of the main developers can do it themselves, then it could be possible to do the second option listed above without them having to handle over their Apple ID credentials while also being able to link to it as a (more or less) official source for macOS binaries. However, it would require them to at least keep one of the branches in their fork as up to date with the upstream as possible and without modifications. Info on how to use the macOS SDK CLI tools to notarize an app can be found [here](https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution).

KhoraLee:
@Macdu 
Screenshot 2022-10-30 at 11 35 15 PM
Screenshot 2022-10-30 at 11 34 38 PM
Dylibs are copied but still tries to load OpenSSL from `/usr/local/opt/openssl@3/lib` and crash.

How about statically link OpenSSL on macOS?
`set(OPENSSL_USE_STATIC_LIBS TRUE)`

merryhime:
>  Dylibs are copied but still tries to load OpenSSL from `/usr/local/opt/openssl@3/lib` and crash.

Would it not be easier to fixup the dylib lookup paths with either `install_name_tool` or a helper like `macpack`?

KhoraLee:
> Would it not be easier to fixup the dylib lookup paths with either `install_name_tool` or a helper like `macpack`?

Changed to @rpath/libssl.dylib but still tries to load from original path.
image

merryhime:
@KhoraLee You'll also need to modify the incorrect references inside `libssl.dylib` and `libcrypto.dylib`.

Macdu:
> @Macdu Screenshot 2022-10-30 at 11 35 15 PM Screenshot 2022-10-30 at 11 34 38 PM Dylibs are copied but still tries to load OpenSSL from `/usr/local/opt/openssl@3/lib` and crash.
> 
> How about statically link OpenSSL on macOS? `set(OPENSSL_USE_STATIC_LIBS TRUE)`

@KhoraLee Does openssl homebrew also has static libraries?

KhoraLee:
IIRC, It does.

KhoraLee:
Its working now without installing OpenSSL through homebrew

PR #2042, opened 2022-08-27
vulkan: Fix all vulkan warnings

EXtremeExploit:


vita3k/renderer/src/vulkan/pipeline_cache.cpp
```diff
@@ -172,7 +172,11 @@ void PipelineCache::read_pipeline_cache() {
 
 void PipelineCache::save_pipeline_cache() {
     size_t pipeline_size;
-    state.device.getPipelineCacheData(pipeline_cache, &pipeline_size, nullptr);
+    auto result = state.device.getPipelineCacheData(pipeline_cache, &pipeline_size, nullptr);
+    if (result != vk::Result::eSuccess) {
+        assert(false);
```
Macdu, line 177: Can you put LOG_ERRORs instead of assert_false ?

vita3k/renderer/src/vulkan/screen_renderer.cpp
```diff
@@ -374,7 +378,11 @@ void ScreenRenderer::swap_window() {
         .pImageIndices = &swapchain_image_idx,
     };
     try {
-        state.general_queue.presentKHR(present_info);
+        auto result = state.general_queue.presentKHR(present_info);
+        if (result != vk::Result::eSuccess) {
```
Macdu, line 382: vk::Result::eSuboptimalKHR is also considered a success

Macdu (dismissed):

vita3k/renderer/src/vulkan/pipeline_cache.cpp
EXtremeExploit: i prefer to leave the assert false in case one of these goes unnoticed and its a lot easier to catch and debug

Macdu: Can you at least add a LOG_ERROR then? if these lines cause issues, it will most likely be driver specific and assert don't do anything in release builds.

Macdu (approved):

PR #2043, opened 2022-08-27
Use the file browser provided by the user's desktop environment in Linux

hobyst:
This PR abstracts the file dialog calls into a C++ API to make porting and handling easier and changes the library used to implement them from nativefiledialog to nativefiledialog-extended, which allows friendly names for file filters and makes use of the XDG desktop portals in Linux to call the file browser of the desktop environment the user is using instead of always calling the GTK file browser.


vita3k/gui/src/archive_install_dialog.cpp
```diff
@@ -33,7 +32,7 @@ static bool delete_archive_file;
 static std::string state, type, title;
 static std::map<fs::path, std::vector<ContentInfo>> contents_archives;
 static std::vector<fs::path> invalid_archives;
-static nfdchar_t *archive_path;
+static std::filesystem::path archive_path = "";
```
Macdu, line 35: Can you use fs::filesystem::... instead for all the fs filesystem objects?

hobyst: Why? `fs::filesystem` is actually a nickname for `boost::filesystem` which is at the same time a C++ filesystem implementation previous to the creation of the official C++17 `std::filesystem`? Also `std::filesystem` plays way better with C++ and C code due to using all C++ types instead of custom Boost types.

Macdu: As you want, but everything else uses fs::filesystem so it was more for consistency.

hobyst: Is there any particular reason why the project still uses `boost::filesystem` instead of `std::filesystem`?

EXtremeExploit: #1912

EXtremeExploit (approved):

EXtremeExploit (approved):
thanks hobyst for the code, good code, good hobyst

PR #2045, opened 2022-08-28
shader: Improve texture instructions

Macdu:
This PR contains 4 commits, each one containing different improvements to the shader recompiler and in particular texture sampling.

- First, rewrite the hint system used by the shader recompiler. The current system is not enough, for some instructions we need to know the texture format (to be more precise the component type and the component count).
- Apply the color swizzle in the shader. Our current implementation didn't work when writing to a texture with different swizzles and didn't support some swizzles like SWIZZLE1_A, this is now fixed.
- Take into account the component count of a texture when sampling. If the texture has only one component (SWIZZLE1_R, for example a depth buffer), do not overwrite the memory at locations that would have matched the 2nd to 4th components.
- Implement missing texture info and texture gather instructions.

This fixes the graphics of many games, to name a few: Persona 4, Freedom War, Dragon Quest Builder, Trails of Cold Steel, Urban Trial...

I'm putting it as a draft because it seems there are still some issues with the shadows.

### DIFF
--- a/vita3k/renderer/include/renderer/functions.h
+++ b/vita3k/renderer/include/renderer/functions.h
@@ -183,7 +183,6 @@ uint16_t get_upload_mip(const uint16_t true_mip, const uint16_t width, const uin
 
 void upload_bound_texture(const TextureCacheState &cache, const SceGxmTexture &gxm_texture, const MemState &mem);
 void cache_and_bind_texture(TextureCacheState &cache, const SceGxmTexture &gxm_texture, MemState &mem);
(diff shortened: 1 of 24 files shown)


Macdu:
I'm also using this PR to fix the vtst_mask instruction which was returning the opposite result of what it should.

Macdu:
I removed the part which applies color swizzle in the shader as it breaks blending... I'll implement it later but only on surfaces with 1 or 2 components as it is necessary for some texture formats = SWIZZLE_G or SWIZZLE_AR.

Zangetsu38 (approved):

PR #2047, opened 2022-08-30
kernel: Refactor SELF loader to remove dependency on ELFIO

CreepNT:
Removes dependency on the external library ELFIO, only used in the SELF loader.
The refactored loader performs many more sanity checks on the input file and is aware of more SCE ELF specifics. This notably fixes `Unknown segment type 0x6FFFFF0x` in Ratchet & Clank and probably other games.
Some changes are commented out as TODO/FIXME because they require broader changes but are not critical.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -17,9 +17,6 @@
 [submodule "external/dlmalloc"]
 	path = external/dlmalloc
 	url = https://github.com/Vita3K/dlmalloc
-[submodule "external/elfio"]
-	path = external/elfio
-	url = https://github.com/serge1/ELFIO
 [submodule "external/ffmpeg"]
 	path = external/ffmpeg
 	url = https://github.com/Vita3K/ffmpeg-core.git
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -49,7 +49,6 @@ endif()
 add_library(printf INTERFACE)
 target_include_directories(printf INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}/printf")
 
-add_subdirectory(elfio)
 
 add_subdirectory(fmt)
 add_library(fmt::fmt ALIAS fmt)
--- a/vita3k/interface.cpp
+++ b/vita3k/interface.cpp
@@ -495,6 +495,8 @@ static ExitCode load_app_impl(Ptr<const void> &entry_point, EmuEnvState &emuenv,
         emuenv.kernel.export_nids.emplace(var.nid, addr);
     }
 
+    //FIXME: The application EBOOT should be the first module ever loaded in the address space!
+
     // Load pre-loaded libraries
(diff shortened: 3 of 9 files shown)


EXtremeExploit:
oh man i really love this changes, hopefully no more critical errors when booting any game at all, all segments would be exported as expected even the special ones like sce version and such. I do wonder tho if you are gonna do something about the TODOs you left, some look somewhat easy enough to do if you are familiar with this part of the code, which seems you are, is it gonna be a draft or once format check is fixed it is ready for merge?

CreepNT:
@EXtremeExploit 
> oh man i really love this changes, hopefully no more critical errors when booting any game at all, all segments would be exported as expected even the special ones like sce version and such. I do wonder tho if you are gonna do something about the TODOs you left, some look somewhat easy enough to do if you are familiar with this part of the code, which seems you are, is it gonna be a draft or once format check is fixed it is ready for merge?

The TL;DR is that either they're minor changes that probably won't improve compatibility but require some RE or a huge rework and I'd rather ship as-is in a mostly-functional-with-a-little-polish-needed than waste hours in RE for very little (if any) gain.

In more details:
1. `FIXME: The application EBOOT should be the first module ever loaded in the address space!` (and its counterpart `TODO: when the virtual process bringup is fixed, uncomment this`) require to rework the whole SELF loading process (most importantly: decouple loading from linking, as `SceKernelModulemgr` links when a module is *started*, not when it is *loaded*). This means touching lots of stuff in `kernel`, and since there are other areas of it I want to work on first (i.e. unified UID/object system) which would probably require rewriting at least a part of the loader+linker, so I'd rather postpone until `kernel` is "more stable".
1. `TODO: use raw I/O from path when io becomes less bad`: I hate the fact that load_self is given a path only to print it, but the `io` module is in a... questionable state, and needs a good refresh too imo. I previously discussed my vision for how it could be improved on Discord. This comment is mostly a reminder to make `io` suck less.
1. `TODO: is OSABI always 0?`, `TODO: is ABI_VERSION always 0?`, `TODO: this may be important and require being loaded` and `TODO: possible values for p_flags`: Needs some Modulemgr RE to check how the SCE loader reacts to those conditions. The latter *might* break C++ exceptions since that's what EXIDX is used for (along with EXTAB) but old Modulemgr straight up ignores the segment and newer one may, too; again, just a reminder there's a little RE to be done so we can polish the fine details.

Zangetsu38 (approved):

Macdu (approved):

.gitmodules
```diff
@@ -95,4 +92,4 @@
 	url = https://github.com/wolfpld/tracy.git
 [submodule "external/nativefiledialog-extended"]
 	path = external/nativefiledialog-extended
-	url = https://github.com/btzy/nativefiledialog-extended.git
+	url = https://github.com/btzy/nativefiledialog-extended.git
```
Zangetsu38, line 95: fail here :D

CreepNT: 💀

Zangetsu38 (approved):

PR #2048, opened 2022-08-31
external: remove rpcs3 bitfields

EXtremeExploit:
we were only using the rpcs3 for SceGxmProgramParameter, c++ has bitfield support, its better to use native c++ bitfield than include an external dependency just for a single bitfield when the other bitfields already use the native version of this. Needs some testing to make sure it is right


EXtremeExploit:
done, also changed uint16 to uint8 since the width of the number is 4 bits anyways

PR #2066, opened 2022-09-07
config: Add strict rendering mode.

Zangetsu38:
# About 
- config: Add strict rendering mode.
force using default value of improve render.

### DIFF
--- a/vita3k/config/include/config/config.h
+++ b/vita3k/config/include/config/config.h
@@ -60,6 +60,7 @@ enum PerfomanceOverleyPosition {
     code(bool, "archive-log", false, archive_log)                                                       \
     code(std::string, "backend-renderer", "OpenGL", backend_renderer)                                   \
     code(int, "gpu-idx", 0, gpu_idx)                                                                    \
+    code(bool, "strict-rendering", false, strict_rendering)                                             \
     code(int, "resolution-multiplier", 1, resolution_multiplier)                                        \
     code(bool, "disable-surface-sync", false, disable_surface_sync)                                     \
     code(bool, "enable-fxaa", false, enable_fxaa)                                                       \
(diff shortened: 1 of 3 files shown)


Macdu:
The name strict rendering doesn't make a lot of sense. I had to look at the files to understand what this was about.
And I'm not really sure if this is necessary as all the possible options can already be set in the gui.

Zangetsu38:
is based of rpcs3 option, is called Strict Rendering Mode
image

EXtremeExploit:
afaik strict rendering mode also does some other stuff to the renderer than just change some settings, maybe a look at https://github.com/RPCS3/rpcs3/pulls?q=strict+rendering can give a hint of what exactly is strict rendering on their end

Zangetsu38:
in majority, it just set scaling/aa/aniso by default, for testing default settin g and look if fix render issue, totaly obligatoryt in old rpcs3 for lot game

Macdu:
Maybe this is useful for rpcs3 but I really don't see the use for Vita3K. FXAA is applied to the swapchain/framebuffer image, it has absolutely no impact on the rendering (which is not the case for MSAA). And anisotropic filtering should also have no impact after a simple PR that I need to do.

Zangetsu38:
yes, I understand, but the goal here is precisely to be able to test the "default" mode with a simple click.

Basically, do you have a problem?
You activate the option and you see if it changes anything, it's just to test quickly, for having spent years testing rpcs3, it's super useful.

Macdu:
Yes, but RPCS3 does a bit more than just changing these settings. And the only setting that could lead to graphical glitches is upscaling. 
So it would be more like "Have you tried to disable upscaling?".

Zangetsu38:
yes i know, but in my rpcs3 test, is very majority testing without upscaling and aniso
other change for strict is very little

is uncharted issue have get me thinks this, and i have so just remember rpcs3 option

Zangetsu38:
deemed unnecessary

PR #2080, opened 2022-09-15
kernel: Rewrite the way svc and callbacks are handled

Macdu:
Rewrite the SVC calls to stop the CPU. This allows hle functions to call callbacks in the same thread without having to use a guest function or a job queue. This is needed in order to have multiple levels of callbacks (like allocating callbacked gxm memory inside a kernel callback).

This allows Trails in the sky games to go ingame.
image


Zangetsu38 (approved):

dabian8999:
Very good!Thank you!

PR #2087, opened 2022-09-25
renderer: Use rgba16 surfaces when rgba8 is requested when needed

Macdu:
Tearaway uses a feature of tiles renderer, which is that the format on the tiled renderer may be different compared to the format used to store the surface. The only way to emulate it accurately with a desktop GPU (immediate renderer) is to use a wider surface format.

As Tearaway is the only game that uses the renderer this way as far as I'm aware and that this causes an increase in memory usage, a decrease in performances and would break many games. I don't think adding it as a setting is a good idea and doing a check for Tearaway only is better.

This partially fixes the graphics in Tearaway.

### DIFF
--- a/vita3k/renderer/src/creation.cpp
+++ b/vita3k/renderer/src/creation.cpp
@@ -35,6 +35,34 @@
 
 namespace renderer {
 
+void State::game_start(const char *base_path, const char *title_id, const char *self_name) {
+    this->base_path = base_path;
+    this->title_id = title_id;
+    this->self_name = self_name;
+
+    static const std::string tearaway_games[] = {
+        "PCSC00048",
+        "PCSC00064",
+        "PCSF00214",
+        "PCSF00476",
+        "PCSF00463",
+        "PCSD00072",
+        "PCSD00077",
+        "PCSA00099",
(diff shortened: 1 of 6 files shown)


Zangetsu38 (approved):

vita3k/renderer/src/creation.cpp
```diff
@@ -35,6 +35,34 @@
 
 namespace renderer {
 
+void State::game_start(const char *base_path, const char *title_id, const char *self_name) {
+    this->base_path = base_path;
+    this->title_id = title_id;
+    this->self_name = self_name;
+
+    static const std::string tearaway_games[] = {
```
CoffeeBrewer64, line 43: Wouldn't it be better to store this in some kind of .xml file and then call it so that it can be edited and added on to easily? This would be helpful for anything like this in the future. idk just a thought

GamingDucking:
Any updates on this or it needs some reworking?

Kitakatarashima:
In my opinion, I don't know what's wrong, but in my opinion there should be an external option for this problem, there is an option to disable and enable it, if that causes many games to have problems when booting, that's what I think.

PR #2095, opened 2022-09-28
pkg: Fix file dialog not showing any file on work.bin

EXtremeExploit:


Zangetsu38:
You are sur ?  I have ’ever get problème before for find work.bin 🤔

EXtremeExploit:
yes, it only affects linux, windows should remain the same

Croden1999:
also we can found the .rif files.

Zangetsu38 (approved):

Issue #2044, opened 2022-08-28
The ability to re-set the keyboard controls doesn't work too well


LOuroboros:
Sup.

Now that Vita3K got a Vulkan renderer merged, I wanted to give things another test in case anything changed in regards to emulating Digimon World: Next 0rder compared to a few weeks back when I tested the emulator for the first time.

Back then I also noticed this, but it's getting annoying enough that I figured I should bring it up in an Issue ticket properly.

So, the emulator won't let me re-bind certain PS Vita buttons. That's the gist of it. I press my NumPad's 8 key to bind it to the emulated PSVita's "Right Stick Up" button, and it just doesn't really work. The emulator doesn't seem to accept that for some reason.

I recorded a quick video after deleting the controls config within the emulator's config.yml file to better show off what I mean.

I can modify the Left Stick keys just fine, but when I reach the Right Stick keys... I'm just not able to bind whatever key I want, though the same happens for the L3 Button which I was trying to set to the "T" key and the emulator was re-setting to the "F" key all on its own.

https://user-images.githubusercontent.com/4485172/187086188-0abb3c4b-a0d8-464c-b3fa-ff6608709dac.mp4

The funniest thing is that the emulator does let me assign the same keyboard key to 2 different PSVita buttons, as I can assign the 5 number to both the PSVita's Select button and its L3 button too, so I'm not sure I understand that "key is reserved!" warning I got in the command prompt.

![Vita3K_20220828_141412017](https://user-images.githubusercontent.com/4485172/187086395-d891f53e-6ad3-4944-8709-d6f3dc6cd267.png)

Maybe having the ability to clear all the set keys would be nice to have, so the user can just change every single button binding to their liking without all this hassle?

scribam:
The issue has been fixed by #2100

PR #2100, opened 2022-10-01
gui: fixed the keyboard remapping user experience

ngeojiajun:
This fix the issues https://github.com/Vita3K/Vita3K/issues/2044


nishinji:
I copied it for review and tried it, but it fails to build ubuntu
There is a problem with setting the size of the warning screen when the same key is typed
https://github.com/nishinji/Vita3K/actions/runs/3250004976

ngeojiajun:
noted, it was the string safety issues. will fit asap

ngeojiajun:
@nishinji Could you please clearify what problem was happened on the warning screen?

nishinji:
I think it would be better to fix the size of the warning dialog
The first time it popped up, it was very long and narrow.
In addition,  the close did not come in the middle.

ngeojiajun:
Ooo I see, will look at it whenever free

ngeojiajun:
Fixed the problem. In case you need the binary for test you can get it from here
https://github.com/ngeojiajun/Vita3K/actions/runs/3256321451#artifacts

vita3k/gui/src/controls_dialog.cpp
```diff
@@ -43,19 +43,68 @@ static char const *SDL_key_to_string[]{ "[unset]", "[unknown]", "[unknown]", "[u
...
+static bool exists_in_array(int *ptr, int val, size_t size) {
+    if (!ptr || !size) {
+        return false;
+    }
+    int *end = ptr + size;
+    return std::find(ptr, end, val) != end;
+}
+
+static void prepare_map_array(EmuEnvState &emuenv, std::array<int, total_key_entries> &map) {
+    map[0] = emuenv.cfg.keyboard_leftstick_up;
```
1whatleytay, line 57: Since this is ordered I would maybe expect this instead. Pretty sure std::array copy constructor is defined nicely here. Alternatively, why don't you return std::array here?
```cpp
std::array entries = {
  emuenv.cfg.keyboard_*,
  ...
}

map = entries;
```

vita3k/gui/src/imgui_impl_sdl_gl3.cpp
```diff
@@ -167,9 +167,10 @@ void ImGui_ImplSdlGL3_RenderDrawData(ImGui_GLState &state) {
             } else {
                 glBindTexture(GL_TEXTURE_2D, (GLuint)(intptr_t)pcmd->TextureId);
                 glScissor((int)pcmd->ClipRect.x, (int)(fb_height - pcmd->ClipRect.w), (int)(pcmd->ClipRect.z - pcmd->ClipRect.x), (int)(pcmd->ClipRect.w - pcmd->ClipRect.y));
-                glDrawElements(GL_TRIANGLES, (GLsizei)pcmd->ElemCount, sizeof(ImDrawIdx) == 2 ? GL_UNSIGNED_SHORT : GL_UNSIGNED_INT, idx_buffer_offset);
+                //fix the offset issues as per instructed in release notes of imgui v1.86
+                glDrawElements(GL_TRIANGLES, (GLsizei)pcmd->ElemCount, sizeof(ImDrawIdx) == 2 ? GL_UNSIGNED_SHORT : GL_UNSIGNED_INT, idx_buffer_offset+pcmd->IdxOffset);
             }
-            idx_buffer_offset += pcmd->ElemCount;
+            //idx_buffer_offset += pcmd->ElemCount;
```
1whatleytay, line 173: Drop comment?

1whatleytay (approved):

vita3k/gui/src/imgui_impl_sdl_gl3.cpp
ngeojiajun: Yeah, I should have do that. Forgot to remove it

PR #2117, opened 2022-10-05
renderer:Increase the size of surface sync by 4 times

sans-2233:
将surface sync的尺寸增加4倍，可以还原littlebigplanet的图像。
a


Croden1999:
this branch is based on Vita3K's branch, not ur branch, and also will break master. if u want submit ur issue, here's https://github.com/Vita3K/Vita3K/issues

Zangetsu38:
Wtf

Ristellise:
What the heck are you smoking. Use master branch, also what/where are you commits for this if you are submitting a PR?

*"But it works for me!"* I hear you say...

And inevitably break dozen of other games in the process so don't do this.

PR #2123, opened 2022-10-07
tracy: Transfer the old implementations to new

EXtremeExploit:
Transfer all the old hardcoded implementations of tracy into the new one made by bookmist, it should even fix some potential segfaults because before it was dereferencing pointers without checking if they were valid


vita3k/modules/include/modules/tracy.h
```diff
@@ -28,6 +29,14 @@ std::string to_debug_str(const MemState &mem, T data) {
...
 }
 
+// Override for guest pointers, we want to print the guest address
+template <typename U>
+std::string to_debug_str(const MemState &mem, Ptr<U> data) {
+    std::stringstream datass;
+    datass << data.address();
+    return datass.str();
+}
+
```
bookmist, line 39: also this (or something like) can be added to log addresses of pointers
`template <typename U>
std::string to_debug_str(const MemState &mem, U* data) {
    return log_hex(Ptr<U>(data, mem).address());
}`

bookmist:
Great change thank you very much

vita3k/modules/include/modules/tracy.h
EXtremeExploit: hmm should i use log_hex or log_hex_full? just saying because tracy will execute this pointer template on most exports because most of them have pointers in them, log_hex uses fmt::format and log_hex_full uses stringstreams, im saying this because of performance, idk if it even really matters :thinking:

EXtremeExploit: did a test and log_hex_full is double the efficient as log_hex, taking always half the time than log_hex, ig i will use log_hex_full

EXtremeExploit: added log_hex_fast, which is the same implementation as log_hex_full but without the padding, so in theory it should be on par or a tiny bit faster and should have same output as log_hex, i added a TODO to replace the fmt one to this faster one in case its needed, this is how it gets displayed on tracy:
image

bookmist: Thank you for checking performance, I did not know about log_hex_full
userAgent value on screenshot looks too long for 32 bit value, it seems wrong to me, maybe the code has an error, i'll check.
I forgot about char* It's needed a separate template specification for this. Something like 
template <> std::string to_debug_str(const MemState &mem, char* data) { return std::string(data ? data : "0x0 NULL"); }

bookmist: I see now. Pointers in Sce functions are usually guest pointers converted to host pointers. So for log I convert them back to guest pointers (Ptr<U>(data, mem).address()). I don't know what version is more usable (host or guest) but someone asks for guest pointer values

EXtremeExploit: Added a template for char*, printing both the address and the contents, For pointers i decided its more consistent to print the host address instead as its more easy to handle imo, unless there is some easy way to convert pointers back to guest i guess its just better to choose host in the meantime

bookmist: way to convert pointers back to guest is Ptr(host_pointer, mem).address()
But i think it's good enough now.

bookmist (approved):

Macdu:
In `tracy.h`, can you define TRACY_FUNC to be empty if TRACY_ENABLE is not defined. Right now if TRACY_ENABLE is not defined, Vita3K won't compile I believe (I might be wrong, I only looked at the code quickly).

EXtremeExploit:
> In `tracy.h`, can you define TRACY_FUNC to be empty if TRACY_ENABLE is not defined? Right now if TRACY_ENABLE is not defined, Vita3K won't compile I believe (I might be wrong, I only looked at the code quickly).
It is like that right now :+1: 
image

vita3k/modules/include/modules/tracy.h
```diff
@@ -28,6 +30,36 @@ std::string to_debug_str(const MemState &mem, T data) {
...
+    // extreme case that the string is actually "0x0 NULLPTR" and be confusing
+    return std::string(data ? log_hex_fast(data) + " " + data : "0x0 NULLPTR");
+}
+
+// Override for char pointers as the contents are readable
+template <typename U>
+std::string to_debug_str(const MemState &mem, const char *data) {
+    return std::string(data ? log_hex_fast(data) + " " + data : "0x0 NULLPTR");
+}
+
```
bookmist, line 62: I don't know what is better: unused <typename U> or inline

EXtremeExploit: inline, turns out the other way doesn't work and  the template gets ignored, now it should work as expected
image
image
image

bookmist (approved):

PR #2129, opened 2022-10-09
tracy: Use binary search when possible

EXtremeExploit:
as the comments say, if the user has some tracy modules enabled, ALL the modules that just support tracy will get some performance hit because of std::find, but if the list is ordered then we can use binary search so the time it takes to find if a specific module is enabled or not is considerably low, there aren't many tracy modules but in some near future there may be +50 and that may put some overhead, especially on kernel or gxm exports

### DIFF
--- a/vita3k/config/src/config.cpp
+++ b/vita3k/config/src/config.cpp
@@ -26,6 +26,7 @@
 #include <CLI11.hpp>
 #include <vector>
 
+#include <algorithm>
 #include <exception>
 #include <iostream>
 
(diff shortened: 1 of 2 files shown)


hobyst (approved):

Macdu:
In this case, I think you can use a std::set to contain the modules instead of an array.

EXtremeExploit:
i cant :c there's some template error on yaml-cpp when using sets

bookmist (approved):

PR #2131, opened 2022-10-11
tracy: Add sceLibc & sceSysModule

EXtremeExploit:


bookmist:
For completely unimplemented modules (like sceNgsInternal) primitive implementation is enough. I think, no need to markup such modules.

bookmist (approved):

PR #2132, opened 2022-10-11
[WIP] host/core: Introduce core OS interaction library

hobyst:
This library from the host abstraction layer is meant to contain useful functions and data structures that while might not really fit into a specific category are crucial for Vita3K to work.

Feedback about the implementation and ideas on what more things could go in such library are more than welcome.


Macdu:
Can you use `fs::path` instead of `std::filesystem::path` please? At least to remain coherent with the rest of the emulator.

hobyst:
> Can you use `fs::path` instead of `std::filesystem::path` please? At least to remain coherent with the rest of the emulator.

Isn't the goal in the long term to move away from `boost::filesystem` to `std::filesystem`? I'm already using `std::filesystem` in everything I write so that no modification is needed afterwards.

Macdu:
The goal is to move away from `std::filesystem` but we can still use `fs::path` as an alias for the time being and transition everything (or not and keep the alias) when needed. 
By doing this you add some unnecessary `fs::path path = fs::path(std_path.wstring())` in the code. I also had to rewrite your file dialog part to use `fs` instead of `std::filesystem` in the Android to match the rest of the code because of some Android quirks.

hobyst:
> The goal is to move away from `std::filesystem` but we can still use `fs::path` as an alias for the time being and transition everything (or not and keep the alias) when needed. By doing this you add some unnecessary `fs::path path = fs::path(std_path.wstring())` in the code. I also had to rewrite your file dialog part to use `fs` instead of `std::filesystem` in the Android to match the rest of the code because of some Android quirks.

Got it. I'll change all references to use `fs::path`. Why the Boost implementation is preferred over the one from the C++ standard library?

Macdu:
Don't really know, I made the pr #1912 to switch to c++ filesystem and I think it was almost ready to be merged but it got kind of delayed of forgotten. But anyway, I think it's better to either only have std::filesystem or fs::path everywhere at least for consistency.

hobyst:
> Don't really know, I made the pr #1912 to switch to c++ filesystem and I think it was almost ready to be merged but it got kind of delayed of forgotten. But anyway, I think it's better to either only have std::filesystem or fs::path everywhere at least for consistency.

Then wouldn't it be better for the time being to just use `std::filesystem` on new code so that less code has to be changed for the transition?

PR #2138, opened 2022-10-12
cmake: Only copy update script if updater is enabled

EXtremeExploit:


Macdu:
You can also only copy the bat script on Windows and the sh script on linux while we are at it.

Macdu (approved):

PR #2157, opened 2022-10-21
renderer/vulkan/surface_cache: fix crash

bookmist:
Fix crash in Ar Nosurge using vulkan renderer


Zangetsu38 (approved):

Macdu:
Yes, this works (I think I did something similar for another part). However, it is inefficient and is more like trying to fix the consequences than the issue itself (this is caused by the set_texture call in `sceGxmSetFragmentTexture`). I'm going to do a PR that makes this fix no longer needed.

bookmist:
Great. Thank you very much for going to Real fix this issue. (This PR is a workaround and I've known it)

PR #2159, opened 2022-10-22
modules: Add Tracy to all implemented modules. No enums.

bookmist:
Instrument almost all implemented and many unimplemented functions. 
Implemented and non-instrumented: 
SceAvPlayer.cpp - now HLE implementation is not used, lle implementation is used
Non-implemented and non-instrumented (due almost all functions in this modules are unimplemented):
SceFios2Kernel.cpp
SceFios2Kernel02.cpp
SceLibm.cpp
SceLibstdcxx.cpp
ScePafStdc.cpp

module::vargs function parameters are not send to tracy

Also: 
Add TRACY_MODULE_NAME macro to shorten text and make more easy to change tracy_module_name variable name in future (if needed)
drop unused hack from SceMp4 (lle implementation is used)
mark unimplemented ScePgf/sceFontNewLib (no idea why unimplemented macro was not used here)

Small optimization in macros TRACY_FUNC_COMMANDS TRACY_FUNC_COMMANDS_SET_STATE (if function get std::string and got char* then std::string constructor is called on every function call)


Macdu:
Wouldn't it be possible to redefine the EXPORT macro in the files with a Tracy implementation to make it call TRACY_FUNC?

bookmist:
I tryed to redefine export, but no success. { after export made it impossible for me.

Macdu:
Ah indeed

EXtremeExploit:
some functions for http, ssl (and i can't remember if theres another one) are missing tracy impl, would be nice to add them

EXtremeExploit:
i've seen some modules now, some functions in the same module have tracy while others not, an example would be in http,ssl and `sceAudiodecGetInternalError` specifically, same goes for the the c++ std libs too, i will keep adding the ones that are complete to the tracker

bookmist:
I don't care about UNIMPLEMENTED() functions. Also not included tracy in implemented functions in 
SceAvPlayer.cpp - now HLE implementation is not used, lle implementation is used
SceDisplayUser.cpp - it just call SceDisplay

Zangetsu38:
Need correct rebase, using merge master is ugly and no can see réal change

bookmist:
I've finished this PR. Also I put description in the first comment.

EXtremeExploit (approved):
cool job bookmist

PR #2160, opened 2022-10-22
renderer: Add tracy support to all commands

EXtremeExploit:
had to add a new define specifically for commands, hope bookmists approves it and macdu the commands part

### DIFF
--- a/vita3k/renderer/CMakeLists.txt
+++ b/vita3k/renderer/CMakeLists.txt
@@ -8,7 +8,7 @@ add_library(
 	src/gl/fence.cpp
 	src/gl/renderer.cpp
 	src/gl/ring_buffer.cpp
-    src/gl/screen_render.cpp
+	src/gl/screen_render.cpp
 	src/gl/surface_cache.cpp
 	src/gl/sync_state.cpp
 	src/gl/texture_formats.cpp
--- a/vita3k/util/include/util/tracy.h
+++ b/vita3k/util/include/util/tracy.h
@@ -17,6 +17,7 @@
 
 #pragma once
 
+#include <config/functions.h> // for is_tracy_advanced_profiling_active_for_module
 #include <mem/ptr.h>
 #include <mem/state.h>
 #include <sstream>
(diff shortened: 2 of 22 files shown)


vita3k/renderer/CMakeLists.txt
```diff
@@ -42,7 +42,7 @@ add_library(
 )
 
 target_include_directories(renderer PUBLIC include)
-target_link_libraries(renderer PUBLIC crypto display dlmalloc mem stb shader glutil threads config util vkutil)
+target_link_libraries(renderer PUBLIC crypto display dlmalloc mem modules stb shader glutil threads config util vkutil)
```
Macdu, line 45: Sorry, but we really shouldn't have to make renderer dependant on modules. Maybe put the tracy files somewhere else like util ?

EXtremeExploit: done

bookmist:
I think, separate tracy.h in separate projects are better. It look like there is tracy_command don't use anything from modules/tracy.h

vita3k/util/include/util/tracy.h
```diff
@@ -110,8 +123,12 @@ inline std::string to_debug_str(const MemState &mem) {
 #else
 #define TRACY_FUNC(name, ...) __TRACY_FUNC(tracy_module_name, name, ##__VA_ARGS__, , , , , , , , )
 #define TRACY_FUNC_M(module_name_var, name, ...) __TRACY_FUNC(module_name_var, name, ##__VA_ARGS__, , , , , , , , )
+#define TRACY_FUNC_COMMANDS(name) __TRACY_FUNC_COMMANDS(name)
+#define TRACY_FUNC_COMMANDS_SET_STATE(name) __TRACY_FUNC_COMMANDS_SET_STATE(name)
```
bookmist, line 127: There is no __VA_ARGS__ in TRACY_FUNC_COMMANDS, so workaround for __VA_ARGS__ is not needed (and also TRACY_FUNC_COMMANDS... should be defined in both branches of #if)

EXtremeExploit: does last push fix it?

bookmist: yes.

bookmist (approved):

PR #2166, opened 2022-10-27
[WIP] vita3K: Support building on AArch64(ARM64)

KhoraLee:
Latest dynarmic now supports AArch64 backend. So AArch64 Mac can now build vita3K natively.

Still need some modification and #2038 should merge first.

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -1,5 +1,9 @@
 include(${CMAKE_ROOT}/Modules/ExternalProject.cmake)
 
+# Arch detection using dynamics CMake module
+list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/external/dynarmic/CMakeModules")
+include(DetectArchitecture)
+
 # Make option() in all external projects do nothing when a normal variable of the
 # same name exists. The normal variable is not removed. The cache entry is not
 # created or updated and is ignored if it exists.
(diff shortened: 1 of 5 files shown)


Macdu:
This would need a lot more work, all the prebuilt librairies (mainly sdl, libtomcrypt and ffmpeg) would need to use aarch64 prebuilt versions. I had to do this for the android version.

And right now I think the aarch64 build would actually end up slower. Even though the aarch64 branch of dynarmic is merged, it is still inferior to the x86_64 version (no fastmem and many parts are not as well optimized).

KhoraLee:
I already made it to build on my local m1 pro mac and saw the very low performance. But I was thinking that it's better to have aarch64 option for future development.

vita3k/modules/SceGxm/SceGxm.cpp
```diff
@@ -24,7 +24,9 @@
 #include <gxm/functions.h>
 #include <gxm/state.h>
 #include <gxm/types.h>
+#ifndef __aarch64__
 #include <immintrin.h>
+#endif
```
bookmist, line 29: As far as I see, it's possible to remove this header completely

Macdu:
By the way, a lot of what you're doing is repetitive with what I have already done for the android port (like implementing the signal handler in aarch64). If you're interested I have a NEON accelerated f32 to f16 code to replace the AVX one used by x86_64.

KhoraLee:
Because you've already done what I was trying to do, I will close this and wait your works being PRed.

external/CMakeLists.txt
```diff
@@ -208,7 +214,11 @@ if(USE_DISCORD_RICH_PRESENCE)
 		${CMAKE_BINARY_DIR}/external/discord_game_sdk/cpp/voice_manager.cpp)
 
 	if(APPLE)
-		target_link_libraries(discord-rpc PUBLIC "${CMAKE_BINARY_DIR}/external/discord_game_sdk/lib/x86_64/discord_game_sdk.dylib")
+		if(ARCHITECTURE STREQUAL "arm64")
+			target_link_libraries(discord-rpc PUBLIC "${CMAKE_BINARY_DIR}/external/discord_game_sdk/lib/x86_64/discord_game_sdk.dylib")
```
cobalt2727, line 218: I realize this is obsolete and will probably never get merged anyways, but it looks like x86_64 and aarch64 are reversed here

KhoraLee: I already aware of it and gonna make PR for arm(aarch64) macOS after macdu PRs aarch64 things

Macdu: I'll try to make a PR with my aarch64 modifications next week, sorry for the delay.

PR #2167, opened 2022-10-28
gui: Make pref path setting manually editable

EXtremeExploit:
# Linux systems
image

# Windows systems
image

* may need some testing
* The field is big on non windows systems because it has the case sensitive option and it makes the dialog a lot wider


Zangetsu38:
is bad idea here, i have removed this for good reason, for using only file explorer, for no set bad path with edit, no see interest here 🤔

PR #2172, opened 2022-10-29
shader: Fix UB on ext_vec_predicate_to_ext

EXtremeExploit:
very IMO, its far better to return NONE instead of just letting it in the hands of undefined behaviour, as reaching function end without returning anything is actually undefined behaviour and ANYTHING can happen, it will still trigger the assert if it happens as theres no break

### DIFF
--- a/vita3k/shader/include/shader/usse_types.h
+++ b/vita3k/shader/include/shader/usse_types.h
@@ -17,6 +17,7 @@
 
 #pragma once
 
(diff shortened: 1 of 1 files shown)


Macdu:
Would be even better to add a LOG_ERROR or a LOG_CRITICAL I think.

bookmist (approved):

PR #2183, opened 2022-11-07
Fix compiler warnings

bookmist:
Fix a lot of compiler (and other analyze tool) warnings:
clang: reordered constructor
clang: loop variable creates a copy, use reference type to prevent copying
unused variables. not all. There is a lot of unused variables.
string->pchar->string conversion
change single character string to character (find(var,"*") -> find(var,'*') )
drop already checked checks and simplify logical expressions. like if(!x) return; if(_!x &&_ expression) {code};
some type changes to avoid type conversion. like uint x = -1;
set type to unsigned int to GXP enums when its values are set.
change empty constructor/destructor to default
inline constants in headers
fix format strings
function not returning value in all cases. (usually there is assert(false) in this case, so I return something after assert)
reorder class fields to more compact classes


Macdu:
Do you think that writing `std::uint32_t` is necessary? From my point of view, `uint32_t` has absolutely nothing to do with the C++ standard library (it's defined in <stdint.h>) and doing so adds some useless `std::` everywhere.

bookmist:
> Do you think that writing `std::uint32_t` is necessary?
No. I just don't know usual c++ way to declare unsigned 32 bit integer. It's great if just "unsigned int" is normal.
And thank you!

Macdu:
Not 'unsigned int' (it technically can be something other than 32 bits although it is never the case). Just use 'uint32_t' without the std:: at the beginning

Macdu (approved):

Macdu:
Really nice PR, this also removes many unnecessary copies in for loops.

PR #2188, opened 2022-11-09
glutil: fix mvsc error

nn9dev:
I set up Vita3K to build on MSVC 2022 entirely fresh, and it doesn't build without this include.
It doesn't fix it if I add it in the .cpp file, only the header. Either way, I could be doing something wrong if nobody has encountered this issue before.

### DIFF
--- a/vita3k/glutil/include/glutil/shader.h
+++ b/vita3k/glutil/include/glutil/shader.h
@@ -18,6 +18,7 @@
 #pragma once
 
 #include <glutil/object.h>
+#include <string>
 
 namespace gl {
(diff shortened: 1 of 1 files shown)


Zangetsu38:
I totaly no have this issue

nn9dev:
It doesn't seem like anyone else does, but this is my screenshot. I don't know how, I don't know why, but it does. again, perhaps I'm doing something wrong, but all I can tell you is that Vita3K doesn't want to build without it.
image

Zangetsu38:
@Macdu you can review it please ?

Macdu:
I also get this error after updating to the latest version of visual studio. Thank you for the contribution!

Macdu (approved):

PR #2198, opened 2022-11-20
modules/SceKernelForMono: Reduce cpu consumption and fix strange crash in (STAY)

bookmist:
Real infinite sleep is OS specific (as far as I know), so I just use some big value.

### DIFF
--- a/vita3k/modules/SceLibKernel/SceKernelForMono.cpp
+++ b/vita3k/modules/SceLibKernel/SceKernelForMono.cpp
@@ -17,7 +17,7 @@
 
 #include "SceKernelForMono.h"
 #include "../SceKernelThreadMgr/SceThreadmgr.h"
-
+#include <kernel/state.h>
 #include <util/tracy.h>
 TRACY_MODULE_NAME(SceKernelForMono);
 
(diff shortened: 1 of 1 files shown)


Macdu:
I think implementing the correct behavior of this function instead should not be too hard (a call to `thread->update_status(ThreadStatus::suspend);` should be all that is needed I believe).

bookmist:
Fixed. Not a really correct implementation in fact, but anything is better then infinite loop.

vita3k/modules/SceLibKernel/SceKernelForMono.cpp
```diff
@@ -49,9 +49,8 @@ EXPORT(int, sceKernelWaitExceptionCBForMono) {
 EXPORT(int, sceKernelWaitExceptionForMono) {
     TRACY_FUNC(sceKernelWaitExceptionForMono);
     STUBBED("Inifinite wait");
```
Macdu, line 51: I think this stub can be modified now

bookmist: It's still infinite wait, just by other way.

Macdu: Ah indeed you're right.

PR #2204, opened 2022-11-24
Theoretical improvements

bookmist:
Here I collect all improvements which do not fix any game. It make emulation more accurate, implement unimplemented and so on.
* audio/audio: fix theoretically possible sound freeze
* io/io: workaround of exception on file delete
* kernel/sync_primitives: set timeout if needed and minor event improvement
* vita3k/manifest: fix xml to build with msvc-clang
* modules/SceFiber: functions return errors instead of crash
* modules/SceTouch: Stubbed implementation of sceTouchGetPixelDensity
* modules/SceHttp: improved hard to understand check in sceHttpUriParse

### DIFF
--- a/vita3k/kernel/src/sync_primitives.cpp
+++ b/vita3k/kernel/src/sync_primitives.cpp
@@ -83,7 +83,7 @@ inline int handle_timeout(const ThreadStatePtr &thread, std::unique_lock<std::mu
     const char *export_name, SceUInt *const timeout) {
     if (timeout) {
         bool status = false;
-
+        auto start = std::chrono::steady_clock::now();
         if (*timeout > 0) {
             status = thread->status_cond.wait_for(primitive_lock, std::chrono::microseconds{ *timeout }, [&] { return thread->status == ThreadStatus::run; });
         }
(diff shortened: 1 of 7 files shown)


vita3k/kernel/src/sync_primitives.cpp
```diff
@@ -98,6 +98,14 @@ inline int handle_timeout(const ThreadStatePtr &thread, std::unique_lock<std::mu
             queue->erase(data_it);
 
             return RET_ERROR(SCE_KERNEL_ERROR_WAIT_TIMEOUT);
+        } else {
+            auto end = std::chrono::steady_clock::now();
+            uint32_t real_timeout = (end - start) / std::chrono::microseconds(1);
```
Macdu, line 103:
```suggestion
            uint32_t real_timeout = static_cast<uint32_t>(std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count());
```

bookmist: Thank you. I changed duration in your suggestion to microseconds

Macdu (approved):

PR #2205, opened 2022-11-24
shader: Fix vtstmsk and or/xor, resize viewport/scissor when the image is downscaled

Macdu:
Fix some issues with vtstmsk (the test type does not determine the dest type) and or/xor with a 0 which stored the result as a float instead of an integer.
Also, when a game downscales a surface without using msaa (why would you do that??), the viewport and scissor must have their dimensions divided by 2 as we are directly rendering to the downscaled surface.

This fixes most of the graphical issues of ys memories of Celceta.

image

### DIFF
--- a/vita3k/renderer/include/renderer/types.h
+++ b/vita3k/renderer/include/renderer/types.h
@@ -215,6 +215,7 @@ struct ShadersHash {
 
 struct RenderTarget {
     int holder;
+    uint16_t multisample_mode;
     virtual ~RenderTarget() = default;
 };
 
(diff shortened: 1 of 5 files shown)


EXtremeExploit (approved):

bookmist:
"or 0" was usually used as something like this
_295.z = uintBitsToFloat(floatBitsToUint(sa[21].x) | 0u);
(opengl code from random shader before "or 0" optimization)
so I change data type to float to avoid double conversion. Sorry if it makes bugs.

Zangetsu38 (approved):

PR #2212, opened 2022-12-03
modules/SceGxm: fix rare case when notification is created already notified

bookmist:
Fix crash in Batman™: Arkham Origins Blackgate [PCSB00353]


Macdu:
I find it kind of weird to arbitrarily change the value of the notification to NOT it when it is already set, do you think your fix is accurate enough?

bookmist:
I'm completely not sure. I don't change value field of notification, only change data which address field pointer reference on, before it is used in renderer. Without this change game crash almost immediately. I checked some games, and only in Batman such situation happens (Notification object create already notified). I use NOT because ~value always not equal to value, maybe zero is also a good value (For Batman zero is ok)

PR #2226, opened 2022-12-14
CMake: disable unknown compiler flags for sccache on windows on github ci

bookmist:
Ready for review and merge. It fix a problem with sccache in github ci. Not completely fix, there are still many non-cacheable calls (558 cacheable 148 non-cacheable)


Macdu:
According to the log, it is not working.
The reason being that the windows ci does not use ccache (and CCACHE_PROGRAM is not defined) but instead uses sccache that is passed though CMAKE_CXX_COMPILER as a flag.

bookmist:
@Macdu It seems, now it's fixed.

Macdu (approved):

PR #2232, opened 2022-12-20
renderer/vulkan/pipeline_cache: fix regression on commit 2779298

bookmist:
Fix regression an AMD GPU from commit 2779298d70616b2f5d89c7cb9813d2d44df77fbc
Now it checks for each possibly unsupported format separately and use workaround only if needed. It fix regression for PCSE00707 - Ar nosurge... and fix crash on PCSB00102 - Chronovolt

### DIFF
--- a/vita3k/renderer/include/renderer/vulkan/pipeline_cache.h
+++ b/vita3k/renderer/include/renderer/vulkan/pipeline_cache.h
@@ -20,6 +20,7 @@
 #include <array>
 #include <limits>
 #include <map>
+#include <set>
 #include <unordered_map>
 
 #include <vkutil/objects.h>
@@ -42,7 +43,7 @@ private:
 
     // does the GPU support vertex attributes with 3 components (like R16G16B16_UNORM), some (like AMD GPUs) don't
     // this is needed when creating the input state
-    bool support_rgb_vertex_attribute;
+    std::set<vk::Format> unsupported_rgb_vertex_attribute_formats{};
 
     // how much time should we wait after the last shader compilation
     // to update the disk-saved shader cache (in seconds)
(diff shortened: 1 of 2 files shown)


vita3k/renderer/include/renderer/vulkan/pipeline_cache.h
```diff
@@ -42,7 +43,7 @@ class PipelineCache {
 
     // does the GPU support vertex attributes with 3 components (like R16G16B16_UNORM), some (like AMD GPUs) don't
     // this is needed when creating the input state
-    bool support_rgb_vertex_attribute;
+    std::unordered_set<vk::Format> unsupported_rgb_vertex_attribute_formats{};
```
Macdu, line 46: A set should be faster than an unordered_set when it contains only a few values I believe.

Macdu:
By the way, what was causing the regression in the previous commit? Setting support_rgb_vertex_attribute to false, even if the GPU supports it, is not supposed to cause graphical issues.

bookmist:
If I change component_count from 3 to 4 in all cases, then I got 
изображение
in PCSE00707 - Ar nosurge...
No idea, why. This game have some more graphic problems, but they are not so big and ugly.

vita3k/renderer/include/renderer/vulkan/pipeline_cache.h
bookmist: I can change it to set of bit flags in one byte, It will be fastest way, but code may be hard to understand and contains a lot of с style casts

Macdu: I was more thinking of using a std::set instead of a std::unordered_set. Using a set of bit flags would indeed be faster but I don't think this is necessary for the added complexity (also in this case using an std::bitset would remove the c casts).

bookmist: I fix set, Is there needed some more changes?

Macdu (approved):

PR #2233, opened 2022-12-20
vita3k updater: fix the auto updater on Windows

NabsiYa:
As the title says it fixes the auto updater on Windows by starting the batch file on it's own process.


EXtremeExploit (approved):
sure thing. looks good enough

PR #2235, opened 2022-12-21
Fix ccache for MacOS

NabsiYa:
This should be the fix for ccache on MacOS CI builds it seemed to work when I tested it.

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -36,7 +36,7 @@ jobs:
             cmake_preset: windows-ninja
           - os: macos-latest
             cache_path: ~/Library/Caches/ccache
-            extra_cmake_args: 
+            extra_cmake_args: -DCMAKE_C_COMPILER_LAUNCHER=sccache -DCMAKE_CXX_COMPILER_LAUNCHER=sccache
             cmake_preset: macos-ninja
 
     steps:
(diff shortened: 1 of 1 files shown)


NabsiYa:
Didn't solve the problem.

Macdu:
From the ccache stat at the end of the MacOS build, it actually looks like it worked

NabsiYa:
Interesting... I may then reopen this pr.

.github/workflows/c-cpp.yml
```diff
@@ -70,6 +74,7 @@ jobs:
           key: cache-${{ matrix.os }}-${{ matrix.config }}-${{ github.sha }}
           restore-keys: |
             cache-${{ matrix.os }}-${{ matrix.config }}-
+        if: matrix.os == 'ubuntu-latest'
```
Macdu, line 77:
```suggestion
        if: matrix.os != 'macos-latest'
```

The Windows CI also uses the cache action for the vcpkg build cache.

NabsiYa:
Looks good to me.

NabsiYa:
I probably should put everything into one commit?

.github/workflows/c-cpp.yml
Macdu: You modified the wrong line.

Macdu:
Not necessary, I can "squash and merge" to only have one commit in the end.

NabsiYa:
Ok, now it should be fine... Let's wait till the CI is done.

PR #2236, opened 2022-12-22
Improve the Windows update script yet again.

NabsiYa:
When I last time I wrote it it was 3 am so there was bound to be some mistake it worked fine on my local machine but now it doesn't work at all...

### DIFF
--- a/vita3k/gui/src/vita3k_update.cpp
+++ b/vita3k/gui/src/vita3k_update.cpp
@@ -190,7 +190,7 @@ static void download_update() {
     if (!fs::exists(path)) {
         std::thread download([path]() {
 #ifdef WIN32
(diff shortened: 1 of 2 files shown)


cascardian:
Hi!

I don't know if this has been discussed before, but maybe the script could be improved further by hiding the download progress bar. [This bug](https://github.com/PowerShell/PowerShell/issues/2138) in PowerShell 5.1, which is still the standard version shipped with Windows, means that download speed is massively slowed down when the progress bar is shown.

So `ProgressPreference` to would have to be set to `SilentlyContinue`. This could perhaps be done by adding `Set-Variable ProgressPreference SilentlyContinue ; ` before `Invoke-WebRequest`, at least for downloading the zip.

NabsiYa:
I did some testing and it indeed was marginally faster, I will it include this in this PR.

EXtremeExploit (approved):
k

PR #2242, opened 2022-12-23
gui: Use SDL_OpenURL in search

nishinji:
Macdu's advice helped me fix it.
Japanese and other languages(like chinese) can be handled.

### DIFF
--- a/vita3k/gui/src/app_context_menu.cpp
+++ b/vita3k/gui/src/app_context_menu.cpp
@@ -24,6 +24,7 @@
 #include <util/log.h>
 #include <util/safe_time.h>
 
+#include <SDL_misc.h>
 #include <pugixml.hpp>
 
 namespace gui {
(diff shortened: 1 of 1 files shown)


Zangetsu38:
Thx, i have never things how fix this

Zangetsu38:
@nishinji already yiu need run format, probably order for header added

nishinji:
@Zangetsu38 fixed

Macdu:
Wouldn't it be better for open_path to use SDL_OpenUrl if the path starts with http ?

nishinji:
Is it preferable to replace open_path with sdl_openurl or rewrite the contents of open_path?

Macdu:
I would say rewrite the content of open_path so that if it starts with http, use SDL_OpenUrl and otherwise use the previous behavior.

vita3k/gui/src/app_context_menu.cpp
```diff
@@ -256,17 +257,22 @@ void delete_app(GuiState &gui, EmuEnvState &emuenv, const std::string &app_path)
 }
 
 void open_path(const std::string &path) {
+    std::string s = "http";
+    if (path.size() >= s.size() && std::equal(std::begin(s), std::end(s), std::begin(path))) {
```
Zangetsu38, line 261: no understund first compare here, path is obligatory more size http, no have sense

Macdu, line 261:
```suggestion
    if (path.starts_with("http")) {
```

nishinji:
Fixed.
Sorry for the trouble (I'm still a newbie).
I didn't use starts_with because the document I referred to said it was only available for C++20 or later.

Zangetsu38 (approved):

Macdu:
No problem. Vita3K has C++20 as a requirement.

PR #2253, opened 2022-12-28
net: Fix return value of sceNetSend/sceNetRecv

edwinr:
Functions `sceNetSend` and `sceNetRecv` should return the number of bytes sent/received. Currently they always return `0` on success.

This PR changes `translate_errorcode` to fix this. It also changes the name to `translate_return_value` to more clearly state the intention (since not all return values are errors). Error handling for other socket functions such as bind, listen, connect etc. should not be affected because they only return either `0` or `-1`.

This vitasdk sample can be used as a test case for this change: https://github.com/vitasdk/samples/tree/master/net_http_bsd
Before: it hangs forever
After: it works.

### DIFF
--- a/vita3k/net/src/posixsocket.cpp
+++ b/vita3k/net/src/posixsocket.cpp
@@ -29,8 +29,8 @@
         return SCE_NET_ERROR_##errname;
 #endif
 
-static int translate_errorcode(int error) {
-    if (error < 0) {
+static int translate_return_value(int retval) {
+    if (retval < 0) {
 #ifdef WIN32
         switch (WSAGetLastError()) {
 #else
@@ -115,7 +115,9 @@ static int translate_errorcode(int error) {
         }
         return SCE_NET_ERROR_EINTERNAL;
     }
-    return 0;
+
+    // zero and positive values do not need to be translated, as they indicate success or bytes sent/received
+    return retval;
 }
 
 static void convertSceSockaddrToPosix(const struct SceNetSockaddr *src, struct sockaddr *dst) {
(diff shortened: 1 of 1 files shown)


vita3k/net/src/posixsocket.cpp
```diff
@@ -143,17 +145,17 @@ static void convertPosixSockaddrToSce(struct sockaddr *src, struct SceNetSockadd
...
 int PosixSocket::bind(const SceNetSockaddr *addr, unsigned int addrlen) {
     struct sockaddr addr2;
     convertSceSockaddrToPosix(addr, &addr2);
-    return translate_errorcode(::bind(sock, &addr2, sizeof(struct sockaddr_in)));
+    return translate_return_value(::bind(sock, &addr2, sizeof(struct sockaddr_in)));
 }
 
 int PosixSocket::listen(int backlog) {
-    return translate_errorcode(::listen(sock, backlog) < 0);
+    return translate_return_value(::listen(sock, backlog) < 0);
```
EXtremeExploit, line 158: i don't think the `< 0` is needed anymore here, now it would already be returning the number of bytes it listened for right?

Macdu: The listen function returns only 0 or -1 if there is an error. Indeed the <0 should now be removed as this function will now return 1 if there is an error.

edwinr: Agreed, actually I think it already should not have been there. Hmm, does `< 0` result in a negative value when cast to an `int`? otherwise it may have been returning success (`0`) for failing calls to `listen`.

edwinr: But listen does not return a number of bytes, it just starts waiting for incoming connections. Its only return values should be `0` or `-1`.

edwinr: @EXtremeExploit I changed it. I don't have a test case for this one but I think it should be ok.

EXtremeExploit (approved):
lgtm

Macdu (approved):

PR #2258, opened 2023-01-02
modules/module_parent: More correct LLE import path. Fix "Import function for NID 0xE340* not found"

bookmist:
I was very curious why "Import function for NID 0xE340* not found" appears in log so often and completely disappear when I comment out lle call stub. Now I found why and how to fix it.
Jit cache invalidate is intentional.


frangarcj (approved):
You are right, pc is at the nid but the three instructions are from the previous pc

PR #2259, opened 2023-01-02
renderer/vulkan: Clear the bound textures at the start of each renderpass

Macdu:
Fix an issue where a texture could still stay bound in the next scene if it was unused, but a texture bound to a slot with a higher index was used. The bound texture could then be invalid which caused the vulkan driver to crash (even though the texture itself was not used).

This fixes the crashes in Uncharted.


EXtremeExploit (approved):
ok macdu-sama

PR #2260, opened 2023-01-03
renderer/vulkan: Restride vertex streams to be aligned by 4 bytes on macOS

KhoraLee:
Metal requires strides to be multiple of 4 bytes.
This will extend stream's data to make stride multiple of 4 bytes.

### DIFF
--- a/vita3k/renderer/src/vulkan/pipeline_cache.cpp
+++ b/vita3k/renderer/src/vulkan/pipeline_cache.cpp
@@ -27,8 +27,9 @@
 #include <gxm/types.h>
 #include <renderer/shaders.h>
 #include <shader/spirv_recompiler.h>
-#include <util/fs.h>
 
+#include <util/align.h>
+#include <util/fs.h>
 #include <util/log.h>
 
 namespace renderer::vulkan {
@@ -478,9 +479,14 @@ vk::PipelineVertexInputStateCreateInfo PipelineCache::get_vertex_input_state(Mem
 
         const bool is_instanced = gxm::is_stream_instancing(static_cast<SceGxmIndexSource>(stream.indexSource));
 
+#ifdef __APPLE__
+        const uint32_t stride = align(stream.stride, 4);
+#else
+        const uint32_t stride = stream.stride;
+#endif
         binding_descr.push_back(vk::VertexInputBindingDescription{
             .binding = stream_index,
-            .stride = stream.stride,
+            .stride = stride,
             .inputRate = is_instanced ? vk::VertexInputRate::eInstance : vk::VertexInputRate::eVertex });
     }
 
--- a/vita3k/renderer/src/vulkan/scene.cpp
+++ b/vita3k/renderer/src/vulkan/scene.cpp
@@ -22,6 +22,8 @@
 
 #include <config/state.h>
 #include <spdlog/fmt/bin_to_hex.h>
+
+#include <util/align.h>
 #include <util/log.h>
 
 namespace renderer::vulkan {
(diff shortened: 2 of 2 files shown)


vita3k/renderer/src/vulkan/pipeline_cache.cpp
```diff
@@ -478,9 +478,14 @@ vk::PipelineVertexInputStateCreateInfo PipelineCache::get_vertex_input_state(Mem
 
         const bool is_instanced = gxm::is_stream_instancing(static_cast<SceGxmIndexSource>(stream.indexSource));
 
+#ifdef __APPLE__
+        const uint32_t stride = (stream.stride + 3) & -4;
```
Macdu, line 482:
```suggestion
        const uint32_t stride = align(stream.stride, 4);
```
You may need to include `util/align.h`

vita3k/renderer/src/vulkan/scene.cpp
```diff
@@ -99,6 +99,18 @@ void update_sync_signal(SceGxmSyncObject *sync) {
     }
 }
 
+#ifdef __APPLE__
+void restride_stream(GXMStreamInfo &stream, uint32_t old_stride) {
+    const uint32_t new_stride = (old_stride + 3) & -4;
```
Macdu, line 104:
```suggestion
    const uint32_t new_stride = align(old_stride, 4);
```

vita3k/renderer/src/vulkan/scene.cpp
```diff
@@ -99,6 +99,18 @@ void update_sync_signal(SceGxmSyncObject *sync) {
...
+    stream.size = (stream.size / old_stride) * new_stride;
```
Macdu, line 105: I believe it should be
```suggestion
    stream.size = ((stream.size + old_stride- 1) / old_stride) * new_stride;
```

Moreover, I believe it will be cleaner to compute `nb_vertex_input =  ((stream.size + old_stride- 1) / old_stride)` and then use it for the rest of this function.

vita3k/renderer/src/vulkan/scene.cpp
```diff
@@ -99,6 +99,18 @@ void update_sync_signal(SceGxmSyncObject *sync) {
...
+    uint8_t *new_data = new uint8_t[stream.size]();
```
Macdu, line 106:
```suggestion
    uint8_t *new_data = new uint8_t[stream.size];
```

No need to initialize new_data

vita3k/renderer/src/vulkan/scene.cpp
KhoraLee: I think original code is correct.

As far as my understand, data structrue looks like below.
```
Original
┌────────────┬────────────┐
│    Data    │    Data    │
└────────────┴────────────┘
├───Stride───┼───Stride───┤
├───────stream.size───────┤

-------------------------------------------

After restriding
┌────────────────────┬────────────────────┐
│    Data    │Padding│    Data    │Padding│
└────────────────────┴────────────────────┘
├───────Stride───────┼───────Stride───────┤
├───────────────stream.size───────────────┤
```
So dividing total data size with old_stride will give number of data counts.
And multipling aligned stride to data counts will create new data array size.

Macdu: The stream.size value is computed here https://github.com/Vita3K/Vita3K/blob/cc7f2847dcb3a7878d0b04c75b04345bea03f2cb/vita3k/modules/SceGxm/SceGxm.cpp#L2093 , as you can see this value is tight, meaning that if there is some already existing padding at the end of the data, the padding bytes at the end of the last data won't be counted in. So if there is some padding at the end of the data (before adding this new padding to make the stride a multiple of 4), the last vertex index will be missing without rounding up instead of down.

vita3k/renderer/src/vulkan/scene.cpp
```diff
@@ -235,10 +247,21 @@ static void bind_vertex_streams(VKContext &context, MemState &mem) {
 
     for (std::size_t i = 0; i < max_stream_idx; i++) {
         if (state.vertex_streams[i].data) {
+#ifdef __APPLE__
+            const bool restride = vertex_program.streams[i].stride % 4 != 0;
```
Macdu, line 251: Could you add a comment saying this is needed because MoltenVK forces the vertex stride to be a multiple of 4.

Macdu (approved):
lgtm

PR #2264, opened 2023-01-05
modules/module_parent: fix regression on STREET FIGHTER X TEKKEN

bookmist:
modules/module_parent: invalidate jit cache only for current thread
fix regression on STREET FIGHTER X TEKKEN


Macdu:
The TODO isn't even that mandatory, at worse this lle function will be hit once for each thread.
However I realized there is a race condition, if two threads call the lle function with a really slight delay,
and the first thread is halfway done modifying stub while the second is JIT-recompiling the half-rewritten code.

This is really a really tight race condition and hopefully won't happen.

PR #2266, opened 2023-01-07
gui/settings_dialog: Used vita font

Croden1999:
Change Settings Dialog to use vita font, and adjust text of SetTooltip.


Zangetsu38:
Sorry, i no want using vita font for settings dialogusing basic fine is better for this part

CoffeeBrewer64:
I think there should be an option to switch between fonts, to get the best of both worlds

Zangetsu38:
> I think there should be an option to switch between fonts, to get the best of both worlds

no needed ^^, setting is basic text with only english, no needed vita font

PR #2272, opened 2023-01-14
renderer/vulkan: Implement Memory Mapping

Macdu:
Implement Memory mapping in Vulkan, this is a first implementation and requires the following extensions:
- VK_KHR_external_memory_host to make the buffers gxm wants to map host gpu accessible
- VK_KHR_buffer_device_address to access buffers using their address instead of a ssbo
- VK_KHR_uniform_buffer_standard_layout so that the uniform buffer passed to shaders are not twice as big as they should be, because in the default std140 layout, the stride of an array of uvec2 is 16 bytes...

This requires a lot of modification, in particular how we handle gxm notifications because now the host gpu must be done when the notification is signaled.

What will be done in other PR:
- Add another memory mapping implementation using a page table for devices not supporting VK_KHR_external_memory_host 
- Implement the shader store operation because gpu buffers are no longer read only (should fix T-Posing in 3D unity games)
- Lots of speed improvements can be made

### DIFF
--- a/vita3k/renderer/src/renderer.cpp
+++ b/vita3k/renderer/src/renderer.cpp
@@ -109,8 +109,8 @@ void transfer_fill(State &state, uint32_t fillColor, const SceGxmTransferImage *
     renderer::send_single_command(state, nullptr, renderer::CommandOpcode::TransferFill, false, fillColor, dest);
 }
 
-void sync_surface_data(State &state, Context *ctx) {
-    renderer::add_command(ctx, renderer::CommandOpcode::SyncSurfaceData, nullptr);
+void sync_surface_data(State &state, Context *ctx, const SceGxmNotification vertex_notification, const SceGxmNotification fragment_notification) {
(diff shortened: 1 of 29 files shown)


vita3k/renderer/src/renderer.cpp
```diff
@@ -21,6 +21,8 @@
 
 #include <gxm/functions.h>
 
+#include <util/float_to_half.h>
+
```
KhoraLee, line 25: This include should be on vita3k/renderer/src/vulkan/renderer.cpp not here

Macdu: Indeed, ty

ArbestRi:
This PR broke LPBV with Vulkan

Macdu:
LBP should not crash anymore with the latest update of this PR.

Zangetsu38 (approved):

Zangetsu38 (approved):

Zangetsu38 (approved):

PR #2275, opened 2023-01-19
cmake: allow building on Ubuntu 22.04

cobalt2727:
I've tested this on my 22.04 laptop and it worked perfectly - I'd be willing to guess this number can go even lower, but I don't have a 20.04 system to test CMake 3.16 currently.


Macdu (approved):

lizziermon19-web:
Rekfkd

PR #2281, opened 2023-01-24
renderer/vulkan: Check texture_type to retrieve S24D8 texture

hiroshiyuri:
Check texture_type to retrieve S24D8 texture

This fixes some texture bugs in games built with the Kaleido ADV Workshop engine


hiroshiyuri:
Remove a force-pushed 👍

PR #2282, opened 2023-01-24
renderer/vulkan: Check texture_type to retrieve S24D8 texture

hiroshiyuri:
Check texture_type to retrieve S24D8 texture

This fixes some texture bugs in games built with the Kaleido ADV Workshop engine


Macdu:
Actually that doesn't sound right, a texture type of 0 is for swizzled textures. It is true that depth sentil surfaces can't be swizzled. But the color surface can.

What is the issue? Looking at the code it looks like a normal texture is at the position of an old color surface. But the PR title kind of implies that the texture is a depth stencil, which I don't see how it could with your modifications.

hiroshiyuri:
> Actually that doesn't sound right, a texture type of 0 is for swizzled textures. It is true that depth sentil surfaces can't be swizzled. But the color surface can.
> 
> What is the issue? Looking at the code it looks like a normal texture is at the position of an old color surface. But the PR title kind of implies that the texture is a depth stencil, which I don't see how it could with your modifications.

Ok its not right
I will close this PR and open a new

Thanks 👍

PR #2284, opened 2023-01-26
renderer: Retrieve from texture cache when surface was changed

hiroshiyuri:
I open a new PR with fixes from the last https://github.com/Vita3K/Vita3K/pull/2282

### DIFF
--- a/vita3k/renderer/src/gl/surface_cache.cpp
+++ b/vita3k/renderer/src/gl/surface_cache.cpp
@@ -406,6 +406,11 @@ GLuint GLSurfaceCache::retrieve_color_surface_texture_handle(const State &state,
         } else if (used_iterator != last_use_color_surface_index.end()) {
             last_use_color_surface_index.erase(used_iterator);
         }
+
+        if (surface_stat_changed && purpose == SurfaceTextureRetrievePurpose::READING) {
+            // The surface was changed, so let the texture cache handle it
+            return 0;
+        }
     }
 
     std::unique_ptr<GLColorSurfaceCacheInfo> info_added = std::make_unique<GLColorSurfaceCacheInfo>();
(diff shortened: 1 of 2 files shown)


Zangetsu38:
Nol open new pr for that, if you want change code, edit commit and force push

Macdu:
This will still break many games as surfaces can be swizzled.

hiroshiyuri:
> This will still break many games as surfaces can be swizzled.

Thanks for you reply. 👍

There are few games for PS Vita using the Kaleido ADV Workshop engine.
Listed here: https://vndb.org/r?q=&s=title&f=02fwKaleido_0ADV_0Workshop-4opsv

When I have some time, I will look if theres a better way to do this without hurting other games with a swizzled surfaces. 🤔

hiroshiyuri:
> This will still break many games as surfaces can be swizzled.

Now it does not break many games but needs improvement

I have not found a better approach to do this yet 🤔

hiroshiyuri:
All right now 😁

@Macdu sensei, can you review please?

Macdu:
That's still a really bad hack, there is no reason for a swizzle u8u8u8u8 surface not to be sampled as u8u8u8u8. My u8u8u8 PR was to allow this texture to be sampled, you try to do the opposite only in a specific case. 
What you want to do is find a discrepency between the texture you want to sample and the surface to be sampled. Looking only at the parameters of the texture or the surface won't work.

hiroshiyuri:
No bad now but is a hack yet.

It remains to be seen why the need region is being extracted out incorrectly when surface cache was changed. 🤔

nishinji:
On master, it fixed.
image
image

hiroshiyuri:
> On master, it is fixed. 

closed in this PR #2996 maybe

good job! 🤝

PR #2288, opened 2023-02-01
compat: add timeout for check and download update.

Zangetsu38:
# About
- Add timeout 2 second for check version.
- fix error on download compat db when path have space.
- little reworks code for fix some strange issue and add check if command is successfully executed.

### DIFF
--- a/vita3k/compat/src/compat.cpp
+++ b/vita3k/compat/src/compat.cpp
@@ -91,22 +91,28 @@ bool load_compat_app_db(GuiState &gui, EmuEnvState &emuenv) {
 }
 
 static std::string get_string_output(const std::string cmd) {
+    std::string result;
     std::string tmpres = std::tmpnam(nullptr);
     std::string res_cmd = cmd + " >> " + tmpres;
-    std::system(res_cmd.c_str());
-    std::ifstream res(tmpres, std::ios::in | std::ios::binary);
-    std::remove(tmpres.c_str());
 
-    std::string result;
-    if (res.is_open()) {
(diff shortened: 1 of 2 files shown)


vita3k/compat/src/compat.cpp
```diff
@@ -160,12 +166,15 @@ bool update_compat_app_db(GuiState &gui, EmuEnvState &emuenv) {
     const auto app_compat_db_link = "https://github.com/Vita3K/compatibility/releases/download/compat_db/app_compat_db.xml";
 
 #ifdef WIN32
-    const auto download_command = fmt::format("powershell Invoke-WebRequest {} -Outfile {}", app_compat_db_link, compat_db_path.string());
+    const auto download_command = fmt::format(R"(powershell Invoke-WebRequest {} -Timeout 2 -Outfile \"{}\")", app_compat_db_link, compat_db_path.string());
 #else
-    const auto download_command = fmt::format("curl -L {} -o {}", app_compat_db_link, compat_db_path.string());
+    const auto download_command = fmt::format(R"(curl -m 2 -L {} -o "{}")", app_compat_db_link, compat_db_path.string());
```
1whatleytay, line 171: The point of this is to make a network request? There are library to do this consistently (libcurl/cpr) and I think we already have a couple net request libraries. I'll assume this works well for the time being if you don't want to tackle one of these libraries, I don't have the time right now to rework this.

1whatleytay (approved):

1whatleytay (approved):

PR #2344, opened 2023-02-15
gui: Add first info message replace log for android user.

Zangetsu38:
- gui: Add first message information replace log for android user.
is also better for other OS. vita3k updater: set init update inside thread, fix freeze in boot.
- vita3k updater: set init update inside thread, fix freeze in boot.

# Result:
image


bookmist:
Oh, no! Clicking ok day by day is so annoying. Could you make it auto closing like "N shaders compiled" or (and this will be even better, make it shown by request, some menu item "Show compatibility DB state")

bookmist (approved):

PR #2347, opened 2023-02-16
modules/SceVideodecUser: set frame horizontalSize and verticalSize

lybxlpsv:
Setting horizontalSize and verticalSize values is needed for MGF and Project Diva games to not get stuck when handling AVC videos.

Another thing probably worth mentioning is that, these game may also depend on a stopped video for its game object textures, if the video is stopped, the related object will have green color. Potential fix would be for sceAvcdecDecodeStop to return 0 frames or return the last decoded frame. This however, is not the scope of the current PR.

### DIFF
--- a/vita3k/modules/SceVideodec/SceVideodecUser.cpp
+++ b/vita3k/modules/SceVideodec/SceVideodecUser.cpp
@@ -190,6 +190,7 @@ EXPORT(int, sceAvcdecDecode, SceAvcdecCtrl *decoder, const SceAvcdecAu *au, SceA
     const auto send = decoder_info->send(reinterpret_cast<uint8_t *>(au->es.pBuf.get(emuenv.mem)), au->es.size);
     if (send && decoder_info->receive(output)) {
         decoder_info->get_res(pPicture->frame.frameWidth, pPicture->frame.frameHeight);
+        decoder_info->get_res(pPicture->frame.horizontalSize, pPicture->frame.verticalSize);
         decoder_info->get_pts(pPicture->info.pts.upper, pPicture->info.pts.lower);
         picture->numOfOutput++;
     }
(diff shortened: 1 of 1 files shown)


Zangetsu38:
I no sur here, you using my get_res function, but is diferent value 🤔 need change
For Green screen is régression caused by decode stop rewriting, aot have it and is very bad

Macdu:
First of all thank you very much for this PR! It solves a long standing issue with Project Diva X.

Indeed, it looks like you are right: the user provide the frameWidth, frameHeight and framePitch info and these function calls update the horizontalSize and verticalSize data (also if these differs, the resulting frame should be cropped to the smallest dimensions, which we are not doing yet but anyway).

vita3k/modules/SceVideodec/SceVideodecUser.cpp
```diff
@@ -190,6 +190,7 @@ EXPORT(int, sceAvcdecDecode, SceAvcdecCtrl *decoder, const SceAvcdecAu *au, SceA
     const auto send = decoder_info->send(reinterpret_cast<uint8_t *>(au->es.pBuf.get(emuenv.mem)), au->es.size);
     if (send && decoder_info->receive(output)) {
         decoder_info->get_res(pPicture->frame.frameWidth, pPicture->frame.frameHeight);
```
Macdu, line 192: Can you remove this line? The frameWidth is provided by the user and should not be updated by this function call.

vita3k/modules/SceVideodec/SceVideodecUser.cpp
```diff
@@ -296,6 +297,7 @@ EXPORT(int, sceAvcdecDecodeStop, SceAvcdecCtrl *decoder, SceAvcdecArrayPicture *
         memset(output, 0, H264DecoderState::buffer_size(size));
         // we get the values from the last frame, maybe we should slightly increase the pts value?
         decoder_info->get_res(pPicture->frame.frameWidth, pPicture->frame.frameHeight);
```
Macdu, line 299: Same as above.

Zangetsu38 (changes requested):
this code t'ot'aly broken what i have fixed before, need found real fix for divaX
image

Macdu:
Ok, I apologize, ignore what Zangetsu was saying. 
It looks like the `decoder_info->get_res(pPicture->frame.frameWidth, pPicture->frame.frameHeight);` lines were added as a hack to allow fmvs to display correctly on some games (like StreetFighter), because we do not compare yet the frameWidth and horizontalWidth.

Could you add them back (revert what I said in the review), then this PR should be good to merge.

PR #2364, opened 2023-02-20
gui: merge Japanese and Extra ranges

Zangetsu38:
# About
- gui: merge merge Japanese and Extra ranges
free some memory.
- small refactor of monospaced font.

# Result
- before after
image


EXtremeExploit (approved):
yesu-desu

PR #2365, opened 2023-02-20
app/app init:  Add dpi support of Android.

Zangetsu38:
# About
- app/app init:  Add dpi support of Android.
just for prepare futur


Macdu:
`#ifdef ANDROID` should not be added before the Android PR is merged, this PR has no reason to be made before the Android PR and should be part of it.

PR #2371, opened 2023-02-23
compat: move from curl and powershell to openssl.

Zangetsu38:
# About:
- compat: move from curl and powershell to openssl.
should fix problem for some users.


EXtremeExploit (approved):
ok

PR #2382, opened 2023-02-27
vita3k update: Add progress bar on download new build.

Zangetsu38:
# About
- Add progress bar on download new build.
- https: Add check for null result from getaddrinfo when resolving host address.
should fix crash when internet is no available.

# Result
image


bookmist (approved):
LGTM
I think, It would be good to make a typedef for process callback function type, but it's ok as is.

PR #2384, opened 2023-02-27
lang: Update language strings

Croden1999:
To looks more reasonable.

image
image

### DIFF
--- a/lang/da.xml
+++ b/lang/da.xml
@@ -27,6 +27,7 @@
 	<app_context>
 		<boot>Boot</boot>
 		<check_app_compatibility>Check App Compatibility</check_app_compatibility>
+		<copy_vita3k_summary>Copy Vita3K Summary</copy_vita3k_summary>
 		<open_issue>Open Issue</open_issue>
 		<create_issue>Create Issue</create_issue>
 		<update_database>Update Database</update_database>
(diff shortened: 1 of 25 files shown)


Zangetsu38:
Remove the "comp", it's short for Compatibility
so there is no possible translation in itself
but is good for other

Croden1999:
> Remove the "comp", it's short for Compatibility so there is no possible translation in itself but is good for other

okay, i got it.

Zangetsu38:
> > Remove the "comp", it's short for Compatibility so there is no possible translation in itself but is good for other
> 
> okay, i got it. Of course, i knew the "comp" is short for compatibility, maybe it is good for a small amount of text, like of jp/ch/zh
> 
> image

hmm is writed what like word here ? for comp ?

Croden1999:
> > > Remove the "comp", it's short for Compatibility so there is no possible translation in itself but is good for other
> > 
> > 
> > okay, i got it. Of course, i knew the "comp" is short for compatibility, maybe it is good for a small amount of text, like of jp/ch/zh
> > image
> 
> hmm is writed what like word here ? for comp ?

yes, the "compatibility" in Simplified Chinese is "兼容性"

Zangetsu38:
hmm oki i see

Zangetsu38 (approved):

PR #2388, opened 2023-03-01
gui: improve size and positon of some window for keep proportion.

Zangetsu38:
# About:
- gui: improve size and positon of some window for keep proportion.
fix window abd icon size deformed when when vita3k window is resized.


EXtremeExploit (dismissed):
oke

EXtremeExploit (approved):

PR #2394, opened 2023-03-03
update zh-s.xml

Xenolies:


Croden1999:
do not touch this, revert all

Xenolies:
Okay，I 'm  understand

PR #2395, opened 2023-03-03
mem: Improve protection

Macdu:
Fix many issues concerning our memory protection, mainly:
- Out of bound iterator reads and writes
- remove all the const_casts
- some speed optimizations
- better debugging and logging of windows / unix errors
- fix memory unprotecting happening when it shouldn't
- replace all size_t by uint32_t: this is confusing and we are emulating a 32-bit address range anyway

### DIFF
--- a/vita3k/mem/src/mem.cpp
+++ b/vita3k/mem/src/mem.cpp
@@ -19,8 +19,8 @@
 #include <mem/state.h>
 
 #include <util/align.h>
-#include <util/log.h>
 #include <util/float_to_half.h>
+#include <util/log.h>
 
 #include <algorithm>
 #include <cassert>
@@ -36,7 +36,7 @@
 #include <unistd.h>
 #endif
 
-constexpr size_t STANDARD_PAGE_SIZE = 4096;
+constexpr uint32_t STANDARD_PAGE_SIZE = KiB(4);
 constexpr size_t TOTAL_MEM_SIZE = GiB(4);
 constexpr bool LOG_PROTECT = false;
 constexpr bool PAGE_NAME_TRACKING = false;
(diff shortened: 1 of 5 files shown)


vita3k/mem/src/mem.cpp
```diff
@@ -202,23 +215,13 @@ Address alloc(MemState &state, size_t size, const char *name, unsigned int align
     return align_addr;
 }
 
-static void align_to_page(MemState &state, ProtectSegmentInfo &protect) {
-    const Address end = align(protect.addr + protect.size, state.page_size);
-    const Address addr = align_down(protect.addr, state.page_size);
-    const size_t size = end - addr;
-    protect.addr = addr;
-    protect.size = size;
+static void align_to_page(MemState &state, Address &addr, Address &size) {
```
bookmist, line 218: It's a bit confusing that addr and size have same type, it looks like size is also a guest pointer. (My suggestion is Address &addr, uint32_t &size)

vita3k/mem/src/mem.cpp
```diff
@@ -227,54 +230,47 @@ void unprotect_inner(MemState &state, Address addr, size_t size) {
...
 #endif
 }
 
-void protect_inner(MemState &state, Address addr, size_t size, const std::uint32_t perm) {
+void protect_inner(MemState &state, Address addr, uint32_t size, const uint32_t perm) {
     uint8_t *addr_ptr = state.use_page_table ? state.page_table[addr / KiB(4)] : state.memory.get();
 
 #ifdef WIN32
     DWORD old_protect = 0;
     const BOOL ret = VirtualProtect(&addr_ptr[addr], size - 1, (perm == MEM_PERM_NONE) ? PAGE_NOACCESS : ((perm == MEM_PERM_READONLY) ? PAGE_READONLY : PAGE_READWRITE), &old_protect);
```
bookmist, line 245: It could be good to make MEM_PERM_* enum named and use its type instead of uint32_t

bookmist (dismissed):
Great. I hope it fix a lot of uint32/uint64 conversion warnings.

vita3k/mem/src/mem.cpp
Macdu: Indeed, you are right

vita3k/mem/src/mem.cpp
Macdu: done

bookmist (approved):

Issue #2400, opened 2023-03-06
Latest build on Steam Deck has no/corrupted text


ds8k:
Running the latest build (3041) has issues rendering fonts on the Steam Deck. Looks to be related to #2396 as the commit before it runs fine.

![steamuserimages-a akamaihd](https://user-images.githubusercontent.com/4458812/223045437-fc850e5d-9d3d-4a83-af86-979a74ba5026.jpg)

Zangetsu38:
if you can check this pr for look if fix your issue
https://github.com/Vita3K/Vita3K/pull/2401

Zangetsu38:
i try some things on pr, you can test build when finished, want already see if this is good solution

Zangetsu38:
@ds8k test again build, and show me result

Zangetsu38:
should be fixed in master now.

PR #2401, opened 2023-03-06
app/app init: Fix calcule dpi for no x86_64 arch.

Zangetsu38:
# About:
- app/app init: Fix calcule dpi for no x86_64 arch.
- fix some format.


Macdu:
I don't think the dpi scale has anything to do with the processor architecture

ds8k:
This does not resolve #2400

Zangetsu38:
> This does not resolve #2400

Yes i know

ds8k:
Latest build

image

Zangetsu38:
@ds8k test it https://github.com/Vita3K/Vita3K/actions/runs/4348552377
is apparently fixed now

Macdu:
Disabling High DPI on all devices that have high dpi because the font does not render is not a solution, only an ugly hack...

Zangetsu38:
The use of High dpi is an obligation on PC, it also fixed problems on Linux, we only deactivate it on a specific device

bookmist (approved):

Macdu:
Why was this merged? This is a font rendering bug that need to be solved. It has nothing to do with Linux, the steam deck or windows. This will happen on any device provided that their dpi is high enough (which is only the case for HD small format screens). You just fixed it on a specific device using a hack that make the UI blurry instead of fixing the font rendering bug...

Zangetsu38:
> Why was this merged? This is a font rendering bug that need to be solved. It has nothing to do with Linux, the steam deck or windows. This will happen on any device provided that their dpi is high enough (which is only the case for HD small format screens). You just fixed it on a specific device using a hack that make the UI blurry instead of fixing the font rendering bug...

I see, except that it was urgent, to find a temporary even solution, you must have seen, that for another I used it completely blocked the launch, and had nothing to do with the fonts

PR #2419, opened 2023-03-12
renderer/texture cache: add fmt U8U8 in unswizzled without decode.

Zangetsu38:
# About
- Add texture format U8U8 on can_texture_be_unswizzled_without_decode

# Result
- fix render on limbo
image


EXtremeExploit (approved):
yes zang-sama :pray:

PR #2420, opened 2023-03-13
util: Fix log craches again

bookmist:
Now (I hope) crashes will be fixed. I finally found how to call spdlog flush.


Macdu:
Have you checked that this is thread safe ?

bookmist:
flush_all use same locks as log message. So I think it's thread-safe

Macdu (approved):

PR #2440, opened 2023-03-23
https: fix crash during a simultaneous call of init.

Zangetsu38:
# About
- https: fix crash during a simultaneous call of init.
fix crash when call few instance in same time of init, like for update compat and check update on vita3K


EXtremeExploit (approved):
:pray: it actually fixes the crash

PR #2467, opened 2023-03-31
shader: Implement the shader store operation

Macdu:
Implement the shader store operation, which is used by shaders to modify the content of buffers.
This obviously only works with memory mapping enabled, as without it, shaders only have access to a copy of the buffer that gets recreated every draw.

This fixes games using this features, mainly 3D Unity games.
However, only half of them are fixed as we still don't implement properly the repeat count for shader stores. This is a bit harder than I thought it would be to reverse engineer, so I will do it in another PR.


Zangetsu38 (dismissed):

Macdu:
This requires some testing before merging.

Zangetsu38:
Request rebase a’d fix conflict

ArbestRiAgain:
Any more info? Hello?

Zangetsu38:
GOD, please rebase 😭, this all conflict everywhere

Macdu:
I rebased, I just need to adapt some of my code to the modifications I made for shader interlock and macrotile emulation before pushing again.

Zangetsu38:
> I rebased, I just need to adapt some of my code to the modifications I made for shader interlock and macrotile emulation before pushing again.

I see, i speak current conflict on pr here, because yes, master have get very lot change lol

Zangetsu38 (approved):

Zangetsu38 (changes requested):
need remove f commit here

Zangetsu38 (approved):

PR #2469, opened 2023-04-01
renderer/vulkan: Implement surface sync

Macdu:
Implement surface sync on the Vulkan renderer.
For many reasons (it is a lot easier to code and a lot more efficient), surface sync can be enabled only with memory mapping.

This implementation is far better than the opengl one. First, if upscaling is enabled, the surface is downscaled on the GPU instead of the CPU. Second, memory trapping is used so that surface sync is done only on surfaces that really need it.

### DIFF
--- a/vita3k/config/include/config/config.h
+++ b/vita3k/config/include/config/config.h
@@ -63,7 +63,7 @@ enum PerfomanceOverleyPosition {
     code(std::string, "backend-renderer", "OpenGL", backend_renderer)                                   \
     code(int, "gpu-idx", 0, gpu_idx)                                                                    \
     code(int, "resolution-multiplier", 1, resolution_multiplier)                                        \
-    code(bool, "disable-surface-sync", false, disable_surface_sync)                                     \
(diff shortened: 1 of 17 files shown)


ArbestRi:
This might help LBPV as it needs surface sync to show graphics.

Foxlum:
> This might help LBPV as it needs surface sync to show graphics.

@Macdu Its not completely there yet but looks so much more correct than the current OpenGL rendering. 
Vulkan:
image
OpenGL:
image

ArbestRi:
> 

Can relate @Foxlum , OpenGL and LBPV looks like 💩

ArbestRi:
Yep, it definitely fixed the graphics
Screenshot 2023-04-02 173139

Macdu:
This PR is ready for review.
As it is not possible to protect mapped buffer with Intel and AMD drivers on Linux, all surfaces are surface synced when this option is enabled when this happened (can't really do better).

Zangetsu38 (approved):

PR #2470, opened 2023-04-02
net/http: Some improvements

EXtremeExploit:
~Allows highschool dxd to go past the PSN screen, but then servers are down so i have to find a workaround for when servers of a game are down. i saw that games that use this compare to 0 and values above it, so -1 is kinda out of place, still tho, i prefer if more games were tested just to make sure this doesn't break any game that require PSN connectivity~

EDIT: 
* Make `sceNetCheckDialogGetResult` result be a valid number
* A bit better of http handling when its HLE and http disabled
* Implement IP OBTAINED callback 
* Fix http read body data always returning 0


CoffeeBrewer64:
you could always just make it a setting, allowing the user to keep the old system. albeit there is still a better way around this, I can feel it

EXtremeExploit:
I think now its ready for review, had some people test some games, no regressions known afaik

bookmist (approved):

PR #2497, opened 2023-04-15
renderer/scene: Implement convert linear to swizzled for transfercopy.

Zangetsu38:
# About
- Implement convert linear to swizzled for transfercopy.

# Result
- fix transition screen of angry birst stars wars
image
image
- fix render of background on &: Sora no Mukou de Saki Masuyou ni 
image

### DIFF
--- a/vita3k/renderer/src/scene.cpp
+++ b/vita3k/renderer/src/scene.cpp
@@ -223,44 +223,109 @@ COMMAND(handle_transfer_copy) {
     const SceGxmTransferType src_type = helper.pop<SceGxmTransferType>();
     const SceGxmTransferType dst_type = helper.pop<SceGxmTransferType>();
 
+    const auto src_is_linear = src_type == SCE_GXM_TRANSFER_LINEAR;
+    const auto dest_is_linear = dst_type == SCE_GXM_TRANSFER_LINEAR;
+    const auto src_is_swizzled = src_type == SCE_GXM_TRANSFER_SWIZZLED;
+    const auto dest_is_swizzled = dst_type == SCE_GXM_TRANSFER_SWIZZLED;
+
+    // Get bits per pixel of source and destination
+    const auto src_bpp = gxm::get_bits_per_pixel(src->format);
+    const auto dest_bpp = gxm::get_bits_per_pixel(dest->format);
+
+    // Set bytes per pixel of source and destination
+    const uint32_t src_bytes_per_pixel = (src_bpp + 7) >> 3;
+    const uint32_t dest_bytes_per_pixel = (dest_bpp + 7) >> 3;
+
+    // Set width and height of source and destination
+    const uint32_t src_width = src->width;
+    const uint32_t src_height = src->height;
+    const uint32_t dest_width = dest->width;
+    const uint32_t dest_height = dest->height;
+
+    const auto copy_pixel = [&](const uint32_t src_offset, const uint32_t dest_offset) {
+        // Set pointer of source and destination
+        const auto src_ptr = (uint8_t *)src->address.get(mem) + src_offset;
+        auto dest_ptr = (uint8_t *)dest->address.get(mem) + dest_offset;
+
+        // Set color of source
+        const auto src_color = *(uint32_t *)src_ptr;
+
+        // Copy result in destination depending color Key
+        switch (colorKeyMode) {
+        case SCE_GXM_TRANSFER_COLORKEY_NONE:
+            std::memcpy(dest_ptr, src_ptr, dest_bytes_per_pixel);
+            break;
+        case SCE_GXM_TRANSFER_COLORKEY_PASS:
+            if ((src_color & colorKeyMask) == colorKeyValue)
+                std::memcpy(dest_ptr, src_ptr, dest_bytes_per_pixel);
+            break;
+        case SCE_GXM_TRANSFER_COLORKEY_REJECT:
+            if ((src_color & colorKeyMask) != colorKeyValue)
+                std::memcpy(dest_ptr, src_ptr, dest_bytes_per_pixel);
+            break;
+        default: break;
+        }
+    };
+
     if (src_type == dst_type) {
-        const auto src_bpp = gxm::get_bits_per_pixel(src->format);
-        const auto dest_bpp = gxm::get_bits_per_pixel(dest->format);
-        const uint32_t src_bytes_per_pixel = (src_bpp + 7) >> 3;
-        const uint32_t dest_bytes_per_pixel = (dest_bpp + 7) >> 3;
-
-        for (uint32_t x = 0; x < src->width; x++) {
-            for (uint32_t y = 0; y < src->height; y++) {
+        for (uint32_t x = 0; x < src_width; x++) {
+            for (uint32_t y = 0; y < src_height; y++) {
                 // Set offset of source and destination
-                const auto src_offset = ((x + src->x) * src_bytes_per_pixel) + ((y + src->y) * src->stride);
-                const auto dest_offset = ((x + dest->x) * dest_bytes_per_pixel) + ((y + dest->y) * dest->stride);
+                const uint32_t src_offset = ((x + src->x) * src_bytes_per_pixel) + ((y + src->y) * src->stride);
+                const uint32_t dest_offset = ((x + dest->x) * dest_bytes_per_pixel) + ((y + dest->y) * dest->stride);
 
-                // Set pointer of source and destination
-                const auto src_ptr = (uint8_t *)src->address.get(mem) + src_offset;
-                auto dest_ptr = (uint8_t *)dest->address.get(mem) + dest_offset;
-
-                // Set color of source
-                const auto src_color = *(uint32_t *)src_ptr;
-
-                // Copy result in destination depending color Key
-                switch (colorKeyMode) {
-                case SCE_GXM_TRANSFER_COLORKEY_NONE:
-                    memcpy(dest_ptr, src_ptr, dest_bytes_per_pixel);
-                    break;
-                case SCE_GXM_TRANSFER_COLORKEY_PASS:
-                    if ((src_color & colorKeyMask) == colorKeyValue)
-                        memcpy(dest_ptr, src_ptr, dest_bytes_per_pixel);
-                    break;
-                case SCE_GXM_TRANSFER_COLORKEY_REJECT:
-                    if ((src_color & colorKeyMask) != colorKeyValue)
-                        memcpy(dest_ptr, src_ptr, dest_bytes_per_pixel);
-                    break;
-                default: break;
-                }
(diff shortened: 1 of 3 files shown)


vita3k/renderer/src/scene.cpp
```diff
@@ -223,44 +223,93 @@ COMMAND(handle_transfer_copy) {
     const SceGxmTransferType src_type = helper.pop<SceGxmTransferType>();
     const SceGxmTransferType dst_type = helper.pop<SceGxmTransferType>();
```
Macdu, line 225: The whole code below is valid if and only if src->format == dst->format (and the format is byte-aligned and not YUV)

vita3k/renderer/src/scene.cpp
```diff
@@ -223,44 +223,93 @@ COMMAND(handle_transfer_copy) {
...
+    // Set width and height of source and destination
+    const uint32_t src_width = src->width;
+    const uint32_t src_height = src->height;
+    const uint32_t dest_width = dest->width;
+    const uint32_t dest_height = dest->height;
+
+    const auto copy_pixel = [&](const uint32_t src_offset, const uint32_t dest_offset) {
+        // Set pointer of source and destination
+        const auto src_ptr = (uint8_t *)src->address.get(mem) + src_offset;
+        auto dest_ptr = (uint8_t *)dest->address.get(mem) + dest_offset;
```
Macdu, line 243: It would be better to get the pointer to src and dest (the src->adress.get(mem) and dest->address.get(mem)) at the beginning of the function and not inside the loop<;

vita3k/renderer/src/scene.cpp
```diff
@@ -223,44 +223,93 @@ COMMAND(handle_transfer_copy) {
...
+        case SCE_GXM_TRANSFER_COLORKEY_NONE:
+            std::memcpy(dest_ptr, src_ptr, dest_bytes_per_pixel);
+            break;
+        case SCE_GXM_TRANSFER_COLORKEY_PASS:
+            if ((src_color & colorKeyMask) == colorKeyValue)
+                std::memcpy(dest_ptr, src_ptr, dest_bytes_per_pixel);
+            break;
+        case SCE_GXM_TRANSFER_COLORKEY_REJECT:
+            if ((src_color & colorKeyMask) != colorKeyValue)
+                std::memcpy(dest_ptr, src_ptr, dest_bytes_per_pixel);
```
Macdu, line 259: This only works for U8U8U8U8_ABGR format

vita3k/renderer/src/scene.cpp
```diff
@@ -223,44 +223,93 @@ COMMAND(handle_transfer_copy) {
...
+                copy_pixel(src_offset, dest_offset);
+            }
+        }
+    } else if ((src_type == SCE_GXM_TRANSFER_LINEAR) && (dst_type == SCE_GXM_TRANSFER_SWIZZLED)) {
+        for (uint32_t y = 0; y < src_height; y++) {
+            for (uint32_t x = 0; x < src_width; x++) {
+                const uint32_t i = y * src_width + x;
+
+                const size_t min = src_width < src_height ? src_width : src_height;
+                const size_t k = static_cast<size_t>(log2(min));
```
Macdu, line 282: put this outside the loop

vita3k/renderer/src/scene.cpp
```diff
@@ -223,44 +223,93 @@ COMMAND(handle_transfer_copy) {
...
+
+                const size_t sx = x + dest->x;
+                const size_t sy = y + dest->y;
+                const size_t dx = texture::decode_morton2_x(i);
+                const size_t dy = texture::decode_morton2_y(i);
```
Macdu, line 287: Use `uint32_t` and not `size_t` please

vita3k/renderer/src/scene.cpp
Macdu, line 282:
```suggestion
                const uint32_t k = std::bit_width(min);
```

vita3k/renderer/src/scene.cpp
```diff
@@ -223,44 +223,93 @@ COMMAND(handle_transfer_copy) {
...
+
+                if (dx >= src_width || dy >= src_height)
+                    continue;
+
+                size_t j;
+                const size_t mask = (1 << k) - 1;
```
Macdu, line 293:
```suggestion
                const uint32_t mask = min - 1;
```
min = 1 << k (width and height must be a power of 2 for swizzled textures)

vita3k/renderer/src/scene.cpp
```diff
@@ -223,44 +223,93 @@ COMMAND(handle_transfer_copy) {
...
+    } else if ((src_type == SCE_GXM_TRANSFER_LINEAR) && (dst_type == SCE_GXM_TRANSFER_SWIZZLED)) {
+        for (uint32_t y = 0; y < src_height; y++) {
+            for (uint32_t x = 0; x < src_width; x++) {
+                const uint32_t i = y * src_width + x;
+
+                const size_t min = src_width < src_height ? src_width : src_height;
+                const size_t k = static_cast<size_t>(log2(min));
+
+                const size_t sx = x + dest->x;
+                const size_t sy = y + dest->y;
```
Macdu, line 285: rename this dest_x and dest_y

vita3k/renderer/src/scene.cpp
```diff
@@ -223,44 +223,93 @@ COMMAND(handle_transfer_copy) {
...
+                const uint32_t dest_offset = ((x + dest->x) * dest_bytes_per_pixel) + ((y + dest->y) * dest->stride);
+
+                // Copy pixel from source to destination
+                copy_pixel(src_offset, dest_offset);
+            }
+        }
+    } else if ((src_type == SCE_GXM_TRANSFER_LINEAR) && (dst_type == SCE_GXM_TRANSFER_SWIZZLED)) {
+        for (uint32_t y = 0; y < src_height; y++) {
+            for (uint32_t x = 0; x < src_width; x++) {
+                const uint32_t i = y * src_width + x;
```
Macdu, line 279:
```suggestion
                const uint32_t swizzled_location = y * src_width + x;
```

vita3k/renderer/src/scene.cpp
```diff
@@ -223,44 +223,93 @@ COMMAND(handle_transfer_copy) {
...
+                const size_t min = src_width < src_height ? src_width : src_height;
+                const size_t k = static_cast<size_t>(log2(min));
+
+                const size_t sx = x + dest->x;
+                const size_t sy = y + dest->y;
+                const size_t dx = texture::decode_morton2_x(i);
+                const size_t dy = texture::decode_morton2_y(i);
+
+                if (dx >= src_width || dy >= src_height)
+                    continue;
```
Macdu, line 290: Wrong (you are comparing the dest coordinate with the source coordinate). And even with the right coordinates, this check will always be true.

vita3k/renderer/src/scene.cpp
```diff
@@ -223,44 +223,93 @@ COMMAND(handle_transfer_copy) {
...
+                const uint32_t src_offset = ((x + src->x) * src_bytes_per_pixel) + ((y + src->y) * src->stride);
+                const uint32_t dest_offset = ((x + dest->x) * dest_bytes_per_pixel) + ((y + dest->y) * dest->stride);
+
+                // Copy pixel from source to destination
+                copy_pixel(src_offset, dest_offset);
+            }
+        }
+    } else if ((src_type == SCE_GXM_TRANSFER_LINEAR) && (dst_type == SCE_GXM_TRANSFER_SWIZZLED)) {
+        for (uint32_t y = 0; y < src_height; y++) {
+            for (uint32_t x = 0; x < src_width; x++) {
```
Macdu, line 278: Invert these two lines to improve cache usage

vita3k/renderer/src/scene.cpp
```diff
@@ -223,44 +223,93 @@ COMMAND(handle_transfer_copy) {
...
+                const uint32_t i = y * src_width + x;
+
+                const size_t min = src_width < src_height ? src_width : src_height;
```
Macdu, line 281: You must compare dst_width and dst_height, no src (the destination is swizzled, not the source)

vita3k/renderer/src/scene.cpp
```diff
@@ -223,44 +223,93 @@ COMMAND(handle_transfer_copy) {
...
+                size_t j;
+                const size_t mask = (1 << k) - 1;
+                if (src_height < src_width) {
+                    j = (dy << k) | (dx & mask);
+                } else {
+                    j = (dx << k) | (dy & mask);
                 }
+
+                const size_t src_x = j % src_width;
+                const size_t src_y = j / src_width;
```
Macdu, line 301: Wrong, this only works because will == height. A mask must be applied before the call to decode_morton.

vita3k/renderer/src/scene.cpp
```diff
@@ -223,44 +223,93 @@ COMMAND(handle_transfer_copy) {
...
+
+                // Set offset of source and destination
+                const uint32_t src_offset = static_cast<uint32_t>(((src_y + src->y) * src_width + (src_x + src->x)) * src_bytes_per_pixel);
```
Macdu, line 304:
```suggestion
                const uint32_t src_offset = static_cast<uint32_t>(((src_y + src->y) * src_stride + (src_x + src->x)) * src_bytes_per_pixel);
```

Macdu (changes requested):

vita3k/renderer/src/scene.cpp
Zangetsu38: ? no understund for that, is point and offset, no have sense set point alone

vita3k/renderer/src/scene.cpp
Zangetsu38: is master code, i change nothing

vita3k/renderer/src/scene.cpp
Zangetsu38: yes, i know, i have thinks of this too

vita3k/renderer/src/scene.cpp
Zangetsu38: is good, dx is tcnically source

vita3k/renderer/src/scene.cpp
Zangetsu38: is not dest here, x and y is for of source

vita3k/renderer/src/scene.cpp
Zangetsu38: using src->stride here broken texture

Macdu (approved):
Well it only works if the whole swizzled texture is copied, the two formats are the same and using a colormode other than none only works with rgba8 format. Also some comments are wrong...
But yeah, I guess I'll rewrite this function later...

PR #2500, opened 2023-04-16
savedata: Fix requiredKiB not being set to 0 when all executed nicely

EXtremeExploit:
So turns out that requiredKiB is not always initialized to 0 ingame, so it ended up being a pointer to garbage and stayed garbage cause we dont do nothing to it, so we have to set it to 0 it ourselves, this fixes minecraft saying that theres no free space

### DIFF
--- a/vita3k/modules/SceAppUtil/SceAppUtil.cpp
+++ b/vita3k/modules/SceAppUtil/SceAppUtil.cpp
@@ -262,6 +262,11 @@ EXPORT(int, sceAppUtilSaveDataDataSave, SceAppUtilSaveDataFileSlot *slot, SceApp
     TRACY_FUNC(sceAppUtilSaveDataDataSave, slot, files, fileNum, mountPoint, requiredSizeKiB);
     SceUID fd;
 
+    if (requiredSizeKiB)
+        // requiredSizeKiB must be set to 0 if there is enough space available
+        // else we need to set it to the space needed in KiB (TODO)
+        *requiredSizeKiB = 0;
+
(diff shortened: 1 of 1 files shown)


Zangetsu38 (approved):

vita3k/modules/SceAppUtil/SceAppUtil.cpp
```diff
@@ -262,6 +262,9 @@ EXPORT(int, sceAppUtilSaveDataDataSave, SceAppUtilSaveDataFileSlot *slot, SceApp
     TRACY_FUNC(sceAppUtilSaveDataDataSave, slot, files, fileNum, mountPoint, requiredSizeKiB);
     SceUID fd;
 
+    if (requiredSizeKiB)
+        *requiredSizeKiB = 0; // Set this to 0 as its not always initialized to 0 and 0 is expected when all is good (afternoon)
```
Macdu, line 266: requiredSizeKiB should be set to the size required for the save file if there is not enough space available.
If there is enough space available, it must be set to 0.
We always assume there is enough space available, so this is the correct behavior, please just update the comment to make it easier to understand, like:
`// requiredSizeKiB must be set to 0 if there is enough space available`

EXtremeExploit: Is what i pushed better now?

Macdu: Fine for me, although I don't think the TODO is necessary: I don't see a reason to emulate the fact that there is no more storage space available on the PS Vita.

PR #2501, opened 2023-04-17
net/types: fix compile error for msvc-clang

bookmist:
I got an error if s_addr macro is not defined back on msvc-clang-16


EXtremeExploit (approved):
definetly this is much better, thanks bookmist

Issue #2499, opened 2023-04-16
The app hangs with infinite loop of "Invalid instruction" errors with Gravity Rush


v-fox:
When using [path-patched](https://github.com/Vita3K/Vita3K/issues/664#issuecomment-1495576124) 8623b51653f85f34766fe51499a9f622bc2cf26b [snapshot on Linux](https://build.opensuse.org/package/show/home:X0F:branches:Emulators/vita3k) if I try to run Gravity Rush (PCSA00011) with all its DLCs, vita3k hangs permanently and infinitely craps into log-file the stream of:
```
[21:40:07.029] |W| [log_error_details]: Executing: .byte 0xee, 0xdb
[21:40:07.030] |W| [ExceptionRaised]: Undefined instruction at address 0x2000, instruction 0xDEADBEE (.byte 0xee, 0xdb)
[21:40:07.030] |E| [log_error_details]: Unicorn error 0xA. Invalid instruction (UC_ERR_INSN_INVALID)
PC: 0x00002000,   SP: 0x00125508,   LR: 0x81002993
r0 : 0x00125d58   r6 : 0x492e789d
r1 : 0x00000000   r7 : 0x1c5c36fb
r2 : 0x00125d68   r8 : 0xcca9471d
r3 : 0x00125552   r9 : 0x81528294
r4 : 0x00125d58   r10: 0x8104cfbb
r5 : 0x81683eb0   r11: 0x00000000
r12: 0x00000000
Thumb: false

[21:40:07.030] |W| [log_error_details]: Executing: .byte 0xee, 0xdb
[21:40:07.030] |W| [ExceptionRaised]: Undefined instruction at address 0x2000, instruction 0xDEADBEE (.byte 0xee, 0xdb)
[21:40:07.030] |E| [log_error_details]: Unicorn error 0xA. Invalid instruction (UC_ERR_INSN_INVALID)
PC: 0x00002000,   SP: 0x00125508,   LR: 0x81002993
r0 : 0x00125d58   r6 : 0x492e789d
r1 : 0x00000000   r7 : 0x1c5c36fb
r2 : 0x00125d68   r8 : 0xcca9471d
r3 : 0x00125552   r9 : 0x81528294
r4 : 0x00125d58   r10: 0x8104cfbb
r5 : 0x81683eb0   r11: 0x00000000
r12: 0x00000000
Thumb: false
```

[...]

PR #2516, opened 2023-04-22
vita3k/gui: fix crash on wrong app icon

bookmist:
It could happens on badly installed game and no firmware, or just wrong icon
should fix SIGABRT here #2499


Zangetsu38:
Ah yes, causing crash on Android 💀
Eu you can juste set commit name in two line ? For no have stupid ... in end 🙏

Zangetsu38 (approved):

PR #2533, opened 2023-04-28
modules/kernel: Better logging on stack corruption

EXtremeExploit:
corruptions are VERY BAD, it should be better to have them be a critical log and then log the current state of the cpu just so its easier to see when the corruption happened c:


bookmist (approved):
It would be great to log thread_id too.

EXtremeExploit:
donedid

bookmist (approved):

PR #2540, opened 2023-05-01
renderer/vulkan/sync state: stub SCE_GXM_REGION_CLIP_INSIDE.

Zangetsu38:
# About
- renderer/vulkan/sync state: stub SCE_GXM_REGION_CLIP_INSIDE.
should fix Ridge racer render in vulkan backend.

# Result
- fix yellow screen issue in vulkan
image
image


Macdu:
This is doing exactly the same as SCE_GXM_REGION_CLIP_OUTSIDE, so exactly the opposite of what it should be doing. There is a reason this command is unimplemented: Vulkan does not support it. 
In order to support it, the extension VK_NV_scissor_exclusive or a shader implementation would be needed.
Setting it as SCE_GXM_REGION_CLIP_NONE would be a better stub than doing the opposite of what it is supposed to do.

Zangetsu38:
I see, done and works too

Macdu (approved):

PR #2551, opened 2023-05-05
shader/renderer: Implement DEPTHF instruction and some depth buffer fixes

Macdu:
Implement the DEPTHF shader instruction, which allows modifying the depth in the fragment buffer.
Also contain some fixes concerning  depth buffer used when MSAA is enabled as their size is not the one expected.


EXtremeExploit (approved):
trust the macdu :pray:

Zangetsu38 (approved):

PR #2552, opened 2023-05-05
Vita3K: code cleanup and optimisations

bookmist:
* drop unused variables (sometimes comment if it looks useful sometimes in future)
* replace const objects to const references
* replace insert to emplace, push_back to emplace_back
* Macdu, look on vita3k/renderer/src/vulkan/context.cpp rt is used later
* adjust types to avoid unnecessary conversions
* other


CoffeeBrewer64:
The guidelines (in da wiki) say to remove code instead of commenting it out

vita3k/shader/src/usse_translator_entry.cpp
```diff
@@ -35,7 +35,7 @@ using USSEMatcher = shader::decoder::Matcher<Visitor, uint64_t>;
 
 template <typename V>
 static std::optional<const USSEMatcher<V>> DecodeUSSE(uint64_t instruction) {
-    static const std::vector<USSEMatcher<V>> table = {
+    static const std::array<USSEMatcher<V>, 33> table = {
```
Macdu, line 38: you would need to update the matching line in tools/usse-decoder-gen/autogen.py

Macdu:
Really nice, that's a lot of small improvements/optimisations. I tend to agree with Plutia in saying that unused lines should be removed instead of commented out. The only exception would be the code in sce_utils.cpp as we still could use it later. I have only one small comment about it.

vita3k/shader/src/usse_translator_entry.cpp
bookmist: done

Macdu (approved):

PR #2582, opened 2023-05-11
lang: update japanese

nishinji:
- update japanese
~~PS TV is Vita TV in jpn~~
- fix small about gui

`※` is not in ltn, makes display
image


lang/ja.xml
```diff
@@ -38,10 +38,10 @@
 	</about>
 
 	<app_context>
-		<check_app_state>アプリの状態をチェック</check_app_state>
+		<check_app_state>アプリの互換性を確認</check_app_state>
```
Zangetsu38, line 41: same revert

nishinji:
> but now is PS TV (https://manuals.playstation.net/document/jp/products.html)

I see
but I believe PS Vita TV is the official abbreviation.
In Japan, the emphasis was on being a vita.

Zangetsu38 (approved):

PR #2585, opened 2023-05-15
misc: Allow toggling of memory mapping on PC

Lupiax:


Zangetsu38:
need squash all, and no need pass config inside it normally, you can using featur for that

Zangetsu38:
Close it, no really needed on pc

mrcmunir:
@Zangetsu38  it's needed for arm devices under PC system  will cause problem in some socs with memory mapping ON   and probably old GPU structure in PC too .
On arm devices due fragmentation drivers you may find that memory mapping via table page works better but introduces new graphical bugs.
Memory mappng via external_memory is quite driver dependent and won't always work depending on the hardware it's a lottery and complete broken if extension detected comming  unstable vita3k compared without it.

vita3k/renderer/src/vulkan/gxm_to_vulkan.cpp
```diff
@@ -480,6 +480,7 @@ namespace texture {
 
 static const vk::ComponentMapping translate_swizzle1(SceGxmTextureSwizzle1Mode mode) {
     switch (mode) {
+    /**/
```
Zangetsu38, line 483: i juste see here but ???

PR #2605, opened 2023-05-25
modules/SceMotion: stub SceMotionGetState with set value to zero.

Zangetsu38:
- modules/SceMotion: stub SceMotionGetState with default values.
Fixes head movement of Sackboy occurring independently in LBP.
Also fixes the issue in Gravity Rush where the screen turns black when using gravity or during specific moments involving the motion sensor.
- motion: Prepares for future motion support.

# Result
- looking camera without moving
image

- FINALY fix gravity rush, move to playable state
image


vita3k/modules/SceDriverUser/SceMotion.cpp
```diff
@@ -48,28 +51,76 @@ EXPORT(int, sceMotionGetDeadbandExt) {
...
 }
 
-EXPORT(int, sceMotionGetState) {
-    TRACY_FUNC(sceMotionGetState);
+EXPORT(int, sceMotionGetState, SceMotionState *motionState) {
+    TRACY_FUNC(sceMotionGetState, motionState);
+    if (!motionState)
+        return RET_ERROR(SCE_MOTION_ERROR_NULL_PARAMETER);
+
+    STUBBED("Set some default values");
```
Macdu, line 79:
```suggestion
    STUBBED("Set to zero");
```

vita3k/motion/include/motion/state.h
```diff
@@ -0,0 +1,30 @@
...
+
+#pragma once
+
+#include <motion/motion.h>
+
+#include <SDL_gamecontroller.h>
+
+struct _SDL_GameController;
+
+typedef std::shared_ptr<_SDL_GameController> GameControllerPtr;
```
Macdu, line 26: remove

gen-windows.bat
```diff
@@ -1,4 +1,5 @@
 @echo off
+title Generate project solution VS
```
Macdu, line 2: remove, not part of this PR

vita3k/motion/include/motion/state.h
Zangetsu38: hmm is used for motion suport in DS/4, is just for prepare next works

gen-windows.bat
Zangetsu38: is small change, i no like create pr for very small things like that is not important

vita3k/modules/SceDriverUser/SceMotion.cpp
Zangetsu38: i no set all to zero, is more default value in default position

vita3k/motion/include/motion/state.h
Macdu: Something starting with an underscore is definitely not meant to be used, remove

Zangetsu38: ??? this line is litteraly same inside ctrl ? 🤔

Zangetsu38: image

Macdu: Ah ok, the c typedef issue. Anyway, no need to add this now and we may have to include Ctrl state which defines it anyway

Zangetsu38: no need include ctrl state, is full diferent; only need this typdef, no any other things inside ctrl, so is no sense add full moiule for one things

Macdu: Either you use it, either you do not add these lines

Zangetsu38: done, but no removed, just disable for now, like pr said is stuff for prepare futur implementation, like that when start it juste enable again

ArbestRiAgain:
FINALLY! GRAVITY RUSH NO LONGER SOFTLOCKS AFTER THAT ONE PART!

vita3k/motion/include/motion/state.h
```diff
@@ -0,0 +1,32 @@
...
+
+#include <motion/motion.h>
+
+/** TODO: Enable this when start implementing motion support with SDL2
+#include <SDL_gamecontroller.h>
+
+// struct _SDL_GameController;
+
+// typedef std::shared_ptr<_SDL_GameController> GameControllerPtr;
+**/
```
Macdu, line 28: What's the point of putting this, it'll most likely get deleted when correctly implementing it for many reasons. Remove.

Zangetsu38: done

Macdu (approved):

PR #2608, opened 2023-05-26
modules/sceJpegEncUser: Implements most of functions.

saturnsky:
- MJpeg and sceJpegEncoderSetValidRegion functions are not yet supported.
- Ignores the CompressionRatio parameter.

### DIFF
--- a/vita3k/codec/src/mjpeg.cpp
+++ b/vita3k/codec/src/mjpeg.cpp
@@ -19,6 +19,7 @@
 
 extern "C" {
 #include <libavcodec/avcodec.h>
+#include <libavutil/imgutils.h>
 #include <libswscale/swscale.h>
 }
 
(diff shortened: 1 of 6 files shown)


vita3k/codec/src/mjpeg.cpp
```diff
@@ -112,6 +224,7 @@ bool MjpegDecoderState::receive(uint8_t *data, DecoderSize *size) {
                     std::memcpy(&channels[a][b * frame->width / (a ? 2 : 1)], &frame->data[a][b * frame->linesize[a]], frame->width / (a ? 2 : 1));
                 }
             }
+            convert_yuv420_to_yuv444(temporary.data(), data, frame->width, frame->height);
```
Macdu, line 227: Do not do that (even though Zangetsu told you you could) for a few reasons:
- It's not efficient (even though that's the least of our issues)
- YUV444 format takes more space than YUV420, if the user allocates the buffer without using sceJpegGetOutputInfo (which returns a way too big value btw), then you'll have a buffer overflow
- The renderer can sample directly from textures in YUV420 format. If you were to supply a YUV444 texture instead, you would get some really bad output.

vita3k/jpegenc/include/jpegenc/types.h
```diff
@@ -0,0 +1,59 @@
...
+
+struct SceJpegEncoderInitParam {
+    uint32_t size;
+    int32_t inWidth;
+    int32_t inHeight;
+    int32_t pixelFormat;
+    Ptr<uint8_t> outBuffer;
+    uint32_t outSize;
+    SceJpegEncoderInitParamOption option;
+};
```
Macdu, line 59: add a newline

vita3k/jpegenc/include/jpegenc/types.h
```diff
@@ -0,0 +1,59 @@
...
+    SCE_JPEGENC_ERROR_INVALID_HEADER_MODE = 0x80650204,
+    SCE_JPEGENC_ERROR_INVALID_POINTER = 0x80650205,
+    SCE_JPEGENC_ERROR_NOT_PHY_CONTINUOUS_MEMORY = 0x80650206
+};
+
+enum SceJpegEncoderPixelFormat : int32_t {
+    SCE_JPEGENC_PIXELFORMAT_ARGB8888 = 0,
+    SCE_JPEGENC_PIXELFORMAT_YCBCR420 = 8,
+    SCE_JPEGENC_PIXELFORMAT_YCBCR422 = 9,
+    SCE_JPEGENC_PIXELFORMAT_CSC_ARGB_YCBCR = 16,
```
Macdu, line 38:
```suggestion
    SCE_JPEGENC_PIXEL_RGBA8888 = 0,
    SCE_JPEGENC_PIXEL_BGRA8888 = 4,
    SCE_JPEGENC_PIXEL_YCBCR420 = 8,
    SCE_JPEGENC_PIXEL_YCBCR422 = 9,
    SCE_JPEGENC_PITCH_HW_CSC = 16,
```

Macdu (changes requested):

vita3k/codec/src/mjpeg.cpp
saturnsky: You are correct. However, there is currently no way to know the image format outside of the codec.
In order for the MjpegDecoderState::receive function to return both YUVJ444/YUVJ420, it needs to return the image format of the current input as well.
What would be a good way to approach this?

saturnsky: It is my current idea

bool MjpegDecoderState::receive(uint8_t *data, DecoderSize *size, AVPixelFormat fmt = AV_PIX_FMT_NONE)

If fmt has a value of AV_PIX_FMT_NONE, bypass the decoding result; otherwise, return the converted result according to the format of fmt.
This approach is easily extensible if other return formats are needed in the future.

I want feedback on this approach. If it looks good, I'll approach it this way.

Macdu: First, try not to include FFmpeg types in aheader as that would require linking ffmpeg to the modules library, which is huge and I'd like to avoid it. You can use a boolean or a custom defined enum instead.
Can you tell me in which case the Mjpeg decoder would have to decode to a YUV format that is not the one used for the encoding? I don't see when this is needed.
However what would be needed is to add a function get_output_size (or similar) to MjpegDecoderState, It's not impossible that a game deduces the YUV format of a JPEG using a call to sceJpegGetOutputInfo, using the fact that YUV420/422/444 formats all have different size and the value returned right now is only an upper bound.

However, all of this is for the Jpeg decoder. Correct me if I'm wrong but I don't think it impacts the Jpeg encoder. You can do a first PR for the Jpeg encoder and then an other one to fix the JPEG decoder If you want.

saturnsky: I'll try to find another workaround that doesn't utilize FFMpeg's headers.

Currently, the biggest problem is that the sceJpegDecodeMJpegYCbCr function should return the result as YUV444 even if it decodes a YUV420 encoded Jpeg (at least for the games I've tested).
And within the sceJpegDecodeMJpegYCbCr function, there is no way to tell if the Jpeg file is YUV420 or YUV444. Furthermore, there may be a third type (grayscale Jpeg, YUV422, etc.) that is neither.
Therefore, the sceJpegDecodeMJpegYCbCr function should be able to ask the MJpegDecoder to return the decoding result as YUV444.

Similarly, the sceJpegDecodeMJpeg function needs to return a Jpeg file encoded in multiple formats such as YUV420, YUV444, YUV422, etc. as RGB. Therefore, this function should also be able to ask the MJpegDecoder to return the decoding result in RGB. The existing code is written with the assumption that the Jpeg file will always be YUV420, which can lead to incorrect behavior if a Jpeg file encoded in a different format is received as input.

But as you say, these are all decoder issues. I discovered the decoder's misbehavior during the development of the encoder and wanted to fix it together, but this can be fixed independently.

Therefore, I will revise the PR to include only the encoder implementation, as you suggest, and send a separate PR for the decoder with additional fixes.

saturnsky:
As commented, I've split the PR to include only the fix for SceJpegEncUser.

vita3k/modules/SceJpegEnc/SceJpegEncUser.cpp
```diff
@@ -17,43 +17,143 @@
...
+    if ((context->pixelFormat & SCE_JPEGENC_PIXEL_YCBCR422) == SCE_JPEGENC_PIXEL_YCBCR422) {
+        is_yuv420 = false;
+    } else if ((context->pixelFormat & SCE_JPEGENC_PIXEL_YCBCR420) == SCE_JPEGENC_PIXEL_YCBCR420) {
+        is_yuv420 = true;
+    } else {
+        return SCE_JPEGENC_ERROR_INVALID_PIXELFORMAT;
+    }
+
+    if (inPixelFormat != SCE_JPEGENC_PIXEL_RGBA8888) {
+        return STUBBED("Only ARGB8888 to YCbCr is implemented.");
```
Macdu, line 68:
```suggestion
        return STUBBED("Only RGBA8888 to YCbCr is implemented.");
```

saturnsky:
Oops, I missed that the name change should also change the STUB. Thanks for the detailed check.

Macdu (approved):

PR #2612, opened 2023-05-27
lang: Remove useless words.

HuanJiCanShang:
Vita3K is an emulator, it isn't a system.


Zangetsu38:
Is based on réal psvita, is not problem

PR #2622, opened 2023-06-01
modules/sceJpegUser: Improved YUV format compatibility

saturnsky:
- Improved compatibility for YUV420 and YUV422 formats
- Fix to handle color range of YUVs as full range

I made a separate DecoderColorSpace and JpegColorSpace because I needed to define a color space inside the Decoder to solve problems caused by other color_spaces, such as NV12 issues in video.


vita3k/modules/SceJpeg/SceJpegUser.cpp
```diff
@@ -150,7 +197,7 @@ EXPORT(int, sceJpegMJpegCsc, uint8_t *rgba, const uint8_t *yuv,
     uint32_t width = size >> 16u;
     uint32_t height = size & (~0u >> 16u);
 
-    convert_yuv_to_rgb(yuv, rgba, width, height, false);
+    convert_yuv_to_rgb(yuv, rgba, width, height, convert_color_space_jpeg_to_decoder((SceJpegColorSpace)(SCE_JPEG_COLORSPACE_YUV | sampling)));
```
Macdu, line 200:
```suggestion
    convert_yuv_to_rgb(yuv, rgba, width, height, convert_color_space_jpeg_to_decoder(static_cast<SceJpegColorSpace>(SCE_JPEG_COLORSPACE_YUV | sampling)));
```
Small nitpick, can you use c++ casts instead of c cast?

vita3k/modules/SceJpeg/SceJpegUser.cpp
```diff
@@ -112,14 +152,21 @@ EXPORT(int, sceJpegGetOutputInfo, const uint8_t *jpeg_data, uint32_t jpeg_size,
...
     DecoderSize size = {};
 
     state->decoder->send(jpeg_data, jpeg_size);
     state->decoder->receive(nullptr, &size);
 
     output->width = size.width;
     output->height = size.height;
-    output->color_space = 2u << 16u;
+    output->color_space = convert_color_space_decoder_to_jpeg(state->decoder->get_color_space());
     output->output_size = size.width * size.height * 3;
```
Macdu, line 170: It looks like this value is also wrong. Format looks like it can only be equal to 0 or -1 (otherwise it will return an error). If format is 0, then the output is the rgba size (so size.width * size.height * 4), otherwise (format is -1) it is the YUV size, which is less.

vita3k/codec/include/codec/state.h
```diff
@@ -68,6 +76,7 @@ struct DecoderState {
     virtual bool send(const uint8_t *data, uint32_t size) = 0;
     virtual bool receive(uint8_t *data, DecoderSize *size = nullptr) = 0;
     virtual uint32_t get_es_size();
+    virtual DecoderColorSpace get_color_space();
```
Macdu, line 79: If this is only used by the Mjpeg decoder, you can instead make it a function of the Mjpeg decoder and modify the MjpegState to have a MjpegDecoderState pointer instead of a DecoderState pointer (not like it can be any of type of decoder anyway).

saturnsky: I thought the same function could be used for color space issues in other decoders, but that's just a guess, and I'm sure the implementation details will be different when working with it.
As suggested, I will move this function to the Mjpeg Decoder. If there is a need for a similar function in other decoders, I'll think about it later.

vita3k/modules/SceJpeg/SceJpegUser.cpp
saturnsky: I've updated the code to reflect your advice. There appears to be additional behavior based on mode, but I haven't identified it correctly yet.

Macdu (approved):
lgtm, nice work!

PR #2623, opened 2023-06-02
kernel/sync_primitives: more accurate implementation of message pipes

bookmist:
Not exactly correct implementation, but a bit better. Fix crash in Sword Art Online -Hollow Fragment-


Macdu (commented):
I'm not gonna make any comment because I would end up asking you to reimplement the whole msgpipe. I plan on doing that later.
Have you checked that it does not lead to regression in other games that use msgpipes (like ratchet and clank) ?

Macdu (approved):

bookmist:
>Have you checked that it does not lead to regression in other games that use msgpipes (like ratchet and clank) ?
No, I don't. Thank you for game names, I'll check.

bookmist:
Ratchet and clank seems working, but slow.

PR #2628, opened 2023-06-04
motion: Add motion support

Macdu:
Add a way to emulate the PS Vita gyro / accelerometer using a gamepad gyro / accelerometer.
Not all functions are implemented but the main one are. I only tested it on gravity rush, there may be a few mistakes when converting vectors from the SDL to Vita dimension space.

### DIFF
--- a/vita3k/modules/SceDriverUser/SceMotion.cpp
+++ b/vita3k/modules/SceDriverUser/SceMotion.cpp
@@ -17,6 +17,7 @@
 
 #include "SceMotion.h"
 
+#include <ctrl/state.h>
 #include <motion/functions.h>
 #include <motion/motion.h>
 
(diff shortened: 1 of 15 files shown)


vita3k/modules/SceDriverUser/SceMotion.cpp
```diff
@@ -68,6 +77,18 @@ EXPORT(SceBool, sceMotionGetMagnetometerState) {
...
+    std::lock_guard<std::mutex> guard(emuenv.motion.mutex);
+    sensorState->accelerometer = get_acceleration(emuenv.motion);
+    sensorState->gyro = get_gyroscope(emuenv.motion);
+
+    sensorState->timestamp = emuenv.motion.last_accel_timestamp;
+    sensorState->counter = emuenv.motion.last_counter;
+    sensorState->hostTimestamp = sensorState->timestamp;
+    sensorState->dataInfo = 0;
+
     return UNIMPLEMENTED();
```
EXtremeExploit, line 92: forgor to change this to 0 or its just a stub? :thinking:

ArbestRiAgain:
Does this work on controllers with motion support?

ArbestRiAgain:
Works really well with Gravity Rush and LBPV!

Zangetsu38:
have probleme on flower/flow and wipeout

vita3k/motion/src/motion.cpp
```diff
@@ -19,4 +19,111 @@
...
+    Util::Vec3f gyroscope = state.motion_data.GetGyroscope() * static_cast<float>(2.f * M_PI);
+    return {
+        gyroscope.x,
+        gyroscope.y,
+        gyroscope.z,
+    };
+}
+
+SceFVector3 get_gyro_bias(const MotionState &state) {
+    Util::Vec3f bias = state.motion_data.GetGyroscopeBias();
```
german77, line 45: You converted gyro input into degrees. Shouldn't this be converted as well since the bias value is in radians?

v-fox:
This seems to be intended to be used while holding DualShock parallel to your body which is quite exhausting. When used perpendicularly, with gamepad resting on lap, [counter-]clockwise movement are reversed for x and y axes. Don't remember that happening with motion controls on cemu & yuzu. Maybe there is some kind of auto-switch function or they are designed for that type of hold.

Zangetsu38 (approved):
LGTM

AndeanPuppy:
Hey, apologies for the necropost -- will there be an option to use the device itself as the gyroscope sensor in the Android builds? I'm not sure if it's there because tilting the smartphone by itself with no gamepad other than the touchscreen doesn't do anything in LBPV.

Mahim2000:
Will the PC Version have it as well via. Keyboard Buttons? I was thinking of using the Numpad for Gyro Emulation.

PR #2635, opened 2023-06-09
SceVideodec: Fix resolution of video decode.

Zangetsu38:
# About
- using resolution send by app for decode video
- should fix video resolution in lego game
- remove memset to 0 on last frame in sceAvcdecDecodeStop, should fix green screen regression.

# Result
- fix video broken in lego marvel
image

- fix green screen when video finish
image
image


saturnsky (approved):
It looks good to me.

Macdu (approved):

PR #2636, opened 2023-06-09
external: fix the path to openssl from the extracted archive.

Zangetsu38:
# About
spotted by @german77, and see again by myself, archive downloaded have now diferent name for directory inside, move from openssl-3 to openssl-3.0

- fix cmake gen if using this download zip.


EXtremeExploit (approved):
aight

PR #2637, opened 2023-06-09
vita3k/CMakeLists: fix path of openssl on copy command.

Zangetsu38:
# About:
- fix my forget mistake change also this on path.


EXtremeExploit (approved):
true

PR #2640, opened 2023-06-11
gui: Improve font atlas

Macdu:
- Make the font 4 times more detailed on default settings. They should look a lot less blurry on a high-definition display and the atlas still fits in a 4096x4096 texture.
- Remove the whole Unified ideograms CJC characters from the default configuration. They take more space than everything else combined and should not be here to begin with
- Use ImGui fonctions for korean, simplified and full chinese ranges

### DIFF
--- a/vita3k/gui/src/gui.cpp
+++ b/vita3k/gui/src/gui.cpp
@@ -153,8 +153,18 @@ static void init_style(EmuEnvState &emuenv) {
 static void init_font(GuiState &gui, EmuEnvState &emuenv) {
     ImGuiIO &io = ImGui::GetIO();
 
+    const auto sys_lang = static_cast<SceSystemParamLang>(emuenv.cfg.sys_lang);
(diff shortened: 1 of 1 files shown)


vita3k/gui/src/gui.cpp
```diff
@@ -189,19 +199,6 @@ static void init_font(GuiState &gui, EmuEnvState &emuenv) {
...
-
-    static const ImWchar korean_range[] = {
-        0x3131, 0x3163, // Korean alphabets
-        0xAC00, 0xD79D, // Korean characters
-        0,
-    };
-
-    static const ImWchar chinese_range[] = {
-        0x2000, 0x206F, // General Punctuation
-        0x4e00, 0x9FAF, // CJK Ideograms
```
Croden1999, line 204: we should keep this, it is helpful to Hanzi/Chinese characters (though not all).

imWatchful:
Closing as "resolved"

PR #2655, opened 2023-06-17
modules/SceNet: Stub SceNetResolverCreate with fake id.

Zangetsu38:
# About
- Fix regression on some game when calling this when Control function try look this id inside map


vita3k/modules/SceNet/SceNet.cpp
```diff
@@ -190,7 +190,9 @@ EXPORT(int, sceNetEpollControl, int eid, SceNetEpollControlFlag op, int id, SceN
     if (!epoll) {
         return RET_ERROR(SCE_NET_ERROR_EBADF);
     }
-
+    if (id == emuenv.net.resolver_id) {
+        return 0;
```
bookmist, line 194: I would add STUBBED herу (STUB("Async DNS resolve is not supported");)

Zangetsu38: Done

bookmist (approved):

PR #2658, opened 2023-06-18
renderer: Apply surface sync before signaling notifications

Macdu:
As found out by @saturnsky , surface sync was applied after notifications were signaled, leading to a race condition.

### DIFF
--- a/vita3k/renderer/src/scene.cpp
+++ b/vita3k/renderer/src/scene.cpp
@@ -94,9 +94,16 @@ COMMAND(handle_sync_surface_data) {
 
     const SceGxmNotification vertex_notification = helper.pop<SceGxmNotification>();
     const SceGxmNotification fragment_notification = helper.pop<SceGxmNotification>();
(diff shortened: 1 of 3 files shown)


vita3k/renderer/src/scene.cpp
```diff
@@ -108,9 +115,14 @@ COMMAND(handle_sync_surface_data) {
...
         renderer.notification_ready.notify_all();
-    }
+    };
+
+    if (renderer.disable_surface_sync)
+        // do it as soon as possible
+        signal_notifications();
 
     if (renderer.current_backend == Backend::Vulkan) {
+        ;
```
bookmist, line 125: ?

Zangetsu38: Yes i have already signaled is just forget

Macdu:
This PR is ready to be reviewed, I also fixed another race condition happening when using surface sync with memory mapping.

PR #2662, opened 2023-06-19
ngs: Big refractor and stub all audio modules

Macdu:
This pull request does in practice ... absolutely nothing (except add a few warnings to the log output).
It refractors most of NGS to make it easier and nicer to use with many modules.

It also stubs all the modules and template definitions. So that I can now focus on implementing all the missing audio modules one by one.


bookmist (approved):
LGTM. Good work, a lot of cleanup. (Special thanks for removing logs from sceNgsPatchCreateRouting)

bookmist (approved):

PR #2663, opened 2023-06-20
gui/user management: Small reworks and improvement.

Zangetsu38:
# About:
- gui/user management: Small reworks and improvement.
Remove scroll bar.
automatic scrol in center pos when click on element if is not centered.


EXtremeExploit (approved):
trust the zang

Issue #1411, opened 2021-06-22
Playback rate unimplemented and crackling audio


pent0:
- Audio playback rate in two decoding modules (PCM and AT9) is ignored, results in some games having the Alvin Simon Theodore symptoms (chipmunking).
- Audio that is decoded with AT9 somehow has weird cracklings decoded from FFMPEG. This is unsure if a backend issue or emulator issue.

hobyst:
Testing FFmpeg `.at9` decoding side by side with [VGAudio](https://github.com/Thealexbarney/VGAudio) decoder implementation which is based on [`libatrac9`](https://github.com/Thealexbarney/LibAtrac9) evidences that FFmpeg has timing issues decoding Atrac9 files as it adds silence both at the beginning and the end of decoded files.
![imagen](https://user-images.githubusercontent.com/48522412/123000071-5abceb00-d3af-11eb-88db-a97ac3a0e64a.png)
And not only that, but seems that invalid data is quite common in Atrac9 audio files, as pointed on a [release note from VGAudio](https://github.com/Thealexbarney/VGAudio/releases/tag/v2.2.1) and FFmpeg seems to struggle a lot with this.

Leaving that aside, I think the crackling on audio has more to do with an issue with the audio stream (ex. audio latency due to SDL) and NGS rather than with audio decoding. The HLE audio out module as of today has functions such as [`sceAudioOutSetConfig`](https://docs.vitasdk.org/group__SceAudioUser.html#ga11a4df4d66bfaaf4a50f15d1d4d74a2a) and [`sceAudioOutGetConfig`](https://docs.vitasdk.org/group__SceAudioUser.html#gadfae507c89fda662493734dcb8b51bbd) which allow to check and change audio out port configuration while the port is still open unimplemented.

With all of this I would say that:
- SDL audio has been known for a while to cause several audio latency issues that might eclipse other audio issues and getting rid of it should help to test and refine things on the audio backend. Libraries like [PortAudio](https://github.com/PortAudio/portaudio) should do the job for desktop target platforms while for mobile native audio extensions could be used via C++ (Android NDK) and Swift's interop with C++ (for iOS). This should be fixed the first, as the crackling issues might be at fault of latency on the audio output rather than from Atrac9 decoding.
- FFmpeg Atrac9 decoder has issues and using `libatrac9` might make an improvement.
[...]

PR #2668, opened 2023-06-21
ngs: Save Atrac9 decoder state

Macdu:
Even when decoding a new atrac9 superframe, the atrac9 decoder needs the data from the previous frame to decode the next one. We were using the same atrac9 decoder in ngs for multiple voice, causing the previous frame data to be incorrect when switching between voices and leading to some audio noise when two or more atrac9 voices overlaps.

This PR adds an Atrac9 state saved by each voice to prevent this issue.
This fixes the audio issues in Persona dancing, Dragon's crown and many over games.

### DIFF
--- a/vita3k/codec/include/codec/state.h
+++ b/vita3k/codec/include/codec/state.h
@@ -123,6 +123,10 @@ struct MjpegDecoderState : public DecoderState {
     MjpegDecoderState();
 };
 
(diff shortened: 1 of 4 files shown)


Zangetsu38 (approved):

mdsmendes94:
#1411 also fixed!

Thank you!

PR #2670, opened 2023-06-22
gui/live area: Some small refactor and improvement.

Zangetsu38:
- gui/live area: Some small refactor and improvement.
Add search in navigate by controller/kb.
Fix navigate when press right if have nothing available on it.
Refactor and improve position and size of gate/frames for all style.
Respect aspect ratio for position and size.

# Result
- Compare master/pr for respect aspect ration, for keep good position and size 
image

### DIFF
--- a/vita3k/gui/src/live_area.cpp
+++ b/vita3k/gui/src/live_area.cpp
@@ -57,108 +57,101 @@ struct STR {
 static std::map<std::string, std::vector<FRAME>> frames;
 static std::map<std::string, std::map<std::string, std::vector<STR>>> str;
 static std::map<std::string, std::map<std::string, std::map<std::string, std::map<std::string, std::pair<SceInt32, std::string>>>>> liveitem;
-static std::map<std::string, std::map<std::string, std::map<std::string, std::map<std::string, ImVec2>>>> items;
-static std::map<std::string, std::map<std::string, std::map<std::string, ImVec2>>> items_pos;
+static std::map<std::string, std::map<std::string, std::map<std::string, ImVec2>>> items_size;
 static std::map<std::string, std::map<std::string, std::string>> target;
 static std::map<std::string, std::map<std::string, uint64_t>> current_item, last_time;
 static std::map<std::string, std::string> type;
 static std::map<std::string, int32_t> sku_flag;
 
-void init_live_area(GuiState &gui, EmuEnvState &emuenv, const std::string &app_path) {
-    // Init type
-    if (items_pos.empty()) {
-        // Content manager
-        items_pos["content_manager"]["gate"]["pos"] = ImVec2(620.f, 364.f);
-        // A1
-        items_pos["a1"]["gate"]["pos"] = ImVec2(620.f, 364.f);
-        items_pos["a1"]["frame1"]["pos"] = ImVec2(900.f, 415.f);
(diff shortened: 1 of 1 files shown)


Macdu:
Add an option to stretch the livearea to be like it was before.

Zangetsu38:
> Add an option to stretch the livearea to be like it was before.

?? for what  ? is not stretch, is just totaly broken render, i set this for fix big problem, i no want broken it again

Zangetsu38:
for explain
that is pr and master
image

on the master it's not just stretched
it is stretched and deformed in all directions depending on the resolution of the window
which makes everything deformed and just horrible

i pr it for can fix this very old issue, and get all time perfect result in any resolution

exactly like works render during playing

Macdu:
When playing it is important to keep the aspect ratio correct. When navigating in the menus, it depends. That's a lot of empty space on the screen which does not necessarily look nice on some screens. Add an option to keep the live area stretched to fit the whole screen.

Zangetsu38:
> When playing it is important to keep the aspect ratio correct. When navigating in the menus, it depends. That's a lot of empty space on the screen which does not necessarily look nice on some screens. Add an option to keep the live area stretched to fit the whole screen.

But, I just explained that I wouldn't touch it, think about it, if one day we arive emulated the vsh, it will be EXACTLY the same as my pr.

It makes no sense to add an option that would break everything again, it's stupid, I've already planned to change everything to do it like that I told you, it's already the case on the sweat and on the home eh.

All the live area and the rest is planned to work in an aspect ratio of 960x544
Like in games, if you touch this, it deforms everything and there's nothing left in the right place or the right size
You can clearly see on my screen that it's a disaster...
So no, there will be no options, it makes no sense

and for infol, it's not a classic gui, it's a reproduction of the psvita interface, so using icons and images that are made for psvita

So yes, gui here is SAME as when you playing

Macdu:
But I personally find it ugly to have to keep the PS Vita aspect ratio for the menu on a non 16:9 screen. That's why I want the option to stretch the screen. This is not emulation, that's an UI made using imgui so most of the UI can scale.

Also on your exemple it's indeed bad, but you took an extreme exemple. For a 20 or 21:9 aspect ratio, I think stretching would the images and using correct imgui code for the rest almost like what is done currently would look better

Zangetsu38:
Exactly, you haven't seen the result if my phone or tablet, with the live area which has a horrible result, or it's all stretched is distorted in height or width, it's not more beautiful, it's just terrible to have everything deformed.

So yes this is an extreme example, my pc screen is split in 2 with the 2 stretched in max width, 0 but just tell yourself that even less stretched like on my phone or tablet, the result is really ugly, and that I started wanting to fix everything just for that

even if I had already wanted to do it more than 3 years ago, but I didn't know how to do it at the time...

So no, is not horrible, is juste better

Macdu:
I don't want to force it to be stretched, I just want an option to allow it to be.

Zangetsu38:
Live area on my Phone is horrible, all is deformed, text on widget is wrong place and i very need fix also info bar, is just shit here
Screenshot_20230622_153134_Vita3K

EXtremeExploit (approved):
trust the zang part 3

EXtremeExploit:
:NanaThink:

PR #2672, opened 2023-06-22
modules/sceJpegEncUser: Improve JPEG encoding more strictly

saturnsky:
- Fixes issue where sceJpegEncoderEncode would generate a still MJPEG instead of a JPEG
- Implement the behavior of sceJpegEncoderSetCompressionRatio

### DIFF
--- a/vita3k/codec/src/mjpeg.cpp
+++ b/vita3k/codec/src/mjpeg.cpp
@@ -16,9 +16,11 @@
 // 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
 #include <codec/state.h>
+#include <codec/types.h>
 
 extern "C" {
 #include <libavcodec/avcodec.h>
+#include <libavcodec/bsf.h>
 #include <libavutil/imgutils.h>
 #include <libswscale/swscale.h>
 }
@@ -130,7 +132,7 @@ void convert_rgb_to_yuv(const uint8_t *rgba, uint8_t *yuv, uint32_t width, uint3
     assert(error == height);
 }
 
-int convert_yuv_to_jpeg(const uint8_t *yuv, uint8_t *jpeg, uint32_t width, uint32_t height, uint32_t max_size, const DecoderColorSpace color_space) {
+int convert_yuv_to_jpeg(const uint8_t *yuv, uint8_t *jpeg, uint32_t width, uint32_t height, uint32_t max_size, const DecoderColorSpace color_space, int32_t compress_ratio) {
     AVPixelFormat format = AV_PIX_FMT_YUV444P;
     int strides_divisor = 1;
     int slice_position = 8;
@@ -158,6 +160,9 @@ int convert_yuv_to_jpeg(const uint8_t *yuv, uint8_t *jpeg, uint32_t width, uint3
     context->pix_fmt = format;
     context->time_base.num = 1;
     context->time_base.den = 25;
+    context->flags |= AV_CODEC_FLAG_QSCALE;
+
+    context->qmin = 1;
 
     int ret = avcodec_open2(context, codec, NULL);
     assert(ret >= 0);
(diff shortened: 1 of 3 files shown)


Zangetsu38 (approved):

vita3k/codec/src/mjpeg.cpp
```diff
@@ -168,6 +173,7 @@ int convert_yuv_to_jpeg(const uint8_t *yuv, uint8_t *jpeg, uint32_t width, uint3
     frame->format = context->pix_fmt;
     frame->width = context->width;
     frame->height = context->height;
+    frame->quality = FF_QP2LAMBDA * ((compress_ratio) * (31 - 1) / 255 + 1);
```
Macdu, line 176: FF_QP2LAMBDA is a bit arbitrary (related to H.263 quality), is there a reason you didn't use FF_LAMBDA_MAX? Does it compress the image too much with the default compression?

saturnsky: Unlike the annotation, FFMPEG's FF_QP2LAMBDA is used to convert FFMPEG's qscale option to internal value without distinguishing between encodings.
(and many encoders that don't need the value multiplied by FF_QP2LABMDA, divide that value by FF_QP2LAMBDA again).

https://github.com/FFmpeg/FFmpeg/blob/af3f12078563d8119bd59771793994f651842ff2/fftools/ffmpeg_mux_init.c#L1219-L1222

Current code attempts to mimic the behavior of the FFMPEG CLI's qscale option. MJPEG's qscale is in the range of 1-31, and a 1 would be treated the same as a 2.
However, FFMPEG is overly complicated, so it is possible that my understanding is wrong.

Macdu: Ok, thank you for the explanation.

Macdu:
Btw, is there a reason you encode the yuv to MJPEG then extract the JPEG frame from it and not encode it directly to JPEG. If this article is right https://www.loc.gov/preservation/digital/formats/fdd/fdd000063.shtml, MJPEG forces the encoded format to be YUV 422 which is not the case for JPEG.

Also about the quality, according to this https://stackoverflow.com/questions/10225403/how-can-i-extract-a-good-quality-jpeg-image-from-a-video-file-with-ffmpeg, a good quality is between 2 and 5 while the default quality for SceJpegEnc is 64, giving a value of 8, which looks quite high.

saturnsky:
> Btw, is there a reason you encode the yuv to MJPEG then extract the JPEG frame from it and not encode it directly to JPEG.
As far as I know, FFMPEG does not support encoding directly to JPEG, only to still MJPEG. If I want an image to be converted to JPEG, I will have to do an additional MJPEG to JPEG conversion (but FFMPEG is very complex, so my knowledge is possibly wrong).
To avoid this, I would have to use libjpeg, but I also wanted to avoid adding more dependencies to the project, I thought it would be more better idea to stick with FFMPEG.
Encoding methods for JPEG and still MJPEG are essentially the same, with only header differences between the two, there is no additional loss in converting MJPEG to JPEG if the implementation is correct, so this is probably not be an issue.

> If this article is right https://www.loc.gov/preservation/digital/formats/fdd/fdd000063.shtml, MJPEG forces the encoded format to be YUV 422 which is not the case for JPEG.

The lack of a standard for MJPEG despite it being a widely used video encoding. Some MJPEG implementations only use YUV422, others support other image formats such as YUV420, and the PS Vita's MJPEG implementation appears to support both YUV422 and YUV420 encoding.
FFMPEG's MJPEG implementation also supports YUV420 encoding, so this isn't be a problem.

> Also about the quality, according to this https://stackoverflow.com/questions/10225403/how-can-i-extract-a-good-quality-jpeg-image-from-a-video-file-with-ffmpeg, a good quality is between 2 and 5 while the default quality for SceJpegEnc is 64, giving a value of 8, which looks quite high.

This can be problematic. I simply linearly mapped the PS Vita's compress level to FFMPEG's qscale, which may have been the wrong decision. There is a possibility that an option that assumes sufficient image quality on PS Vita could produce a poor quality image in the current implementation.
It would be good to come up with a better mapping heuristic for this.

Macdu:
Ok, thank you for the explanation. Indeed, I did not see anything older than JPEG2000 for jpeg encoding on ffmpeg and adding a new dependency is not really worth it.

Macdu (approved):

PR #2674, opened 2023-06-22
modules/SceNpTrophy: Fix platinum trophy unlock.

Zangetsu38:
# About
- modules/SceNpTrophy: Fix platinum trophy unlock.
Unlock platinum trophy compare if trophies count is same as total unlocked, but that's not possible, because trophy counting includes platinum.
So need remove 1 from the trophy count to check that only one is missing.


EXtremeExploit (approved):
alr

PR #2676, opened 2023-06-23
modules/SceCommonDialog: Fix allocate EmptyParam for slot upper 0.

Zangetsu38:
# About 
- modules/SceCommonDialog: Fix allocate EmptyParam for slot upper 0.
Empty param is not only in slot 0, is for all slot available, so move on vector.
Should fix all missing new save icon.
- gui/common dialog/save dialog: fix position of text in new save.
Fix pos and the size of the elements regardless of the size of the window.
Fix scroll bar missing.

# Result
- master 
image

- pr
image


EXtremeExploit (approved):
trust the zang part 2

PR #2679, opened 2023-06-24
gui/live area: Some small refactor and improvement.

Zangetsu38:
# About
- gui/live area: Some small refactor and improvement.
Add search in navigate by controller/kb.
Fix navigate when press right if have nothing available on it.
Refactor and improve position and size of gate/frames for all style.
Respect proportion for position and size of item.
Remove border of window for live area/User manager.


Macdu (changes requested):
Please add an option to keep the old behavior.

Zangetsu38:
> Please add an option to keep the old behavior.

i have already explain, you no ask option for keep old behavior
here you want option for disable my fix, have no sense, i FIX one probleme in gui.
i have already apply save fix in home and in user and more, no have sense want this option here.

i repeat again i fix proportion for no get deformed and keep good size/place

but if you no want black font when is outside viewport and only get blue in full screen is possible

Zangetsu38:
> Please add an option to keep the old behavior.

is done by @nishinji

Macdu (approved):

PR #2684, opened 2023-06-26
mem: Register signal handler also to SIGBUS on macOS

KhoraLee:
On macOS, accessing PROT_NONE region raises SIGBUS not SIGSEGV. So registers same signal handler also to SIGBUS.

This will fix crash on PCSB01099 on macOS.
image

### DIFF
--- a/vita3k/mem/src/mem.cpp
+++ b/vita3k/mem/src/mem.cpp
@@ -646,6 +646,13 @@ static void register_access_violation_handler(AccessViolationHandler handler) {
     if (sigaction(SIGSEGV, &sa, NULL) == -1) {
         LOG_CRITICAL("Failed to register an exception handler");
     }
+#ifdef __APPLE__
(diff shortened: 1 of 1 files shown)


Macdu (dismissed):

vita3k/mem/src/mem.cpp
```diff
@@ -646,6 +646,12 @@ static void register_access_violation_handler(AccessViolationHandler handler) {
     if (sigaction(SIGSEGV, &sa, NULL) == -1) {
         LOG_CRITICAL("Failed to register an exception handler");
     }
+#ifdef __APPLE__
+    // When accessing PROT_NONE region on macOS, it is rasing SIGBUS not SIGSEGV. So apply same signal_handler to SIGBUS
```
Macdu, line 650:
```suggestion
    // When accessing PROT_NONE region on macOS, it is raising SIGBUS not SIGSEGV. So apply same signal_handler to SIGBUS
```
small typo

PR #2699, opened 2023-07-01
modules/SceLibKernel: implement SceIoRename.

Zangetsu38:
# About
- Implment SceIoRename
Tested in both case, rename in same folder, and rename to other folder, is fine.

### DIFF
--- a/vita3k/io/src/io.cpp
+++ b/vita3k/io/src/io.cpp
@@ -619,6 +619,47 @@ int remove_file(IOState &io, const char *file, const std::wstring &pref_path, co
     return 0;
 }
 
+int rename(IOState &io, const char *old_name, const char *new_name, const std::wstring &pref_path, const char *export_name) {
+    auto device = device::get_device(old_name);
+    if (device == VitaIoDevice::_INVALID) {
+        LOG_ERROR("Cannot find device for path: {}", old_name);
+        return IO_ERROR(SCE_ERROR_ERRNO_ENOENT);
+    }
+
+    const auto translated_old_path = translate_path(old_name, device, io.device_paths);
+    if (translated_old_path.empty()) {
+        LOG_ERROR("Cannot translate path: {}", translated_old_path);
+        return IO_ERROR(SCE_ERROR_ERRNO_ENOENT);
+    }
+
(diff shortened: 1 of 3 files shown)


vita3k/io/src/io.cpp
```diff
@@ -619,6 +619,47 @@ int remove_file(IOState &io, const char *file, const std::wstring &pref_path, co
...
+        LOG_ERROR("File does not exist at path: {} (target path: {})", emulated_old_path.string(), old_name);
+        return IO_ERROR(SCE_ERROR_ERRNO_ENOENT);
+    }
+
+    const auto emulated_new_path = device::construct_emulated_path(device, translated_new_path, pref_path, io.redirect_stdio);
+
+    LOG_TRACE_IF(log_file_op, "{}: Renaming file {} to {} ({} to {})", export_name, old_name, new_name, emulated_old_path.string(), emulated_new_path.string());
+
+    boost::system::error_code error_code{};
+    fs::detail::rename(emulated_old_path, emulated_new_path, &error_code);
```
Macdu, line 652:
```suggestion
    fs::rename(emulated_old_path, emulated_new_path, &error_code);
```

fs::detail is not portable (boost only)

Macdu: Other part of the code using it will have to be changed.

Zangetsu38: ah i see hmm but i no can check error code so ? 🤔

Macdu: image
You should be able to

Zangetsu38: ah yes.. i hgave no understund, it said me no can, i remove, and after i re set and said is ok, visual studio drink

Zangetsu38: ah no çok, it no want i using &error_code but pass with just error_code

Zangetsu38: ok, so done

Macdu (approved):
Make sure for the CI to be done before merging, just in case.

Zangetsu38:
oki 👌

PR #2706, opened 2023-07-02
gui: Add option to allow disable display compile shaders dialog.

Zangetsu38:
# About
Request by lot user
- Add option in setting dialog for can disable display dialog of compile ;shaders

### DIFF
--- a/vita3k/gui/src/settings_dialog.cpp
+++ b/vita3k/gui/src/settings_dialog.cpp
@@ -741,6 +741,10 @@ void draw_settings_dialog(GuiState &gui, EmuEnvState &emuenv) {
         ImGui::Checkbox("Texture Cache", &emuenv.cfg.texture_cache);
         if (ImGui::IsItemHovered())
             ImGui::SetTooltip("Uncheck the box to disable texture cache.");
+        ImGui::SameLine();
+        ImGui::Checkbox("Show Compile Shaders", &emuenv.cfg.show_compile_shaders);
+        if (ImGui::IsItemHovered())
+            ImGui::SetTooltip("Uncheck the box to disable the display of the compile shaders dialog.");
         ImGui::Separator();
         const auto perfomance_overley_size = ImGui::CalcTextSize("Performance overlay").x;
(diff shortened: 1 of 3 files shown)


vita3k/gui/src/settings_dialog.cpp
```diff
@@ -741,6 +741,10 @@ void draw_settings_dialog(GuiState &gui, EmuEnvState &emuenv) {
         ImGui::Checkbox("Texture Cache", &emuenv.cfg.texture_cache);
         if (ImGui::IsItemHovered())
             ImGui::SetTooltip("Uncheck the box to disable texture cache.");
+        ImGui::SameLine();
+        ImGui::Checkbox("Show Compile Shaders", &emuenv.cfg.show_compile_shaders);
+        if (ImGui::IsItemHovered())
+            ImGui::SetTooltip("Uncheck the box to disable display of compile shaders dialog.");
```
Macdu, line 747:
```suggestion
            ImGui::SetTooltip("Uncheck the box to disable the display of compile shaders dialogs.");
```
I'm not entirely sure about this.

Zangetsu38: is wrong, is only one dialog ^^

Zangetsu38: but i have pushed fix typo for the missing

Macdu: It's one dialog which shows up multiple time so both are right I guess. If it's only one dialog it should be "display of the compile shader dialog."  I think.

Zangetsu38: is because it show count of shaders :) so shaers but dialog
one dialog for few shaders :)

Macdu: Ah sorry, that was a typo, I did not meant to tell you to replace shaders by shader.

Macdu (approved):

vita3k/gui/src/settings_dialog.cpp
Zangetsu38: no problem :)

Issue #2707, opened 2023-07-02
[macOS] Updater: Update fails when there is a space in the path


shinra-electric:
I keep Vita3k in `/Applications/Games/Emulators/Playstation Vita/`. 

When trying to use the updater, it fails: 

[Vita3K Updater Fail.txt](https://github.com/Vita3K/Vita3K/files/11928904/Vita3K.Updater.Fail.txt)

It is due to the space in the path name. `update-vita3k.sh` doesn't handle spaces correctly.

shinra-electric:
Fixed by #2708

PR #2708, opened 2023-07-02
[Needs Testing] Fix macOS updater

shinra-electric:
I just add a few quotation marks to try to get `update-vita3k.sh` to recognise spaces in the path. 

Needs testing on macOS, especially when there are spaces in the path to the app bundle. 

Shouldn't affect Windows or Linux


shinra-electric:
Should hopefully fix #2707

Macdu:
@KhoraLee Does this look fine for you?

KhoraLee (approved):
@Macdu I've tested and it works. So looks good to go.

PR #2709, opened 2023-07-03
modules/SceDriverUser/SceAppMgrUser: Stub sceAppMgrReceiveEventNum.

Zangetsu38:
# About
after reverse in ghidra, see this function send enventNum with value particular very big, causing big freeze in very long loop of call recieve event function
image

So Accorded with @CreepNT , event is not implemented to Vita3K, so just set 0 for can skip it and stop freeze

### DIFF
--- a/vita3k/modules/SceDriverUser/SceAppMgrUser.cpp
+++ b/vita3k/modules/SceDriverUser/SceAppMgrUser.cpp
@@ -522,9 +522,13 @@ EXPORT(int, sceAppMgrReceiveEvent) {
     return UNIMPLEMENTED();
 }
 
-EXPORT(int, sceAppMgrReceiveEventNum) {
-    TRACY_FUNC(sceAppMgrReceiveEventNum);
-    return UNIMPLEMENTED();
+EXPORT(int, sceAppMgrReceiveEventNum, SceUInt32 *eventNum) {
+    TRACY_FUNC(sceAppMgrReceiveEventNum, eventNum);
+
+    // Vita3K does not yet manage events
+    *eventNum = 0;
+
+    return STUBBED("Set eventNum to 0");
 }
 
 EXPORT(int, sceAppMgrReceiveNotificationRequestForShell) {


vita3k/modules/SceDriverUser/SceAppMgrUser.cpp
```diff
@@ -517,14 +517,16 @@ EXPORT(int, sceAppMgrPspSaveDataRootMount) {
     return UNIMPLEMENTED();
 }
 
-EXPORT(int, sceAppMgrReceiveEvent) {
-    TRACY_FUNC(sceAppMgrReceiveEvent);
+EXPORT(int, sceAppMgrReceiveEvent, SceAppMgrEvent *mgrEvent) {
+    TRACY_FUNC(sceAppMgrReceiveEvent, mgrEvent);
```
Macdu, line 521: Is this needed, and shouldn't an error be returned in this case, as setting the event num to 0 in sceAppMgrReceiveEventNum means this function does not have any event to receive

vita3k/modules/SceAppMgr/SceAppMgr.h
```diff
@@ -29,6 +29,21 @@ enum SceAppMgrErrorCode {
     SCE_APPMGR_ERROR_BGM_PORT_BUSY = 0x80803000 //!< BGM port was occupied and could not be secured
 };
 
+
+enum SceAppMgrEventType {
```
Macdu, line 33: Isn't SceAppMgrSystemEventType exactly this?

Zangetsu38: is little diiferent, is @CreepNT have send me this enum and struct

vita3k/modules/SceDriverUser/SceAppMgrUser.cpp
Zangetsu38: yes, you right, i have just set it because is same change but is not obligatory for now

Zangetsu38:
i have add comment and add arg for tracy, i forget it

PR #2716, opened 2023-07-06
renderer/vulkan: Implement shader interlock

Macdu:
Implement shader interlock to emulate more accurately programmable blending on desktop GPUs.
This PR is overly complicated for the single reason that Vulkan does not allow a surface to be used both as a color attachment and a storage image (and barrier for this not supported case do not work).
So I have to duplicate all the vulkan structures (render pass, framebuffer...) for them to work with shader interlock.

This has been tested with a NVidia GPU on Windows and an AMD GPU on linux with a dev mesa build.

Without
image

With
image

I also improved the shader cache to be cleared when using a different GPUs as shader will now most likely not be compatible anymore.
There shouldn't be any graphical regressions, but shader interlock may have a moderate performance impact.

This should fix the 'boxes' issues caused by geometry overlapping in a single draw in all games which had it.

### DIFF
--- a/vita3k/renderer/include/renderer/state.h
+++ b/vita3k/renderer/include/renderer/state.h
@@ -77,6 +77,9 @@ struct State {
         const GxmState &gxm, MemState &mem)
         = 0;
     virtual void swap_window(SDL_Window *window) = 0;
+    virtual uint32_t get_device_id() {
+        return 0;
+    }
     // return a bitmask with the Filter enum values of the supported enum filters
     virtual int get_supported_filters() = 0;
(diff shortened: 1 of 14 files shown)


Zangetsu38 (approved):

ArbestRiAgain:
This will also fix the text overlap issue in LBPV

nn9dev:
Will an option to toggle this be added? From reports in the Discord, some games seem to suffer a fairly substantial performance hit (10-15% decrease in FPS).

Macdu:
Yes, I'll do it. Although the 10%-15% decrease is only if the GPU performance is the bottleneck, which is really rarely the case.

PR #2721, opened 2023-07-08
modules/SceGxm: Handle calling sceGxmExecuteCommandList with empty immediate list

Macdu:
SceMidSceneFlush empties the command list, and sceGxmExecuteCommandList  does not except the command list to be empty. handle this case.

This allows Zanki zero to reach intro (was previously nothing).


EXtremeExploit (approved):
ok

Zangetsu38 (approved):

PR #2725, opened 2023-07-09
modules/SceAppMgr: partial Impelment sceAppMgrGameDataMount for app path.

Zangetsu38:
# About
- modules/SceAppMgr: Impelment sceAppMgrGameDataMount.
create first virtual mountpoint of Appmgr. 
- regmgr: small change.


Zangetsu38:
reworks pr, remove other path, is more complicated i have thinks, it can using this 3 path in same time
for njow i have only see app path used, so do it later when it try using other

PR #2726, opened 2023-07-09
renderer/vulkan: Emulate macroblock tile sync

Macdu:
Emulate macroblock tile sync basically by stopping and starting again the recording at every draw when the render target has the macroblock flag. This is accurate but kind of slow, I'll implement soon something better.


Zangetsu38 (dismissed):

Zangetsu38 (dismissed):

Zangetsu38 (approved):

Degerz:
VK_EXT_attachment_feedback_loop_layout might be a better way to express self dependency barriers between each draws.

Zangetsu38 (approved):

PR #2728, opened 2023-07-09
regmgr: Move value from string to vector char.

Zangetsu38:
# About
- regmgr: Move value from string to vector char.
reworks GetBin
- Stub sceRegMgrGetKeysInfo.

### DIFF
--- a/vita3k/regmgr/src/regmgr.cpp
+++ b/vita3k/regmgr/src/regmgr.cpp
@@ -135,8 +135,8 @@ void init_reg_template(RegMgrState &regmgr, const std::string &reg) {
     }
 }
 
+static constexpr uint32_t spaceSize = 32;
 static uint32_t get_space_size(const uint32_t str_size) {
-    const uint32_t spaceSize = 32;
     const uint32_t line = 16;
 
     return spaceSize > str_size ? spaceSize - str_size : ((str_size / line) + 1) * line - str_size;
@@ -149,7 +149,7 @@ static bool load_system_dreg(RegMgrState &regmgr, const std::wstring &pref_path)
         char buf = 0;
 
         // Skip the space
-        for (uint32_t i = 0; i < 33; i++) {
+        for (uint32_t i = 0; i < (spaceSize + 1); i++) {
             file.read(&buf, 1);
         }
 
@@ -189,10 +189,9 @@ static bool load_system_dreg(RegMgrState &regmgr, const std::wstring &pref_path)
 
                 // Read the value
                 const uint32_t value_size = entry.size;
-                std::vector<char> value(value_size);
+                auto &value = regmgr.system_dreg[reg.first][entry.name];
+                value.resize(value_size);
                 file.read(value.data(), value_size);
-                const auto value_str = std::string(value.begin(), value.end());
-                regmgr.system_dreg[reg.first][entry.name] = value_str;
 
                 // Skip the space
                 const auto diff_value = get_space_size(value_size);
@@ -203,8 +202,7 @@ static bool load_system_dreg(RegMgrState &regmgr, const std::wstring &pref_path)
         }
 
         file.close();
-    } else
-        LOG_WARN("Registry file not found: {}, attempting to create it", system_dreg_path.string());
+    }
 
     return !regmgr.system_dreg.empty();
 }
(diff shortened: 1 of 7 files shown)


vita3k/regmgr/src/regmgr.cpp
```diff
@@ -271,49 +268,57 @@ static void init_system_dreg(RegMgrState &regmgr, const std::wstring pref_path)
 
     for (const auto &reg : reg_template) {
         for (const auto &entry : reg.second) {
-            regmgr.system_dreg[reg.first][entry.name] = entry.value;
+            const std::vector<char> default_value(entry.value.begin(), entry.value.end());
+            auto &value = regmgr.system_dreg[reg.first][entry.name];
+            value.resize(entry.size);
+            value = default_value;
```
Macdu, line 274:
```suggestion
            auto &value = regmgr.system_dreg[reg.first][entry.name];
            value.assign(entry.value.begin(), entry.value.end());
```

vita3k/regmgr/src/regmgr.cpp
```diff
@@ -271,49 +268,57 @@ static void init_system_dreg(RegMgrState &regmgr, const std::wstring pref_path)
...
+    regmgr.system_dreg[fix_category(category)][name].assign(data, data + bufSize);
 
     save_system_dreg(regmgr, pref_path);
 }
 
 int32_t get_int_value(RegMgrState &regmgr, const std::string &category, const std::string &name) {
     std::lock_guard<std::mutex> lock(regmgr.mutex);
 
-    return std::stoi(regmgr.system_dreg[fix_category(category)][name]);
+    const auto sys = regmgr.system_dreg[fix_category(category)][name];
```
Macdu, line 297:
```suggestion
    const auto& sys = regmgr.system_dreg[fix_category(category)][name];
```

vita3k/regmgr/src/regmgr.cpp
```diff
@@ -271,49 +268,57 @@ static void init_system_dreg(RegMgrState &regmgr, const std::wstring pref_path)
...
+
+    const auto str_value = std::to_string(value);
+    regmgr.system_dreg[fix_category(category)][name].assign(str_value.begin(), str_value.end());
 
     save_system_dreg(regmgr, pref_path);
 }
 
 std::string get_str_value(RegMgrState &regmgr, const std::string &category, const std::string &name) {
     std::lock_guard<std::mutex> lock(regmgr.mutex);
+    const auto sys = regmgr.system_dreg[fix_category(category)][name];
```
Macdu, line 314:
```suggestion
    const auto& sys = regmgr.system_dreg[fix_category(category)][name];
```

vita3k/regmgr/src/regmgr.cpp
Zangetsu38: for why ?? value is just read, no need & 🤔

vita3k/regmgr/src/regmgr.cpp
Zangetsu38: forget this idea :) i have try is shit and get crazy shit inside file
only works fine like this

Zangetsu38: and ho, it very not like too if you set value inside vector without resize
resize is request before assign too

Zangetsu38: for more explain
that is when i using this code
image
i have no used it for this reason

Zangetsu38: ok is done, sorry i have bad read, but just important, all is empty here, so is obligatory to resize for using assign
if not get big crash

vita3k/regmgr/src/regmgr.cpp
Zangetsu38: No sur of utility but done

vita3k/regmgr/src/regmgr.cpp
Zangetsu38: same

vita3k/regmgr/src/regmgr.cpp
saturnsky: @Zangetsu38 The original code takes extra execution time to allocate memory for the sys variable and copy its value. Macdu's modification uses a reference, which means that the variable will only be read, so it doesn't need to be copied like the original.

Zangetsu38: oh, i see

Macdu (approved):

PR #2735, opened 2023-07-11
ctrl: Implement button rebinds

nn9dev:
This PR is an attempt at implementing rebinds for controller buttons. Binding buttons to sticks (which, I imagine would also be requested by some) is not implemented here.

As this is my first (serious) commit to this project, feedback is greatly appreciated since I'm almost certain some of this is sub-par.


nn9dev:
I have once again gotten ahead of myself. This PR is not ready, as it does not work in-game. I understand where the offending code is, and will take another stab at it in the following days. I would still appreciate criticism on the rest of it, though.

bookmist:
It would be good to make a spelling fixes in separated commit (or separated PR)

Zach-star1:
It would also be very cool to have the possibility to swap bumpers with triggers, especially in games that make heavy use of the L1/R1 buttons :)

Issue #2733, opened 2023-07-10
Trails in the Sky FC Evolution [PCSG00488] red screen when entering/leaving a battle/building


othmanead:
Game has a red screen you get whenever you transition between one scene to another (like existing a room/building) or exiting/entering a battle, here is an example of it
![image](https://github.com/Vita3K/Vita3K/assets/106798110/990fe154-e7f3-43dd-9604-61684bdd8562)

joshua0705:
This issue still exists in the latest version, I'm not sure how to fix it. (Win11, i9-13900HX, RTX 4060; I've tried change OpenGL or Vulkan mode, but no difference, still exists)

joshua0705:
![image](https://github.com/Vita3K/Vita3K/assets/38355432/f03f18ef-f2e6-4ad7-a266-75cc56d2b2c6)

srw20166:
same issue

tsukina22:
Is not better now? I played and haven't see this issue so far?

shlgf:
Using steam deck still have this issue.

PR #2736, opened 2023-07-11
renderer/vulkan: Improve u1u5u5u5 swizzle

Macdu:
U1U5U5U5 texture swizzle is not fully supported right now. However we are using an argb texture when rendering to it (the rgba version is not compatible with all GPUs). Therefore it is more accurate to use a argb swizzle.

This fixes the screen transition in trails games.


othmanead:
fixes https://github.com/Vita3K/Vita3K/issues/2733

Zangetsu38 (approved):

joshua0705:
Sorry, I don't quite understand as a player, what I should do to fix this bug, as I'm confronting a same one. Could you please give me a more detailed guide? Thx!

joshua0705:
image

PR #2741, opened 2023-07-11
regmgr: Add more id on get category and name by id.

Zangetsu38:
# About
- regmgr: Add more id on get category and name by id.
Fix crash when category/name is empty.
Fix crash when category/name is empty.
Fix init regmgr when change pref path.
Fix crash if value is not int for get int.
Add Set default value if value is corrupted.
Using vector for reg category template for keep order instead unordered map for fix broken file when switch pref path directory.

### DIFF
--- a/vita3k/regmgr/src/regmgr.cpp
+++ b/vita3k/regmgr/src/regmgr.cpp
@@ -22,7 +22,7 @@
 
 namespace regmgr {
 
-const std::array<unsigned char, 16> xorKey = { 0x89, 0xFA, 0x95, 0x48, 0xCB, 0x6D, 0x77, 0x9D, 0xA2, 0x25, 0x34, 0xFD, 0xA9, 0x35, 0x59, 0x6E };
+constexpr std::array<unsigned char, 16> xorKey = { 0x89, 0xFA, 0x95, 0x48, 0xCB, 0x6D, 0x77, 0x9D, 0xA2, 0x25, 0x34, 0xFD, 0xA9, 0x35, 0x59, 0x6E };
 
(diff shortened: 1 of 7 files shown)


Zangetsu38:
Is ready for review

vita3k/regmgr/src/regmgr.cpp
```diff
@@ -273,62 +278,117 @@ static void init_system_dreg(RegMgrState &regmgr, const std::wstring pref_path)
...
+    const auto &sys = regmgr.system_dreg[cat][name];
+    auto value_str = std::string(sys.begin(), sys.end());
+    if (!std::all_of(value_str.begin(), value_str.end(), [](unsigned char c) { return std::isdigit(c) || (c == '\0'); })) {
+        LOG_ERROR("Invalid value for {}{}: {}, attempt using default value!", cat, name, value_str);
+        value_str = set_default_value(regmgr, cat, name);
+        if (value_str.empty())
+            return 0;
+    }
 
     return std::stoi(value_str);
```
Macdu, line 355: Use a try catch block around std::stoi looking for std::invalid_argument instead, that would be much cleaner.

Zangetsu38: For here no, using try catch juste get succès or no, and here if first char of value is number that works, but is wrong...
Because if string is 0rue it pass and succes std::stoi with take just 0
Here my code check every char , for can reset any wrong int value set in file
For exemple: if vakue is 0ru it detect is not number and reset value by default

Macdu (approved):

PR #2747, opened 2023-07-15
modules: Describe and partial implememt SceRtabi module

bookmist:
used by driver_us.suprx
Return value of `__rtabi_ldivmod` is very unusual but it seems correct.
Behavior on division by zero is taken from original disasmed code.
original names of functions are `__aeabi_*` but they was renamed to not mix with same functions in other module with other nids

### DIFF
--- a/vita3k/module/src/write_return_value.cpp
+++ b/vita3k/module/src/write_return_value.cpp
@@ -37,6 +37,18 @@ void write_return_value(CPUState &cpu, uint64_t ret) {
     write_reg(cpu, 1, ret >> 32);
 }
 
+void write_return_value(CPUState &cpu, std::div_t ret) {
+    write_reg(cpu, 0, ret.quot);
+    write_reg(cpu, 1, ret.rem);
+}
+
+void write_return_value(CPUState &cpu, std::lldiv_t ret) {
+    write_reg(cpu, 0, ret.quot & UINT32_MAX);
+    write_reg(cpu, 1, ret.quot >> 32);
+    write_reg(cpu, 2, ret.rem & UINT32_MAX);
+    write_reg(cpu, 3, ret.rem >> 32);
+}
+
 void write_return_value(CPUState &cpu, bool ret) {
     write_reg(cpu, 0, ret);
 }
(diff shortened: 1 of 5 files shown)


vita3k/module/include/module/write_return_value.h
```diff
@@ -21,10 +21,22 @@
...
 
+struct uint_uint {
+    uint32_t value_01;
+    uint32_t value_02;
+};
+
+struct uint64_uint64 {
+    uint64_t value_01;
+    uint64_t value_02;
+};
```
Macdu, line 32: Where are these used?

bookmist: Nowhere. tnx.

vita3k/module/src/write_return_value.cpp
```diff
@@ -37,6 +37,18 @@ void write_return_value(CPUState &cpu, uint64_t ret) {
...
+void write_return_value(CPUState &cpu, std::div_t ret) {
+    write_reg(cpu, 0, ret.quot);
+    write_reg(cpu, 1, ret.rem);
+}
+
+void write_return_value(CPUState &cpu, std::lldiv_t ret) {
+    write_reg(cpu, 0, ret.quot & UINT32_MAX);
+    write_reg(cpu, 1, ret.quot >> 32);
+    write_reg(cpu, 2, ret.rem & UINT32_MAX);
+    write_reg(cpu, 3, ret.rem >> 32);
```
Macdu, line 49: The `& UINT32_MAX` is not necessary. If you decide to put it, also add it to `ret.quot >> 32` and `ret.rem >> 32` as their upper bits will be set if they are negative (they are signed integer).

bookmist: There is the same code upper for int64_t and uint64_t. So change nowhere or everywhere?

Macdu: Ah indeed, oh well I guess even though this is slightly misleading, the code is correct.

Macdu (approved):

PR #2749, opened 2023-07-17
utils: implement stoi_def and use it instead of std::stoi

bookmist:


vita3k/gui/src/information_bar.cpp
```diff
@@ -122,7 +123,7 @@ static bool set_notice_info(GuiState &gui, EmuEnvState &emuenv, const NoticeList
         init_notice_icon(gui, emuenv, content_path / "sce_sys/icon0.png", info);
     } else {
         auto common = gui.lang.common.main;
-        switch (static_cast<np::trophy::SceNpTrophyGrade>(std::stoi(info.group))) {
+        switch (static_cast<np::trophy::SceNpTrophyGrade>(string_utils::stoi_def(info.group, 0, "trophy group"))) {
```
Zangetsu38, line 126: no sur for 0 here, 0 litteraly no exist in SceNpTrophyGrade is mini 1 to 4

bookmist: so we go to default branch of next switch statement. and `name` variable is empty afaik. But you know trophies better, so what value should be here?

Zangetsu38: is fine i thinks, but maybe add log for default: 🤔 but hmm anyway, go merge sorry

PR #2753, opened 2023-07-17
net: Some implement some rare net funcs

EXtremeExploit:
* Implement both htonll and ntohll, some platforms/systems dont have the syscall, so we have to check for ourselves the endianness and do the operation on our side
* ethernet addr translation implemented (sceNetEtherNtostr, sceNetEtherStrton), sceNetEtherStrton is good as long as the input string is not too short or has some unexpected stuff, but that just doesnt happen, so should still be good
* add some param checks sceNetInit <:
~~smol pr~~

### DIFF
--- a/vita3k/modules/SceNet/SceNet.cpp
+++ b/vita3k/modules/SceNet/SceNet.cpp
@@ -17,6 +17,7 @@
 
 #include "SceNet.h"
 
+#include <cstdio>
 #include <kernel/state.h>
 #include <net/functions.h>
 #include <net/state.h>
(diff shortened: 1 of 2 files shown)


Macdu:
You can assume the system is little endian. Vita3K would go very far anyway on a big endian system

vita3k/modules/SceNet/SceNet.cpp
```diff
@@ -258,14 +258,37 @@ EXPORT(Ptr<int>, sceNetErrnoLoc) {
...
+    if (!n || !str)
+        return RET_ERROR(SCE_NET_ERROR_EINVAL);
+
+    if (0x11 < len) {
+        snprintf(str, len, "%02x:%02x:%02x:%02x:%02x:%02x",
+            n->data[0], n->data[1], n->data[2], n->data[3], n->data[4], n->data[5]);
+        return 0;
+    }
+    return RET_ERROR(SCE_NET_ERROR_EINVAL);
 }
```
bookmist, line 275: code flow looks something strange for me
```suggestion
    if (len <=0x11) 
        return RET_ERROR(SCE_NET_ERROR_EINVAL);
        
    snprintf(str, len, "%02x:%02x:%02x:%02x:%02x:%02x",
        n->data[0], n->data[1], n->data[2], n->data[3], n->data[4], n->data[5]);
    return 0;
```

vita3k/modules/SceNet/SceNet.cpp
```diff
@@ -258,14 +258,37 @@ EXPORT(Ptr<int>, sceNetErrnoLoc) {
...
+
+    if (!n || !str)
+        return RET_ERROR(SCE_NET_ERROR_EINVAL);
+
+    if (0x11 < len) {
+        snprintf(str, len, "%02x:%02x:%02x:%02x:%02x:%02x",
+            n->data[0], n->data[1], n->data[2], n->data[3], n->data[4], n->data[5]);
+        return 0;
+    }
+    return RET_ERROR(SCE_NET_ERROR_EINVAL);
```
bookmist, line 274: code flow looks something strange for me
```suggestion
    if (len <=0x11) 
        return RET_ERROR(SCE_NET_ERROR_EINVAL);
        
    snprintf(str, len, "%02x:%02x:%02x:%02x:%02x:%02x",
        n->data[0], n->data[1], n->data[2], n->data[3], n->data[4], n->data[5]);
    return 0;
```

EXtremeExploit: should be better now

vita3k/modules/SceNet/SceNet.cpp
```diff
@@ -258,14 +258,34 @@ EXPORT(Ptr<int>, sceNetErrnoLoc) {
...
+        return RET_ERROR(SCE_NET_ERROR_ENOTINIT);
+
+    if (!str || !n)
+        return RET_ERROR(SCE_NET_ERROR_EINVAL);
+
+    for (int i = 0; i < sizeof(n->data); i++) {
+        const char *currStr = str + (i * 3);
+        printf("currStr = %s\n", currStr);
+        n->data[i] = std::stoul(currStr, nullptr, 16);
+    }
```
Macdu, line 286: You can use sscanf for this

EXtremeExploit: now using sscanf :D

bookmist (approved):

Macdu (approved):

PR #2759, opened 2023-07-19
modules/SceKernelThreadMgr: Implement sceKernelOpenEventFlag.

Zangetsu38:


vita3k/kernel/src/sync_primitives.cpp
```diff
@@ -1358,6 +1368,25 @@ SceUID eventflag_create(KernelState &kernel, const char *export_name, SceUID thr
...
+    const std::lock_guard<std::mutex> kernel_lock(kernel.mutex);
+
+    const auto it = std::find_if(kernel.eventflags.begin(), kernel.eventflags.end(), [=](const auto &eventflag) {
+        return strncmp(eventflag.second->name, pName, KERNELOBJECT_MAX_NAME_LENGTH) == 0;
+    });
+
+    if (it != kernel.eventflags.end())
+        return it->first;
+
+    return RET_ERROR(SCE_KERNEL_ERROR_UNKNOWN_EVF_ID);
```
Macdu, line 1387: Are you sure this error is right ?

Zangetsu38: Yes is same error other, it return unknow (type) id

Macdu: I'm quite sure it would return SCE_KERNEL_ERROR_UID_CANNOT_FIND_BY_NAME

Zangetsu38: Hmm i have juste used error already pressent for o’e already here 🤔 but maybe yes

Zangetsu38: i have done and changed for all

Macdu (approved):

PR #2763, opened 2023-07-21
gui: allow circle for confirm in app close dialogue

nishinji:
i hope this is the last :D

### DIFF
--- a/vita3k/interface.cpp
+++ b/vita3k/interface.cpp
@@ -591,12 +591,30 @@ bool handle_events(EmuEnvState &emuenv, GuiState &gui) {
             app::switch_state(emuenv, !emuenv.kernel.is_threads_paused());
         }
     };
+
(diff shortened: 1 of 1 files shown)


vita3k/interface.cpp
```diff
@@ -591,12 +591,28 @@ bool handle_events(EmuEnvState &emuenv, GuiState &gui) {
...
+                cancel();
+            else
+                confirm();
+            break;
+        case SCE_CTRL_CROSS:
+            if (emuenv.cfg.sys_button == 1)
+                confirm();
+            else
+                cancel();
+        }
```
Zangetsu38, line 615: you have forget break
and in end switch need using default: break;

nishinji: done

Zangetsu38 (approved):

PR #2764, opened 2023-07-21
net: Fix max epoll nfd number on wait

EXtremeExploit:
turns out that the `nfds` param needs to be the highest fd PLUS ONE, 
linux says this in its documentation, but on windows its missing this tiny detail, i've seen other people use select and having it only work if adding 1 to nfds, so its the same as linux
https://man7.org/linux/man-pages/man2/select.2.html
this fixes a forever waiting/loading on some homebrews with epoll waits (idk about commercial games)


Macdu (approved):
Lgtm, although according to the doc you linked, this function is not recommended and epoll_wait should be used instead. As this is for implementing the Epoll::wait function, this would kind of make sense.

EXtremeExploit:
true, ig i could do that once we find a game that uses more than a thousand sockets and this becomes a problem, but considering not many games even use networking at all i think this wont be a problem

PR #2813, opened 2023-07-25
interface, kernel: Fixes

Macdu:
Fix a regression caused by the fact that preloaded modules were not registered, causing some game to see them unloaded and try to load them again, causing a freeze.

Also put the thread callbacks at the right location (run_callback can only be called in the thread and run_guest_function is not appropriate here).

Also some fixes, there is no reason to pass the kernel state to all the thread functions.

This fixes regression in spiderman.

### DIFF
--- a/vita3k/interface.cpp
+++ b/vita3k/interface.cpp
@@ -493,23 +493,28 @@ static ExitCode load_app_impl(SceUID &main_module_id, EmuEnvState &emuenv, const
     const auto is_app = fs::exists(module_app_path) && !fs::is_empty(module_app_path);
     std::vector<std::string> lib_load_list = {};
     // todo: check if module is imported
-    auto add_preload_module = [&](uint32_t code, const std::string &name, bool load_from_app) {
-        if ((process_preload_disabled & code) == 0 && is_lle_module(name, emuenv)) {
-            if (load_from_app)
-                lib_load_list.emplace_back(fmt::format("app0:sce_module/{}.suprx", name));
-            else
-                lib_load_list.emplace_back(fmt::format("vs0:sys/external/{}.suprx", name));
+    auto add_preload_module = [&](uint32_t code, SceSysmoduleModuleId module_id, const std::string &name, bool load_from_app) {
+        if ((process_preload_disabled & code) == 0) {
+            if (is_lle_module(name, emuenv)) {
+                if (load_from_app)
+                    lib_load_list.emplace_back(fmt::format("app0:sce_module/{}.suprx", name));
+                else
+                    lib_load_list.emplace_back(fmt::format("vs0:sys/external/{}.suprx", name));
+            }
+
+            if (module_id != SCE_SYSMODULE_INVALID)
+                emuenv.kernel.loaded_sysmodules.push_back(module_id);
         }
     };
-    add_preload_module(0x00010000, "libc", is_app);
-    add_preload_module(0x00020000, "libdbg", false);
-    add_preload_module(0x00080000, "libshellsvc", false);
-    add_preload_module(0x00100000, "libcdlg", false);
-    add_preload_module(0x00200000, "libfios2", is_app);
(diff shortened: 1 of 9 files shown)


vita3k/interface.cpp
```diff
@@ -493,23 +493,28 @@ static ExitCode load_app_impl(SceUID &main_module_id, EmuEnvState &emuenv, const
...
-    add_preload_module(0x00100000, "libcdlg", false);
-    add_preload_module(0x00200000, "libfios2", is_app);
-    add_preload_module(0x00400000, "apputil", false);
-    add_preload_module(0x00800000, "libSceFt2", false);
-    add_preload_module(0x01000000, "libpvf", false);
-    add_preload_module(0x02000000, "libperf", false); // if DEVELOPMENT_MODE dipsw is set
+    add_preload_module(0x00010000, SCE_SYSMODULE_INVALID, "libc", is_app);
+    add_preload_module(0x00020000, SCE_SYSMODULE_DBG, "libdbg", false);
+    add_preload_module(0x00080000, SCE_SYSMODULE_INVALID, "libshellsvc", false);
+    add_preload_module(0x00100000, SCE_SYSMODULE_INVALID, "libcdlg", false);
```
Zangetsu38, line 512: With invalide ? 🤔

bookmist (approved):

Macdu:
@bookmist I fixed a bug with the thread callbacks, it should now be ready.

vita3k/kernel/src/thread.cpp
```diff
@@ -156,52 +156,79 @@ int ThreadState::start(KernelState &kernel, SceSize arglen, const Ptr<void> &arg
...
     returned_value = static_cast<uint32_t>(status);
 }
 
-void ThreadState::exit_delete(KernelState &kernel, bool exit) {
-    if (exit && kernel.thread_event_end) {
-        int ret = run_callback(kernel.thread_event_end.address(), { SCE_KERNEL_THREAD_EVENT_TYPE_END, static_cast<uint32_t>(id), 0, kernel.thread_event_end_arg });
-        if (ret != 0)
-            LOG_WARN("Thread end event handler returned {}", log_hex(ret));
-    }
+void ThreadState::exit_delete(bool exit) {
```
bookmist, line 175: Why we need this `exit` flag now? I think we should not run end callback twice or don't run at all.

vita3k/kernel/src/thread.cpp
```diff
@@ -156,52 +156,79 @@ int ThreadState::start(KernelState &kernel, SceSize arglen, const Ptr<void> &arg
...
-void ThreadState::exit(KernelState &kernel, SceInt32 status) {
+void ThreadState::exit(SceInt32 status) {
     if (kernel.thread_event_end) {
         int ret = run_callback(kernel.thread_event_end.address(), { SCE_KERNEL_THREAD_EVENT_TYPE_END, static_cast<uint32_t>(id), 0, kernel.thread_event_end_arg });
         if (ret != 0)
             LOG_WARN("Thread end event handler returned {}", log_hex(ret));
     }
 
     std::lock_guard<std::mutex> guard(mutex);
+    run_end_callback = true;
```
bookmist, line 170: We already run end callback. Why run twice?

Macdu: Sorry forgot to delete the run_callback above.

vita3k/kernel/src/thread.cpp
Macdu: thread->exit_delete is also called by sceKernelThreadDelete, but this shouldn't call the callback a second time

Macdu:
From what I understand, the start callback should only be called if a thread is started using sceKernelThreadStart and the end callback only if sceKernelThreadExit or sceKernelThreadExitDelete is called. It may be needed to check it to be sure in the future if necessary.

bookmist (approved):

PR #2815, opened 2023-07-26
cpu: Remove unicorn fallback for dynarmic

Macdu:
A few years ago, this may have been useful. However, now dynarmic is more accurate and feature-complete compared to unicorn. If we reach an invalid opcode, it means for 99% of the cases that we did something wrong previously and even in the 1% left switching to unicorn won't help.
Also it produces a huge unreadable backlog.


bookmist (approved):

nn9dev:
Is it on purpose that this doesn't completely gut unicorn from the repository? I imagine so, but personally I don't know what else it's used for that would warrant keeping it around. Just curious is all.

Macdu:
Well that's not the goal of this PR.
But indeed it should be considered if Unicorn should be removed from the project as Dynarmic is better in almost every way now.

PR #2818, opened 2023-07-27
kernel, modules: Improve nid performance

Macdu:
Resolving the NID during an HLE (or first LLE) call takes a non-negligeable amount of time (on android the current implementation was taking almost 10-15% of many threads processing time), here are two performance improvements (the first being already used on Android):
- using a shared-mutex is not a good idea because of its implementation, it shouldn't be used to lock something which is quite fast (like nid resolving) as it amounts basically to two consecutive mutex locks. Replace it by a standard mutex, it is in practise much faster.
- Use a faster data structure for the nid map: a boost::unordered_flat_map is around 3 times faster than a std::unordered_map.

Also updates the boost requirement to 1.81.0 as boost::unordered_flat_map was added at this version.

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -51,8 +51,9 @@ jobs:
 
       - name: Set up build environment (ubuntu-latest)
         run: |
+          sudo add-apt-repository -y ppa:mhier/libboost-latest
           sudo apt update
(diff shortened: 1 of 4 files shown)


Zangetsu38 (dismissed):

Macdu:
Ok, this PR should be ready. I disabled the hard requirement for boost 1.81 and instead if a boost version lower than 1.81 is being used, it will fallback to std::unordered_map.
I also modify the CI for linux to use boost 1.81.

Zangetsu38 (approved):

PR #2823, opened 2023-07-29
modules/SceSysmodule: Implement sceSysmoduleIsLoadedInternal.

Zangetsu38:


bookmist:
All modules can be checked with this function, not only system. But there are some module flags checks in disasmed code. (is it game program, is it system program, is it development mode)

Zangetsu38:
Is see 🤔

Zangetsu38:
> All modules can be checked with this function, not only system. But there are some module flags checks in disasmed code. (is it game program, is it system program, is it development mode)

Changed, good like this

bookmist:
Good. It would be good to also add check if internal_module_id is valid (-module_id between 0x01 and 0x29)

Zangetsu38:
done

bookmist (approved):

PR #2830, opened 2023-07-31
gui: allow stretching the display area

nishinji:


Zangetsu38:
Stretch render no have sense, that just full broken render and totally déform texture

you need found way for using game in other aspect ration with exemple hack eboot
one user jave already done this

Macdu:
That's an option that is disabled by default, and some people prefer to have it stretched, seem fine to me.

Zangetsu38:
> That's an option that is disabled by default, and some people prefer to have it stretched, seem fine to me.

hmm if you said...
but using stretch just deforme all in render, is obligatory bad with get very ugly result

for that i speak need found way for apply patch in eboot, for can using diferent aspect ratio

Macdu:
Yes, I agree that together with this option, the best for a game would be to be patched to support the output ratio.

Zangetsu38:
yes, in current stater, juust get all deformed etired in up or down and so ugly things lol

vita3k/renderer/include/renderer/state.h
```diff
@@ -88,6 +89,9 @@ struct State {
     void set_surface_sync_state(bool disable) {
         disable_surface_sync = disable;
     }
+    void set_display_area(bool disable) {
```
Macdu, line 92: this function name is not really appropriate, something like set_stretch_display would be better.

nishinji: exactly

vita3k/app/src/app_init.cpp
```diff
@@ -74,7 +74,13 @@ void update_viewport(EmuEnvState &state) {
     if (h > 0) {
         const float window_aspect = static_cast<float>(w) / h;
         const float vita_aspect = static_cast<float>(DEFAULT_RES_WIDTH) / DEFAULT_RES_HEIGHT;
-        if (window_aspect > vita_aspect) {
+        if (state.cfg.stretch_the_display_area) {
+            // Match the aspect ratio to the screen size.
+            state.viewport_size.x = static_cast<SceFloat>(w);
+            state.viewport_size.y = static_cast<SceFloat>(h);
+            state.viewport_pos.x = 0;
+            state.viewport_pos.y = 0;
```
Zangetsu38, line 82: hmm i have juste see here, i have problem with this, you can found way for no touching it ?
for my gui code i want can keep viewport_(pos/size) like here also with enable this option

Zangetsu38: maybe create other variable for gui only with same code current 🤔 for separe gui/render

Zangetsu38 (changes requested):
after testing and viewing in my side, this commit broken all my stuff of gui for all dialog when i have apply fix with using viewport size and pos for can all time keep in in good position/size.
Get instant all broken again and shit result, look save dialog if you stretch render, is full broken...
No touch of viewport_size and pos, need found solution for no touching it

PR #2833, opened 2023-07-31
kernel: Implement sceClibMemcpy_safe

EXtremeExploit:
some games use HLE libc, the ones that do sometimes use the safe variant, and the safe variant is just not a simple memcpy, it does some checks to alert memory overlaps and has breakpoints for games that are in the development process, so i decided to replicate this, also technically the memcpy still happens whatever the overlap is, should be 1:1 to real hw (afaik), mainly pr just in case a game actually has memory overlaps due to bugs on emu side and other devs dont pull their hairs out because of memory corruption. should have no effect for game compatibility

### DIFF
--- a/vita3k/modules/SceLibKernel/SceLibKernel.cpp
+++ b/vita3k/modules/SceLibKernel/SceLibKernel.cpp
@@ -41,6 +41,7 @@
 #include <util/tracy.h>
 
 #include <algorithm>
+#include <cmath>
 #include <cstdlib>
 
 enum class TimerFlags : uint32_t {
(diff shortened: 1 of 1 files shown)


vita3k/modules/SceLibKernel/SceLibKernel.cpp
```diff
@@ -164,9 +167,30 @@ EXPORT(int, sceClibMemcpyChk) {
...
 
-EXPORT(Ptr<void>, sceClibMemcpy_safe, Ptr<void> dst, const void *src, SceSize len) {
+EXPORT(Ptr<void>, sceClibMemcpy_safe, Ptr<void> dst, const Ptr<void> src, SceSize len) {
     TRACY_FUNC(sceClibMemcpy_safe, dst, src, len);
-    memcpy(dst.get(emuenv.mem), src, len);
+    /*
+    should be 1:1 to real hw, asserts are in the same position as decompiled code,
+    still call memcpy when the breakpoint happens just like in real hw, should practically
+    be the same as just calling memcpy, but with a bit more checks for the developers,
+    asserts do get annoying if the game is tend to do memcpy on overlapping pointers, which is weird,
```
Macdu, line 176:
```suggestion
    asserts do get annoying if the game tends to do memcpy on overlapping pointers, which is weird,
```

Macdu (approved):
Don't forget to squash when merging.

PR #2836, opened 2023-08-02
renderer/vulkan: Fix vertex explosion on Mac

Macdu:
This fixes the vertex explosion when using non-multiple of 4 stride with MoltenVK, was caused by a typo.


EXtremeExploit (approved):
true

PR #2843, opened 2023-08-07
mem: Improve memory issue when page size is not 4KiB

KhoraLee:
On devices that have larger page size than 4KiB(such as aarch64 macOS), sometimes page is overlapped when allocating at specific page(`alloc_at`) and caused memory issue because overlapped page's memory is zero-fiiled. So if region contains already allocated page then it will only allocate remain parts and update allocation table's info to largest one.

 If there is better way then I'll close this PR.

### DIFF
--- a/vita3k/mem/include/mem/state.h
+++ b/vita3k/mem/include/mem/state.h
@@ -25,6 +25,11 @@
 #include <mutex>
 #include <set>
 
+#if defined(__APPLE__) && defined(__aarch64__)
+#define PAGE_NOT_4KIB
+typedef std::map<Address, std::pair<uint32_t, uint32_t>, std::greater<Address>> ForceAllocPage;
+#endif
+
 struct AllocMemPage {
     uint32_t allocated : 4;
     uint32_t size : 28;
@@ -78,4 +83,10 @@ struct MemState {
     bool use_page_table = false;
     PageTable page_table;
     std::map<uint64_t, MemExternalMapping, std::greater<uint64_t>> external_mapping;
+
+#ifdef PAGE_NOT_4KIB
+    ForceAllocPage force_alloc;
+    BitmapAllocator seg_allocator;
+    uint32_t multiplier;
+#endif
 };
--- a/vita3k/mem/src/mem.cpp
+++ b/vita3k/mem/src/mem.cpp
@@ -103,7 +103,10 @@ bool init(MemState &state, const bool use_page_table) {
     memset(state.alloc_table.get(), 0, sizeof(AllocMemPage) * table_length);
 
     state.allocator.set_maximum(table_length);
-
+#ifdef PAGE_NOT_4KIB
+    state.multiplier = state.page_size / KiB(4);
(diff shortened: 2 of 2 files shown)


Macdu:
I think it would be better to make the whole allocator use blocks which are the host page size instead of trying to use workaround like that.

KhoraLee:
> I think it would be better to make the whole allocator use blocks which are the host page size instead of trying to use workaround like that.

I don't understand what you mean. Would you describe it a bit more?

Macdu:
> > I think it would be better to make the whole allocator use blocks which are the host page size instead of trying to use workaround like that.
> 
> I don't understand what you mean. Would you describe it a bit more?

Ah, I kind of see, so the issue is that some games are trying to allocate two contiguous memory segment with the second one not being 8KB-aligned, right?

KhoraLee:
> Ah, I kind of see, so the issue is that some games are trying to allocate two contiguous memory segment with the second one not being 8KB-aligned, right?

For example, on Uncharted 
```
[16:32:25.181] |D| [load_self]: Loading SELF at app0:eboot.bin... (ELF type: 0xfe00, self_filesize: 0x325b, self_offset: 0x4, module_info_offset: 0x4a4)
[16:32:25.181] |D| [load_self]:     [LOAD] (p_type: 0x1): p_offset: 0x1000, p_vaddr: 0x81000000, p_paddr: 0x0, p_filesz: 0x838, p_memsz: 0x838, p_flags: 0x5, p_align: 0x1000
[16:32:25.181] |D| [load_self]:     [LOAD] (p_type: 0x1): p_offset: 0x2000, p_vaddr: 0x81001000, p_paddr: 0x0, p_filesz: 0xbc, p_memsz: 0xc0, p_flags: 0x6, p_align: 0x1000
[16:32:25.181] |C| [alloc_inner]: Failed to allocate at specific page
```

```
4KiB Page
┌──────────────────────┬──────────────────────┐
│    Seg0    │ Padding │    Seg1    │ Padding │
└──────────────────────┴──────────────────────┘
├──── Page 528,384 ────┼──── Page 528,385 ────┤

16KiB Page
┌───────────────────────────────────────────────...──┐
│    Seg0    │ Padding │    Seg1    │ Padding │ ...
└───────────────────────────────────────────────...──┘
├────────────────── Page #132,096 ──────────────...──┤

```
On 4KiB page size, seg0 and seg1 are allocated in different page.
On 16KiB, Seg0 and Seg1 is in same page and it is calling `alloc_inner` with forcing same page, Seg0 region is zero-filled when allocating Seg1 and caused memory issue.

KhoraLee:
@macdu How about current(93e9e5b4df089147207fa2754d781d48819a3fee) implementation?

Macdu:
Given that your fix only affects aarch64 mac, which for the time being does not work because of the underlying problem, I am fine with it.

EXtremeExploit:
Superseded by https://github.com/Vita3K/Vita3K/pull/3862

PR #2847, opened 2023-08-10
ci: add macOS universal build

KhoraLee:
This will add additional build of universal macOS binary which can run natively on both Apple silicon and Intel-based macOS.

TODO:
- [ ] Make FFMPEG binaries as universal binary
- [ ] Build unicorn library on CI not using pre-built universal binary


Macdu:
A universal build is a nice idea. However, the longer compile time is kind of an issue. I think the MacOs compiler cache (ccache / scache) in the CI should be fixed before merging this PR.

Macdu:
Also it would be better not to add prebuilts to the main repo. The best would be to do like for Windows and download the prebuilts only when necessary (about unicorn)

KhoraLee:
> Also it would be better not to add prebuilts to the main repo. The best would be to do like for Windows and download the prebuilts only when necessary (about unicorn)

I'm aware about unicorn. But how about FFMPEG? It is downloading all prebuilts regardless of OS.
Also I saw android version of Vita3K uses Unicorn 2, so if you don't mind could you upstream Unicorn 2 things?

Macdu:
> > Also it would be better not to add prebuilts to the main repo. The best would be to do like for Windows and download the prebuilts only when necessary (about unicorn)
> 
> I'm aware about unicorn. But how about FFMPEG? It is downloading all prebuilts regardless of OS.
> Also I saw android version of Vita3K uses Unicorn 2, so if you don't mind could you upstream Unicorn 2 things?

I was only talking about unicorn here. For ffmpeg I'll need to change later the way prebuilts are downloaded. It is not a concern for this PR.

About unicorn, it is not included in the Android builds as it is GPL. But indeed when I first tried building Vita3K for Android, I was using Unicorn and updated to Unicorn 2 in the process. I'll upstream the changes.

KhoraLee:
Now it builds unicorn while CI process. But due to Unicorn 1 bug, it will not work on arm64 macOS machines.

CI run time will be about 20~25 min

Macdu:
I'll do the PR to update to unicorn2 today or tomorrow.

shinra-electric:
Is this blocked by anything anymore? 

A HiDPI PR was merged recently in #3473, and Unicorn was updated in #3477

KhoraLee:
We still need to deal with 4KiB page size things.
Also, I haven’t been able to get Unicorn working yet. (This can be resolved by enabling only dynarmic, like on Android.)

PR #2848, opened 2023-08-10
gui/settings dialog: Fix refresh option stretch the display area.

Zangetsu38:
# About
- Fix allow Apply change stretch display area state in real time.
Fix refresh module list when open setting dialog.

### DIFF
--- a/vita3k/gui/src/settings_dialog.cpp
+++ b/vita3k/gui/src/settings_dialog.cpp
@@ -18,6 +18,7 @@
 #include "imgui.h"
 #include "private.h"
 
+#include <app/functions.h>
 #include <audio/state.h>
 #include <config/functions.h>
 #include <config/state.h>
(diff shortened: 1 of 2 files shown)


vita3k/gui/src/settings_dialog.cpp
```diff
@@ -228,6 +229,8 @@ void init_config(GuiState &gui, EmuEnvState &emuenv, const std::string &app_path
         config.ngs_enable = emuenv.cfg.ngs_enable;
         config.psn_status = emuenv.cfg.psn_status;
     }
+    get_modules_list(gui, emuenv);
```
Macdu, line 232: What does this have to do with the PR?

vita3k/renderer/include/renderer/state.h
```diff
@@ -89,8 +89,8 @@ struct State {
     void set_surface_sync_state(bool disable) {
         disable_surface_sync = disable;
     }
-    void set_stretch_display(bool disable) {
-        stretch_the_display_area = disable;
+    void set_stretch_display(bool state) {
+        stretch_the_display_area = state;
```
Macdu, line 93: Give it a meaningful name like stretch (also in the header) and not state

vita3k/gui/src/settings_dialog.cpp
Zangetsu38: Is fix

vita3k/renderer/include/renderer/state.h
Zangetsu38: State is perfect, why stretch ? Is strange
Is bool indicate litteraly state ?

Macdu: state is perfect? If I call set_stretch_display with state = false, are you able to deduce what it does? Does it stretch the display or not?

Zangetsu38: I no understund ? 
State false = disable ? So is clear 🤔 I no understund what you wabt said here 🤷‍♂️

Zangetsu38: done

Macdu (approved):

PR #2853, opened 2023-08-13
renderer: Add virtual destructor to base filter class

EXtremeExploit:
mainly to get rid of a warning about `delete called on 'ScreenFilter' that is abstract but has non-virtual destructor`


Macdu:
Don't add it yet, there are some issues to be fixed with filter destruction and it can cause a crash.

EXtremeExploit:
alrighty

PR #2858, opened 2023-08-16
ci: fix bad ccache hit rate on macOS

KhoraLee:
macOS CI's ccache hit rate was bad(about 0.00%) because Compiler mtime was changed on every CI run. So changing ccache's compiler check from mtime to content will make cache work on macOS.

I did 5 CI run on my forked repo and it took average 10 min to finish. Surprisingly, on all 5 run macOS finished faster than Windows.

First CI run after this PR merged will take time because previous cache is not usable.


Macdu:
Nice, thank you very much!

Macdu (approved):

PR #2866, opened 2023-08-18
config & gui/settings dialog: Add Option to display system apps.

Zangetsu38:
- config & gui/setting dialogs: Add Option to display system apps.
For can allow disable disable display system apps from the home screen to the main menu bar only.


Macdu (approved):
nice

PR #2878, opened 2023-08-22
gui/initial setup: Fix respect aspect ratio.

Zangetsu38:
- gui/initial setup: Fix respect aspect ratio.
Improve size of column of select language.
Save config any time press on next.
Add two gui setting.

### DIFF
--- a/vita3k/gui/src/initial_setup.cpp
+++ b/vita3k/gui/src/initial_setup.cpp
@@ -17,6 +17,7 @@
 
 #include "private.h"
 
+#include <config/functions.h>
 #include <config/state.h>
 #include <gui/functions.h>
 #include <host/dialog/filesystem.hpp>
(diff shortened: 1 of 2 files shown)


Macdu:
The two settings have absolutely nothing to do with aspect ratio?

Zangetsu38:
> The two settings have absolutely nothing to do with aspect ratio?

No, i juste adding it in same time, no need separe commit if change is inside same file

Macdu:
Tell why would I expect a commit whose name is "Fix aspect ratio" to add config options to the initial setup?

Also why would the initial setup need to be 16:9? It looks nothing like the PS Vita initial setup anyway, is some simple text, buttons and scrolling list and this just reduces the size for the initial setup gui and add black bars to the side.

Zangetsu38:
> Tell why would I expect a commit whose name is "Fix aspect ratio" to add config options to the initial setup?
> 
> Also why would the initial setup need to be 16:9? It looks nothing like the PS Vita initial setup anyway, is some simple text, buttons and scrolling list and this just reduces the size for the initial setup gui and add black bars to the side.

Eu ? Is litteraly writed in second line

And same other... i want all respect psvita size, is no complicated here...
My all gui code writed is full based on interface of pscita for all, so is full normal
No idea why you have problèm with just respect original proportion of élément for keep good render and proper
Currently is broken, i fix it, that all

PR #2899, opened 2023-09-06
renderer: Rewrite texture cache

Macdu:
Rewrite most of the rendering API agnostic texture cache (except texture uploading):
- Put most of the format-related functions in the texture_format.cpp file
- Replace the C-style inheritance with proper C++ classes and inheritance.
- Use 64 bits hashes, no reason to truncate them to 32 bits
- Last but not least, replace the linear lookup when using a texture (and finding the least recently used texture) with a hashmap lookup when looking for a texture and a doubly linked list to always have a pointer to the least recently used texture.

The last point optimizes a function called thousands of times per second from a O(N) complexity to O(1) (well the O(1) contains a hashmap lookup so it's not completely negligeable but still a lot faster) for N=1024 when the cache is full (the time for it to happen depends on the game).

This should fix the emulator slowing down after a while (mostly noticeable on devices with weak CPUs) as the texture cache gets more and more full and texture lookups were taking more and more time.


Zangetsu38 (approved):

Macdu:
This PR should be ready

Zangetsu38 (approved):

PR #2900, opened 2023-09-06
vita3k: specify minimum version of the target on macOS

KhoraLee:
Due to discord game SDK and homebrew, possible macOS minimum target version is macOS 11, but CI is automatically setting target version to 12.6 which is OS version of CI environment. So it will manually specify minimum target version for macOS.

But if homebrew remove support for big_sur bottles, we have to change target version corresponding to brew bottles, or we have to build OpenSSL and boost ourselves.

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -45,7 +45,11 @@ jobs:
 
       - name: Set up build environment (macos-latest)
         run: |
-          brew install boost ccache ninja molten-vk openssl create-dmg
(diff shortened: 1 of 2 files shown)


Macdu:
Can the current ffmpeg binaries be used with this PR or do you need to update the ffmpeg repositories?
If possible I'd like not to update the ffmpeg repository before #2847 is merged.

Macdu:
Ok, thank you for the explanation, that's what I thought.

Macdu (approved):

PR #2901, opened 2023-09-07
external: switch to unicorn2

Macdu:
Update unicorn to unicorn2.
Benefits:
- About 5 times faster performance-wise (still a lot slower than dynarmic)
- On windows, build statically instead of using mingw prebuilts. Allow 3 dlls to be removed from the final zip.

I had to implement cache invalidation for Vita3K to work with unicorn2 (even though I don't understand how unicorn1 was working without cache invalidation after bookmist's changes anyway).

We are still using a custom repository for this as I had to do some modifications to unicorn CMakeList for it to build with our CI.

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -150,37 +150,6 @@ if(WIN32)
 	target_link_libraries(winsock INTERFACE WSOCK32 WS2_32 IPHLPAPI)
 endif()
 
-if(WIN32)
-	set(UNICORN_VER "4b52942b")
-
-	set(UNICORN_PREFIX "${CMAKE_BINARY_DIR}/external/unicorn")
-
-	if(NOT EXISTS "${UNICORN_PREFIX}")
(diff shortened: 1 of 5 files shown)


Zangetsu38 (dismissed):

external/CMakeLists.txt
```diff
@@ -200,6 +169,9 @@ if(MSVC)
 	endif()
 endif()
 
+set(UNICORN_ARCH "arm")
+add_subdirectory(unicorn EXCLUDE_FROM_ALL)
+
```
bookmist, line 174: I don't know if it's needed, but if this add new project into solution then please add `set_property(TARGET unicorn PROPERTY FOLDER externals)` to put it into `externals` folder in MSVC

Zangetsu38 (approved):

external/CMakeLists.txt
Zangetsu38: Ah yes, good catch, i have forget it

PR #2905, opened 2023-09-08
modules/SceGxm, renderer: Improve accuracy of depth stencil surface

Macdu:
- Fix an issue where the 4 upper bits of the stencil were always cleared with force_load disabled
- Update the SceGxmDepthStencilSurface to make it nicer and have the same layout as SceGxm
- Handle disabled stencil surfaces in the renderer, can potentially slightly improve performance on tiled renderers

### DIFF
--- a/vita3k/renderer/include/renderer/gxm_types.h
+++ b/vita3k/renderer/include/renderer/gxm_types.h
@@ -55,22 +55,58 @@ struct SceGxmColorSurface {
 
 static_assert(sizeof(SceGxmColorSurface) == (32 + sizeof(SceGxmTexture)), "Incorrect size.");
 
-struct SceGxmDepthStencilControl {
-    uint32_t content;
-
-    static constexpr uint32_t format_bits = ~0xFFF;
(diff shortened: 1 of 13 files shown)


vita3k/renderer/include/renderer/gxm_types.h
```diff
@@ -56,22 +56,72 @@ struct SceGxmColorSurface {
 static_assert(sizeof(SceGxmColorSurface) == (32 + sizeof(SceGxmTexture)), "Incorrect size.");
 
 struct SceGxmDepthStencilControl {
```
xerpi, line 58: This struct seems not to be used anymore. Can't it be removed?

Macdu: You're right, I forgot to remove it

Zangetsu38 (approved):

Macdu:
This PR is ready

Zangetsu38 (approved):

Zangetsu38:
This pr have broken render on retro city rampage in ogl, causing black screen 🤔

PR #2909, opened 2023-09-11
modules/SceCommonDialog: Stub SceCommonDialog_E37069D5

dima-xd:
Allows to boot https://github.com/Vita3K/compatibility/issues/1577 in-game (maybe it's playable now, needs to be tested).

I don't know where do I get the actual name of this function since in eboot elf I can only get `SceCommonDialog_E37069D5` name


Zangetsu38:
> 



> That works but it's not the proper fix for a function with an unknown name and whom content is empty. It's better in this case to fix the behavior of unkown NIDs and make them return 0 (instead of what is in r0, most likely the function first arg if it has one). I thought I PRed it but looks like I didn't, sorry.

yes i confirm, but other dev have jkust said me it return random value when is uniknown nid

PR #2913, opened 2023-09-12
vita3k: Prefer contains() instead of find() + end()

scribam:


vita3k/modules/SceFiber/SceFiber.cpp
```diff
@@ -87,7 +87,7 @@ void set_thread_fiber(FiberState &state, const SceUID &tid, SceFiber *fiber) {
 }
 
 SceFiber *get_thread_fiber(FiberState &state, const SceUID &tid) {
-    if (state.thread_fibers.find(tid) == state.thread_fibers.end()) {
+    if (!state.thread_fibers.contains(tid)) {
         return nullptr;
     }
     return state.thread_fibers[tid];
```
bookmist, line 93:
```suggestion
    auto fiber = state.thread_fibers.find(tid);
    if (fiber == state.thread_fibers.end()) {
        return nullptr;
    }
    return fiber->second;```

File: vita3k/net/src/epoll.cpp [hunk 2]
```diff
@@ -1,8 +1,7 @@
 #include <net/epoll.h>
 
 int Epoll::add(int id, abs_socket sock, SceNetEpollEvent *ev) {
-    auto it = eventEntries.find(id);
-    if (it != eventEntries.end()) {
+    if (eventEntries.contains(id)) {
```
Macdu, line 4: Replace that function by:
`if(!eventEntries.try_emplace(id, EpollSocket{ ev->events, ev->data, sock }).second)
    return SCE_NET_ERROR_EEXIST;
return 0; `

Macdu (commented):
Pretty nice PR

vita3k/modules/SceFiber/SceFiber.cpp
scribam: Thanks for the suggestion, applied.

vita3k/gui/src/gui.cpp
```diff
@@ -524,7 +524,7 @@ void init_home(GuiState &gui, EmuEnvState &emuenv) {
     regmgr::init_regmgr(emuenv.regmgr, emuenv.pref_path);
 
     const auto is_cmd = emuenv.cfg.run_app_path || emuenv.cfg.content_path;
-    if (!gui.users.empty() && (gui.users.find(emuenv.cfg.user_id) != gui.users.end()) && (is_cmd || emuenv.cfg.auto_user_login)) {
+    if (!gui.users.empty() && (gui.users.contains(emuenv.cfg.user_id)) && (is_cmd || emuenv.cfg.auto_user_login)) {
```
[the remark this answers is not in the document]
scribam: Indeed

vita3k/net/src/epoll.cpp
scribam: Nice one 👍

Macdu (commented):
Some things I missed. 
There are still a few times where a contains is followed by an access (in which case it would be better performance-wise to call find and get an iterator), but the current code does not do it, the syntax contains + operator[] looks a lot nicer than find and they are not done on a performance critical path, so no need to change them if you don't feel like it.

vita3k/gui/src/app_context_menu.cpp
```diff
@@ -241,7 +241,7 @@ void delete_app(GuiState &gui, EmuEnvState &emuenv, const std::string &app_path)
         if (fs::exists(SHADER_LOG_PATH))
             fs::remove_all(SHADER_LOG_PATH);
 
-        if (gui.app_selector.user_apps_icon.find(app_path) != gui.app_selector.user_apps_icon.end()) {
+        if (gui.app_selector.user_apps_icon.contains(app_path)) {
             gui.app_selector.user_apps_icon[app_path] = {};
             gui.app_selector.user_apps_icon.erase(app_path);
         }
```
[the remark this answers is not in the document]
Zangetsu38: No touch this 💀 is here for good things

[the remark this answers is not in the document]
Macdu: Can you explain?

vita3k/shader/src/usse_utilities.cpp
```diff
@@ -678,13 +678,13 @@ spv::Id shader::usse::utils::unpack_one(spv::Builder &b, SpirvUtilFunctions &uti
     case DataType::UINT8:
     case DataType::UINT16:
     case DataType::INT16: {
-        if (utils.unpack_funcs.find(type) == utils.unpack_funcs.end()) {
+        if (!utils.unpack_funcs.contains(type)) {
             utils.unpack_funcs[type] = make_unpack_func(b, features, type);
         }
         return b.createFunctionCall(utils.unpack_funcs.at(type), { scalar });
```
bookmist, line 684:
```suggestion
        auto iter = utils.unpack_funcs.find(type);
        if (iter == utils.unpack_funcs.end()) {
            iter = utils.unpack_funcs.emplace(type,make_unpack_func(b, features, type)).first;
        }
        return b.createFunctionCall(iter->second, { scalar });
```
same suggestion for two next changes

bookmist: it's not critical, corresponding to https://github.com/Vita3K/Vita3K/pull/2913#pullrequestreview-1622179494

vita3k/gui/src/app_context_menu.cpp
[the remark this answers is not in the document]
Zangetsu38: Is not my fix, is other dev have sud-est me using this for fix crash, for can clear texture and after erase map, like that it clear texture so no used anymore, and can erase map in safe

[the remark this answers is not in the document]
Macdu: using `gui.app_selector.user_apps_icon[app_path] = {};` does nothing (just sets a pointer to null), so this just leaks the texture memory.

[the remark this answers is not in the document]
Macdu: Putting aside the fact that you don't speak english, thank you for explaining that you have memory leaks all over the GUI, not that I am really surprised by this anyway...
However, this does not really concern this PR, so for Scribam, I guess you can keep this part of the code the way it is currently in your PR

vita3k/shader/src/usse_program_analyzer.cpp
```diff
@@ -473,7 +473,7 @@ void analyze(USSEBlockNode &root, USSEOffset end_offset, AnalyzeReadFunction rea
                         baddr = current_code->offset - 1;
                     }
                 }
-            } else if ((branches_to_back.find(baddr) != branches_to_back.end()) && (request.block_node->start_offset() != baddr)) {
+            } else if ((branches_to_back.contains(baddr)) && (request.block_node->start_offset() != baddr)) {
```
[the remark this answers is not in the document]
scribam: Fixed

vita3k/gui/src/gui.cpp
```diff
@@ -538,7 +538,7 @@ void init_user_app(GuiState &gui, EmuEnvState &emuenv, const std::string &app_pa
     const auto APP_INDEX = get_app_index(gui, app_path);
     if (APP_INDEX != gui.app_selector.user_apps.end()) {
         gui.app_selector.user_apps.erase(APP_INDEX);
-        if (gui.app_selector.user_apps_icon.find(app_path) != gui.app_selector.user_apps_icon.end())
+        if (gui.app_selector.user_apps_icon.contains(app_path))
             gui.app_selector.user_apps_icon.erase(app_path);
```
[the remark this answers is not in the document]
scribam: Done

vita3k/shader/src/usse_utilities.cpp
scribam: I used `try_emplace` instead

bookmist: `try_emplace` run `make_pack_func(b, features, source_type)` every time and then throw away its result.

scribam: Are you sure? :thinking: 
From https://rules.sonarsource.com/cpp/RSPEC-6030/:
> C++17 introduced try_emplace that does not construct the value if the key is already present in the map and constructs the value in place if necessary.

bookmist: arguments of try_emplace (and any other function) are evaluated before function call

bookmist: just log calls of `make_*_func` and check it yourself

scribam: My bad, fixed. Thanks for the review 👍

Macdu (approved):

PR #2915, opened 2023-09-12
renderer/screen_filters: fix memory leak

bookmist:
It seems that destructors of ScreenFilter and it's descendants are newer called. (`vao.destroy` is called automatically on call `vao` destructor, so double `destroy()` caused crash)

### DIFF
--- a/vita3k/renderer/src/vulkan/screen_filters.cpp
+++ b/vita3k/renderer/src/vulkan/screen_filters.cpp
@@ -44,7 +44,6 @@ SinglePassScreenFilter::~SinglePassScreenFilter() {
     vk::Device device = screen.state.device;
     // this will only happen when the user changes the option in the GUI, we can afford to waitIdle
     device.waitIdle();
-    vao.destroy();
     device.destroy(pipeline);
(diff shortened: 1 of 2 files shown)


vita3k/renderer/src/vulkan/screen_filters.cpp
```diff
@@ -44,7 +44,6 @@ SinglePassScreenFilter::~SinglePassScreenFilter() {
     vk::Device device = screen.state.device;
     // this will only happen when the user changes the option in the GUI, we can afford to waitIdle
     device.waitIdle();
-    vao.destroy();
     device.destroy(pipeline);
     device.destroy(pipeline_layout);
     device.freeDescriptorSets(descriptor_pool, descriptor_sets);
```
Macdu, line 49:
```suggestion
```
Can you remove this line too? It is not necessary and causing a validation error.

bookmist:
This PR cause validation errors (even with all suggested changes) and crash when I try to change screen filter when game is running. I don't know how to make it working and problem is not critical at all, so I close this PR.

PR #2926, opened 2023-09-16
external: Update submodules

scribam:


Zangetsu38:
Please squash all in one commit 🙏
Set commit name
External: update lot submodules
- detail1
- detail2
- etc...

external/CMakeLists.txt
```diff
@@ -315,7 +300,7 @@ if(NOT OPENSSL_FOUND)
 				WORKING_DIRECTORY "${CMAKE_BINARY_DIR}/external/openssl")
 		endif()
 
-		set(OPENSSL_ROOT_DIR "${CMAKE_BINARY_DIR}/external/openssl/openssl-3.0/x64")
+		set(OPENSSL_ROOT_DIR "${CMAKE_BINARY_DIR}/external/openssl/openssl-3/x64")
```
Zangetsu38, line 303: Revert, i have changed it for fix problem, that broken build

Zangetsu38:
you can add in clang format commit fix format for interface.cpp
i have ping you in my pr for this

Zangetsu38:
Sorry, you can rebase and skip last commit

Zangetsu38 (approved):

Macdu:
Why was this merged so quickly? The Vulkan-Header submodule must only be updated along the VMA submodule or it can break it.
Also was there a reason to update the glad submodule?

scribam:
>  The Vulkan-Header submodule must only be updated along the VMA submodule or it can break it.

We can update https://github.com/Macdu/VulkanMemoryAllocator-Hpp with https://github.com/YaaZ/VulkanMemoryAllocator-Hpp/commit/d9b2147d5e78b947ea20011b50645df67e1815a6 if necessary

> Also was there a reason to update the glad submodule?

Not really. I did it to batch all the updates in one PR. glad2 provides more options but it probably doesn't matter for Vita3K. Do you have issues with the updated version?

Macdu:
> We can update https://github.com/Macdu/VulkanMemoryAllocator-Hpp with [YaaZ/VulkanMemoryAllocator-Hpp@d9b2147](https://github.com/YaaZ/VulkanMemoryAllocator-Hpp/commit/d9b2147d5e78b947ea20011b50645df67e1815a6) if necessary

The issue is that the Vulkan-Header module is not retro-compatible between versions (there is no guarantee for it and it happens every few months that this is not the case), so it is a bad idea for the Vulkan-Header submodule not to match exactly the VMA submodule.



> Not really. I did it to batch all the updates in one PR. glad2 provides more options but it probably doesn't matter for Vita3K. Do you have issues with the updated version?

Well updating the submodule (exept when adding a new extension we would like to use) is kind of useless. It changes absolutely nothing (being performance-wise or in the code) and just appears as a big 16K lines changed as it is not a submodule. But now it would be even worse to revert this commit.

PR #2928, opened 2023-09-16
ngs: workaround crashes on invalid configurations

bookmist:
util/log: LOG_*_ONCE macros
modules/SceNgs: implement sceNgsVoiceGetModuleType
ngs/modules/player: workaround crash if required playback frequency is invalid
ngs: stub for incomplete output paths (for Smart As [PCSF00216])
ngs: stub for null voice definition

### DIFF
--- a/vita3k/modules/SceNgsUser/SceNgs.cpp
+++ b/vita3k/modules/SceNgsUser/SceNgs.cpp
@@ -62,6 +62,7 @@ enum SceNgsErrorCode : uint32_t {
     SCE_NGS_ERROR_INVALID_ARG = 0x804A0002,
     SCE_NGS_ERROR_INVALID_STATE = 0x804A0010,
     SCE_NGS_ERROR_PARAM_OUT_OF_RANGE = 0x804A0009,
+    SCE_NGS_ERROR_INVALID_HANDLE = 0x804A000C,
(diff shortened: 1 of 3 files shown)


Macdu:
I think if a rack is created with an invalid voice definition, there is something that already went pretty wrong previously. Are you sure that the game expects the given definition to be null?

bookmist:
> I think if a rack is created with an invalid voice definition, there is something that already went pretty wrong previously. Are you sure that the game expects the given definition to be null?

Something definitely went pretty wrong previously. Most of this PR is stubs to not crash completely with AV on such problems.

Macdu:
I think if these functions were to be called with an invalid voice definition on a real PS Vita, it would almost surely cause a crash. From my point of view, fixing what went wrong at the beginning should be the focus instead of trying to handle some behavior that can't happen on a real PS Vita.

bookmist:
@Macdu Now I move most of checks to sceNgs module and return appropriate error in most cases. It's the same behavior as on real psvita. Proper fixing playback frequency and returning error to caller needs refactoring, which I'm too lazy to do now.

Macdu (approved):
This is more of a hack than a real fix but fine I guess (also I like the LOG_*_ONCE functions)

bookmist:
> also I like the LOG_*_ONCE functions
It was inspired by your code. Thanks for approve.

PR #2929, opened 2023-09-16
shader: use `std::bit_set`,  return `nullptr` after `assert`

xerpi:
Fixes ` warning: dereferencing type-punned pointer will break
strict-aliasing rules [-Wstrict-aliasing]` when compiling with `-Wall`.

`assert()` is ignored on release builds.
Fixes many `may be used uninitialized` warnings.

### DIFF
--- a/vita3k/shader/include/shader/usse_constant_table.h
+++ b/vita3k/shader/include/shader/usse_constant_table.h
@@ -17,6 +17,8 @@
 
 #pragma once
 
+#include <util/bit_cast.h>
+
 #include <cstdint>
 
 namespace shader::usse {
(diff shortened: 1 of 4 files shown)


Zangetsu38:
Missing include for bit cast for mac

xerpi:
> Missing include for bit cast for mac

It's still failing after adding the missing include, and I don't know why.

xerpi:
It seems like Apple Clang does not support `std::bit_cast`: https://github.com/AcademySoftwareFoundation/Imath/issues/297
https://en.cppreference.com/w/cpp/compiler_support

Zangetsu38 (approved):

PR #2930, opened 2023-09-16
modules/sceMotion: Fix wrong sizeof usage on memset

EXtremeExploit:
forgor the *, pretty sure it didn't mean to set only a few bytes to 0 :D


Zangetsu38:
Fix commit name, is wrong here
Modules/SceMotion: fix.....

Zangetsu38 (approved):

PR #2932, opened 2023-09-18
gui/initial setup: Add two gui setting and little more.

Zangetsu38:
Split my other pr for separe Fix aspect ratio of other.
# About
- gui/initial setup: Add two gui setting and little more
Fix size of column of select language.
Save config any time press on next.
Small clean code.
Change bg color

### DIFF
--- a/vita3k/gui/src/initial_setup.cpp
+++ b/vita3k/gui/src/initial_setup.cpp
@@ -17,6 +17,7 @@
 
 #include "private.h"
 
+#include <config/functions.h>
 #include <config/state.h>
 #include <dialog/state.h>
 #include <gui/functions.h>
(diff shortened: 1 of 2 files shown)


Zangetsu38:
It is keep ready

vita3k/gui/src/initial_setup.cpp
```diff
@@ -210,6 +213,14 @@ void draw_initial_setup(GuiState &gui, EmuEnvState &emuenv) {
     case SELECT_INTERFACE_SETTINGS:
         title_str = lang["select_interface_settings"];
         ImGui::SetCursorPosY((WINDOW_SIZE.y / 2.f) - ImGui::GetFontSize());
+        ImGui::Checkbox("Display Info Message", &emuenv.cfg.display_info_message);
```
Macdu, line 216:
```suggestion
        ImGui::Checkbox("Display Compatibility Update Notification", &emuenv.cfg.display_info_message);
```

vita3k/gui/src/initial_setup.cpp
```diff
@@ -210,6 +213,14 @@ void draw_initial_setup(GuiState &gui, EmuEnvState &emuenv) {
...
+        if (ImGui::IsItemHovered())
+            ImGui::SetTooltip("Uncheck the box to display info message in log only.");
```
Macdu, line 218:
```suggestion
            ImGui::SetTooltip("Check the box to display a notification every time the compatibility database is updated.");
```

vita3k/gui/src/initial_setup.cpp
Zangetsu38: Setting added is just same of setting dialog, if i change here i need change on setting dialog too 
Also is really info message, i have code it in base for can set every info message inside, not only compat

vita3k/gui/src/initial_setup.cpp
Zangetsu38: Same, code is just copy pasted from setting dialog, i no want change it, like i said, is message d'or any info, just for now is onky set for compat

vita3k/gui/src/initial_setup.cpp
Macdu: Well for the time being it only gives notifications about the compatibility database update, feel free to update it once you make it display something else.
Also "info message" is way too broad, it is one of the first thing displayed and some user can let it checked thinking it shows important information.

Zangetsu38: Hmm important no, juste said is info message, no santé said is important
And your suggestion is too long, info message is short

bookmist (approved):

Issue #2922, opened 2023-09-14
Crash on startup (c++ reference failed)


NerosTie:
When I launch vita3k I have a black window then it crashes.

The log:

```
vita3k
[00:40:46.690] |I| [init_config]: backend-renderer: Vulkan
[00:40:46.690] |I| [init_config]: log-level: 0
[00:40:46.992] |I| [main]: Vita3K v0.1.9 3372-54b276de-;home;main-builder;pkgsrc/master
[00:40:46.992] |I| [main]: Number of logical CPU cores: 12
[00:40:46.992] |I| [main]: Available ram memory: 32018 MiB
[00:40:47.018] |I| [create]: Enabling vulkan validation layers (has a performance impact but allows better error messages)
[00:40:47.051] |I| [create]: Vulkan device: AMD Radeon RX 6600 XT (RADV NAVI23)
[00:40:47.051] |I| [create]: Driver version: 23.1.7
[00:40:47.052] |I| [create]: Memory mapping is enabled
[00:40:47.052] |E| [debug_callback]: Validation layer: terminator_CreateDevice: Failed in ICD /usr/lib/libvulkan_radeon.so vkCreateDevice call
[00:40:47.052] |E| [debug_callback]: Validation layer: vkCreateDevice:  Failed to create device chain.
[00:40:47.064] |I| [setup]: Present mode: Mailbox
/usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/13.2.1/../../../../include/c++/13.2.1/array:202: reference std::array<renderer::TextureCacheInfo, 1024>::operator[](size_type) [_Tp = renderer::TextureCacheInfo, _Nm = 1024]: Assertion '__n < this->size()' failed.
[1]    99675 IOT instruction (core dumped) vita3k
```

Infos:
- Vita3K v0.1.9 3372-54b276de
- Mesa 23.1.7
- GPU: AMD RX 6600 XT
- CPU: AMD Ryzen 5 3600
- OS: Arch
- Kernel: 6.5.3

EXtremeExploit:
Was this compiled using debug or release?, if it was using debug then attaching the coredump file would be nice, else you will have to compile debug yourself and try running it there to get debugging symbols

EXtremeExploit:
same error as https://aur.archlinux.org/packages/vita3k-git#comment-934046

NerosTie:
[...]

PR #2935, opened 2023-09-18
renderer: Fix OOB access on texture cache

EXtremeExploit:
This is my last resort of fixing this weird bug that only happens on the AUR, AND I DONT KNOW WHY asserts are being compiled, i checked every compiler flag and it still doesn't make sense to why it happens on a release build, but anyways, i didnt want to modify upstream but here i am, doing exactly that, i give up, this is only needed for that one specific case of being compiled using makepkg

fixes https://github.com/Vita3K/Vita3K/issues/2922

### DIFF
--- a/vita3k/renderer/src/texture_cache.cpp
+++ b/vita3k/renderer/src/texture_cache.cpp
@@ -117,8 +117,10 @@ bool TextureCache::init(const bool hashless_texture_cache) {
 
     // initialize the doubly linked list
     for (int i = 0; i < TextureCacheSize; i++) {
-        infoes[i].prev = &infoes[i - 1];
-        infoes[i].next = &infoes[i + 1];
+        if (i > 0)
+            infoes[i].prev = &infoes[i - 1];
+        if (i < TextureCacheSize - 1)
+            infoes[i].next = &infoes[i + 1];
     }
     // fix the first and last elements
     infoes[0].prev = &infoes[TextureCacheSize - 1];


vita3k/renderer/src/texture_cache.cpp
```diff
@@ -117,8 +117,15 @@ bool TextureCache::init(const bool hashless_texture_cache) {
...
-        infoes[i].next = &infoes[i + 1];
+        if (i == 0) // This is annoying
+            infoes[i].prev = nullptr;
+        else
+            infoes[i].prev = &infoes[i - 1];
+
+        if (i == TextureCacheSize - 1) // this is annoying too
+            infoes[i].next = nullptr;
+        else
+            infoes[i].next = &infoes[i + 1];
```
Macdu, line 128:
```suggestion
        if (i > 0)
            infoes[i].prev = &infoes[i - 1];

        if (i < TextureCacheSize - 1) 
            infoes[i].next = &infoes[i + 1];
```

EXtremeExploit: done

Macdu:
Look nice, it's me Macdu2 on discord

xerpi:
Good catch!
However, any reason why we are rolling our own linked-list implementation instead of using something C++'s `std` provides?

Macdu (approved):

xerpi:
> > Good catch! However, any reason why we are rolling our own linked-list implementation instead of using something C++'s `std` provides?
> 
> Well a doubly-linked list is really easy to implement. Moreover, using the std (I think you mean using std::list) has many drawbacks:
> 
> * the code would actually be kind of less clean, with iterators everywhere
> * the api for std::list is quite limited, like for taking an element and putting it in the front you need to erase it and insert it, performance wise that's really bad
> * The current implementation is more cache-efficient than a list (we already know the size of the list and allocate everything in a contiguous memory range)
>   For this case, performance really matters as a texture lookup (and lru update) must be performed each time a texture is used.

Makes sense, thank you for the detailed explanation!

PR #2937, opened 2023-09-20
ctrl: Drop inputs while rebinding menu is open

nn9dev:
Make it so that when the rebinds menu is open that it does not affect the GUI and in-game events.

Currently I have it set up so that events get dropped when the controller BIND menu is up, and not the controller LIST menu, but this can be easily changed.

### DIFF
--- a/vita3k/ctrl/include/ctrl/state.h
+++ b/vita3k/ctrl/include/ctrl/state.h
@@ -60,6 +60,7 @@ struct CtrlState {
     ControllerList controllers;
     int controllers_num = 0;
     bool has_motion_support = false;
+    bool drop_inputs = false;
     const char *controllers_name[SCE_CTRL_MAX_WIRELESS_NUM];
     bool free_ports[SCE_CTRL_MAX_WIRELESS_NUM] = { true, true, true, true };
     SceCtrlPadInputMode input_mode = SCE_CTRL_MODE_DIGITAL;
--- a/vita3k/ctrl/src/ctrl.cpp
+++ b/vita3k/ctrl/src/ctrl.cpp
@@ -268,6 +268,7 @@ static void retrieve_ctrl_data(EmuEnvState &emuenv, int port, bool is_v2, bool n
     axes.fill(0);
 
     const auto reset_axes = [&]() {
+        // Re-center joysticks to (128,128). Range is (0-255,0-255).
         SceCtrlPadInputMode mode = from_ext_function ? state.input_mode_ext : state.input_mode;
         if (mode == SCE_CTRL_MODE_DIGITAL) {
             lx = 0x80;
@@ -282,7 +283,7 @@ static void retrieve_ctrl_data(EmuEnvState &emuenv, int port, bool is_v2, bool n
         }
     };
 
-    if (emuenv.common_dialog.status == SCE_COMMON_DIALOG_STATUS_RUNNING) {
+    if (emuenv.common_dialog.status == SCE_COMMON_DIALOG_STATUS_RUNNING || emuenv.ctrl.drop_inputs) {
         if (negative)
             buttons ^= ~0;
         reset_axes();
--- a/vita3k/gui/src/controllers_dialog.cpp
+++ b/vita3k/gui/src/controllers_dialog.cpp
@@ -361,6 +361,8 @@ void draw_controllers_dialog(GuiState &gui, EmuEnvState &emuenv) {
         reset_controller_binding(emuenv);
 
     ImGui::End();
+
+    ctrl.drop_inputs = rebinds_is_open; // if the dialog is active, drop inputs from passing to ctrl
 }
 
 } // namespace gui
--- a/vita3k/gui/src/controls_dialog.cpp
+++ b/vita3k/gui/src/controls_dialog.cpp
@@ -20,6 +20,7 @@
 #include <SDL_scancode.h>
 #include <config/functions.h>
 #include <config/state.h>
+#include <ctrl/state.h>
 #include <emuenv/state.h>
 #include <gui/functions.h>
 #include <interface.h>
@@ -104,6 +105,7 @@ static void remapper_button(GuiState &gui, EmuEnvState &emuenv, int *button, con
         std::array<int, total_key_entries> original_state;
         prepare_map_array(emuenv, original_state);
         while (gui.is_capturing_keys) {
+            emuenv.ctrl.drop_inputs = false; // set ctrl.drop_inputs to false, so they don't get discarded while rebinding
             handle_events(emuenv, gui);
             *button = gui.captured_key;
             if (*button < 0 || *button > 231)
@@ -204,6 +206,8 @@ void draw_controls_dialog(GuiState &gui, EmuEnvState &emuenv) {
     }
 
     ImGui::End();
+
+    emuenv.ctrl.drop_inputs = gui.controls_menu.controls_dialog;
 }
 
 } // namespace gui
--- a/vita3k/interface.cpp
+++ b/vita3k/interface.cpp
@@ -619,7 +619,7 @@ bool handle_events(EmuEnvState &emuenv, GuiState &gui) {
             return false;
 
         case SDL_KEYDOWN: {
-            if (ImGui::GetIO().WantTextInput || gui.is_key_locked)
+            if (ImGui::GetIO().WantTextInput || gui.is_key_locked || emuenv.ctrl.drop_inputs)
                 continue;
 
             const auto get_sce_ctrl_btn_from_scancode = [&emuenv](const SDL_Scancode scancode) {
@@ -717,7 +717,7 @@ bool handle_events(EmuEnvState &emuenv, GuiState &gui) {
             if (!emuenv.kernel.is_threads_paused() && (event.cbutton.button == SDL_CONTROLLER_BUTTON_TOUCHPAD))
                 toggle_touchscreen();
 
-            if (ImGui::GetIO().WantTextInput)
+            if (ImGui::GetIO().WantTextInput || emuenv.ctrl.drop_inputs)
                 continue;
 
             for (const auto &binding : get_controller_bindings_ext(emuenv)) {


vita3k/ctrl/src/ctrl.cpp
```diff
@@ -258,6 +258,11 @@ static void apply_controller(EmuEnvState &emuenv, uint32_t *buttons, float axes[
 static void retrieve_ctrl_data(EmuEnvState &emuenv, int port, bool is_v2, bool negative, bool from_ext_function, SceUInt32 &buttons, SceUInt8 &lx, SceUInt8 &ly, SceUInt8 &rx, SceUInt8 &ry) {
     std::lock_guard<std::mutex> guard(emuenv.ctrl.mutex);
 
+    if(emuenv.ctrl.drop_inputs){
```
Zangetsu38, line 261: Just moving it with other if in 285 line, no have sense here

vita3k/interface.cpp
```diff
@@ -622,6 +622,9 @@ bool handle_events(EmuEnvState &emuenv, GuiState &gui) {
             if (ImGui::GetIO().WantTextInput || gui.is_key_locked)
                 continue;
 
+            if (emuenv.ctrl.drop_inputs)
```
Zangetsu38, line 625: Same ? With no using if just before ?

vita3k/interface.cpp
```diff
@@ -717,6 +720,9 @@ bool handle_events(EmuEnvState &emuenv, GuiState &gui) {
             if (!emuenv.kernel.is_threads_paused() && (event.cbutton.button == SDL_CONTROLLER_BUTTON_TOUCHPAD))
                 toggle_touchscreen();
 
+            if (emuenv.ctrl.drop_inputs)
```
Zangetsu38, line 723: Same... usinv if set juste after

vita3k/ctrl/src/ctrl.cpp
nn9dev: Could you tell me what `SCE_COMMON_DIALOG_RUNNING` does so that i have more information as to if it is an appropriate checking point or not?

Zangetsu38: ? Just read code

Zangetsu38:
You need fix format

vita3k/ctrl/src/ctrl.cpp
nn9dev: Zang, I am asking you because I do not understand what the code is telling me. There is no documentation, no comments, no nothing for that enum and it's constants. I am grasping at straws. Best I can tell is that it's supposed to be for if an in-game dialogue is pulled up, which WOULD make sense to drop all inputs for, so I could `OR` it with my check, but I have asked for information from you because I am not sure, and would like to be certain. It won't kill you to help someone out.

Zangetsu38: I said that because there was nothing to say, just read the code, it is clear and clear, no need for comments, and very simple to understand.  if the commong dialog is in progress it deactivates the ingame input, easy to understand

nn9dev:
nn9dev vs format-check (it only took 4 tries)

vita3k/gui/src/controls_dialog.cpp
```diff
@@ -104,6 +105,7 @@ static void remapper_button(GuiState &gui, EmuEnvState &emuenv, int *button, con
         std::array<int, total_key_entries> original_state;
         prepare_map_array(emuenv, original_state);
         while (gui.is_capturing_keys) {
+            emuenv.ctrl.drop_inputs = false; // set ctrl.drop_inputs to false, so they don't get discarded while rebinding
```
Zangetsu38, line 108: i no understund here ?

vita3k/ctrl/src/ctrl.cpp
```diff
@@ -282,11 +282,11 @@ static void retrieve_ctrl_data(EmuEnvState &emuenv, int port, bool is_v2, bool n
...
     };
 
-    if (emuenv.common_dialog.status == SCE_COMMON_DIALOG_STATUS_RUNNING) {
+    if (emuenv.common_dialog.status == SCE_COMMON_DIALOG_STATUS_RUNNING || emuenv.ctrl.drop_inputs) {
         if (negative)
             buttons ^= ~0;
-        reset_axes();
-        return;
+        reset_axes(); // keep sticks centered
+        return; // drop everything else
```
Zangetsu38, line 289: revert, no need this is already clear

vita3k/gui/src/controls_dialog.cpp
nn9dev: If you remove this line, Vita3K locks up when you input a keyboard bind.

Zangetsu38: Want said your code is wrong here
Looking your code inside interface

vita3k/interface.cpp
```diff
@@ -619,7 +619,7 @@ bool handle_events(EmuEnvState &emuenv, GuiState &gui) {
             return false;
 
         case SDL_KEYDOWN: {
-            if (ImGui::GetIO().WantTextInput || gui.is_key_locked)
+            if (ImGui::GetIO().WantTextInput || gui.is_key_locked || emuenv.ctrl.drop_inputs)
```
Zangetsu38, line 622: Moving this after 663 line

nn9dev: No. If we are dropping inputs, there only needs to be one check at the very beginning of the case, and if the check succeeds, we should continue. there is no point in even reading the input. If you want me to move the `ImGui::GetIO().WantTextInput || gui.is_key_locked` part down, I can, but the ctrl.drop_inputs should stay at the top, both for readability and practicality.

Zangetsu38: Just done what i said, is togaly good, and remove your false in other file, have no sense

vita3k/ctrl/src/ctrl.cpp
```diff
@@ -282,10 +282,10 @@ static void retrieve_ctrl_data(EmuEnvState &emuenv, int port, bool is_v2, bool n
         }
     };
 
-    if (emuenv.common_dialog.status == SCE_COMMON_DIALOG_STATUS_RUNNING) {
+    if (emuenv.common_dialog.status == SCE_COMMON_DIALOG_STATUS_RUNNING || emuenv.ctrl.drop_inputs) {
         if (negative)
             buttons ^= ~0;
-        reset_axes();
+        reset_axes(); // keep sticks centered
```
Zangetsu38, line 288: Remove comment, is not needed, function name already explain what it done

vita3k/gui/src/controls_dialog.cpp
Zangetsu38: Remove it here

vita3k/interface.cpp
nn9dev: No, your suggestion has no sense. There is absolutely ZERO reason to read the input if we are dropping the inputs. they ONLY need to be un-dropped if we are in the binds loop.

vita3k/ctrl/src/ctrl.cpp
nn9dev: Quite frankly, it was not obvious to me, someone who rarely touches the code. Documentation is helpful.

Zangetsu38: Just remove 🙏 is full clear

vita3k/interface.cpp
Zangetsu38: Is inversed here, is you current code have no sense... 
Juste donné what i said

vita3k/ctrl/src/ctrl.cpp
nn9dev: image

vita3k/interface.cpp
Zangetsu38: You set your bool on false, for can read gui captured key, jyst donne what i said, onoy that, can also move code on more top for read captured key

vita3k/ctrl/src/ctrl.cpp
Zangetsu38: Th8s  ode is not me, i have hyst create function, so yes number is magic, but function of name is clear

nn9dev: name of function is not clear. if you renamed it to, say, `recenter_joysticks`, it would be clear.

Zangetsu38: Here you only adding comment in one time, no have sense, if you want, adding comment inside lambda function, but is already clear

nn9dev: i'll do that then.

Zangetsu38: > name of function is not clear. if you renamed it to, say, `recenter_joysticks`, it would be clear.

Name reset axis is clear
If tiu nocknow what is axis 🤷‍♂️

vita3k/interface.cpp
Zangetsu38: I thinks you wrong view, you no read any input with moving this full if contrition in 664 line
You just only allow édit keyboard input

vita3k/gui/src/controls_dialog.cpp
Zangetsu38: Your current code is full wrong, you set truc in end for set false here, is so never on truc is no sense
Juste done what i said in interface, and delete this

vita3k/interface.cpp
nn9dev: image

Zangetsu38: You read your code ? 
Just move this  if with continue before if (allow...
And fine

nn9dev: I just drew you a whole diagram. `if(gui.is_capturing_keys...` does NOT need executed if we are dropping inputs. It ONLY needs executed when we click on the button to remap, which is the point of the false statement in the other file.

vita3k/gui/src/controls_dialog.cpp
Zangetsu38: you no understund why your code here have no sense..
you get not responding here because you r code inside interface is on wrong place
here you set code to false for can passe and captured key, have no sense, set it true for set it false in same time

Move code in interface wjere i said, and delete this line and works fine

vita3k/gui/src/controls_dialog.cpp
```diff
@@ -204,6 +206,8 @@ void draw_controls_dialog(GuiState &gui, EmuEnvState &emuenv) {
     }
 
     ImGui::End();
+
+    emuenv.ctrl.drop_inputs = gui.controls_menu.controls_dialog;
```
Zangetsu38, line 210: is same for this, this line no have sense
if this function it set obligatory to true...
you need change this

vita3k/interface.cpp
Zangetsu38: drop input is for drop action of controller, no for block remap

vita3k/gui/src/controls_dialog.cpp
nn9dev: What the fuck? this is the line that SETS drop_inputs to true when the menu is open. Did you even read it? Remove the line and you can very clearly see that if you press up and down while the keyboard menu is open, you can still navigate the UI in the background.

vita3k/interface.cpp
nn9dev: No, it's for dropping action of both keyboard and controller so that you CANNOT navigate the UI while the rebinds menu is open.

vita3k/gui/src/controls_dialog.cpp
nn9dev: No. Your solution has zero sense to execute more code than necessary. I understand I am arguing over a cycle or two, but my solution is valid and your solution arguably makes it less readable. I have asked two other people if your solution is good, and they have both said no. I am sick of arguing.

Zangetsu38: you litteraly set one value to true for ste to false here... is you execute more code are needed compare me
is not my solution, is just logical

vita3k/gui/src/controls_dialog.cpp
Zangetsu38: you have bad understund me here, i said only you need CHANGE this way, here have no sense
this line just set true all time, you check value of controller dialog bool but it is obligatory on true here
i perfectly more understund you who code works
you code here have no sense, is totaly not logical

vita3k/gui/src/controllers_dialog.cpp
```diff
@@ -361,6 +361,8 @@ void draw_controllers_dialog(GuiState &gui, EmuEnvState &emuenv) {
         reset_controller_binding(emuenv);
 
     ImGui::End();
+
+    ctrl.drop_inputs = rebinds_is_open; // if the dialog is active, drop inputs from passing to ctrl
```
Zangetsu38, line 365: need change this, more good ir probably check in other place  if controll dialog or controller dialog is true set drop to true

nn9dev: That's the job of ImGui::Begin(). The p_open parameter that gets passed to ImGui::Begin() is set to TRUE when the window is open and FALSE when the window is closed. placing this parameter outside of ImGui::End() makes certain that when the window is OPEN it will always be TRUE and when it is CLOSED it will always be false. setting drop_binds to the p_open parameter ensures that when the window is open, drop_binds is TRUE and when it is closed, drop binds is FALSE. Same goes for the other place i set drop_binds like this.

vita3k/interface.cpp
Zangetsu38: yes i know ? and is what i explain you ? 
what i said why move code after if gui.captured is full normal

vita3k/gui/src/controls_dialog.cpp
nn9dev: No, the line gets set to false when the window closes. see my other comment. ImGui::Begin() makes sure of this.

Zangetsu38: i thinks you no understund how works this here
controll dialog is set to true by click on button on menubar.
and it waiting you close dialog it for can false

no idea why you speak of ImGui::Begin() totaly no relate it 
you no need set this here

best way is move this drop input outside both 
with check in other part, like inside handle event if one of both is true set drop to true, is very simple

Zangetsu38 (changes requested):
Just change how i said
We must change the logic of the code, what is being done currently makes no sense.

vita3k/gui/src/controllers_dialog.cpp
Zangetsu38: ??? when you close dialog it move value to false
bu no need set this inside function

vita3k/ctrl/include/ctrl/state.h
```diff
@@ -60,6 +60,7 @@ struct CtrlState {
     ControllerList controllers;
     int controllers_num = 0;
     bool has_motion_support = false;
+    bool drop_inputs = false;
```
Zangetsu38, line 63: remove this here
after look code, set this here is totaly not needed, set it inside emuenv state directly

Zangetsu38:
Pr refactored and fixed in my proper pr.

PR #2940, opened 2023-09-22
shader, renderer: Optimize texture access

Macdu:
One of the drawbacks of the current implementation is that it has to do a lot of texture copies, when reading only a part of a rendered texture. I saw two ways to solve this:
- Use memory aliasing: the most accurate, however this does not work for many cases because of alignment constraints, also vulkan does not allow a stride to be specified for linear tiled images.
- Use a texture viewport: does not need any kind of extension, but is not entirely accurate because the border addressing mode won't be respected.
I used the second approach: using a texture viewport, but because this adds 16 additional bytes for each texture used in the uniform, a few changes had to be made

- Set the default uniform buffer at position 0 instead of 14, I can't optimize anything buffer-related in the uniform info if the most used buffer is at the maximal position
- Improve the Uniform info builder to support different kind of fields depending on the features enabled
- Add texture viewport to prevent most of the texture copies 
- Add a renderer accuracy option to the GUI, I had to make some changes to allow this option to be specified per game.

For the time being, the renderer accuracy only affects shader interlock and texture viewport, as they are two options that are more accurate but have a performance impact.


Zangetsu38 (dismissed):

Macdu:
This PR should be ready, I just fixed a small issue and removed some lines that are already in #2941.

vita3k/modules/SceGxm/SceGxm.cpp
```diff
@@ -3054,12 +3067,12 @@ EXPORT(int, sceGxmPrecomputedVertexStateSetAllUniformBuffers, SceGxmPrecomputedV
...
 
-    UniformBuffers *uniform_buffers = precomputedState->uniform_buffers.get(emuenv.mem);
+    UniformBuffer *uniform_buffers = precomputedState->uniform_buffers.get(emuenv.mem);
     if (!uniform_buffers)
         return RET_ERROR(SCE_GXM_ERROR_INVALID_POINTER);
 
-    for (auto b = 0; b < SCE_GXM_MAX_UNIFORM_BUFFERS; b++)
-        (*uniform_buffers)[b] = bufferDataArray[b];
+    for (int i = 0; i + 1 < precomputedState->buffer_count; i++)
+        uniform_buffers[i + SCE_GXM_UNIFORM_BUFFER_OFFSET] = bufferDataArray[i];
```
bookmist, line 3075: `i+1` and `i + SCE_GXM_UNIFORM_BUFFER_OFFSET` maybe in previous line also should be `i + SCE_GXM_UNIFORM_BUFFER_OFFSET`

vita3k/modules/SceGxm/SceGxm.cpp
```diff
@@ -3101,16 +3117,18 @@ EXPORT(int, sceGxmPrecomputedVertexStateSetUniformBuffer, SceGxmPrecomputedVerte
...
     if (!bufferData) {
         return RET_ERROR(SCE_GXM_ERROR_INVALID_POINTER);
     }
 
-    UniformBuffers &uniform_buffers = *precomputedState->uniform_buffers.get(emuenv.mem);
-    uniform_buffers[bufferIndex] = bufferData;
+    if (bufferIndex + 1 < precomputedState->buffer_count) {
+        UniformBuffer *uniform_buffers = precomputedState->uniform_buffers.get(emuenv.mem);
+        uniform_buffers[bufferIndex + SCE_GXM_UNIFORM_BUFFER_OFFSET] = bufferData;
+    }
```
bookmist, line 3131: same as prevoius

vita3k/renderer/src/vulkan/texture.cpp
```diff
@@ -92,19 +117,16 @@ void sync_texture(VKContext &context, MemState &mem, std::size_t index, SceGxmTe
...
+            renderer::SurfaceTextureRetrievePurpose::READING, swizzle, nullptr, nullptr, &texture_viewport);
     }
 
-    vk::ImageLayout layout = vk::ImageLayout::eShaderReadOnlyOptimal;
-    if (image) {
-        layout = vk::ImageLayout::eGeneral;
-    } else {
-        // Try to retrieve S24D8 texture
+    if (!image && is_depth_stencil_compatible_format(base_format)) {
+        // Try to retrieve depth/stencil
```
bookmist, line 124: Maybe add a warning if format is not compatible?

vita3k/shader/src/usse_program_analyzer.cpp
```diff
@@ -150,26 +150,25 @@ bool is_buffer_fetch_or_store(const std::uint64_t inst, int &base, int &cursor,
     return false;
 }
 
-int match_uniform_buffer_with_buffer_size(const SceGxmProgram &program, const SceGxmProgramParameter &parameter, const shader::usse::UniformBufferMap &buffers) {
-    assert(parameter.component_count == 0 && parameter.category == SCE_GXM_PARAMETER_CATEGORY_UNIFORM_BUFFER);
+int get_uniform_buffer_sizes(const SceGxmProgram &program, UniformBufferSizes &sizes) {
+    memset(sizes.data(), 0, sizeof(UniformBufferSizes));
```
bookmist, line 154: are you sure if `sizeof` here is correct. It can be true, but it looks strange for me

vita3k/renderer/src/vulkan/texture.cpp
Macdu: No, this path is taken by all textures (except color surfaces). This is just an optimisation. For example, if the game tries to read a BC3 or RGBA16 texture, we know that it can't be a depth or stencil surface so no need to check if its address matches something in the depth-stencil cache.

vita3k/shader/src/usse_program_analyzer.cpp
Macdu: From my points of view, this is correct. UniformBufferSizes is an alias for std::array<std::uint32_t, 15>, so sizeof(UniformBufferSizes) returns 60 and the memset has the expected behavior of setting all the entries of sizes to 0.

bookmist: I see. It just looks strange for me (` memset(&sizes, 0, sizeof(UniformBufferSizes));` or ` memset(sizes.data(), 0, sizes.size()*sizeof(sizes.data()));`)

bookmist (approved):

vita3k/shader/src/usse_program_analyzer.cpp
Macdu: Yes, &sizes points to the same location as sizes.data(), I agree that this can look a bit weird.

Zangetsu38 (approved):

PR #2941, opened 2023-09-23
vkutil: hide some warnings

bookmist:
Just hide some annoying warnings in VulkanMemoryAllocator-Hpp for clang

### DIFF
--- a/vita3k/vkutil/include/vkutil/vkutil.h
+++ b/vita3k/vkutil/include/vkutil/vkutil.h
@@ -24,7 +24,17 @@
 #include <vulkan/vulkan.hpp>
 #define VMA_STATIC_VULKAN_FUNCTIONS 0
 #define VMA_DYNAMIC_VULKAN_FUNCTIONS 1
+
+#if defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wnullability-completeness"
+#endif
+
 #include <vk_mem_alloc.hpp>
+
+#if defined(__clang__)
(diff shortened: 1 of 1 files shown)


vita3k/vkutil/include/vkutil/vkutil.h
```diff
@@ -24,8 +24,18 @@
...
+#if (defined(_WIN64) && defined(__clang__))
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wnullability-completeness"
+#endif
+
 #include <vk_mem_alloc.hpp>
 
+#if (defined(_WIN64) && defined(__clang__))
+#pragma clang diagnostic pop
+#endif
```
Macdu, line 37: Is the `defined(_WIN64)` necessary?

bookmist: This check was used by default for macos and ios (now it's on for all targets), So I think switching it on on windows is a bug in the library, and for macos it's desired behavior. No idea about linux. so maybe `!__APPLE__` is better

Macdu: The nullability warnings also appear on Linux (you can look at the log of the github ubuntu action), so I think if it does not cause any issue, just putting `#ifdef __clang__` would be fine (that's not an issue for it to be already defined on MacOS)

bookmist: you are right. done.

Macdu (approved):

Issue #1418, opened 2021-07-04
Flatpak build


LeandroStanger:
Please, consider supporting [Flatpak](http://docs.flatpak.org/en/latest/introduction.html#reasons-to-use-flatpak) and publishing Vita3K on [Flathub](https://github.com/flathub/flathub/wiki/App-Submission).

Doomsdayrs:
Second, An ubuntu-tar is irrelevant and a flatpak bundle can be used on more systems

Doomsdayrs:
Here is the way https://docs.flatpak.org/en/latest/single-file-bundles.html

qurious-pixel:
It is more than just bundling the app. 
flatpak has some read-only portions that Vita3K currently does not support.
It will fail to launch.

Doomsdayrs:
> It is more than just bundling the app. flatpak has some read-only portions that Vita3K currently does not support. It will fail to launch.

What read-only portions?

qurious-pixel:
https://docs.flatpak.org/en/latest/sandbox-permissions.html
Sandbox permissions prevent the creation of `shaderlog, config.yml, and vita3k.log`
The root folder of `data, shaders, shaders-builtin and Vita3k` are in a read-only sandbox when packaged as a flatpak.

Doomsdayrs:
You can drop a payload into the user data directory.

This is how the flatpak for WarThunder works

EXtremeExploit:
> https://docs.flatpak.org/en/latest/sandbox-permissions.html
> Sandbox permissions prevent the creation of `shaderlog, config.yml, and vita3k.log`
> The root folder of `data, shaders, shaders-builtin and Vita3k` are in a read-only sandbox when packaged as a flatpak.

This could be implemented to go into ./config or ./local with the right permissions and PR, still its not very good to have the emularor "unbox" just for these files

qurious-pixel:
This can be accomplished by placing the appropriate files/folders in `$XDG_DATA_HOME` and `$XDG_CONFIG_HOME`, but this will also affect Windows and MacOS.
I prefer this for Linux (it follows the XDG standard), but would require Windows users to look in `%LOCALAPPDATA%`

Doomsdayrs:
If there is a requirement for having it be portable for windows / MacOS, then there can just be a different variant with it set to portable mode.

P-C-R:
+1 for a flatpak build. For all the SteamDeck users out there

j8r:
[...]

PR #2944, opened 2023-09-26
cmake: some refactor and set on link time optimization

bookmist:
This changes were inspired by this https://github.com/Vita3K/Vita3K/issues/1418#issuecomment-1267773600 comment

I'm not sure if link time optimization is worth it, but it seems to have no problems.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,21 +63,23 @@ endmacro(b2_build)
 
 # Macro to compile Boost
 macro(boost_compile)
+    list(TRANSFORM BOOST_MODULES_TO_FIND PREPEND --with- OUTPUT_VARIABLE b2_cmd_line_common)
+    set(b2_cmd_line_common ${b2} ${b2_cmd_line_common} -j${CPU_COUNT} --build-dir=${BOOST_INSTALLDIR} --stagedir=${BOOST_INSTALLDIR} cxxflags=${BOOST_CXX_FLAGS})
 	if (CMAKE_SYSTEM_NAME STREQUAL "Windows")
 		execute_process(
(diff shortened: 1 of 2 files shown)


Macdu:
Nice PR, I would advise LTO to be enabled only for the release build (`if(CMAKE_BUILD_TYPE STREQUAL "Release")`) as it increases in a non-negligible way the compile time even when modifying only one file.

bookmist:
@Macdu ready for review.
I hope it will compile faster after first rebuild.
cmake variable for link time optimization supports 3 values: ALWAYS RELEASE_ONLY NEVER default value is RELEASE_ONLY

boost builds with LTO for modes ALWAYS and RELEASE_ONLY. I think it's not a problem to make non-LTO build with LTO static libraries. It's not an easy way to force rebuild boost in current environment, so this will be checked sometimes in future by a new developer.

Macdu:
> boost builds with LTO for modes ALWAYS and RELEASE_ONLY. I think it's not a problem to make non-LTO build with LTO static libraries. It's not an easy way to force rebuild boost in current environment, so this will be checked sometimes in future by a new developer.

Actually it is a problem, you can use non-lto libs in a lto project but not the opposite. The linker should crash if you were to try that.
The github CI does not use the boost executable (it uses boost from vcpkg instead), so I think you can make the boost build non-lto, it should be fine from my point of view (gcc also has the flag `-ffat-lto-objects` to build a lib containing both the lto and non-lto code), but if you are another idea you can try it.

Macdu:
Also one think that could be done (in another PR I guess) would be for the C++ packages used by the github CI to be built with LTO. I believe this is easily doable.

bookmist:
> Actually it is a problem, you can use non-lto libs in a lto project but not the opposite. The linker should crash if you were to try that. The github CI does not use the boost executable (it uses boost from vcpkg instead), so I think you can make the boost build non-lto, it should be fine from my point of view (gcc also has the flag `-ffat-lto-objects` to build a lib containing both the lto and non-lto code), but if you are another idea you can try it.

@Macdu I make the boost build non-lto. Now it's ready for review.

Macdu (approved):
If you want you can (in another PR) easily configure vcpkg to build LTO versions of the librairies.

PR #2946, opened 2023-09-26
renderer/vulkan, shader: Fix texture viewport without memory mapping

### LINKED ISSUE #2943: [Regression] macOS: PR #2940 broke FFX
Broken build: 1a912280 (#2940)
Last working build: 810c3aea 

1a912280: 
Screenshot 2023-09-26 at 10 18 04
Screenshot 2023-09-26 at 10 18 40

810c3aea: 

Screenshot 2023-09-26 at 10 24 56
Screenshot 2023-09-26 at 10 25 17

Macdu:
Fix an issue where the offset in the generated shader was wrong when using texture viewport without memory mapping. Also add memory mapping support in the feature mask. This should fix #2943 .

The second commit sets by default the renderer accuracy to high. On desktop, performance is already fine with the accuracy set to high (except when upscaling a lot I guess), so it does not make too much sense to set it to standard by default.

### DIFF
--- a/vita3k/config/include/config/config.h
+++ b/vita3k/config/include/config/config.h
@@ -68,7 +68,7 @@ enum PerfomanceOverleyPosition {
     code(bool, "archive-log", false, archive_log)                                                       \
     code(std::string, "backend-renderer", "OpenGL", backend_renderer)                                   \
     code(int, "gpu-idx", 0, gpu_idx)                                                                    \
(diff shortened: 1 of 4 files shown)


Zangetsu38 (approved):

KhoraLee:
Screenshot 2023-09-27 at 1 24 28 PM
#2943 looks fixed

PR #2954, opened 2023-09-30
renderer/vulkan: Add support for macroblock sync when a scissor is not specified

Macdu:
The game flower does not specify a scissor matching the current macroblock, so as a fallback this PR implement a workaround (quite inefficient, but it can't really do better, render each draw in its own render pass).

This fixes the graphical regressions in flower and also the blur issues (like other games using macroblock sync).


ArbestRiAgain:
Will this also fix other games?

Macdu:
It shouldn't, flower had a really specific regression were the background was turning black.

Zangetsu38 (approved):

PR #2955, opened 2023-10-01
https: fix resume download with small refactor.

Zangetsu38:


Macdu (approved):
Too bad there isn't any library available which would allow you to download a file from a URL in one line instead of writing a 500-line file for it which sometimes does not work.

Zangetsu38:
I lknow, i have just wanted using one things already inside project

PR #2957, opened 2023-10-02
vita3k: Drop inputs while any settings or controls dialog is open.

Zangetsu38:
Based on https://github.com/Vita3K/Vita3K/pull/2937 but totaly refactored and fixed with lower change
# About
- vita3k: Drop inputs while any settings or controls dialog is open.
Allow app and gui no take any input anymore when this dialogs is open.

### DIFF
--- a/vita3k/ctrl/src/ctrl.cpp
+++ b/vita3k/ctrl/src/ctrl.cpp
@@ -279,6 +279,7 @@ static void retrieve_ctrl_data(EmuEnvState &emuenv, int port, bool is_v2, bool n
     axes.fill(0);
 
     const auto reset_axes = [&]() {
+        // Re-center joysticks to (128,128). Range is (0-255,0-255).
         SceCtrlPadInputMode mode = from_ext_function ? state.input_mode_ext : state.input_mode;
         if (mode == SCE_CTRL_MODE_DIGITAL) {
             lx = 0x80;
@@ -291,11 +292,11 @@ static void retrieve_ctrl_data(EmuEnvState &emuenv, int port, bool is_v2, bool n
             rx = float_to_byte(axes[2]);
             ry = float_to_byte(axes[3]);
         }
-    };
-
-    if (emuenv.common_dialog.status == SCE_COMMON_DIALOG_STATUS_RUNNING) {
         if (negative)
             buttons ^= ~0;
+    };
+
+    if ((emuenv.common_dialog.status == SCE_COMMON_DIALOG_STATUS_RUNNING) || emuenv.drop_inputs) {
         reset_axes();
         return;
     }
@@ -311,9 +312,6 @@ static void retrieve_ctrl_data(EmuEnvState &emuenv, int port, bool is_v2, bool n
     }
 
     reset_axes();
-
-    if (negative)
-        buttons ^= ~0;
 }
 
 int ctrl_get(const SceUID thread_id, EmuEnvState &emuenv, int port, SceCtrlData2 *pData, SceUInt32 count, bool negative, bool is_peek, bool is_v2, bool from_ext) {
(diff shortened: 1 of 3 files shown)


vita3k/emuenv/include/emuenv/state.h
```diff
@@ -137,6 +137,7 @@ struct EmuEnvState {
     IVector2 drawable_size = { 0, 0 };
     FVector2 viewport_pos = { 0, 0 };
     FVector2 viewport_size = { 0, 0 };
+    bool drop_inputs{};
```
Macdu, line 140: Put this in the ctrl state

vita3k/ctrl/src/ctrl.cpp
```diff
@@ -280,11 +281,11 @@ static void retrieve_ctrl_data(EmuEnvState &emuenv, int port, bool is_v2, bool n
...
             ry = float_to_byte(axes[3]);
         }
-    };
-
-    if (emuenv.common_dialog.status == SCE_COMMON_DIALOG_STATUS_RUNNING) {
         if (negative)
             buttons ^= ~0;
+    };
+
+    if ((emuenv.common_dialog.status == SCE_COMMON_DIALOG_STATUS_RUNNING) || emuenv.drop_inputs) {
```
Macdu, line 288: Why the case (emuenv.common_dialog.status == SCE_COMMON_DIALOG_STATUS_RUNNING) is not taken into account in drop_inputs?

Zangetsu38: ? Is just normal formating code ?
If have one == and adding other bool with OR i need separe it

vita3k/emuenv/include/emuenv/state.h
Zangetsu38: No relate ctrl state att all, it only relate emuenv

vita3k/ctrl/src/ctrl.cpp
Macdu: drop_inputs only purpose is to be set to true when you drop the input. But when there is a common dialog running you drop the input but don't set it to true.

vita3k/emuenv/include/emuenv/state.h
Macdu: I don't know, this variable is read only once in the function retrieve_ctrl_data in the file ctrl.cpp, it kind of seems like it is related to the controller state.

vita3k/ctrl/src/ctrl.cpp
Zangetsu38: Ah i understund sorry
So yes is normal, common dialog is diferent of other dialog relate drop input
Need 2 separate for good reason
No forget common dialog using (for save dialog only for now)  input for controle it with input
If you look interface, i only drop input for settings dialog, not for common dialog, is normal

vita3k/emuenv/include/emuenv/state.h
Zangetsu38: no no at all look better, this variable is used and defined at the emulator level, in the interface file, the ctrl is only for reading, I mean this variable is assigned for vita3k itself, and not not related to the ctrl state

vita3k/ctrl/src/ctrl.cpp
Macdu: Don't name a variable drop_input if it being set to false doesn't mean we don't drop input.

vita3k/emuenv/include/emuenv/state.h
Macdu: Can you tell me what part of a variable meaning that we need to return nothing when the game asks for controller inputs is not related to the controller state?
Also the emuenv/state.h file is included by 400 files, it should contain nothing but pointers to other states.

Zangetsu38: It include already very lot bool and very lot variable, again, this variable not relate ctrl att all, relate only emu env side for drop jnout when gui dialog is open

vita3k/ctrl/src/ctrl.cpp
Zangetsu38: Eu what ? Is not my name and what ? 🤔
Ifcis false is normal it no drop 🤔

bookmist (approved):

bookmist:
I think macdu is right and `drop_inputs` should not be in `emuenv`, But it will be easier to change it myself when I'll refactor emuenv.

PR #2970, opened 2023-10-06
github_cache: cleanup cache on windows

bookmist:
Try to fix not using cache on windows. step one - delete cache and create it again


Macdu:
Fixed by #3031

PR #2974, opened 2023-10-08
https: Add MD5 checksum verification file using Content-MD5 header.

Zangetsu38:
# About
- https: Add MD5 checksum verification file using Content-MD5 header.
should fix corrupted file download with checksum integrity before try using it.

### DIFF
--- a/vita3k/https/src/https.cpp
+++ b/vita3k/https/src/https.cpp
@@ -30,6 +30,7 @@ typedef int abs_socket;
 #endif
 
 #include <openssl/err.h>
+#include <openssl/md5.h>
 #include <openssl/ssl.h>
 
 #include <util/log.h>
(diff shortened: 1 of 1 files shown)


Macdu:
Seriously, instead of writing your own implementation of something that is (indeed) very hard (communicating with a server and downloading a file) and having to fix it every two weeks, just take one of the 20 available header-only library already doing this and replace all of that with a single function call.

Zangetsu38:
> Seriously, instead of writing your own implementation of something that is (indeed) very hard (communicating with a server and downloading a file) and having to fix it every two weeks, just take one of the 20 available header-only library already doing this and replace all of that with a single function call.

I see what you mean, but I don't want to burden the project with another submodule, this code does the same thing, and even with libcurl, it's not as simple as you make it think, you don't just need a single call function, far from it, there's almost as much code in the end, and even there I should add code to verify the md5.

And also, this commit has no relation to my previous fix.
I've been wanting to do this md5 check for a long time.

Macdu:
https://github.com/yhirose/cpp-httplib

This is a header only library using openssl and it takes less than 5 lines to download a file.

Macdu (approved):

PR #2975, opened 2023-10-08
gxm, renderer: Rewrite texture uploading

Macdu:
Rewrite the texture uploading part of the renderer, among other things:
- Fix the offset between faces for cube textures 
- Fix unsupported textures not being untiled or unswizzled if necessary
- Better logging of cases not (yet) handled by the renderer
- Fix Mip offset for linear textures
- Clean up the upload part of the code
- Remove some useless functions, making use of opengl/vulkan capabilities
- Fix offset for YUV planes with non-zero mip
- Upload 2x2 and 1x1 mip for BCn textures (they were previously ignored)
- Have a function which returns the exact texture size (will be necessary for texture dumping/replacement)

This should fix many games which had graphical issues previously. I know this fixes graphical issues for Tales Of Hearts and FIFA.

As a side not, the current code decompresses YUV420P2 textures as YUV420P3 which is wrong. I wanted to fix this, however, this caused all the fmv to be corrupted, because the video decoder always output a YUV420P3 frame instead of what the game asks for... So I'll fix it in another PR.


vita3k/gxm/src/textures.cpp
```diff
@@ -17,47 +17,280 @@
...
 
-size_t get_height(const SceGxmTexture *texture) {
-    if ((texture->type << 29 != SCE_GXM_TEXTURE_SWIZZLED) && (texture->type << 29 != SCE_GXM_TEXTURE_CUBE)) {
-        return texture->height + 1;
+uint32_t get_height(const SceGxmTexture &texture) {
+    if (texture.texture_type() != SCE_GXM_TEXTURE_SWIZZLED && texture.texture_type() != SCE_GXM_TEXTURE_CUBE) {
+        return texture.height + 1;
     }
-    return 1ull << (texture->height_base2 & 0xF);
+    return 1ull << texture.height_base2;
```
bookmist, line 36:
```suggestion
    return 1u << texture.height_base2;
```
to the same code as in ` get_width`

vita3k/renderer/src/gl/texture.cpp
```diff
@@ -50,19 +50,14 @@ void GLTextureCache::select(size_t index, const SceGxmTexture &texture) {
...
     const SceGxmTextureAddrMode vaddr = (SceGxmTextureAddrMode)(gxm_texture.vaddr_mode);
-    auto width = static_cast<uint32_t>(gxm::get_width(&gxm_texture));
-    auto height = static_cast<uint32_t>(gxm::get_height(&gxm_texture));
-    if (gxm::is_block_compressed_format(base_format)) {
-        // align width and height to block size
-        width = (width + 3) & ~3;
-        height = (height + 3) & ~3;
-    }
+    auto width = static_cast<uint32_t>(gxm::get_width(gxm_texture));
+    auto height = static_cast<uint32_t>(gxm::get_height(gxm_texture));
```
bookmist, line 58: It seems like function return types are already uint32_t

bookmist (dismissed):

Macdu:
This should be ready. I replaced the hash used by the texture cache to look only at the first mip. This is more similar to what was being done previously and what other emulators are doing (anyway, higher mips should almost always be scaled down versions of the first mip).

Zangetsu38:
all code warning no need to fix before ?

Macdu:
Not relevant, they are the same code warnings as before (they just poped again because I changed the line) and are not relevant, we know everything will fit inside 32-bit integers.

PR #2981, opened 2023-10-10
vita3k: Use LOG_..._ONCE when appropriate

Macdu:
Replace all previous use of a static variables with a call to LOG_***_ONCE which looks much better.

Also replace the calls to STUBBED, UNIMPLEMENTED and RET_ERROR to use the LOG_***_ONCE version instead of locking a mutex and looking in a map each time they are called (which is very inefficient). This should help performance.


Zangetsu38 (dismissed):

vita3k/util/include/util/log.h
```diff
@@ -63,24 +57,33 @@ ExitCode add_sink(const fs::path &log_path);
...
 
 #define LOG_ONCE(log_function, ...)    \
-    do {                               \
+    {                                  \
         static bool LOG_DONE = false;  \
         if (!LOG_DONE)                 \
             log_function(__VA_ARGS__); \
         LOG_DONE = true;               \
-    } while (0)
+    }
```
bookmist, line 65: then `;` after block become a warning. Such loop is a special kind of expression to use in macros, it is a block after that `;` should be (after `{ ... }` should not be a `;`)

vita3k/module/include/module/module.h
```diff
@@ -21,11 +21,17 @@
...
-int unimplemented_impl(const char *name);
-#define UNIMPLEMENTED() unimplemented_impl(export_name)
-
-int stubbed_impl(const char *name, const char *info);
-#define STUBBED(info) stubbed_impl(export_name, info)
+#define UNIMPLEMENTED()                                                \
+    ([&]() {                                                           \
+        LOG_WARN_ONCE("Unimplemented {} import called.", export_name); \
+        return 0;                                                      \
+    })()
```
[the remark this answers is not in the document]
Macdu: The compiler is clever enough to optimize these two as the same things (also it is already clever enough to completely remove the lambda).

vita3k/util/include/util/log.h
Macdu: Oh, I didn't know ty, I will change that the the way it was previously.

vita3k/renderer/src/vulkan/sync_state.cpp
```diff
@@ -249,10 +249,7 @@ void sync_visibility_index(VKContext &context, bool enable, uint32_t index, bool
     }
 
     if (index >= context.current_visibility_buffer->size) {
-        static bool has_happened = false;
-        LOG_WARN_IF(!has_happened, "Using visibility index {} which is too big for the buffer");
-        has_happened = true;
-
+        LOG_WARN_ONCE("Using visibility index {} which is too big for the buffer");
```
bookmist, line 252: what index? there is a `{}` and no parameter

bookmist (approved):

PR #2991, opened 2023-10-12
gui: check if app is found in index for log in init_app_background.

Zangetsu38:
# About
- gui: check if app is found in index for log in init_app_background.
should fix crash when fw is missing.

### DIFF
--- a/vita3k/gui/src/gui.cpp
+++ b/vita3k/gui/src/gui.cpp
@@ -388,14 +388,16 @@ void init_app_background(GuiState &gui, EmuEnvState &emuenv, const std::string &
     else
         vfs::read_app_file(buffer, emuenv.pref_path, app_path, "sce_sys/pic0.png");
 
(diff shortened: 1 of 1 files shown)


bookmist (approved):

vita3k/gui/src/gui.cpp
```diff
@@ -388,14 +388,16 @@ void init_app_background(GuiState &gui, EmuEnvState &emuenv, const std::string &
     else
         vfs::read_app_file(buffer, emuenv.pref_path, app_path, "sce_sys/pic0.png");
 
+    const auto title = (APP_INDEX != gui.app_selector.sys_apps.end()) && (APP_INDEX != gui.app_selector.user_apps.end()) ? APP_INDEX->title : app_path;
```
Macdu, line 391: That will always be false

Zangetsu38: is not works fine for me

PR #2993, opened 2023-10-12
shader: Increase repeat offset for GPI on mode both

Macdu:
The RepeatMode Both means that both internal and external register must be increased during each step.

This (partially) fixes graphics in Chaos Rings 2.


EXtremeExploit (approved):
oke

Zangetsu38 (approved):

PR #2996, opened 2023-10-15
renderer/vulkan: Rewrite surface cache

Macdu:
The current vulkan surface cache was basically a copy-paste of the opengl surface cache, swapping opengl functions with vulkan functions.
But it had many issues related to the fact it was really complex and to the way vulkan is a lot closer to the harware:
- create two different functions for retrieving surfaces either as a framebuffer or as a texture. Make the code much more readable and easy to modify.
- Refractor the result given when retrieving textures from the surface cache so that a whole new Image field does not have to be created when all we want is a different view
- The sampler state is different compared to the texture state in Vulkan. So create a separate sampler cache for it. This avoids duplication of textures if the only thing which changes is the filter or lod parameter. This will also be needed for my next PR.
- Rewrite the way removing the least recently used texture is done (using a vector is pretty bad for it, even if the cache is small).
- A few other fixes here and there I did while rewriting the cache.

This shouldn't do much on its own (maybe reduce memory usage as it avoids some texture duplication (and a lot of sampler duplication but that shouldn't matter) but is needed for future improvements.

### DIFF
--- a/vita3k/util/include/util/containers.h
+++ b/vita3k/util/include/util/containers.h
@@ -1,0 +1,99 @@
+// Vita3K emulator project
+// Copyright (C) 2023 Vita3K team
+//
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation; either version 2 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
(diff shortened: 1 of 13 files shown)


Zangetsu38 (approved):

vita3k/util/include/util/containers.h
```diff
@@ -0,0 +1,99 @@
...
+    // get the least recently used element
+    T *get_lru() const {
+        // it is at the end of the list (so before the first one)
+        return &head->prev->content;
+    }
+
+    // set an element as the most recently used
+    void set_as_mru(T *ptr) {
+        // get the item from a pointer to its content
+        // removed the sizeof to avoid compilation warnings
```
bookmist, line 73: Maybe "removed offsetof ..."

Macdu: indeed

Zangetsu38 (approved):

Macdu:
This PR should be ready, it contains a fix for persona 4 (the reason persona 4 was rendering correctly previously was because of a surface cache bug which I removed).

Zangetsu38 (approved):

PR #2997, opened 2023-10-15
tracy: autoregister modules, faster check if module is profiling

bookmist:
Main idea of this PR is optimization of tracy usage
* automatic collection of modules list for advanced profiling. Remove available modules list from config.
* faster check if module is active for advanced profiling
* rename some macro not used yet to correspond tracy naming convention
* disable unused tracy option to save screenshots on demand. This option create background thread to compress that images, so it's better to disable this if nobody use it.
* some minor code improvements

I use global variables because I'm lazy.

### DIFF
--- a/vita3k/gui/src/settings_dialog.cpp
+++ b/vita3k/gui/src/settings_dialog.cpp
@@ -1119,25 +1119,25 @@ void draw_settings_dialog(GuiState &gui, EmuEnvState &emuenv) {
 
         // Tracy modules list
         if (ImGui::BeginListBox(tracy_modules_list_label.c_str(), { 0.0f, ImGui::GetTextLineHeightWithSpacing() * 8.25f + ImGui::GetStyle().FramePadding.y * 2.0f })) {
+            // Get all HLE modules available for advanced profiling using Tracy. Do it only once.
+            static std::vector<std::string> tracy_modules = tracy_module_utils::get_available_module_names();
             // For every HLE module available for advanced profiling using Tracy
-            for (auto &module : emuenv.cfg.tracy_available_advanced_profiling_modules) {
-                // Get activation state and position in vector of activated modules
-                bool activation_state = false;
-                int index = -1;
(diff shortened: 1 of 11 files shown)


vita3k/util/src/util.cpp
```diff
@@ -194,6 +195,66 @@ void register_log_exception_handler() {}
...
+typedef std::bitset<max_modules> tracy_module_flags;
+typedef std::vector<std::string> tracy_module_names;
+
+tracy_module_names tracy_available_advanced_profiling_modules{};
+tracy_module_flags tracy_advanced_profiling_modules{};
+
+tracy_module_helper::tracy_module_helper(const char *module_name) {
+    mod_id = tracy_available_advanced_profiling_modules.size();
+    tracy_available_advanced_profiling_modules.emplace_back(module_name);
+    if (mod_id >= tracy_advanced_profiling_modules.size()) {
```
Macdu, line 213:
```suggestion
    if (mod_id >= max_modules) {
```

vita3k/gui/src/settings_dialog.cpp
```diff
@@ -1119,25 +1119,27 @@ void draw_settings_dialog(GuiState &gui, EmuEnvState &emuenv) {
 
         // Tracy modules list
         if (ImGui::BeginListBox(tracy_modules_list_label.c_str(), { 0.0f, ImGui::GetTextLineHeightWithSpacing() * 8.25f + ImGui::GetStyle().FramePadding.y * 2.0f })) {
+            // Get all HLE modules available for advanced profiling using Tracy. Do it only once.
+            static std::vector<std::string> tracy_modules{};
+            if (tracy_modules.empty())
+                tracy_module_utils::get_available_module_names(tracy_modules);
```
Macdu, line 1125:
```suggestion
            static std::vector<std::string> tracy_modules = tracy_module_utils::get_available_module_names(tracy_modules);
```
Will be called only once

bookmist:
Done

Macdu (approved):

PR #3005, opened 2023-10-18
ci: Dont bundle OpenSSL libs anymore

EXtremeExploit:
Older version of OpenSSL (1.1.0/1.1.1 LTS) are not maintained anymore since 9/11 (i swear this is a coincidence). So it makes sense to not bundle them anymore as they are not needed since at this point (1 year later since the switch to 3.x of mostly all distros) should have already updated from it.
Gonna leave as draft to just to make sure there isnt any major distro which still hasnt updated (SteamOS, LTS version of some distro, etc)

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -116,14 +116,6 @@ jobs:
         run: echo "build_variable=$(git rev-list HEAD --count)" >> $GITHUB_ENV
         if: matrix.os == 'ubuntu-latest'
 
-      - name: Bundle OpenSSL shared objects
-        id: bundle_openssl
-        run: |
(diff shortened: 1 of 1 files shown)


scribam:
- Ubuntu LTS (20.04) will stay with 1.x: https://ubuntu.com/blog/running-openssl-1-1-1-after-eol-with-ubuntu-pro. Consumer EOL is 2025.
- NixOS 24.05 has openssl 3.3.1: https://github.com/NixOS/nixpkgs/blob/086b448a5d54fd117f4dc2dee55c9f0ff461bdc1/pkgs/development/libraries/openssl/default.nix#L329

Zangetsu38 (approved):

PR #3009, opened 2023-10-20
renderer: Implement texture replacement

Macdu:
Allows textures to be exported and imported on the fly using Vita3K. This allows texture packs to be created for games running using Vita3K.
There are two formats available for dumping textures:
- png, easier to handle because any image viewer/editor can view/edit it. However it has a few drawbacks: loosing texture quality because of compression and conversion to 8-bit per component format. This format also does not support mipmaps and cubemaps.
- dds, contains the raw texture (but if the raw texture was compressed, it will still be compressed in the dds file). it also supports mipmaps and cubemaps and more formats. But is harder to use (I basically need to use Gimp to look at the dumped textures).

It is possible to import textures using both the dds and png formats. In case both are present, the dds one will be preferred.

Texture exportation does not use any renderer (OpenGL/Vulkan) code and is completely rendering API agnostic. Texture importation works on both OpenGL and Vulkan (and most of the code is rendering API agnostic too).

textures are dumped to the shared folder / textures / export / game_id / hash.dds|png
Any texture in shared folder / textures / import / game_id / hash.dds|png will be imported provided the option is checked in the settings.

The options to enable importation/exportation and choose the export format are in the GPU tab of the settings.


Zangetsu38 (dismissed):

Macdu:
I improved the hashing function for texture dumping to never look at unsused pixels and corrected the streaming hash algorithm to use XXH3 instead of XXH64. The PR should be ready now.

EXtremeExploit (approved):

Zangetsu38 (approved):

PR #3011, opened 2023-10-20
shader: Fix vcomp with multiple dest components

Macdu:
The result of vcomp was only stored in the first possible component, even when the mask had multiple components.
This fixes the menu of flower
image

### DIFF
--- a/vita3k/shader/src/translator/alu.cpp
+++ b/vita3k/shader/src/translator/alu.cpp
@@ -700,7 +700,7 @@ bool USSETranslatorVisitor::vcomp(
     inst.opr.dest.swizzle = SWIZZLE_CHANNEL_4_DEFAULT;
 
     // TODO: Should we do this ?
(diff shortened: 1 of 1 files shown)


Zangetsu38:
Ah finaly fix red color, can maybe fix sale’issue on few atelier game

vita3k/shader/src/translator/alu.cpp
```diff
@@ -726,12 +726,18 @@ bool USSETranslatorVisitor::vcomp(
 
     default: break;
     }
+    // (upper bound on the) number of components to set in the destination
+    const uint32_t nb_components = std::bit_width(write_mask);
 
     m_b.setLine(m_recompiler.cur_pc);
 
     // TODO: Log
```
Zangetsu38, line 734: No delete it if tou set log ?

Zangetsu38 (approved):

PR #3020, opened 2023-10-23
gui/lang: Updated translations and fixed typos.

Croden1999:
- gui/lang: Updated translations and fixed typos.
- data/fonts: Add a font file, and also included license. 

Please devs review this PR.


vita3k/gui/src/gui.cpp
```diff
@@ -213,8 +213,8 @@ static void init_font(GuiState &gui, EmuEnvState &emuenv) {
     ImFontGlyphRangesBuilder builder;
     builder.AddRanges(io.Fonts->GetGlyphRangesJapanese());
     builder.AddRanges(extra_range);
-    ImVector<ImWchar> japanes_and_extra_ranges;
-    builder.BuildRanges(&japanes_and_extra_ranges);
+    ImVector<ImWchar> japanese_and_extra_ranges;
```
Zangetsu38, line 216: Oupsy 🤦‍♂️

nishinji:
Is it too much of a disadvantage to increase the size of the font file by a factor of ten or more just for Chinese?
Besides, it is only used when no official font exists.
up: master 
down: pr
image

Croden1999:
> Is it too much of a disadvantage to increase the size of the font file by a factor of ten or more just for Chinese? Besides, it is only used when no official font exists. up: master down: pr image

the Noto Sans CJK font include Korean characters, but i no test it.

nishinji:
The problem is not that Chinese is the only language that benefits, but that the build is too large
It would be better to consider using the system's fonts

CoffeeBrewer64:
could this be put into an option (a tick box) in the gui? on setup you could ask if you want to download a font with more languages supported, and when the user switches to chinese and korean, ask if they want to download it as well. it could also be kept as a button in the system config. this way, the build size stays the same and the user still gets the font they want/need.

Croden1999:
> could this be put into an option (a tick box) in the gui? on setup you could ask if you want to download a font with more languages supported, and when the user switches to chinese and korean, ask if they want to download it as well. it could also be kept as a button in the system config. this way, the build size stays the same and the user still gets the font they want/need.

no sense, just for full character range

EXtremeExploit:
I kinda agree with nishinji in this case, it adds a whole 10MB of disk space to the build, its nice to fix the full chinese and korean characters, but is there another font option that doesnt change the build size by about 10MB?

Croden1999:
> I kinda agree with nishinji in this case, it adds a whole 10MB of disk space to the build, its nice to fix the full chinese and korean characters, but is there another font option that doesnt change the build size by about 10MB?

If wanna to be compatible with CJK fonts, only need change the file

Croden1999:
Ready for review

Croden1999:
> I've fixed typos in Indonesian and errors in Lang

if you wanna a translators list, no need add your name in lang.xml file

Kitakatarashima:
Oh yes yes i I'll fix it right away

Kitakatarashima:
Where have you been??

Croden1999:
> Where have you been??

i doing other things. by the way, you can fork the Vita3K repo, create a new branch and switch, upload your lang.xml file(s) and push a new PR, it's very easily.

Kitakatarashima:
It doesn't support YML file

Kitakatarashima:
Sorry, I'm still new to GitHub so I don't really know how to do it

Zangetsu38 (approved):

Macdu (approved):
The 10% Vita3K file increase is a bit annoying but I guess it's still reasonable

PR #3022, opened 2023-10-25
motion: Workaround for steam deck

### LINKED ISSUE #3019: (Steam Deck) Game crashes when DualShock 4 is connected without Steam Input
Attempting to play Gravity Rush and have had no issues playing in handheld. (The Deck’s gyroscope doesn’t work, but as far as I understand, it simply isn’t supported and that’s expected)

Playing with a DS4 controller also works “fine”, but Steam Input overrides the native touchpad and gyroscope support. When disabling Steam Input and having Vita3K handle the controller natively, the game crashes directly after the shaders compile.

I know that Vita3K is properly recognizing the controller. I can go in desktop mode and see It listed under Vita3K’s “controller” tab, and I changed the DS4 light color so that I could know when the controller’s recognized. It is properly connected in the emulators UI, and when games are being launched (during shader compilation), in both desktop, and game mode. 

No matter what controller I am using, if a single DS4 without Steam Input is connected, games will not launch. Games still boot perfectly fine otherwise.

If this detail is important, I have installed Vita3K through Emudeck. I am on the latest build.

Macdu:
The steam is using SDL 2.24 while Vita3K relies on SDL_GameControllerGetSensorDataWithTimestamp which is only implemented in SDL 2.26. 
As a workaround, detect at runtime the SDL version and only use this function when it is possible.

Should fix #3019

### DIFF
--- a/vita3k/motion/src/motion.cpp
+++ b/vita3k/motion/src/motion.cpp
@@ -21,6 +21,7 @@
 
 #include <ctrl/state.h>
 
+#include <SDL.h>
 #include <SDL_gamecontroller.h>
 
 SceFVector3 get_acceleration(const MotionState &state) {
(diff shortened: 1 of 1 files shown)


Zangetsu38:
Is on this german said need using sdl event and no function

Macdu:
I'll look at this when we'll update to SDL 3 which won't be the case until it is officially released.

Zangetsu38:
> I'll look at this when we'll update to SDL 3 which won't be the case until it is officially released.

ok but here variable _timestamp is so just ignored 🤔 for < 2.26 so so no can causing probleme on value of gyro and accel ?
@german77

nn9dev:
I can confirm this works on SteamOS 3.4. Gravity Rush boots and gyro is functional with a PS4 controller attached and with Steam open or closed.

Macdu:
Not using the timestamp function is just less precise. The (recurring) issue here is that steam deck is using outdated shared librairies.

Zangetsu38 (approved):

PR #3037, opened 2023-10-30
renderer/vulkan: Improve support of MSAA without a color surface

Macdu:
First, indeed using MSAA without a color surface on the PS Vita is completely useless (the depth stencil is never downscaled). The only thing it changes is the scaling of the fragment coordinate in the fragment shader I guess.

FFX remaster uses it. This fixes the crash when launching it with the vulkan renderer.

### DIFF
--- a/vita3k/renderer/src/vulkan/context.cpp
+++ b/vita3k/renderer/src/vulkan/context.cpp
@@ -111,7 +111,7 @@ void set_context(VKContext &context, MemState &mem, VKRenderTarget *rt, const Fe
 
         // set back default values
         vk_format = vk::Format::eR8G8B8A8Unorm;
-        context.record.color_surface.downscale = false;
+        context.record.color_surface.downscale = static_cast<bool>(rt->multisample_mode);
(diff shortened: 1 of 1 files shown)


vita3k/renderer/src/vulkan/context.cpp
```diff
@@ -125,7 +125,7 @@ void set_context(VKContext &context, MemState &mem, VKRenderTarget *rt, const Fe
     }
 
     SceGxmDepthStencilSurface *ds_surface_fin = &context.record.depth_stencil_surface;
-    if ((ds_surface_fin->depth_data.address() == 0) && (ds_surface_fin->stencil_data.address() == 0)) {
+    if (ds_surface_fin->depth_data.address() == 0 && ds_surface_fin->stencil_data.address() == 0) {
```
Zangetsu38, line 128: Why remove this ? Is totaly good and correct like is it all continue like this request ()

Zangetsu38 (approved):

PR #3043, opened 2023-11-01
mem: Improve accuracy of virtual memory

Macdu:
Improve how the virtual memory is handled to match better what is happening on a real PS Vita.
Right now, the memory is allocated starting from 0 whereas on the PS Vita, the user main memory is allocated starting from 0x81000000. This can have some impact as games can expect addresses to be negative. 
I'm starting the user address at 0x80000000 instead of 0x81000000 because it shouldn't make a difference and games compiled with old SDKs want the elf segment to be exactly as 0x81000000 so doing so makes sure this memory range is not allocated for whatever reason before.

This fixes a crash in Trails in the sky the 3rd.

### DIFF
--- a/vita3k/mem/include/mem/functions.h
+++ b/vita3k/mem/include/mem/functions.h
@@ -24,6 +24,8 @@ struct MemState;
 
 typedef std::function<bool(uint8_t *addr, bool write)> AccessViolationHandler;
 
+constexpr Address user_main_memory_start = 0x80000000U;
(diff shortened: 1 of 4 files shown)


Zangetsu38 (approved):

Moobeeart:
@Macdu Hi Macdu, how and where can I change the user main memory / user address on vita3k that you have mentioned in the post above?

PR #3057, opened 2023-11-08
renderer, gui: Add keyboard shortcut to toggle texture replacement

Macdu:
At the request of a third-party group, add a keyboard shortcut to toggle texture replacement (enable/disable texture importation). 
Also contains some code cleanup for accessing the texture cache.


EXtremeExploit (approved):
sure

PR #3058, opened 2023-11-08
io: Fix export/import textures path being same place as exec in linux

EXtremeExploit:
This fixed imported/exported textures being in the same path as the executable (in linux) by default, however, this come at the super small cost of not being able to change the base path from where gui font, lang and builtin shaders get loaded from, shouldnt be a problem since i really see no reason for someone to use it, now the textures folder is `.local/share/Vita3K/textures` by default, shared_path should never be used for stuff thats meant to be inside the executable folder, and base_path only for reading files, never writing

### DIFF
--- a/vita3k/app/src/app_init.cpp
+++ b/vita3k/app/src/app_init.cpp
@@ -118,22 +118,22 @@ void init_paths(Root &root_paths) {
 #if defined(__linux__) && !defined(__ANDROID__) && !defined(__APPLE__)
     // XDG Data Dirs.
     auto env_home = getenv("HOME");
-    auto env_data_dirs = getenv("XDG_DATA_DIRS");
-    auto env_data_home = getenv("XDG_DATA_HOME");
-    auto env_cache_home = getenv("XDG_CACHE_HOME");
-    auto env_config_home = getenv("XDG_CONFIG_HOME");
+    auto XDG_DATA_DIRS = getenv("XDG_DATA_DIRS");
+    auto XDG_DATA_HOME = getenv("XDG_DATA_HOME");
+    auto XDG_CACHE_HOME = getenv("XDG_CACHE_HOME");
+    auto XDG_CONFIG_HOME = getenv("XDG_CONFIG_HOME");
 
-    if (env_data_home != NULL)
-        root_paths.set_pref_path(fs::path(env_data_home) / app_name / app_name / dir_sep);
+    if (XDG_DATA_HOME != NULL)
+        root_paths.set_pref_path(fs::path(XDG_DATA_HOME) / app_name / app_name / dir_sep);
 
-    if (env_config_home != NULL)
-        root_paths.set_config_path(fs::path(env_config_home) / app_name / dir_sep);
+    if (XDG_CONFIG_HOME != NULL)
+        root_paths.set_config_path(fs::path(XDG_CONFIG_HOME) / app_name / dir_sep);
     else if (env_home != NULL)
         root_paths.set_config_path(fs::path(env_home) / ".config" / app_name / dir_sep);
 
-    if (env_cache_home != NULL) {
-        root_paths.set_cache_path(fs::path(env_cache_home) / app_name / dir_sep);
-        root_paths.set_log_path(fs::path(env_cache_home) / app_name / dir_sep);
+    if (XDG_CACHE_HOME != NULL) {
+        root_paths.set_cache_path(fs::path(XDG_CACHE_HOME) / app_name / dir_sep);
+        root_paths.set_log_path(fs::path(XDG_CACHE_HOME) / app_name / dir_sep);
     } else if (env_home != NULL) {
         root_paths.set_cache_path(fs::path(env_home) / ".cache" / app_name / dir_sep);
         root_paths.set_log_path(fs::path(env_home) / ".cache" / app_name / dir_sep);
@@ -141,21 +141,37 @@ void init_paths(Root &root_paths) {
 
     // Don't assume that base_path is portable.
     if (fs::exists(root_paths.get_base_path() / "data") && fs::exists(root_paths.get_base_path() / "lang") && fs::exists(root_paths.get_base_path() / "shaders-builtin"))
-        root_paths.set_shared_path(root_paths.get_base_path());
+        root_paths.set_static_assets_path(root_paths.get_base_path());
     else if (env_home != NULL)
+        root_paths.set_static_assets_path(fs::path(env_home) / ".local/share" / app_name / dir_sep);
+
+    if (XDG_DATA_DIRS != NULL) {
+        auto env_paths = string_utils::split_string(XDG_DATA_DIRS, ':');
+        for (auto &i : env_paths) {
+            if (fs::exists(fs::path(i) / app_name)) {
+                root_paths.set_static_assets_path(fs::path(i) / app_name / dir_sep);
+                break;
+            }
+        }
+    } else if (XDG_DATA_HOME != NULL) {
+        if (fs::exists(fs::path(XDG_DATA_HOME) / app_name / "data") && fs::exists(fs::path(XDG_DATA_HOME) / app_name / "lang") && fs::exists(fs::path(XDG_DATA_HOME) / app_name / "shaders-builtin"))
+            root_paths.set_static_assets_path(fs::path(XDG_DATA_HOME) / app_name / dir_sep);
+    }
+
+    // shared path
+    if (env_home != NULL)
         root_paths.set_shared_path(fs::path(env_home) / ".local/share" / app_name / dir_sep);
 
-    if (env_data_dirs != NULL) {
-        auto env_paths = string_utils::split_string(env_data_dirs, ':');
+    if (XDG_DATA_DIRS != NULL) {
+        auto env_paths = string_utils::split_string(XDG_DATA_DIRS, ':');
         for (auto &i : env_paths) {
             if (fs::exists(fs::path(i) / app_name)) {
                 root_paths.set_shared_path(fs::path(i) / app_name / dir_sep);
                 break;
             }
         }
-    } else if (env_data_home != NULL) {
-        if (fs::exists(fs::path(env_data_home) / app_name / "data") && fs::exists(fs::path(env_data_home) / app_name / "lang") && fs::exists(fs::path(env_data_home) / app_name / "shaders-builtin"))
-            root_paths.set_shared_path(fs::path(env_data_home) / app_name / dir_sep);
+    } else if (XDG_DATA_HOME != NULL) {
+        root_paths.set_shared_path(fs::path(XDG_DATA_HOME) / app_name / dir_sep);
     }
 #endif
 
@@ -182,6 +198,7 @@ bool init(EmuEnvState &state, Config &cfg, const Root &root_paths) {
     state.config_path = string_utils::utf_to_wide(root_paths.get_config_path_string());
     state.cache_path = string_utils::utf_to_wide(root_paths.get_cache_path_string());
     state.shared_path = root_paths.get_shared_path_string();
+    state.static_assets_path = root_paths.get_static_assets_path();
 
     // If configuration does not provide a preference path, use SDL's default
     if (state.cfg.pref_path == root_paths.get_pref_path() || state.cfg.pref_path.empty())
@@ -193,7 +210,8 @@ bool init(EmuEnvState &state, Config &cfg, const Root &root_paths) {
     }
 
     LOG_INFO("Base path: {}", state.base_path.string());
-    LOG_INFO("Shared assets path: {}", state.shared_path.string());
+    LOG_INFO("Static assets path: {}", state.static_assets_path.string());
+    LOG_INFO("Shared path: {}", state.shared_path.string());
     LOG_INFO("Log path: {}", state.log_path.string());
     LOG_INFO("User config path: {}", state.config_path.string());
     LOG_INFO("User pref path: {}", state.pref_path.string());
--- a/vita3k/renderer/include/renderer/state.h
+++ b/vita3k/renderer/include/renderer/state.h
@@ -48,6 +48,7 @@ struct State {
     std::string cache_path;
     std::string log_path;
     std::string shared_path;
+    fs::path static_assets;
     const char *title_id;
     const char *self_name;
 
@@ -80,7 +81,7 @@ struct State {
 
     bool need_page_table = false;
 
-    virtual bool init(const char *shared_path, const bool hashless_texture_cache) = 0;
+    virtual bool init(const fs::path &static_assets, const bool hashless_texture_cache) = 0;
     virtual void late_init(const Config &cfg, const std::string_view game_id) = 0;
 
     virtual TextureCache *get_texture_cache() = 0;
--- a/vita3k/renderer/src/vulkan/renderer.cpp
+++ b/vita3k/renderer/src/vulkan/renderer.cpp
@@ -176,7 +176,7 @@ VKState::VKState(int gpu_idx)
     , screen_renderer(*this) {
 }
 
-bool VKState::init(const char *shared_path, const bool hashless_texture_cache) {
+bool VKState::init(const fs::path &static_assets, const bool hashless_texture_cache) {
     shader_version = fmt::format("v{}", shader::CURRENT_VERSION);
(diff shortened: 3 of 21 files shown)


Macdu:
I think we already talked about it, with some packinging, one could end up with Vita3K in /usr/bin. As a consequence only the executable will be in base_path and everything else will be put in shared_path.

EXtremeExploit:
Right now the way its done (and should be done from the time being until we have some flatpak or appimage), is to have a link instead of the binary in /usr/share/bin, MANY projects do it like this, if being able to change the read-only assets is really needed, i can try to put it back

Macdu:
I would say to keep it the way it is and just put the texture folder somewhere else but I'm really not sure. @redpolline do you have some opinion about it?

EXtremeExploit:
Tried my best to find all ocurrences of the usage of base_path, i made it so the old implementation stays without changing the value of shared_path, each time a file from base_path is read it checks first if it exists in the shared_path, if it does it will use that but if it isnt it will use the one in base_path instead, i think its nicer because now  lets say that you want to try out some language changes you wont have to copy all the files in base_path, just the lang folder, each file gets treated individually

vita3k/app/src/app_init.cpp
```diff
@@ -118,44 +118,40 @@ void init_paths(Root &root_paths) {
...
+    if (XDG_CACHE_HOME != NULL) {
+        root_paths.set_cache_path(fs::path(XDG_CACHE_HOME) / app_name / dir_sep);
+        root_paths.set_log_path(fs::path(XDG_CACHE_HOME) / app_name / dir_sep);
     } else if (env_home != NULL) {
         root_paths.set_cache_path(fs::path(env_home) / ".cache" / app_name / dir_sep);
         root_paths.set_log_path(fs::path(env_home) / ".cache" / app_name / dir_sep);
     }
 
-    // Don't assume that base_path is portable.
-    if (fs::exists(root_paths.get_base_path() / "data") && fs::exists(root_paths.get_base_path() / "lang") && fs::exists(root_paths.get_base_path() / "shaders-builtin"))
```
redpolline, line 143: The reason that this is here, is to allow the normal release archives to work without any effort from the end-user.

If it's removed, when someone attempts to run Vita3K from the release archive, Vita3K will probably crash due to missing assets. Unless the end user just so happens to have a valid shared_path w/ needed assets in .local/share. The new code assumes this exception is the rule, which may not be the case.

Really there should probably be some validate_shared_path() function that actually checks that a given path _actually_ has all of the needed assets, (Vita3K tends to crash in a very user-unfriendly way when this is not the case), but this was the bare minimum needed to keep the normal release archives working while also allowing the new XDG support to work at the same time.

vita3k/app/src/app_init.cpp
```diff
@@ -118,44 +118,40 @@ void init_paths(Root &root_paths) {
...
+    if (XDG_DATA_DIRS != NULL) {
+        auto env_paths = string_utils::split_string(XDG_DATA_DIRS, ':');
         for (auto &i : env_paths) {
             if (fs::exists(fs::path(i) / app_name)) {
                 root_paths.set_shared_path(fs::path(i) / app_name / dir_sep);
                 break;
             }
         }
-    } else if (env_data_home != NULL) {
-        if (fs::exists(fs::path(env_data_home) / app_name / "data") && fs::exists(fs::path(env_data_home) / app_name / "lang") && fs::exists(fs::path(env_data_home) / app_name / "shaders-builtin"))
```
redpolline, line 157: Same here. If this is removed, and if XDG_DATA_DIRS is not defined, Vita3K will unconditionally override a detected data folder in base_path with the folder in .local/share. (Which may not be valid. Or may be missing required assets.) Breaking the standard release archive builds.

redpolline:
> I would say to keep it the way it is and just put the texture folder somewhere else but I'm really not sure. @redpolline do you have some opinion about it?

I tried to put dumped textures and shaders in HOME/.cache/Vita3K, but the consensus was against that change.

I thought reasoning behind it was due to those folders only being generated as part of the debugging code path and that there wasn't a permanent cache for textures and shaders.

>is to have a link instead of the binary in /usr/share/bin

That location is a system wide path managed by the system's package manager. Putting a link to some user's home directory in there would be a massive violation of the FHS. (Not to mention it probably won't work in a multi-user system due to filesystem permissions.)

You could potentially place it in another path. But the issue there is that there isn't much of a "local sysadmin defined" path in the FHS. Sure you have /usr/local, but as you've already indicated, that gets used by third-party packages a lot and cannot be guaranteed to be free of conflicts. The same is true of /opt and various places under /var (snapd and docker love shoving an entire OS in here....). With even /home now under threat from packages like systemd and steam. (Although in steam's case it tends to jail everything into it's own folder. The real issue comes up when multiple users have steam accounts and want to install the same games. Eating up disk space unless HOME just so happens to be on a de-duplicating filesystem. To say nothing about Vita3K's own installation to the virtual eMMC requirement to run anything....)

Personally I just use a custom folder in the root path for storing stuff that needs to be accessible to all users, but that's not much of an option for an app developer.

As for the link to the exe idea in general, All AppImages do that. Specifically, they have a /apprun.sh that links to SQUASHFS-MOUNT/usr/bin/foo. (foo in this case being whatever was set as the launch target when the AppImage was built.) My own AppImage build of Vita3K uses this to run a script that changes the shared_path location at startup. By prepending the SQUASHFS-MOUNT/usr/share/Vita3K path to the XDG_DATA_DIRS environment variable during Vita3K's invocation.

>Tried my best to find all ocurrences of the usage of base_path, i made it so the old implementation stays without changing the value of shared_path, each time a file from base_path is read it checks first if it exists in the shared_path, if it does it will use that but if it isnt it will use the one in base_path instead, i think its nicer because now lets say that you want to try out some language changes you wont have to copy all the files in base_path, just the lang folder, each file gets treated individually

That seems like it would be cumbersome to keep working. As any change to the files and how they are used would need to be kept in sync with the detection code. Again, a validation function that handles all of that in one place might be a better option. That, or implementing it as some sort of virtual filesystem. Which is needed anyway if you want to be able to load something like a PSV or PKG file / Filesystem mods directly without installing them to the virtual eMMC first.

>but i am not sure what enviroment variable to use for it

As this seems to be an app specific use case, there's really no restriction on defining in init_paths() something to the effect of:
```
ignore_base_path = getenv("VITA3K_PATH_DETECT_IGNORE_BASE_PATH");
if (ignore_base_path == NULL || ignore_base_path[0] != '1') {
    /* Code to detect data in base_path */
}
```
Or any more specific path. (For example data/textures.)

Multiple apps do this in the event that they need to override default behavior at runtime. They just cannot change the default behavior in a way that would break the spec and remain compliant.

EXtremeExploit:
> That seems like it would be cumbersome to keep working. As any change to the files and how they are used would need to be kept in sync with the detection code. Again, a validation function that handles all of that in one place might be a better option. That, or implementing it as some sort of virtual filesystem. Which is needed anyway if you want to be able to load something like a PSV or PKG file / Filesystem mods directly without installing them to the virtual eMMC first.

Added a function in fs_utils that helps choose between the 2 paths, i hope its better, it will still fallback to base_path if the thing the emulator is looking for is not in shared_path, no installations should break in any way. And yes, i agree thats its a cumbersome to keep in sync with more usages of base_path, but at this poin of development i dont think new features that use base_path will get added, or if they are, its probably (i hope) going to be with people that are a bit experienced with how base_path an shared_path are handled, if not, it should be easy to fix with a second PR that fixes the inconsistency. Just for the time being

Macdu:
I think @redpolline created the shared path specially for Vita3K files (data, lang and shader-builtins). As such doing these checks kind of defies the purpose.
The issue here is mostly where to put the textures folder. It can't be in base_path or shared_path because it can be readonly (although not entirely sure about that). It can't be in pref path because it's not a vita file and it can't really be in cache because these files are not temporary and shouldn't be cleared. Do you have some opinion about it @redpolline  and @EXtremeExploit. I guess the cleaner solution would be to create a new path (being the XDG_DATA_HOME on linux) which for the time being would contain only the texture replacement files.

EXtremeExploit:
True, maybe reverting the checks to the definitons of `shared_path` and renaming it to something like `static_assets`, and create a new one with the old name of `shared_path` and that one be .local/share.
So in the end it would be like 
Rename `shared_path` -> `static_assets`
Create new path `shared_path` that points to `XDG_DATA_HOME` or `.local/share/Vita3K` bye default

redpolline:
> I think @redpolline created the shared path specially for Vita3K files (data, lang and shader-builtins).

I did.

>As such doing these checks kind of defies the purpose.

The issue that comes up if we don't do them is that Vita3K won't use the base_path for those files at all. Which means unless the end-users manually move things from the release archive into the correct folders, or manually sets the XDG_* environment variables to base_path themselves, Vita3K _will_ crash during start up before the GUI can initialize.

This isn't a problem if the decision is made to deprecate the release archive builds. As any container based release, (appimage, flatpak, etc.), would have the needed files built in, but existing users would still need to move their existing pref_path to the new location. Which they would need to be notified about. (Release notes, UI notification, etc.)

>The issue here is mostly where to put the textures folder. It can't be in base_path or shared_path because it can be readonly (although not entirely sure about that).

base_path is read only on linux. Assuming otherwise is going to cause problems. (Such as Vita3K failing to start due to not being able to write it's config / log file.)

shared_path (as I re-wrote it) is the same way. That's just static assets that if Vita3K were a Windows only app could be embedded in the exe / dll as a data resource and loaded via [FindResourceEx() / LoadResource()](https://learn.microsoft.com/en-us/windows/win32/menurc/finding-and-loading-resources). The same data can also exist under C:\Program Files under Windows. Which is also not writable by end-users. (Unless they have admin privileges and elevate Vita3K to allow it.)

Most linux distros are more strict about this, they tend to put static assets in /usr/share/APP_NAME as required by the FHS. This path is not writable by anyone except root, so again assuming otherwise will cause problems.

>It can't be in pref path because it's not a vita file

Agreed.

>and it can't really be in cache because these files are not temporary and shouldn't be cleared.
This is where I disagree. Dumped textures and shaders are _always_ temp data. The game doesn't need them to run, as they can be regenerated if they are not there, and in some cases the regeneration is required. (For example an incompatible renderer update.) These can also be removed to free up space on systems with small storage devices. As another example, if a user deletes a game there's no reason to keep the dumped shaders and textures from it. Also, rpcs3 has a dedicated option in the menu for removing shaders and PPU data, and stores said data in ~/.cache.

Yes there is a performance benefit from keeping the data. (Not needing to regenerate it while the game is running.) But there are reasons why that data shouldn't be considered "permanent" and ineligible for placement in ~/.cache.

>I guess the cleaner solution would be to create a new path (being the XDG_DATA_HOME on linux) which for the time being would contain only the texture replacement files.

[XDG_DATA_HOME is currently used as the pref_path](https://github.com/Vita3K/Vita3K/blob/6e75edfc6ba559297529d803be95473883cd6b02/vita3k/app/src/app_init.cpp#L127). Which is vita files only.

But creating a new path would allow for potentially making it a user-configurable option in the GUI. It would also resolve any future issue with needing a spot for renderer specific data.

EXtremeExploit:
the folder for texture dumps is also used for texture replacement, and im pretty sure that if an user were to clean their cache they wouldnt want to have to reinstall the texture replacements they had for a game, think about it like the "mods" folder.
Also the pref_path is already pointing to `XDG_DATA_HOME/Vita3K/Vita3K`, so the main .local/share/Vita3K folder is clean, it only has 1 folder and that is the pref_path, but of course we cant just cd one directory higher from pref_path as its user defined on the config file.

> Perhaps moving the current pref_path to XDG_DATA_HOME/Vita3K/eMMC and then placing the textures in XDG_DATA_HOME/Vita3K/textures would work?

I agree with this, having the pref_path be named Vita3K could be confusing for users, and eMMC seems accurate enough. I want to know what @Macdu opinion is on this

I still stand by my comment about renaming `shared_path` to `static_assets` (and still maintaining the checks that are already there for changing their path)
And making a new path called `shared_path` (just like the one thats renamed) and have that one point to `.local/share/Vita3K` and put the textures in `(shared_path)/textures/`

vita3k/renderer/src/vulkan/renderer.cpp
```diff
@@ -629,6 +629,7 @@ void VKState::late_init(const Config &cfg, const std::string_view game_id) {
 
     pipeline_cache.init();
 
+    // Use log_path as in windows its the same place as executable but .local/share in linux, which is the correct place it should be
```
Macdu, line 632: I think you can remove this line

vita3k/renderer/include/renderer/state.h
```diff
@@ -80,7 +81,7 @@ struct State {
 
     bool need_page_table = false;
 
-    virtual bool init(const char *shared_path, const bool hashless_texture_cache) = 0;
+    virtual bool init(const fs::path static_assets, const bool hashless_texture_cache) = 0;
```
Macdu, line 84:
```suggestion
    virtual bool init(const fs::path &static_assets, const bool hashless_texture_cache) = 0;
```
Always use a reference when passing a const fs::path (or std::string)

Macdu:
Look good to me, is the PR as it is fine for you @redpolline ?

vita3k/app/src/app_init.cpp
```diff
@@ -118,44 +118,60 @@ void init_paths(Root &root_paths) {
...
-    auto env_config_home = getenv("XDG_CONFIG_HOME");
+    auto XDG_DATA_DIRS = getenv("XDG_DATA_DIRS");
+    auto XDG_DATA_HOME = getenv("XDG_DATA_HOME");
+    auto XDG_CACHE_HOME = getenv("XDG_CACHE_HOME");
+    auto XDG_CONFIG_HOME = getenv("XDG_CONFIG_HOME");
 
-    if (env_data_home != NULL)
-        root_paths.set_pref_path(fs::path(env_data_home) / app_name / app_name / dir_sep);
+    if (XDG_DATA_HOME != NULL)
+        root_paths.set_pref_path(fs::path(XDG_DATA_HOME) / app_name / app_name / dir_sep);
```
redpolline, line 127: pref_path needs to move to a subdir (I voted for /eMMC), or the fallback paths for static_assets_path and shared_path below are going to collide with it. (I.e. Non-vita files will end up in pref_path.)

vita3k/app/src/app_init.cpp
```diff
@@ -193,6 +210,7 @@ bool init(EmuEnvState &state, Config &cfg, const Root &root_paths) {
     }
 
     LOG_INFO("Base path: {}", state.base_path.string());
+    LOG_INFO("Static assets path: {}", state.static_assets_path.string());
```
redpolline, line 213: This is confusing. Perhaps Shared assets below should just drop the "assets" part?

vita3k/app/src/app_init.cpp
```diff
@@ -182,6 +198,7 @@ bool init(EmuEnvState &state, Config &cfg, const Root &root_paths) {
     state.config_path = string_utils::utf_to_wide(root_paths.get_config_path_string());
     state.cache_path = string_utils::utf_to_wide(root_paths.get_cache_path_string());
     state.shared_path = root_paths.get_shared_path_string();
+    state.static_assets_path = root_paths.get_shared_path();
```
redpolline, line 201: Is this correct? I thought it should be get_static_assets_path() here?

vita3k/app/src/app_init.cpp
EXtremeExploit: pref_path is already in a subdir (`.local/share/Vita3K/Vita3K`), i like the idea of changing it to eMMC, but i just want to know what the others think about it first

vita3k/app/src/app_init.cpp
EXtremeExploit: indeed, my bad

vita3k/app/src/app_init.cpp
EXtremeExploit: fixed

vita3k/app/src/app_init.cpp
redpolline: Maybe that should be a different PR then?

redpolline:
> Look good to me, is the PR as it is fine for you @redpolline ?

Looks fine to me.

Macdu (approved):

PR #3059, opened 2023-11-08
ci: Bundle curl gnutils lib

EXtremeExploit:
Deck users cant install libraries because the root fs is read only, so just bundle it


Macdu:
I think you can rename the openssl task above Bundle shared object and put it along the openssl so copy command.

EXtremeExploit:
alrighty

Macdu (approved):
Wait for the CI to finish still.

PR #3065, opened 2023-11-09
build: Add support for building AppImages under linux

redpolline:
This is an attempt at packaging Vita3K as an AppImage.

It currently depends on [linuxdeploy](https://github.com/linuxdeploy/linuxdeploy) as it's final packager. (linuxdeploy itself is shipped as an AppImage.)

To build it, simply place the release from linuxdeploy in appimage/linuxdeploy.AppImage and use the linux-ninja-clang-appimage cmake preset.

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -21,12 +21,20 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
         config: [Release]
+        version: [zip, appimage]
         include:
           - os: ubuntu-latest
+            version: appimage
+            cache_path: ~/.ccache
+            extra_cmake_args: -DLINUXDEPLOY_COMMAND=/usr/local/bin/linuxdeploy-x86_64.AppImage
+            cmake_preset: linux-ninja-clang15-appimage
+          - os: ubuntu-latest
+            version: zip
             cache_path: ~/.ccache
             extra_cmake_args: 
             cmake_preset: linux-ninja-clang15
           - os: windows-latest
+            version: zip
             cache_path: |
                 C:\vcpkg\installed
                 C:\vcpkg\packages
@@ -34,9 +42,15 @@ jobs:
             extra_cmake_args: -DCMAKE_TOOLCHAIN_FILE=C:\vcpkg\scripts\buildsystems\vcpkg.cmake 
             cmake_preset: windows-ninja
           - os: macos-latest
+            version: zip
             cache_path: ~/Library/Caches/ccache
             extra_cmake_args: 
             cmake_preset: macos-ninja
+        exclude:
+          - os: macos-latest
+            version: appimage
+          - os: windows-latest
+            version: appimage
 
     steps:
       - uses: actions/checkout@v3
@@ -58,8 +72,9 @@ jobs:
       - name: Set up build environment (ubuntu-latest)
         run: |
           sudo add-apt-repository -y ppa:mhier/libboost-latest
+          sudo add-apt-repository universe
           sudo apt update
-          sudo apt -y install ccache libboost-filesystem1.81-dev libboost-program-options1.81-dev libboost-system1.81-dev libgtk-3-dev libsdl2-dev ninja-build
+          sudo apt -y install ccache libboost-filesystem1.81-dev libboost-program-options1.81-dev libboost-system1.81-dev libgtk-3-dev libsdl2-dev ninja-build libfuse2
         if: matrix.os == 'ubuntu-latest'
 
       - uses: ilammy/msvc-dev-cmd@v1
@@ -95,6 +110,15 @@ jobs:
           fi
           sudo make -C SDL2-${SDL2VER} install
         if: matrix.os == 'ubuntu-latest'
+
+      - name: Set up linuxdeploy (ubuntu-latest, appimage)
+        run: |
+          if [[ ! -e linuxdeploy-x86_64.AppImage ]]; then
+            curl -sLO https://github.com/linuxdeploy/linuxdeploy/releases/latest/download/linuxdeploy-x86_64.AppImage
+          fi
+          sudo cp -f linuxdeploy-x86_64.AppImage /usr/local/bin/
+          sudo chmod +x /usr/local/bin/linuxdeploy-x86_64.AppImage
+        if: matrix.os == 'ubuntu-latest' && matrix.version == 'appimage'
 
       - name: Ccache setup
         run: ccache -z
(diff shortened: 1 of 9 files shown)


EXtremeExploit:
add the `appimage/linuxdeploy.AppImage` in .gitignore

And regarding the updater, how would that work? because right now it downloads a zip and replaces the files in the base folder, would it be possible to not make it possible to compile with the updater and have the user update the appimage with appimageupdater using a zsync instead?, this is just an opinion tho, if you can come up with a better way to handle this it would be awesome

Macdu:
I think we could add the app_image to the CI.

redpolline:
Looks like hdiutil timed out on the mac buildbot. (I don't think this commit is responsible for that.)

redpolline:
> add the `appimage/linuxdeploy.AppImage` in .gitignore

Done.

> And regarding the updater, how would that work? because right now it downloads a zip and replaces the files in the base folder, would it be possible to not make it possible to compile with the updater and have the user update the appimage with appimageupdater using a zsync instead?, this is just an opinion tho, if you can come up with a better way to handle this it would be awesome

I've added the zsync info to the built appimage and moved the zsync file to the cmake output directory. appimageupdater and appimaged should be able to update the AppImage once there's a zsync file on the github releases page.

redpolline:
> I think we could add the app_image to the CI.

I've updated the ci build script. [AppImage and zsync files are included in the build artifact.](https://github.com/Vita3K/Vita3K/suites/18104196432/artifacts/1043574187)

Macdu:
Wouldn't this break the linux updater when updating to this PR? (I was thinking maybe for the time being only have an additional CI run making the AppImage).

redpolline:
> Wouldn't this break the linux updater when updating to this PR? (I was thinking maybe for the time being only have an additional CI run making the AppImage).

It shouldn't. Given that all I did was enable the extra cmake defines for a pre-existing preset. (That had the updater enabled.)

Oh, I see. Because I enabled the extra defines, the else block doesn't execute. so all we get is the built Vita3K binary and the manually copied .so files. We're missing the asset directories and updater script. Hang on, I'll fix it.

redpolline:
Done.

That should enable copying the original data, as well as the new appimage and zsync files to the cmake target directory.

redpolline:
> Wouldn't this break the linux updater when updating to this PR? (I was thinking maybe for the time being only have an additional CI run making the AppImage).

Although, thinking about it, there's also #1418. So it should be a separate CI job so it's easier to add that build variant in the future.

redpolline:
Done.

Separate ci build for appimage is implemented.

j8r:
> Although, thinking about it, there's also https://github.com/Vita3K/Vita3K/issues/1418. So it should be a separate CI job so it's easier to add that build variant in the future.

This can be the case, but usually Flatpaked projects have their repository at https://github.com/flathub.

Macdu:
> Done.
> 
> Separate ci build for appimage is implemented.

The build is separate for the AppImage, however the generated zip (ubuntu-latest) contains both the standard Vita3K files and the AppImage, I would suggest to keep ubuntu-latest as it is (in order not to break the updater) and make a ubuntu-appimage-latest containing only the appimage.

EXtremeExploit:
Also the CI should upload the AppImage alone, not in a zip, it should be able to be downloaded alongside with the zsync so appimageupdater knows what to do and doesnt break

redpolline:
>The build is separate for the AppImage, however the generated zip (ubuntu-latest) contains both the standard Vita3K files and the AppImage, I would suggest to keep ubuntu-latest as it is (in order not to break the updater) and make a ubuntu-appimage-latest containing only the appimage.

There is no "ubuntu-appimage-latest" ci runner on github. As matrix.os is used to set the run-on tag (in addition to others) that won't work without completely restructuring the ci matrix and corresponding job tasks.

The final zip is supposed to be created by create-release, and I've already altered it to separate out the builds, but that task doesn't execute for CI runs initiated by PRs.

What I can do, (without doing a complete restructure), is change the upload name to include the version tag I introduced. That should create a new artifact separate from the zip build, and it's what the new commit does. It will also change the names for the existing artifacts, however. Is this acceptable? Or would you still prefer the complete restructure?

redpolline:
> Also the CI should upload the AppImage alone, not in a zip, it should be able to be downloaded alongside with the zsync so appimageupdater knows what to do and doesnt break

[The code already does that.](https://github.com/Vita3K/Vita3K/blob/7613257a78f18386778bcf2facac6ad79677d2e9/.github/workflows/c-cpp.yml#L207) It's a straight up cp call just like the [macos-latest build does for it's DMG file.](https://github.com/Vita3K/Vita3K/blob/7613257a78f18386778bcf2facac6ad79677d2e9/.github/workflows/c-cpp.yml#L201)

The zip you are downloading from the artifact page is just how github sends the artifact data when it's pulled from the ci summary page. [For example, note the zip extension on the macos-latest artifact from PR 3068.](https://github.com/Vita3K/Vita3K/suites/18106853662/artifacts/1043728945)

.github/workflows/c-cpp.yml
```diff
@@ -96,6 +111,15 @@ jobs:
...
         if: matrix.os == 'ubuntu-latest'
 
+      - name: Set up linuxdeploy (ubuntu-latest)
+        run: |
+          if [[ ! -e linuxdeploy-x86_64.AppImage ]]; then
+            curl -sLO https://github.com/linuxdeploy/linuxdeploy/releases/latest/download/linuxdeploy-x86_64.AppImage
+          fi
+          sudo cp -f linuxdeploy-x86_64.AppImage /usr/local/bin/
+          sudo chmod +x /usr/local/bin/linuxdeploy-x86_64.AppImage
+        if: matrix.os == 'ubuntu-latest'
```
Macdu, line 121: Isn't this only needed for the appimage build?

redpolline: Fixed.

vita3k/CMakeLists.txt
```diff
@@ -201,6 +201,35 @@ if(APPLE)
...
+    endif()
+		add_custom_command(
+		TARGET vita3k
+		POST_BUILD
+		COMMAND ${LINUXDEPLOY_WRAPPER} ${LINUXDEPLOY_COMMAND} --appdir="${APPDIR}"
+		 -e "$<TARGET_FILE:vita3k>" --icon-filename="vita3k"
+		 -i "${CMAKE_CURRENT_SOURCE_DIR}/../data/image/icon.png" -d "${CMAKE_CURRENT_SOURCE_DIR}/../appimage/vita3k.desktop"
+		 --custom-apprun="${CMAKE_CURRENT_SOURCE_DIR}/../appimage/apprun.sh" --output=appimage ${DISCORD_LIB_APPIMAGE}
+		COMMAND ${CMAKE_COMMAND} -E copy_if_different "${CMAKE_CURRENT_BINARY_DIR}/*.AppImage*" "$<TARGET_FILE_DIR:vita3k>/"
+		COMMAND ${CMAKE_COMMAND} -E rm "${CMAKE_CURRENT_BINARY_DIR}/*.AppImage*")
```
Macdu, line 231: image
The indentation is all over the place here.

redpolline:
Should be fixed now.

Macdu:
> Should be fixed now.

Could you use tabs instead of 2 spaces for indentation (like the reste of the CMakeList.txt file), it still looks a bit out of place. Sorry about that, the PR should be ready after this change.

redpolline:
The code before was in tabs. I guess your cmake code has different tab lengths....

redpolline:
How about now?

Macdu:
> The code before was in tabs. I guess your cmake code has different tab lengths....

I looked at the diff on the github website. The rest of the CMakeList file was definitely using tabs while your part was using 2 spaces. Anyway it looks like it is fixed now. Hopefully this doesn't break the release part / updater, in the worst case I'll revert this commit.

Macdu (approved):

PR #3068, opened 2023-11-10
cmake: Enable XXH_X86DISPATCH_ALLOW_AVX by default

EXtremeExploit:
Issue reported in [the AUR](https://aur.archlinux.org/packages/vita3k-git#comment-942974), This enables the option of compiling the AVX implementation of xxh3 even if it will crash on SSE2-Only CPUs, as this seems to only affect few users at the moment i decided to go with OFF as default, once merged the option will be set on the PKGBUILD as ON for the few users as the build generated by it is intended for usage in the same machine it was compiled on, not sharing

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -325,8 +325,12 @@ target_include_directories(libatrac9 PUBLIC LibAtrac9/C/src)
 
 option(BUILD_SHARED_LIBS "Build shared library" OFF)
 option(XXHASH_BUILD_XXHSUM "Build the xxhsum binary" OFF)
+option(XXH_X86DISPATCH_ALLOW_AVX "Allow building XXH3 with AVX even if it crashes on SSE2-Only CPUs" OFF)
 set(DISPATCH 1)
 add_subdirectory(xxHash/cmake_unofficial EXCLUDE_FROM_ALL)
+if(XXH_X86DISPATCH_ALLOW_AVX)
+	target_compile_definitions(xxhash PRIVATE XXH_X86DISPATCH_ALLOW_AVX)
+endif()
 
 # Tracy
 option(TRACY_ENABLE_ON_CORE_COMPONENTS


Macdu:
This sets an option (so a cmake variable), but this variable is never used in XXHash's CmakeListe.
What is needed is to set this as a compile definition instead.

external/CMakeLists.txt
```diff
@@ -325,6 +325,10 @@ target_include_directories(libatrac9 PUBLIC LibAtrac9/C/src)
 
 option(BUILD_SHARED_LIBS "Build shared library" OFF)
 option(XXHASH_BUILD_XXHSUM "Build the xxhsum binary" OFF)
+option(XXH_X86DISPATCH_ALLOW_AVX "Allow building XXH3 with AVX even if it crashes on SSE2-Only CPUs" OFF)
+if(XXHASH_BUILD_XXHSUM)
+	add_definitions(-DXXHASH_BUILD_XXHSUM)
+endif()
 set(DISPATCH 1)
 add_subdirectory(xxHash/cmake_unofficial EXCLUDE_FROM_ALL)
```
Macdu, line 333:
```suggestion
set(DISPATCH 1)
add_subdirectory(xxHash/cmake_unofficial EXCLUDE_FROM_ALL)
if(XXH_X86DISPATCH_ALLOW_AVX)
    target_compile_definition(xxhash PRIVATE XXH_X86DISPATCH_ALLOW_AVX)
 endif()
```

What does XXHASH_BUILD_XXHSUM have to do with AVX?

EXtremeExploit: oh my god i forgot to save the file im so dumb

EXtremeExploit: should be good now

external/CMakeLists.txt
```diff
@@ -325,6 +325,10 @@ target_include_directories(libatrac9 PUBLIC LibAtrac9/C/src)
 
 option(BUILD_SHARED_LIBS "Build shared library" OFF)
 option(XXHASH_BUILD_XXHSUM "Build the xxhsum binary" OFF)
+option(XXH_X86DISPATCH_ALLOW_AVX "Allow building XXH3 with AVX even if it crashes on SSE2-Only CPUs" OFF)
+if(XXH_X86DISPATCH_ALLOW_AVX)
+	add_definitions(-DXXH_X86DISPATCH_ALLOW_AVX)
```
Macdu, line 330:
```suggestion
	target_compile_definition(xxhash PRIVATE XXH_X86DISPATCH_ALLOW_AVX)
```
We don't want to add this definition to every file in the project if it is only used in a single cpp file (xxh_x86dispatch.c)

external/CMakeLists.txt
```diff
@@ -325,6 +325,10 @@ target_include_directories(libatrac9 PUBLIC LibAtrac9/C/src)
 
 option(BUILD_SHARED_LIBS "Build shared library" OFF)
 option(XXHASH_BUILD_XXHSUM "Build the xxhsum binary" OFF)
+option(XXH_X86DISPATCH_ALLOW_AVX "Allow building XXH3 with AVX even if it crashes on SSE2-Only CPUs" OFF)
+if(XXH_X86DISPATCH_ALLOW_AVX)
+	target_compile_definitions(xxhash PRIVATE XXH_X86DISPATCH_ALLOW_AVX)
+endif()
```
Macdu, line 331: xxhash is not defined at this point. I'm not entirely sure but I think doing so here would cause CMake to fail.
This needs to be added after the call to add_subdirectory

Macdu (approved):

PR #3076, opened 2023-11-13
modules/SceGxm, display, renderer: Improve accuracy of sync objects and decrease video latency

Macdu:
First part: There was some flickering in some games (persona 4 dancing, Catherine) when using memory mapping. This was caused by the game using the display callback in order to know when the rendering was done (which is fine without memory mapping because everything is copied, but with the game was corrupting some uniforms / vertex buffers). Making the sync object be signaled after the GPU rendering is done (like what they are supposed to do) fixes this issue but...

Second part: This causes an increase of video latency (the frame needs to be rendered, then the request thread notified, then the display thread notified, then the callback executed, which calls setDisplay, then the main thread finishes its queue, record the render command and swapchain swaps and submit them...). To prevent this and even decrease the video latency on all renderers, I implemented a frame prediction: If Vita3K detects that the swapchain images form a cycle of length at most 6 at least 3 times in a row, then as soon as a display entry is added, the main thread will render the predicted image, bypassing everything above. This should decrease the video latency by around 1 (or even 2) frames (which at 30FPS is a lot) on all renderers.

### DIFF
--- a/vita3k/app/src/app.cpp
+++ b/vita3k/app/src/app.cpp
@@ -63,8 +63,8 @@ void calculate_fps(EmuEnvState &emuenv) {
 
 void set_window_title(EmuEnvState &emuenv) {
     const auto af = emuenv.cfg.current_config.anisotropic_filtering > 1 ? fmt ::format(" | AF {}x", emuenv.cfg.current_config.anisotropic_filtering) : "";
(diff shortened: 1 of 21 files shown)


EXtremeExploit (approved):
awesome

PR #3082, opened 2023-11-19
util: Improve float_to_half code

Macdu:
Improve the float_to_half code such that native __fp16 types are used on aarch64 and on x86_64, the fast path can be used with any compiler (and not just MSVC).


EXtremeExploit (approved):
awesome :D

Issue #3071, opened 2023-11-11
add more Language support for vita3k gui


ikhoeyZX:
we know that someone already make vita3k other language version in illegal way, so can we request make vita3k gui support more language (even it not supported by vita it self)?

Macdu:
If you want to add support for an additional language in Vita3K, you can make a pull request for it.

ikhoeyZX:
how?

Croden1999:
my suggestion is need add console language before add system language, same as likes to rpcs3, yuzu etc.

Macdu:
Yes I agree, that sounds like the most logical idea.

Kitakatarashima:
how to add console language??

Zangetsu38:
> my suggestion is need add console language (for games) before add system language (for gui), same as likes to rpcs3, yuzu etc.

rpcs3 don't have this, rpcs3 have ONLY like viota3k for System language, and gui is only in english
![image](https://github.com/Vita3K/Vita3K/assets/5261759/77b202be-641f-4c17-ab03-4754f45e9196)

Croden1999:
> > my suggestion is need add console language (for games) before add system language (for gui), same as likes to rpcs3, yuzu etc.
> 
> rpcs3 does not have this, rpcs3 only has console language, like vita3k, also called system language, and the GUI is only in English. ![image](https://user-images.githubusercontent.com/5261759/283318417-77b202be-641f-4c17-ab03-4754f45e9196.png)
> 
> > how to add console language??
> 
> This is just not possible with the current code if this language does not exist on the psvita

but rpcs3 support custom lang.qt/ts files right?

Zangetsu38:
> but rpcs3 support custom translations .qt/ts files right?

I have never see that, this option is literally April fool lol
![image](https://github.com/Vita3K/Vita3K/assets/5261759/de4c6f2b-ca01-44dc-bcb0-f06576383641)

Croden1999:
> > but rpcs3 support custom translations .qt/ts files right?
> 
> I have never see that, this option is literally April fool lol, english is hard coded ![image](https://user-images.githubusercontent.com/5261759/283325368-de4c6f2b-ca01-44dc-bcb0-f06576383641.png)

i knew only english have in master, the users can pust the translation files in qt6\translations folder

![image](https://github.com/Vita3K/Vita3K/assets/61804715/a7e00f11-216e-4901-a923-a774eb74ff14)
[...]

PR #3083, opened 2023-11-19
lang: Add support of user custom languague.

Zangetsu38:
# About
- Implement featur request by https://github.com/Vita3K/Vita3K/issues/3071


Macdu:
I'm not sure this approach (making a system and user directory inside lang) will work on linux as the static assets directory can be different compared to the shared folder.

Zangetsu38:
I will admit to you that I haven't thought about Linux, but in itself there shouldn't be a problem, it's inside the lang folder, so whether there is another folder or not, it normally doesn't change anything , like the data folder

Macdu:
I agree that this still works. But in this case creating in the data folder the file "user/PLACE USER LANG HERE.txt" does not make a lot of sense. It would be better in some init function to create the file shared_path / "data/user/PLACE USER LANG HERE.txt" in some init function if does not already exists.

Zangetsu38:
hmm I see, on the other hand I want to leave that in the lang folder, it doesn't make sense otherwise.
The goal is for users to upload and test their translated language file, and then for it to be directly added to the project here via a pr.

so this txt is just here for indicate to user where place lang file

Macdu:
yes, so if you create the shared_path / "data/lang/user/PLACE USER LANG HERE.txt" in some init function, on windows it will be in the right place (in the lang folder) while on Linux, it will be in its correct read-write location.

Zangetsu38:
> yes, so if you create the shared_path / "data/lang/user/PLACE USER LANG HERE.txt" in some init function, on windows it will be in the right place (in the lang folder) while on Linux, it will be in its correct read-write location.

I have said i want it inside lang, but, here you said shared_path/data/lang, and me i wxant so base_path/lang, no data

Zangetsu38:
> yes, so if you create the shared_path / "data/lang/user/PLACE USER LANG HERE.txt" in some init function, on windows it will be in the right place (in the lang folder) while on Linux, it will be in its correct read-write location.

I'm not sure I understand, I'm talking to you about the lang folder at the root of the base_path, but here you're telling me about the data folder, which is a read-only folder by the way

On the Windows side it must be base_path/lang/user
There is no question of data folder here

Macdu:
Ah yes sorry, remove the data part, for some reason I thought lang was in a data folder

Zangetsu38:
> Ah yes sorry, remove the data part, for some reason I thought lang was in a data folder

I was also saying, you lost me there lol, i have coded lang in root so lol, but so the shared path is good for writing on Linux? this all new path lost me0..

Macdu:
Yes, basically, basic_path and static_path can be read-only while shared-path should be guaranteed to be read-write (at least that's how I understood it). On windows, these 3 paths are the same.

Zangetsu38:
i see hmm

Zangetsu38:
is normally done like you request @Macdu

vita3k/lang/src/lang.cpp
```diff
@@ -67,10 +67,19 @@ void init_lang(LangState &lang, EmuEnvState &emuenv) {
...
     }
 
-    pugi::xml_document lang_xml;
+    const auto system_lang_path{ emuenv.static_assets_path / "lang/system" };
+    const auto user_lang_path{ emuenv.shared_path / "lang/user" };
+
+    // Create user lang folder if not exists and create a file to indicate where to place the lang files
+    if (!fs::exists(user_lang_path)) {
+        fs::create_directories(user_lang_path);
+        std::ofstream outfile((user_lang_path / "PLACE USER LANG HERE.txt").string());
```
Macdu, line 76:
```suggestion
        fs::ofstream outfile(user_lang_path / "PLACE USER LANG HERE.txt");
```

Zangetsu38: done

Macdu (approved):

PR #3095, opened 2023-11-22
appimage: Fix desktop integration

EXtremeExploit:
Running the appimage once worked fine, but clicking the option to integrate it into the desktop wasnt working because the desktop entry was wrong, it was missing the `Exec` line.


Macdu (approved):
I'm not entirely sure removing the PATH export won't cause any issue though

PR #3097, opened 2023-11-25
util: Fix static initialization in tracy.cpp

oltolm:
This is a fix for the static initialization problem which can lead to a crash. I compile Vita3K myself with Clang on Windows and it does crash without this fix. See https://isocpp.org/wiki/faq/ctors#static-init-order.


bookmist (approved):
LGTM

PR #3100, opened 2023-11-27
lang: Add two static user languages.

Zangetsu38:
# About
- lang: Add two static user languages.
Indonesia and Malaysia.
- Add current language system set.

special credit to @Kitakatarashima both languages translate.

# Result
image

### DIFF
--- a/vita3k/gui/src/settings_dialog.cpp
+++ b/vita3k/gui/src/settings_dialog.cpp
@@ -250,15 +250,19 @@ void init_config(GuiState &gui, EmuEnvState &emuenv, const std::string &app_path
     }
 
     list_user_lang.clear();
-    const auto LANG_PATH{ emuenv.shared_path / "lang/user" };
-    if (fs::exists(LANG_PATH) && !fs::is_empty(LANG_PATH)) {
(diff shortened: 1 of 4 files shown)


vita3k/gui/src/settings_dialog.cpp
```diff
@@ -250,15 +250,19 @@ void init_config(GuiState &gui, EmuEnvState &emuenv, const std::string &app_path
...
                 list_user_lang.push_back(lang_file_name);
             }
         }
-    }
+    };
+
+    get_list_user_lang(emuenv.static_assets_path);
+#ifndef WIN32
+    get_list_user_lang(emuenv.shared_path);
+#endif
```
Macdu, line 265: Wouldn't this cause a problem (duplicate entry) if emuenv.static_assets_path = emuenv.shared_path (which can also happen on a non-Windows OS) ?

Zangetsu38: Only Linux have it, and Android so for Mac  🤔

Zangetsu38: i have changed it little, normally should fine for all case

Macdu (approved):

PR #3104, opened 2023-11-30
app/app init: Enable the init log path for OS assign only.

Zangetsu38:


vita3k/app/src/app_init.cpp
```diff
@@ -216,12 +216,14 @@ bool init(EmuEnvState &state, Config &cfg, const Root &root_paths) {
     }
 
     LOG_INFO("Base path: {}", state.base_path.string());
+#if defined(__linux__) && !defined(__ANDROID__) && !defined(__APPLE__)
     LOG_INFO("Static assets path: {}", state.static_assets_path.string());
     LOG_INFO("Shared path: {}", state.shared_path.string());
     LOG_INFO("Log path: {}", state.log_path.string());
     LOG_INFO("User config path: {}", state.config_path.string());
     LOG_INFO("User pref path: {}", state.pref_path.string());
```
Macdu, line 224: I think we still want the pref path being logged even on Windows / MacOS

Zangetsu38: Pref path loged here is wrong, it log default pref path on appdata
No have sense

Zangetsu38: hmm i see, sorry, no idea wqhy ihave get this

PR #3107, opened 2023-12-05
packages: Copy self authid when decrypting

Macdu:
This allows self files generated by Vita3K to match exactly self files decrypted using third-party tools.
This is useful when patching file with a hash check (right now it is not possible to patch the eboot file of Tales of Innocence decrypted with Vita3K with the english patch).

### DIFF
--- a/vita3k/packages/src/pkg.cpp
+++ b/vita3k/packages/src/pkg.cpp
@@ -98,9 +98,10 @@ bool decrypt_install_nonpdrm(EmuEnvState &emuenv, std::string &drmlicpath, const
 
     for (const auto &file : fs::recursive_directory_iterator(title_id_src)) {
         if ((file.path().extension() == ".suprx") || (file.path().extension() == ".self") || (file.path().filename() == "eboot.bin")) {
-            self2elf(file.path().string(), file.path().string() + "elf", SCE_KEYS, temp_klicensee.data());
+            uint64_t authid = 0x2F00000000000001ULL;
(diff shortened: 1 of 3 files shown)


vita3k/packages/src/pkg.cpp
```diff
@@ -98,9 +98,10 @@ bool decrypt_install_nonpdrm(EmuEnvState &emuenv, std::string &drmlicpath, const
 
     for (const auto &file : fs::recursive_directory_iterator(title_id_src)) {
         if ((file.path().extension() == ".suprx") || (file.path().extension() == ".self") || (file.path().filename() == "eboot.bin")) {
-            self2elf(file.path().string(), file.path().string() + "elf", SCE_KEYS, temp_klicensee.data());
+            uint64_t authid = 0x2F00000000000001ULL;
+            self2elf(file.path().string(), file.path().string() + "elf", SCE_KEYS, temp_klicensee.data(), &authid);
             fs::rename(file.path().string() + "elf", file.path().string());
-            make_fself(file.path().string(), file.path().string() + "fself");
+            make_fself(file.path().string(), file.path().string() + "fself", authid);
             fs::rename(file.path().string() + "fself", file.path().string());
```
bookmist, line 105: Whole content of this if can be replaced to call of `decrypt_fself`, but of course such refactoring is not a matter of this PR. (Sorry, it seems that I forgot this place when I made my refactoring)

PR #3112, opened 2023-12-07
gui: fix undefined behavior

Macdu:
It is undefined behavior to compare iterators of two different vectors.
It is undefined behavior to use the content of a variable outside of its living scope.

### DIFF
--- a/vita3k/gui/src/content_manager.cpp
+++ b/vita3k/gui/src/content_manager.cpp
@@ -103,7 +103,7 @@ static void get_save_data_list(GuiState &gui, EmuEnvState &emuenv) {
 
     for (const auto &save : fs::directory_iterator(SAVE_PATH)) {
         const auto title_id = save.path().stem().generic_string();
-        if (fs::is_directory(save.path()) && !fs::is_empty(save.path()) && get_app_index(gui, title_id) != gui.app_selector.user_apps.end()) {
+        if (fs::is_directory(save.path()) && !fs::is_empty(save.path()) && get_app_index(gui, title_id)) {
             tm updated_tm = {};
 
             const auto last_writen = fs::last_write_time(save);
@@ -363,7 +363,7 @@ void draw_content_manager(GuiState &gui, EmuEnvState &emuenv) {
                     if (menu == "app") {
                         fs::remove_all(emuenv.pref_path / "ux0/app" / content.first);
                         fs::remove_all(emuenv.pref_path / "ux0/addcont" / content.first);
-                        gui.app_selector.user_apps.erase(get_app_index(gui, content.first));
+                        gui.app_selector.user_apps.erase(gui.app_selector.user_apps.begin() + (get_app_index(gui, content.first) - &gui.app_selector.user_apps[0]));
                         gui.app_selector.user_apps_icon.erase(content.first);
                     }
                     const auto SAVE_PATH{ emuenv.pref_path / "ux0/user" / emuenv.io.user_id / "savedata" / content.first };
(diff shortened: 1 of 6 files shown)


bookmist:
There are other places where get_app_index is compared to some of `end()` iterators. For example: https://github.com/Vita3K/Vita3K/blob/fe7d6f01655b96d5ab1f8b998bbb0562b25a5193/vita3k/gui/src/main_menubar.cpp#L90

vita3k/gui/src/gui.cpp
```diff
@@ -389,16 +388,14 @@ void init_app_background(GuiState &gui, EmuEnvState &emuenv, const std::string &
     else
         vfs::read_app_file(buffer, emuenv.pref_path.wstring(), app_path, "sce_sys/pic0.png");
 
-    const auto title = (APP_INDEX != gui.app_selector.sys_apps.end()) && (APP_INDEX != gui.app_selector.user_apps.end()) ? APP_INDEX->title : app_path;
```
Zangetsu38, line 392: I no see any problèm here ?
Is just fine 🤔, same code is used also in other place

bookmist: It crash on debug build at least.

Zangetsu38: Is strange but ok 🤔

Macdu:
> There are other places where get_app_index is compared to some of `end()` iterators. For example:
> 
> https://github.com/Vita3K/Vita3K/blob/fe7d6f01655b96d5ab1f8b998bbb0562b25a5193/vita3k/gui/src/main_menubar.cpp#L90

Yes you are right, I am going to make something a bit better instead then (return a pointer instead of an iterator)

Macdu:
Ok, I modified the signature of get_app_index. That's not the cleanest thing (because some parts of the code assume get_app_index always return a value in user_apps) but it's better than what is being currently done and allows the debug builds no longer to crash.

vita3k/gui/src/content_manager.cpp
```diff
@@ -103,7 +103,7 @@ static void get_save_data_list(GuiState &gui, EmuEnvState &emuenv) {
 
     for (const auto &save : fs::directory_iterator(SAVE_PATH)) {
         const auto title_id = save.path().stem().generic_string();
-        if (fs::is_directory(save.path()) && !fs::is_empty(save.path()) && get_app_index(gui, title_id) != gui.app_selector.user_apps.end()) {
+        if (fs::is_directory(save.path()) && !fs::is_empty(save.path()) && get_app_index(gui, title_id) != nullptr) {
```
Zangetsu38, line 106:
```suggestion
        if (fs::is_directory(save.path()) && !fs::is_empty(save.path()) && get_app_index(gui, title_id)) {
```

Here like this should works too no ? You using this for title 🤔

vita3k/gui/src/main_menubar.cpp
```diff
@@ -87,7 +87,7 @@ static void draw_emulation_menu(GuiState &gui, EmuEnvState &emuenv) {
                 for (auto i = 0; i < std::min(8, int32_t(gui.time_apps[emuenv.io.user_id].size())); i++) {
                     const auto time_app = gui.time_apps[emuenv.io.user_id][i];
                     const auto app_index = get_app_index(gui, time_app.app);
-                    if ((app_index != gui.app_selector.user_apps.end()) && (app_index != gui.app_selector.sys_apps.end()))
+                    if (app_index != nullptr)
```
Zangetsu38, line 90:
```suggestion
                    if (app_index)
```
Should he same here

vita3k/gui/src/content_manager.cpp
Macdu: Yes, this also works, but I feel like the != nullptr is a bit more expressive (you get to know that get_app_index returns a pointer and not a boolean). Although that's only personal preference.

Zangetsu38: Vita3k more no using it, dev have some time refactor for remove all of this with compare on nullptr 
Is more clean

Zangetsu38 (approved):

bookmist (approved):

PR #3113, opened 2023-12-08
Initialize cache_path when installing packages from the command line

xperia64:
When installing an app or a patch, Vita3k creates a license file based on the zRIF. This currently requires `emuenv.cache_path` to be initialized, as the temporary license is created there, and then moved into the proper location. This path is not currently initialized when installing packages via a command-line argument, as it is only ever initialized in `app::init`.

This patch initializes this path in the CLI pkg install code path in `main`, similar to `pref_path`.


Macdu (approved):
Thank you a lot for this PR and sorry for the delay!

PR #3128, opened 2023-12-15
shader: Fix constant loading, dual absolute + typo

Macdu:
Some shader fixes (the constant table being used for f32 loading could be wrong and 16-bit integer loading was also wrong).
Also the absolute flag was used when it shouldn't for some cases of the vdual instruction.

This fixes all the remaining graphical issues in 2D ubisoft games (rayman and child of light)


EXtremeExploit (approved):
yes macdu-sama

PR #3132, opened 2023-12-17
.github: Add bot labeler

nishinji:
No one is putting any effort into labeling, and this will continue to be the case.
Bots may make mistakes, but it is better than no labels at all.


EXtremeExploit (approved):
nice

PR #3134, opened 2023-12-17
kernel: Page-align the stack size

Macdu:
The kernel rounds up the stack size to the next page size.
Spiderman was allocating a stack of size 0x180c4 and expecting the stack pointer to be 8 byte aligned, which wasn't the case.
This allows Spiderman to go ingame---.


EXtremeExploit (approved):
awesome

PR #3144, opened 2023-12-20
app: fix incorrect AppImage assets handling

nn9dev:
Fixes an issue where if Vita3K is called to run from an AppImage (not AS an AppImage) it will incorrectly believe that it is running AS an AppImage, and try to read its files from a place they do not exist

### DIFF
--- a/vita3k/app/src/app_init.cpp
+++ b/vita3k/app/src/app_init.cpp
@@ -160,7 +160,7 @@ void init_paths(Root &root_paths) {
             root_paths.set_static_assets_path(fs::path(XDG_DATA_HOME) / app_name / dir_sep);
     }
 
-    if (APPDIR != NULL) {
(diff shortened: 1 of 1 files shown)


nn9dev:
i learn something new about git every time i make a commit

vita3k/app/src/app_init.cpp
```diff
@@ -160,9 +159,12 @@ void init_paths(Root &root_paths) {
             root_paths.set_static_assets_path(fs::path(XDG_DATA_HOME) / app_name / dir_sep);
     }
 
+#if defined(WE_ARE_APPIMAGE)
+    auto APPDIR = getenv("APPDIR"); // Used in AppImage
     if (APPDIR != NULL) {
```
Macdu, line 164:
```suggestion
    if (APPDIR != NULL && fs::exists(fs::path(APPDIR) / "usr/share/Vita3K")) {
```
much better solution

nn9dev:
done 👍

Macdu (approved):

PR #3171, opened 2024-01-08
gui/lang: update japanese for settings dialog

nishinji:
Translations added against the settings dialog for Japanese.

### DIFF
--- a/lang/system/ja.xml
+++ b/lang/system/ja.xml
@@ -4,6 +4,7 @@
 	<main_menubar>
 		<file name="ファイル">
 			<open_pref_path>保存先を開く</open_pref_path>
+			<open_textures_path>テクスチャパスを開く</open_textures_path>
 			<install_firmware>ファームウェアをインストール</install_firmware>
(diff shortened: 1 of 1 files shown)


Croden1999 (approved):

Macdu (approved):

lang/system/ja.xml
```diff
@@ -97,6 +102,7 @@
 		<file_corrupted>ファイルが壊れています。</file_corrupted>
 		<microphone_disabled>マイクを有効にしてください。</microphone_disabled>
 		<no>いいえ</no>
+		<ok>OK</ok>
```
Croden1999, line 105: no need add it, the same name exists in english

Croden1999 (changes requested):

lang/system/ja.xml
Zangetsu38: Yes, is Ok for all

nishinji: whichever, but done.

Macdu (approved):

PR #3179, opened 2024-01-14
Typo fixes - Season 2

nishinji:
I don't like this type of commit because it not only doesn't do anything, but could break something. Although there are too many of them, so I'll pr.
@Zangetsu38 Please review this 😄


vita3k/renderer/src/texture/pvrt-dec.cpp
```diff
@@ -400,7 +400,7 @@ static int32_t getModulationValues(int32_t i32ModulationValues[16][8], int32_t i
                 return RepVals0[i32ModulationValues[xPos][yPos]] + modulationFlag;
             }
 
-            // else average from the neighbours
+            // else average from the neighbors
```
Macdu, line 403: This sentence comes from the powervr sdk and is valid american english, no need to change it.

nishinji: Done

Macdu (approved):
Really nice

PR #3180, opened 2024-01-14
.github: Create GItHub issue templates

nishinji:
Created with reference to other emulators.
You can check here https://github.com/nishinji/Vita3K/issues/new/choose.
image

### DIFF
--- a/.github/ISSUE_TEMPLATE/1-bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/1-bug_report.yml
@@ -1,0 +1,63 @@
+name: Bug Report
+description: Template for reporting bugs
+title: "[BUG]: "
+labels: "bug"
+body:
+  - type: markdown
+    attributes:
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/1-bug_report.yml
```diff
@@ -0,0 +1,64 @@
+name: Bug Report
+description: Template for reporting bugs
+title: "[BUG]: "
+labels: "bug"
+body:
+  - type: markdown
+    attributes:
+      value: Tech support does not belong here. You should only file an issue here if you think you have experienced an actual bug with Vita3K.
+             We are **not** accepting bug reports for specific games at this time.
```
Macdu, line 9: I think this line should be removed. It's not that we are not accepting game specific bug reports but indeed as you say below that they must go to the game specific issue on the compatibility repository.

nishinji: If you say so.

Macdu (approved):

PR #3184, opened 2024-01-16
renderer/vulkan: Fix depth/stencil identification

Macdu:
Fix some regressions from my surface cache rewrite. This caused D24S8 (but not only) depth sampling to sample the stencil instead.

This fixes the light/shadow regression in gravity rush and should fix a few other regressions too.

### DIFF
--- a/vita3k/renderer/src/vulkan/surface_cache.cpp
+++ b/vita3k/renderer/src/vulkan/surface_cache.cpp
@@ -780,7 +780,7 @@ std::optional<TextureLookupResult> VKSurfaceCache::retrieve_depth_stencil_as_tex
     if (cached_info.multisample_mode == SCE_GXM_MULTISAMPLE_4X)
         height /= 2;
 
-    const bool is_stencil = cached_info.surface.stencil_data.address() == address;
+    const bool is_stencil = can_be_stencil;
 
     vk::ImageView ds_attachment = reinterpret_cast<VKContext *>(state.context)->current_ds_view;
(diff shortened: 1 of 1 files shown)


vita3k/renderer/src/vulkan/surface_cache.cpp
```diff
@@ -780,7 +780,7 @@ std::optional<TextureLookupResult> VKSurfaceCache::retrieve_depth_stencil_as_tex
     if (cached_info.multisample_mode == SCE_GXM_MULTISAMPLE_4X)
         height /= 2;
 
-    const bool is_stencil = cached_info.surface.stencil_data.address() == address;
+    const bool is_stencil = can_be_stencil;
```
Zangetsu38, line 783: no can juste remove "is_stencil" and using directly "can_be_stencil" ?

Macdu: It doesn't really have the same meaning, there is a can_be_stencil and can_be_color declaration at the beginning meaning the texture may be a stencil or a color and here is_stencil means for sure it is a stencil.

Macdu: Using a variable named can_be_stencil below can lead to a misunderstanding and using a variable named is_stencil above can also lead to a misunderstanding.

Zangetsu38: I see 🤔

EXtremeExploit (approved):

PR #3191, opened 2024-01-19
vita3k: set fs::path type for string paths and refactoring

bookmist:
avoid not need conversion string->wstring->string for linux and macos 
native logging of fs::path
more correct conversion from std path to boost path 
refactor pathes for shaders compiler, shorten parameters lists 
crypto::hex_string: remove not needed null character as last character in string

### DIFF
--- a/vita3k/interface.cpp
+++ b/vita3k/interface.cpp
@@ -84,9 +84,9 @@ static bool is_nonpdrm(EmuEnvState &emuenv, const fs::path &output_path) {
     const auto app_license_path{ emuenv.pref_path / "ux0/license" / emuenv.app_info.app_title_id / fmt::format("{}.rif", emuenv.app_info.app_content_id) };
     const auto is_patch_found_app_license = (emuenv.app_info.app_category == "gp") && fs::exists(app_license_path);
     if (fs::exists(output_path / "sce_sys/package/work.bin") || is_patch_found_app_license) {
-        std::string licpath = is_patch_found_app_license ? app_license_path.string() : output_path.string() + "/sce_sys/package/work.bin";
-        LOG_INFO("Decrypt layer: {}", output_path.string());
-        if (!decrypt_install_nonpdrm(emuenv, licpath, output_path.string())) {
+        fs::path licpath = is_patch_found_app_license ? app_license_path : output_path / "sce_sys/package/work.bin";
+        LOG_INFO("Decrypt layer: {}", output_path);
+        if (!decrypt_install_nonpdrm(emuenv, licpath, output_path)) {
             LOG_ERROR("NoNpDrm installation failed, deleting data!");
             fs::remove_all(output_path);
             return false;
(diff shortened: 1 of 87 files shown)


vita3k/util/include/util/fs.h
```diff
@@ -19,9 +19,156 @@
...
+ * \return First path + second path
+ */
+fs::path path_concat(const fs::path &path1, const fs::path &path2);
+
+/**
+ * \brief Check if a directory exists and create the directory and all its parents if they don't exist.
+ * \param path Directory to check
+ * \return True if the directory exists or was created successfully, false otherwise.
+ */
+bool force_directories(const fs::path &path);
```
Macdu, line 74: fs::create_directories already does exactly this (the call to fs::exists before is useless)

bookmist: I've seen such code (`if (!fs::exists(dir)) create_directories(dir);`) If it's ok to just call `create_directories(dir);` it's great

vita3k/interface.cpp
```diff
@@ -124,15 +124,15 @@ static bool set_content_path(EmuEnvState &emuenv, const bool is_theme, fs::path
...
 }
 
-bool install_archive_content(EmuEnvState &emuenv, GuiState *gui, const fs::path &archive_path, const ZipPtr &zip, const std::string &content_path, const std::function<void(ArchiveContents)> &progress_callback) {
+bool install_archive_content(EmuEnvState &emuenv, GuiState *gui, const ZipPtr &zip, const std::string &content_path, const std::function<void(ArchiveContents)> &progress_callback) {
     std::string sfo_path = "sce_sys/param.sfo";
     std::string theme_path = "theme.xml";
     vfs::FileBuffer buffer, theme;
 
-    const auto is_theme = mz_zip_reader_extract_file_to_callback(zip.get(), (fs::path(content_path) / theme_path).string().c_str(), &write_to_buffer, &theme, 0);
+    const auto is_theme = mz_zip_reader_extract_file_to_callback(zip.get(), (content_path + "/" + theme_path).c_str(), &write_to_buffer, &theme, 0);
```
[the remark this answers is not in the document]
bookmist: We get `content_path` from `mz_zip_reader_file_stat` and put it back into `mz_zip_reader_extract_file_to_callback` I think character encoding here is not important at all. (and also path.string() does not convert to utf-8 on old windows (it convert to application (or system) ANSI codepage))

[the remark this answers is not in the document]
bookmist: But you are right, installing from zip needs re-checking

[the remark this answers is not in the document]
Macdu: My bad, I didn't notice you were not using a fs path for this specific case

vita3k/util/include/util/fs.h
Macdu: On the documentation of `create_directories(dir);` for both std and boost, it is written that this function has no effect if the directory exists. The only case where force_directories would behave differently compared to just calling create_directories is if dir already exists but as a file, create_directories would throw an exception while force_directories will do nothing, I am fine with the exception as this defintely wouldn't be expected.

bookmist: Great. Thank you. I replaced force_directories to create_directories
(Also I checked zip unpacking and now it's ready to review)

Macdu (approved):
Really nice!

PR #3192, opened 2024-01-20
Fix building RelDebug

dima-xd:
Me and nishinji were having problems building RelDebug.
That's because this macro is conflicting with tracy submodule. So I decided to rename it.
If you have better idea for new name or other ways to fix it, feel free to tell.

### DIFF
--- a/vita3k/util/include/util/keywords.h
+++ b/vita3k/util/include/util/keywords.h
@@ -18,7 +18,5 @@
 #pragma once
 
 #if defined(_MSC_VER)
-#define restrict __restrict
-#else
-#define restrict __restrict__
+#define __restrict__ __restrict
 #endif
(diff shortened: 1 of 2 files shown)


vita3k/util/include/util/keywords.h
```diff
@@ -18,7 +18,7 @@
 #pragma once
 
 #if defined(_MSC_VER)
-#define restrict __restrict
+#define __restrict__ __restrict
 #else
-#define restrict __restrict__
+#define __restrict__ __restrict__
 #endif
```
Macdu, line 24:
```
#if defined(_MSC_VER)
#define __restrict__ __restrict
#endif

Macdu (approved):

PR #3193, opened 2024-01-21
vita3k: Improve path support for Linux and macOS.

Steveice10:
Improves path support in the following ways:
* macOS:
  * Config and other files are currently placed inside the app bundle's `Resources` directory, as by default this is what `SDL_GetBasePath` returns. This can be inconvenient as if the app bundle is ever deleted/replaced, config files will be lost. Users will also likely not even realize the files are being stored here as generally app bundles are not used to store user data. Instead, use the `SDL_GetPrefPath` "~/Library/Application Support/Vita3K/Vita3K" by default.
    * We can't just use the path outside the app bundle as the user may install Vita3K to "/Applications/", which should not be used to store emulator data.
* macOS & Linux:
  * Since macOS now uses the user data path by default, and Linux already uses the XDG paths by default, introduce a new "portable" directory concept where, if the user creates a directory named "portable" next to the executable/app bundle, it will use that for all user data by default. This preserves the case where users want to keep all of their data together in a more portable location. Some other emulators like Citra, Yuzu, Ryujinx, Dolphin, etc have the same or similar concepts, so this will be familiar to users who want this.

### DIFF
--- a/vita3k/app/src/app_init.cpp
+++ b/vita3k/app/src/app_init.cpp
@@ -104,83 +104,125 @@ void update_viewport(EmuEnvState &state) {
 
 void init_paths(Root &root_paths) {
     auto sdl_base_path = SDL_GetBasePath();
-    auto sdl_pref_path = SDL_GetPrefPath(org_name, app_name);
     auto base_path = fs_utils::utf8_to_path(sdl_base_path);
-    auto pref_path = fs_utils::utf8_to_path(sdl_pref_path);
     SDL_free(sdl_base_path);
-    SDL_free(sdl_pref_path);
+
     root_paths.set_base_path(base_path);
-    root_paths.set_pref_path(pref_path);
-    root_paths.set_log_path(base_path);
-    root_paths.set_config_path(base_path);
     root_paths.set_static_assets_path(base_path);
-    root_paths.set_shared_path(base_path);
-    root_paths.set_cache_path(base_path / "cache" / "");
+
+#if defined(__APPLE__)
+    // On Apple platforms, base_path is "Contents/Resources/" inside the app bundle.
+    // An extra parent_path is apparently needed because of the trailing slash.
+    auto portable_path = base_path.parent_path().parent_path().parent_path().parent_path() / "portable" / "";
+#else
+    auto portable_path = base_path / "portable" / "";
+#endif
+
+    if (fs::is_directory(portable_path)) {
+        // If a portable directory exists, use it for everything else.
+        // Note that pref_path should not be the same as the other paths.
+        root_paths.set_pref_path(portable_path / "fs" / "");
+        root_paths.set_log_path(portable_path);
+        root_paths.set_config_path(portable_path);
+        root_paths.set_shared_path(portable_path);
+        root_paths.set_cache_path(portable_path / "cache" / "");
+    } else {
+        // SDL_GetPrefPath is deferred as it creates the directory.
+        // When using a portable directory, it is not needed.
+        auto sdl_pref_path = SDL_GetPrefPath(org_name, app_name);
+        auto pref_path = fs_utils::utf8_to_path(sdl_pref_path);
+        SDL_free(sdl_pref_path);
+
+#if defined(__APPLE__)
+        // Store other data in the user-wide path. Otherwise we may end up dumping
+        // files into the "/Applications/" install directory or the app bundle.
+        // This will typically be "~/Library/Application Support/Vita3K/Vita3K/".
+        // Check for config.yml first, though, to maintain backwards compatibility,
+        // even though storing user data inside the app bundle is not a good idea.
+        auto existing_config = base_path / "config.yml";
+        if (!fs::exists(existing_config)) {
+            base_path = pref_path;
+        }
+
+        // pref_path should not be the same as the other paths.
+        // For backwards compatibility, though, check if ux0 exists first.
+        auto existing_ux0 = pref_path / "ux0";
+        if (!fs::is_directory(existing_ux0)) {
+            pref_path = pref_path / "fs" / "";
+        }
+#endif
+
+        root_paths.set_pref_path(pref_path);
+        root_paths.set_log_path(base_path);
+        root_paths.set_config_path(base_path);
+        root_paths.set_shared_path(base_path);
+        root_paths.set_cache_path(base_path / "cache" / "");
 
 #if defined(__linux__) && !defined(__ANDROID__) && !defined(__APPLE__)
-    // XDG Data Dirs.
-    auto env_home = getenv("HOME");
-    auto XDG_DATA_DIRS = getenv("XDG_DATA_DIRS");
-    auto XDG_DATA_HOME = getenv("XDG_DATA_HOME");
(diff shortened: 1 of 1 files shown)


vita3k/app/src/app_init.cpp
```diff
@@ -103,83 +103,116 @@ void update_viewport(EmuEnvState &state) {
...
-    root_paths.set_pref_path(string_utils::utf_to_wide(pref_path));
-    root_paths.set_log_path(string_utils::utf_to_wide(base_path));
-    root_paths.set_config_path(string_utils::utf_to_wide(base_path));
-    root_paths.set_static_assets_path(string_utils::utf_to_wide(base_path));
-    root_paths.set_shared_path(string_utils::utf_to_wide(base_path));
-    root_paths.set_cache_path(fs::path(string_utils::utf_to_wide(base_path)) / "cache" / dir_sep);
-    SDL_free(base_path);
-    SDL_free(pref_path);
+    auto base_path_sdl = SDL_GetBasePath();
+    fs::path base_path(base_path_sdl);
```
bookmist, line 107: May be incorrect for windows earlier then Version 1903 (May 2019 Update) (see here https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page). SDL uses utf8 everywhere, so we need to convert from utf8 to wstring (or current application codepage) directly.

Macdu:
Really nice, your PR will conflict a lot with PR #3191, so we'll have to wait for the mentioned PR to be merged first and for you to fix the conflict before reviewing and merging this PR.

vita3k/app/src/app_init.cpp
Steveice10: Should be fixed, thanks.

nn9dev:
I have two gripes, both very minor. 
- the name "portable". It's logical, but I think it could be better. Maybe "portableFS" or "Vita3K_FS" since that's more in-line with what it *is*? Obviously very up for discussion, and I'm also okay proceeding as-is. Just a nitpick.
- I ***WHOLLY*** understand if you don't want to with ImGui being ImGui (and it being somewhat grueling to work with), but a tickbox in the initial setup would be nice to enable/disable a portable install. Obviously though, someone else could do this if you don't feel like it's within the scope of this PR.

Also, the attention for MacOS is greatly appreciated. I didn't even know we weren't putting stuff there already lol

Steveice10:
> * the name "portable". It's logical, but I think it could be better. Maybe "portableFS" or "Vita3K_FS" since that's more in-line with what it _is_? Obviously very up for discussion, and I'm also okay proceeding as-is. Just a nitpick.

I don’t think it should necessarily involve FS because it’s not just the FS, it includes everything like the config files, etc as well.

> * I _**WHOLLY**_ understand if you don't want to with ImGui being ImGui (and it being somewhat grueling to work with), but a tickbox in the initial setup would be nice to enable/disable a portable install. Obviously though, someone else could do this if you don't feel like it's within the scope of this PR.

Probably out of scope for this although like you say someone could add it after.

Steveice10:
Rebased on top of latest commit.

Macdu:
Your approach, on apple or when a portable directory is used, set the pref path to be the same as the log/shared/cache path. That's not good as this means that Vita3K's folders (shaderlog, cache, ...) will be right next to the PS Vita filesystem's folders (vs0, ux0, ...). This should be avoided (pref_path which contains the ps vita file system should always be different compared to the other paths)

Steveice10:
> Your approach, on apple or when a portable directory is used, set the pref path to be the same as the log/shared/cache path. That's not good as this means that Vita3K's folders (shaderlog, cache, ...) will be right next to the PS Vita filesystem's folders (vs0, ux0, ...). This should be avoided (pref_path which contains the ps vita file system should always be different compared to the other paths)

Addressed by moving `set_pref_path` to a subdirectory in these cases.

j8r:
Can it help solving the `render_main.vert` permission issue of https://github.com/Vita3K/Vita3K/issues/1418?

Steveice10:
> Can it help solving the `render_main.vert` permission issue of #1418?

I don't think so, for Linux this just adds a portable mode, the XDG directory stuff is unchanged.

Macdu:
The only issue I have left with this PR is that it will change the pref path on MacOS for existing users. Without any notice, it will look that they have no games anymore.
I would guess if pref_path / "ux0" exists on MacOS, keep the pref path as it is (don't add "fs" to it) as a workaround for it (although it means vita3k and the vita os folders will get mixed, which was what I wanted to avoid).

Macdu:
Also @EXtremeExploit @KhoraLee do you have any comment about this PR?

Steveice10:
> The only issue I have left with this PR is that it will change the pref path on MacOS for existing users. Without any notice, it will look that they have no games anymore. I would guess if pref_path / "ux0" exists on MacOS, keep the pref path as it is (don't add "fs" to it) as a workaround for it (although it means vita3k and the vita os folders will get mixed, which was what I wanted to avoid).

Implemented logic for this, and also simplified things a bit so there's fewer branching preprocessor logic paths.

Steveice10:
Have also added backwards compatibility for the config.yml inside the app bundle just in case. Although it's not a great idea to have user data inside the app bundle I'll err on the side of not breaking any existing setups.

EXtremeExploit:
> Also @EXtremeExploit @KhoraLee do you have any comment about this PR?

Looks fine to me, i also approved workflows (i thought i already did this)

Macdu (approved):
lgtm, I'll wait one day in case KhoraLee has some feedback to give

vita3k/app/src/app_init.cpp
```diff
@@ -104,83 +104,125 @@ void update_viewport(EmuEnvState &state) {
...
-    root_paths.set_log_path(base_path);
-    root_paths.set_config_path(base_path);
     root_paths.set_static_assets_path(base_path);
-    root_paths.set_shared_path(base_path);
-    root_paths.set_cache_path(base_path / "cache" / "");
+
+#if defined(__APPLE__)
+    // On Apple platforms, base_path is "Contents/Resources/" inside the app bundle.
+    // An extra parent_path is apparently needed because of the trailing slash.
+    auto portable_path = base_path.parent_path().parent_path().parent_path().parent_path() / "portable" / "";
```
KhoraLee, line 116: https://wiki.libsdl.org/SDL2/SDL_GetBasePath

SDL_GetBasePath's behaviour can be overridden by adding a property to the Info.plist file

KhoraLee (changes requested):
I couldn't test it because I'm working as NTO at YOGOC now. But code itself looks fine.

vita3k/app/src/app_init.cpp
Steveice10: I learned about that before, but we actually do still need the Resources path for the static assets and update script, and if I change the SDL behavior to give the directory the bundle is in, we won’t know what the bundle is called to get the Resources directory.

Open to any ideas on how to deal with that if you want.

KhoraLee: Ah okay. Then using current implentation for a while and chaing SDL_GetBasePath's behaviour when deprecating 'Contents/Resources' usage would be better.

PR #3194, opened 2024-01-23
config: Implement global volume percent setting.

Steveice10:
Adds a setting to configure a global audio volume. This works by multiplying each audio port's volume by the global volume at the adapter level. For example, if you set the global volume to 50%, it will multiply each port's volume by 0.5, cutting it in half.

The setting is only usable with the Cubeb backend, as the SDL backend does not provide a means of changing the volume.

### DIFF
--- a/vita3k/lang/include/lang/state.h
+++ b/vita3k/lang/include/lang/state.h
@@ -587,6 +587,8 @@ struct LangState {
             { "boot_apps_full_screen", "Boot apps in full screen" },
             { "audio_backend", "Audio Backend" },
             { "select_audio_backend", "Select your preferred audio backend." },
+            { "audio_volume", "Audio Volume" },
+            { "audio_volume_description", "Adjusts the volume percentage of all audio outputs." },
             { "enable_ngs_support", "Enable NGS support" },
             { "ngs_description", "Uncheck the box to disable support for advanced audio library NGS." },
             { "trace", "Trace" },
--- a/lang/system/en-gb.xml
+++ b/lang/system/en-gb.xml
@@ -667,6 +667,8 @@ Some applications don't use this and get default confirmation button.</enter_but
 			<boot_apps_full_screen>Boot apps in full screen</boot_apps_full_screen>
 			<audio_backend>Audio Backend</audio_backend>
 			<select_audio_backend>Select your preferred audio backend.</select_audio_backend>
+			<audio_volume>Audio Volume</audio_volume>
+			<audio_volume_description>Adjusts the volume percentage of all audio outputs.</audio_volume_description>
 			<enable_ngs_support>Enable NGS support</enable_ngs_support>
 			<ngs_description>Uncheck the box to disable support for advanced audio library NGS.</ngs_description>
 			<trace>Trace</trace>
--- a/lang/system/en.xml
+++ b/lang/system/en.xml
@@ -667,6 +667,8 @@ Some applications don't use this and get default confirmation button.</enter_but
 			<boot_apps_full_screen>Boot apps in full screen</boot_apps_full_screen>
 			<audio_backend>Audio Backend</audio_backend>
 			<select_audio_backend>Select your preferred audio backend.</select_audio_backend>
+			<audio_volume>Audio Volume</audio_volume>
+			<audio_volume_description>Adjusts the volume percentage of all audio outputs.</audio_volume_description>
 			<enable_ngs_support>Enable NGS support</enable_ngs_support>
 			<ngs_description>Uncheck the box to disable support for advanced audio library NGS.</ngs_description>
 			<trace>Trace</trace>
--- a/vita3k/app/src/app_init.cpp
+++ b/vita3k/app/src/app_init.cpp
@@ -348,7 +348,7 @@ bool late_init(EmuEnvState &state) {
             thread->update_status(ThreadStatus::run);
         }
     };
-    if (!state.audio.init(resume_thread, state.cfg.audio_backend)) {
+    if (!state.audio.init(resume_thread, state.cfg.audio_backend, state.cfg.current_config.audio_volume)) {
         LOG_WARN("Failed to initialize audio! Audio will not work.");
     }
 
--- a/vita3k/audio/include/audio/state.h
+++ b/vita3k/audio/include/audio/state.h
@@ -114,11 +114,13 @@ struct AudioState {
     AudioInPort in_port;
     ResumeAudioThread resume_thread;
     std::string audio_backend;
+    float global_volume;
 
-    bool init(const ResumeAudioThread &resume_thread, const std::string &adapter_name);
+    bool init(const ResumeAudioThread &resume_thread, const std::string &adapter_name, float global_volume);
     void set_backend(const std::string &adapter_name);
     AudioOutPortPtr open_port(int nb_channels, int freq, int nb_sample);
     void audio_output(ThreadState &thread, AudioOutPort &out_port, const void *buffer);
     void set_volume(AudioOutPort &out_port, float volume);
+    void set_global_volume(float volume);
     void switch_state(const bool pause);
 };
--- a/vita3k/audio/src/audio.cpp
+++ b/vita3k/audio/src/audio.cpp
@@ -30,7 +30,7 @@
 #include <cassert>
 #include <cstring>
 
-static void mix_out_port(uint8_t *stream, uint8_t *temp_buffer, int len, AudioOutPort &port, const ResumeAudioThread &resume_thread) {
+static void mix_out_port(uint8_t *stream, uint8_t *temp_buffer, int len, float global_volume, AudioOutPort &port, const ResumeAudioThread &resume_thread) {
     ZoneScopedC(0xF6C2FF); // Tracy - Track function scope with color thistle
 
     // How much data is available?
@@ -57,7 +57,7 @@ static void mix_out_port(uint8_t *stream, uint8_t *temp_buffer, int len, AudioOu
     const int bytes_got = SDL_AudioStreamGet(port.stream.get(), temp_buffer, bytes_to_get);
     lock.unlock();
     if (bytes_got > 0) {
-        SDL_MixAudioFormat(stream, temp_buffer, AUDIO_S16LSB, bytes_got, static_cast<int>(port.volume * SDL_MIX_MAXVOLUME));
+        SDL_MixAudioFormat(stream, temp_buffer, AUDIO_S16LSB, bytes_got, static_cast<int>(port.volume * global_volume * SDL_MIX_MAXVOLUME));
     }
 }
 
@@ -77,14 +77,15 @@ void AudioAdapter::audio_callback(uint8_t *stream, int len_bytes) {
     std::memset(stream, state.spec.silence, len_bytes);
 
     for (const AudioOutPortPtr &port : ports) {
-        mix_out_port(stream, temp_buffer.data(), len_bytes, *port.get(), state.resume_thread);
+        mix_out_port(stream, temp_buffer.data(), len_bytes, state.global_volume, *port.get(), state.resume_thread);
     }
 
     FrameMarkNamed("Audio"); // Tracy - End discontinuous frame for audio rendering
 }
 
-bool AudioState::init(const ResumeAudioThread &resume_thread, const std::string &adapter_name) {
+bool AudioState::init(const ResumeAudioThread &resume_thread, const std::string &adapter_name, float global_volume) {
     this->resume_thread = resume_thread;
+    this->global_volume = global_volume;
 
     set_backend(adapter_name);
     if (!adapter)
@@ -135,6 +136,7 @@ AudioOutPortPtr AudioState::open_port(int nb_channels, int freq, int nb_sample) 
     } else {
         // let the adapter open the port
         AudioOutPortPtr port = adapter->open_port(nb_channels, freq, nb_sample);
+        adapter->set_volume(*port, global_volume);
         return port;
     }
 }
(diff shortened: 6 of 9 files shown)


Macdu:
Is this really useful? The code is fine as it is but I personally do not really see a use to it as you can set Vita3K's volume in the OS. Are there other emulators with this feature?

Steveice10:
> Is this really useful? The code is fine as it is but I personally do not really see a use to it as you can set Vita3K's volume in the OS. Are there other emulators with this feature?

* Some OSes do not have a built in per-app volume control, eg Mac.
* Yes, most other emulators I’ve used allow you to set a volume within the emulator settings.

Zangetsu38:
> Is this really useful? The code is fine as it is but I personally do not really see a use to it as you can set Vita3K's volume in the OS. Are there other emulators with this feature?

Litteraly all 🫡

Macdu:
Fine, also it should be quite straightforward to implement it for the sdl audio backend.
Here https://github.com/Vita3K/Vita3K/blob/d3f3038aeeed3edd1904dbbcae6ddc8b95839194/vita3k/audio/src/impl/sdl_audio.cpp#L29

If the volume is not 1.0f, then just multiply each entry in stream (state.spec.samples * 2 unsigned 16-bit entries) by the volume.

Steveice10:
Figured out my way around how the single-stream code works and applied an update to make it work for all single-stream backends: multiplying the volume in when making the `SDL_MixAudioFormat` call. Works on both backends now. I also made it clear that `set_volume` on the adapter should only be used for non single-stream backends, otherwise port volumes could be double-applied if it is implemented.

Zangetsu38:
i have already try implement this little old time, but get little probleme with ngs, because ngs can ajust volume of audio too, so need like set, you set 50 %, and game set 50 %, want sait, ngs need set no 50 % but 50 % of this global 50 % set  hmm
i have try in sdl

Steveice10:
> i have already try implement this little old time, but get little probleme with ngs, because ngs can ajust volume of audio too, so need like set, you set 50 %, and game set 50 %, want sait, ngs need set no 50 % but 50 % of this global 50 % set hmm i have try in sdl

The PR addresses the ability of the emulated game to set volumes by mixing the volume with the global volume, so for example if the global volume is 50% and you have two sources, one at 100% and one at 50%, the ultimate output is 50% and 25% respectively.

Zangetsu38:
> > i have already try implement this little old time, but get little probleme with ngs, because ngs can ajust volume of audio too, so need like set, you set 50 %, and game set 50 %, want sait, ngs need set no 50 % but 50 % of this global 50 % set hmm i have try in sdl
> 
> The PR addresses the ability of the emulated game to set volumes by mixing the volume with the global volume, so for example if the global volume is 50% and you have two sources, one at 100% and one at 50%, the ultimate output is 50% and 25% respectively.

yes, i have looked your last code, and yes, is exactly that, i let macvdu review this 👌

vita3k/interface.cpp
```diff
@@ -440,6 +440,7 @@ static ExitCode load_app_impl(SceUID &main_module_id, EmuEnvState &emuenv, const
 
     LOG_INFO("{}: {}", emuenv.cfg[e_cpu_backend], emuenv.cfg.current_config.cpu_backend);
     LOG_INFO_IF(emuenv.kernel.cpu_backend == CPUBackend::Dynarmic, "CPU Optimisation state: {}", emuenv.cfg.current_config.cpu_opt);
+    LOG_INFO("Audio volume: {}", emuenv.cfg.current_config.audio_volume);
```
Zangetsu38, line 443: exept for this, i no sur is really needed hmm 
i have  set this log for set important log for debug, but so for volume hmm

Steveice10: I put it in here so that if someone comes along with a problem like “I have no sound”, one look at the log would tell you that they might just have their volume set to 0. I can remove the log line though if it’s not wanted.

Zangetsu38: hmm I see, indeed;, well not sure that it happens lol
But I'll let Macdu decide :)

Kitakatarashima:
If you want to apply the global audio setting, my suggestion is to also add audio accuracy so that it is suitable for the global audio setting. I'm afraid it's just the global audio that the audio will crack or even worse for some games like that in my opinion.

Kitakatarashima:
If there is no audio accuracy, I'm afraid, but this is just my fear. If I set it to 50%, I'm afraid that there won't be any audio or the audio will be very small. In some games it can work, but for some other games it won't work. If I set it to 70%, there's no difference. with 100%, it could actually be 100%, it could be broken, I'm afraid that if there is audio accuracy, the Vita3K can adjust itself to the game, so whatever set it is, the audio in the game will still be normal, like that in my opinion.

Steveice10:
> If you want to apply the global audio setting, my suggestion is to also add audio accuracy so that it is suitable for the global audio setting. I'm afraid it's just the global audio that the audio will crack or even worse for some games like that in my opinion.

I’m not sure what you’re referring to but this doesn’t affect audio emulation, this just adjusts the volume of the output.

vita3k/interface.cpp
Macdu: I agree with Zangetsu here, I don't think it's necessary to put it in the log.

Kitakatarashima:
> > Jika Anda ingin menerapkan setting audio global, saran saya tambahkan juga akurasi audio agar sesuai dengan setting audio global. Saya khawatir hanya audio global yang audionya akan retak atau bahkan lebih buruk lagi untuk beberapa game seperti itu menurut saya.
> 
> Saya tidak yakin apa yang Anda maksud tetapi ini tidak memengaruhi emulasi audio, ini hanya menyesuaikan volume keluaran.

Yes, that's what I mean by the output. I'm afraid there will be problems. We don't know what the future will be like if we implement it like that.

Macdu:
You can ignore Kitikata saying random stuff.

Zangetsu38:
> Yes, that's what I mean by the output. I'm afraid there will be problems. We don't know what the future will be like if we implement it like that.

what problem do you want there to be, it only modifies the volume, nothing more, it causes 0 problems so

vita3k/interface.cpp
Steveice10: Okay, removed.

bookmist:
As far as I see, global volume can't be changed when a game started. I think, it's not obvious and it should be fixed or at least explained in the tooltip.

Steveice10:
> As far as I see, global volume can't be changed when a game started. I think, it's not obvious and it should be fixed or at least explained in the tooltip.

Can you elaborate on how this would be fixed? It's implemented in a way that it should be changable but I don't see a way for the config to be opened in a game, which itself would be a larger limitation than this PR seeks to address.

bookmist:
> I don't see a way for the config to be opened in a game
Set "Interface"/"GUI visible" on settings window before game start, then main menu does not hide ingame and it can be used. To show or hide main menu press "G".

Kitakatarashima:
> You can ignore Kitikata saying random stuff.

yeah, it's up to you, I'm just giving my opinion, if you don't want to hear me, yeah, whatever, but don't blame me for my hunch, because I'm always like that when there's something new.

Steveice10:
> > I don't see a way for the config to be opened in a game
> 
> Set "Interface"/"GUI visible" on settings window before game start, then main menu does not hide ingame and it can be used. To show or hide main menu press "G".

Thanks, I missed that option. I just checked and adjusting it does work when a game is running, make sure to hit "Save & Apply".

Kitakatarashima:
You guys know better than me, I can't do anything other than a random person who just gives me an opinion

nishinji:
Can you add more than 100% volume?
And this looks not good
image
Suggest using Imgui::SliderInt or SliderFloat

Macdu:
> Can you add more than 100% volume?
> And this is not good looking
> image
> Suggest using Imgui::SliderInt or SliderFloat

Setting the volume to more than 100% can cause the audio to saturate and also won't work with the current approach. Let's keep it at at most 100% for now.

Steveice10:
Updated to a SliderInt instead of DragInt.

Kitakatarashima:
I already said that the feature is not perfect, it needs some more adjustments. What I mean is that I suggest making the audio accurate so that it suits the game 😑😑 but I know I can't do anything and I'm not anyone here

Kitakatarashima:
> > Can you add more than 100% volume?
> > And this is not good looking
> > image
> > Suggest using Imgui::SliderInt or SliderFloat
> 
> Setting the volume to more than 100% can cause the audio to saturate and also won't work with the current approach. Let's keep it at at most 100% for now.

I already said that the feature is not perfect, it needs some more adjustments. What I mean is that I suggest making the audio accurate so that it suits the game 😑😑 but I know I can't do anything and I'm not anyone here but yes, I know I'm nobody here so it's whatever 😮‍💨

Kitakatarashima:
> > Can you add more than 100% volume?
> > And this is not good looking
> > image
> > Suggest using Imgui::SliderInt or SliderFloat
> 
> Setting the volume to more than 100% can cause the audio to saturate and also won't work with the current approach. Let's keep it at at most 100% for now.

bro, I already said that there must be audio accuracy so that the emulator's output can be in sync with the game's output, it doesn't have to be an option, isn't it, it's more about the coding, indeed for some games it can function well, but it's impossible for it to work for all games, bro. you know better than me, bro, bro, you have to realize that this is an emulator, not just one or two games, you have to think about most games, I'm not talking randomly, but this is my hunch, I'm scared, I'm like that. but yes, I know I'm nobody here so it's whatever 😮‍💨

Zangetsu38:
> 
> bro, I already said that there must be audio accuracy so that the emulator's output can be in sync with the game's output, it doesn't have to be an option, isn't it, it's more about the coding, indeed for some games it can function well, but it's impossible for it to work for all games, bro. you know better than me, bro, bro, you have to realize that this is an emulator, not just one or two games, you have to think about most games, I'm not talking randomly, but this is my hunch, I'm scared, I'm like that. but yes, I know I'm nobody here so it's whatever 😮‍💨

What do you fox said ? 🦊 
What you say makes no sense 🏴‍☠️ 
It's just a simple volume control, absolutely everything you say is off topic...

vita3k/lang/include/lang/state.h
```diff
@@ -587,6 +587,8 @@ struct LangState {
             { "boot_apps_full_screen", "Boot apps in full screen" },
             { "audio_backend", "Audio Backend" },
             { "select_audio_backend", "Select your preferred audio backend." },
+            { "audio_volume", "Audio Volume" },
+            { "audio_volume_description", "Applies a global volume multiplier to all audio outputs." },
```
Zangetsu38, line 591: this is little wrong, is not multipilier, but pourcent ex:
"Applies a certain percentage of the volume to all audio outputs."

Steveice10: I put multiplier as it's internally just multiplying by a float (e.g. 50% -> 0.5) but I agree that it should be reworded to match the user-facing control, will update.

Zangetsu38: yes ^^

Zangetsu38 (approved):

Kitakatarashima:
> > bro, I already said that there must be audio accuracy so that the emulator's output can be in sync with the game's output, it doesn't have to be an option, isn't it, it's more about the coding, indeed for some games it can function well, but it's impossible for it to work for all games, bro. you know better than me, bro, bro, you have to realize that this is an emulator, not just one or two games, you have to think about most games, I'm not talking randomly, but this is my hunch, I'm scared, I'm like that. but yes, I know I'm nobody here so it's whatever 😮‍💨
> 
> What do you fox said ? 🦊 What you say makes no sense 🏴‍☠️ It's just a simple volume control, absolutely everything you say is off topic...

yeah, even though the controls are simple, you still have to synchronize the emulator with the game so that the output is the same so that the output doesn't have any errors. If it's not synchronized, it can still run, but I'm sure it won't work for all games, there will definitely be problems if you keep setting set set. Bro, everyone knows about Android kids, right? That doesn't mean I only think about Android, but on a PC it's the same possibility, right?

Kitakatarashima:
There's a possibility that there won't be any audio, and there's a 50% chance that's very small, right? And there's a 100% chance that the audio will be broken, right? But it's still up to you, you understand better than me, if you don't want to hear me, try updating first. We'll see for ourselves whether my hunch is right or not, but I'm 50% sure of my hunch

Kitakatarashima:
I mean the audio is broken, it's not in the game, no, it's more about the device, whether it's a PC or Android, I mean the output is afraid it won't match the device. If the audio in the game is definitely safe, the problem is the output. That's what I said, I'm sure it's not possible for all games, meaning there must be some games on some devices, whether PC or Android, that don't match the output we set.

Kitakatarashima:
Now I ask you all, can you test it on all types of PC, huh??

nishinji:
@Kitakatarashima Please don't say random stuff about things you don't know. What you are saying is complete nonsense. It interferes with development.
If you continue this kind of behavior, I think you will be banned from this repository.

Kitakatarashima:
Yes, I won't say any more, I'll just be here, if no one believes me, it's up to you, you all know better than me, that's what you want, let's just see.
😁😁

Macdu (approved):

Zangetsu38:
> Hi everyone. Testing the most recent build with this change/implementation. A future option to go beyond 100% would be useful for games that by default can sound quite low compared to others. Example, Gravity Rush (lower volume) vs Persona 4 Golden. You can check a similar implementation on a emulator like Yuzu where you can move the slider till 200%. If you test a game there like the recent Prince of Persia the default volume is quite low but moving it to 200% improves dramatically without any distortion or issues.

Set more of 100 % no works, get no more volume

Enkak:
> > Hi everyone. Testing the most recent build with this change/implementation. A future option to go beyond 100% would be useful for games that by default can sound quite low compared to others. Example, Gravity Rush (lower volume) vs Persona 4 Golden. You can check a similar implementation on a emulator like Yuzu where you can move the slider till 200%. If you test a game there like the recent Prince of Persia the default volume is quite low but moving it to 200% improves dramatically without any distortion or issues.
> 
> Set more of 100 % no works, get no more volume

Not at the moment in Vita3K but please read what I wrote regarding the Yuzu emulator example. 
You have an option (global and per game) where you can move the slider there till 200% and you get more volume. RPCS3 is another example with the same option. Default is 100% with a slider to move lower or higher till 200%.

Zangetsu38:
> > > Hi everyone. Testing the most recent build with this change/implementation. A future option to go beyond 100% would be useful for games that by default can sound quite low compared to others. Example, Gravity Rush (lower volume) vs Persona 4 Golden. You can check a similar implementation on a emulator like Yuzu where you can move the slider till 200%. If you test a game there like the recent Prince of Persia the default volume is quite low but moving it to 200% improves dramatically without any distortion or issues.
> > 
> > 
> > Set more of 100 % no works, get no more volume
> 
> Not at the moment in Vita3K but please read what I wrote regarding the Yuzu emulator example. You have an option (global and per game) where you can move the slider there till 200% and you get more volume. RPCS3 is another example with the same option. Default is 100% with a slider to move lower or higher till 200%.

i know for rpcs3, but like i said, in vita3k set more of 100 % no works att all, and in SDL get just audio full broken

Kitakatarashima:
Turns out my hunch was right so now let's just enjoy it. The problem isn't my fault yes

PR #3198, opened 2024-01-29
gui/settings dialog; Add audio tab & small refactor.

Zangetsu38:
# About
- Add audio tab & small refactor on settings dialog.
- Remove double init of global volume value.

### DIFF
--- a/vita3k/gui/src/settings_dialog.cpp
+++ b/vita3k/gui/src/settings_dialog.cpp
@@ -185,6 +185,13 @@ static bool get_custom_config(GuiState &gui, EmuEnvState &emuenv, const std::str
                 config.fps_hack = gpu_child.attribute("fps-hack").as_bool();
             }
 
+            // Load Audio Config
+            if (!config_child.child("audio").empty()) {
+                const auto audio_child = config_child.child("audio");
+                config.audio_volume = audio_child.attribute("audio-volume").as_int();
+                config.ngs_enable = audio_child.attribute("enable-ngs").as_bool();
+            }
+
             // Load System Config
             const auto system_child = config_child.child("system");
             if (!system_child.empty())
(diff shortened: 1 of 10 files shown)


nishinji:
In SDL case will not work until push save in settings.

vita3k/gui/src/settings_dialog.cpp
```diff
@@ -185,6 +185,13 @@ static bool get_custom_config(GuiState &gui, EmuEnvState &emuenv, const std::str
                 config.fps_hack = gpu_child.attribute("fps-hack").as_bool();
             }
 
+            // Load Audio Config
+            if (!config_child.child("audio").empty()) {
+                const auto audio_child = config_child.child("audio");
+                config.audio_volume = audio_child.attribute("audio-volume").as_float();
```
bookmist, line 191: You changed type of `audio_volume` class field to int, but here you load it as float

Zangetsu38: 😐🤦‍♂️😂
Sorry 😐🤦‍♂️

Zangetsu38: Done 🗿

Zangetsu38:
> In SDL case will not work until push save in settings.

Is normal that, code is set like that 🤔
I no touch this here

nishinji:
> > In SDL case will not work until push save in settings.
> 
> Is normal that, code is set like that 🤔 I no touch this here

No, it works in master.
I don't know the details, but seemingly the volume value is not divided by 100 before press save.

Zangetsu38:
> 
> No, it works in master. I don't know the details, but seemingly the volume value is not divided by 100 before press save.

I no underboobs what you said here, code is exactly same before
And click on save for apply change is totaly normal, i change nothing on this

Zangetsu38:
> No, it works in master. I don't know the details, but seemingly the volume value is not divided by 100 before press save.

I no understund what you said here, code is exactly same before
And click on save for apply change is totaly normal, i change nothing on this

Macdu (approved):

PR #3202, opened 2024-02-02
renderer: Add screenshot support

Macdu:
Add support for taking screenshot on Vita3K.
Screenshot are taken directly from the PS Vita surface, meaning that they use the native resolution (at 8x they will be of size 7860x4352, even if your screen does not support it), it is possible to bind a key to take a screenshot.
This is implemented on both the opengl and vulkan renderer.


vita3k/renderer/src/vulkan/surface_cache.cpp
```diff
@@ -1218,4 +1218,53 @@ vk::ImageView VKSurfaceCache::sourcing_color_surface_for_presentation(Ptr<const
...
+
+    // this will cause a waitIdle, not an issue
+    vkutil::end_single_time_command(state.device, state.general_queue, state.general_command_pool, cmd_buffer);
+
+    memcpy(frame.data(), temp_buff.mapped_data, frame.size() * 4);
+
+    return frame;
+}
+
 }; // namespace renderer::vulkan
```
Zangetsu38, line 1270: if you can fix endline in same time 🙏

Macdu: done

EXtremeExploit (approved):
looks nice

vita3k/gui/src/controls_dialog.cpp
```diff
@@ -195,6 +197,7 @@ void draw_controls_dialog(GuiState &gui, EmuEnvState &emuenv) {
         ImGui::TableSetupColumn("button");
         ImGui::TableSetupColumn("mapped_button");
         remapper_button(gui, emuenv, &emuenv.cfg.keyboard_toggle_texture_replacement, lang["toggle_texture_replacement"].c_str());
+        remapper_button(gui, emuenv, &emuenv.cfg.keyboard_take_screenshot, "Take a screenshot");
```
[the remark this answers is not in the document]
Macdu: Yes, I'll let someone else add some translation and the lang support

eisnerguy1:
Thanks for adding this! 😁 👍🏻 What's the default path where the screenshots are saved? For the life of me I can't seem to find it. I'm on macOS btw.

Issue #3185, opened 2024-01-17
Can't load mp3 files for use in playing Project Diva f/f 2nd edit songs.


Wieurhel:
[22:55:35.365] |W| [export_sceAppUtilMusicMount::< lambda_1>::operator ()]: Unimplemented sceAppUtilMusicMount import called.
[22:55:35.367] |C| [translate_path]: Unknown device with path music0: used. Report this to the developers!
[22:55:35.367] |E| [open_dir]: Directory does not exist at: C:/Users/xxxxx/AppData/Roaming/Vita3K/Vita3K/_INVALID\ (target path: music0:)
[22:55:35.367] |W| [io_error_impl]: open_dir (_sceIoDopen) returned 0x80010002

This is the error code i get when trying to load mp3 files for playing custom songs in Project Diva f 2nd's edit mode.
Would probably get the same thing in Project Diva f.

PR #3204, opened 2024-02-03
io: implement music0, photo0, video0 devices

bookmist:
It should fix https://github.com/Vita3K/Vita3K/issues/3185


Macdu:
The music folder is not created here https://github.com/Vita3K/Vita3K/blob/e6ecae35f0aad08c905c67e725c90cacbe127a5b/vita3k/io/src/io.cpp#L107

bookmist:
> The music folder is not created here
> 
> https://github.com/Vita3K/Vita3K/blob/e6ecae35f0aad08c905c67e725c90cacbe127a5b/vita3k/io/src/io.cpp#L107

fixed

PR #3205, opened 2024-02-04
build: fix openssl pre-build path.

Zangetsu38:
# About
- build: fix openssl pre-build path.
- lang: fix typo of audio settings.


Macdu:
These 2 modifications really have nothing to do with each other but anyway...

Macdu (approved):

PR #3208, opened 2024-02-06
SceCommonDialog: Add common_dialog.savedata.display_type == 0

nishinji:
Tried fix to prevent this strange dialog from appearing.(This dialog don't show up on vita)
image
I don't know if the change is correct, but the dialog will no longer appear.
The dialog is only displayed when creating save data.


Macdu:
That does not look like the proper fix, I would need to take a proper look at this.

PR #3210, opened 2024-02-07
Fix texture export for gamma textures

dima-xd:
Probably fixes most crashes because of texture export in most games.  At least it fixes texture export in Gravity Rush.
This thing was going OOB.


Macdu:
Indeed, nice catch, my bad.

Macdu (approved):

PR #3211, opened 2024-02-08
vita3k: some minor bugfixes

bookmist:
kernel/thread: fix possible crash in log_stack_traceback
modules/SceLibKernel: more correct sceClibSnprintf
kernel/load_self: fix possible crash on incorrect library modules. load process params from first loaded module (eboot.bin) only
modules/SceGxm: fix crash if display_thread not found
kernel/debugger: remove direct access to mem.memory to support memory mapping more complete
vita3k: remove direct access to kernel.threads (it's not thread-safe)
vita3k: implement and fix virtual destructors for class hierarchies to prevent memory leaks
clang-tidy warning: Constructing string from nullptr is undefined behaviour
clang warning: use after std::move

### DIFF
--- a/vita3k/kernel/src/thread.cpp
+++ b/vita3k/kernel/src/thread.cpp
@@ -415,10 +415,12 @@ std::string ThreadState::log_stack_traceback() const {
     std::stringstream ss;
     const Address sp = read_sp(*cpu);
     for (Address addr = sp - START_OFFSET; addr <= sp + END_OFFSET; addr += 4) {
-        const Address value = *Ptr<uint32_t>(addr).get(mem);
-        const auto mod = kernel.find_module_by_addr(value);
-        if (mod)
-            ss << fmt::format("{} (module: {})\n", log_hex(value), mod->module_name);
+        if (Ptr<uint32_t>(addr).valid(mem)) {
+            const Address value = *Ptr<uint32_t>(addr).get(mem);
+            const auto mod = kernel.find_module_by_addr(value);
+            if (mod)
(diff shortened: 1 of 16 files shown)


vita3k/kernel/src/thread.cpp
```diff
@@ -415,10 +415,12 @@ std::string ThreadState::log_stack_traceback() const {
...
-            ss << fmt::format("{} (module: {})\n", log_hex(value), mod->module_name);
+        if (Ptr<uint32_t>(addr).valid(mem)) {
+            const Address value = *Ptr<uint32_t>(addr).get(mem);
+            const auto mod = kernel.find_module_by_addr(value);
+            if (mod)
+                ss << fmt::format("{} (module: {})\n", log_hex(value), mod->module_name);
+        }
     }
     return ss.str();
 }
```
Zangetsu38, line 426: fix New line here 🙏

bookmist: done

Macdu:
Actually there is an issue with the screen filter, if the destructor gets called the emulator crashes, this needs to be fixed before adding the screen filter destructors (although I don't like it right now a small memory leak is better than the emulator crashing).

bookmist:
> Actually there is an issue with the screen filter, if the destructor gets called the emulator crashes, this needs to be fixed before adding the screen filter destructors (although I don't like it right now a small memory leak is better than the emulator crashing).

You are right. I did not check this moment properly. 
Reverted.

Macdu (approved):

PR #3219, opened 2024-02-17
modules/SceRtcUser: No need to subtract since the time is already from the UNIX epoch

nishinji:
master:
image
pr:
image


Zangetsu38:
back to the futur 😎

Macdu:
https://github.com/Vita3K/Vita3K/blob/1b0e65e66d17202b1ae9f155cfecea802dd58778/vita3k/rtc/src/rtc.cpp#L106

RTC_OFFSET is added in __RtcPspTimeToTicks, not removing it in the sce function would definitely cause an overflow.

nishinji:
image
This bug is already caused by an overflow.
After looking at the PPSSPP implementation, it seems that the function itself is not at fault, so it is closed

PR #3223, opened 2024-02-22
renderer/vulkan: Fix MSAA2x

Macdu:
When using MSAA, tiles size is actually 32x16 and not 16x32.
This fixes the blur in Nihon Falcom games.


bookmist (approved):

niekvlessert:
Memories of Celceta still has the blur, i.e. I see everything in the back bigger and mostly transparent at the front.

PR #3226, opened 2024-02-23
vita3k: big code cleanup

bookmist:
Improvements
* lang: a bit more verbose error on wrong lang xml

Bugfixes
* kernel/sync_primitives: fix theoretically possible AV
* modules/SceAudio: a bit more correct sceAudioOutSetConfig
* fix wrong objects initialisation ({0} -> {})

Warning fixes
* clang warning: Macro argument should be enclosed in parentheses [bugprone-macro-parentheses]
* modules: set unsigned type for Sce*ErrorCode enums
* clang warning: variable might be not be initialized
* clang warning: Function returns by const value. Consider returning by non-const value instead.
* clang warning: Prefer transparent functors [modernize-use-transparent-functors]

Refactoring
* io/device: move functions implementation from .h to .cpp file
* vita3k: fix fs::path usage
* clang warning: 'if' statement is unnecessary; deleting null pointer has no effect
* vita3k: use structured bindings and references in for-each-loops
* vita3k: simplify container operations, extract frequent operations to utility functions
* clang:warning: 'contains' member function can be used
* kernel/sync_primitives: refactoring. drop unused parameter of handle_timeout
* vita3k: drop unused function parameters
* vita3k: fix typecasts 
use static_cast instead of reinterpret casts if possible use correct constants (1/1.0/1.0f)
correct variable types to minimise typecasts
use explicit typecasts (not everywhere)
replace c-style typecasts to c++ style typecasts (not everywhere)
clang warning: remove redundant typecasts
clang warning: Cast drops const qualifier

Optimisation
* vita3k: remove unused includes
* clang warning: Constness of returning variable prevents automatic move [performance-no-automatic-move]
* clang warning: String concatenation results in allocation of unnecessary temporary strings [performance-inefficient-string-concatenation]
* vita3k: use std::move
* vita3k: optimization. push_back -> emplace_back
* vita3k: fix datatypes to avoid unnesessary conversions
* clang warning: remove unneeded string->c_str->string conversion
* clang warning: The parameter is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param]
* clang warning: 'find' called with a string literal consisting of a single character [performance-faster-string-find]
* clang warning: Result of a postfix operator is discarded, consider replacing it with a prefix operator
* clang warning: Pass value parameters by const reference

Code cleanup
* io/io: remove once used variables
* net/posixsocket: remove struct keyword
* vita3k: use raw string literals to avoid excessive escaping
* vita3k: refactoring. replace Ptr<> to pointer in Sce functions
* vita3k: drop unused variables and incorrect comments
* clang warning: readability-inconsistent-declaration-parameter-name
* clang warning: Extra ';' after member function definition
* clang warning: Possible misuse of comma operator (, is used instead of ;)
* clang warning: Algorithm accepting an iterator and a count can be used
* clang warning: Declaration and assignment can be joined
* clang warning: Type trait can be simplified using a template alias or a variable template
* vita3k: cleanup class methods 
remove unimplemented (and obviously not used) methods 
move deleted methods to public
change empty body to =default
remove auto constructors
set member default values

Same changes in small, atomic commits: https://github.com/bookmist/Vita3K/tree/misc_04


vita3k/gui/src/vita3k_update.cpp
```diff
@@ -101,7 +101,7 @@ bool init_vita3k_update(GuiState &gui) {
                     boost::replace_all(commits, "\\\"", "&quot;");
 
                     // Using regex to get sha, author and message from commits
-                    const std::regex commit_regex("\"sha\":\"([a-f0-9]{40})\".*?\"author\":\\{\"name\":\"([^\"]+)\".*?\"message\":\"([^\"]+)\"");
+                    const std::regex commit_regex(R"lit("sha":"([a-f0-9]{40})".*?"author":\{"name":"([^"]+)".*?"message":"([^"]+)")lit");
```
Zangetsu38, line 104: hmm you have checked this ?  i not know this lit 🤔

vita3k/mem/include/mem/block.h
```diff
@@ -82,9 +85,6 @@ class Block {
     }
 
 private:
-    Block(const Block &) = delete;
-    const Block &operator=(const Block &) = delete;
-
     Address addr;
     Deleter deleter;
 };
```
nishinji, line 90: Please add a new line here

vita3k/gui/src/vita3k_update.cpp
bookmist: "lit" here is just random set of chаracters, which is not occurs inside the string.
https://en.cppreference.com/w/cpp/language/string_literal  Raw string literals

vita3k/mem/include/mem/block.h
bookmist: done

vita3k/touch/src/touch.cpp
```diff
@@ -282,13 +282,11 @@ int touch_get(const SceUID thread_id, EmuEnvState &emuenv, const SceUInt32 &port
...
             nb_returned_data = 0;
     } else {
-        uint64_t vblank_count = emuenv.display.vblank_count;
-        if (vblank_count <= last_vcount[port_idx]) {
+        if (emuenv.display.vblank_count <= last_vcount[port_idx]) {
             // sceTouchRead is blocking, wait for the next vsync for the buffer to be updated
             auto thread = emuenv.kernel.get_thread(thread_id);
 
             wait_vblank(emuenv.display, emuenv.kernel, thread, last_vcount[port_idx] + 1, false);
-            vblank_count = emuenv.display.vblank_count;
```
Macdu, line 291: same as in ctrl

vita3k/gui/src/user_management.cpp
```diff
@@ -690,7 +687,7 @@ void draw_user_management(GuiState &gui, EmuEnvState &emuenv) {
         ImGui::SetCursorPos(INPUT_NAME_POS);
         ImGui::PushItemWidth(INPUT_NAME_SIZE);
         if (ImGui::InputText("##user_name", temp.name.data(), SCE_NP_ONLINEID_MAX_LENGTH))
-            temp.name = temp.name.data();
+            temp.name = temp.name.data(); // Hack
```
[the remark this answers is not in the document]
Zangetsu38: yes, is not hack here 🤔 
i have juste found this way for can set this name set in data() to inside name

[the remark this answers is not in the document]
bookmist: It's a working hack due to small string optimization, (when length of string buffer (string length+1 for trailing zero) is no more then 16 bytes, then std::string store string data inside itself and don't use heap memory)
But you are right, due to memory buffer here should be 17 bytes (SCE_NP_ONLINEID_MAX_LENGTH + 1 for trailing zero)

vita3k/modules/SceGxm/SceGxm.cpp
```diff
@@ -4173,6 +4170,8 @@ EXPORT(void, sceGxmSetViewport, SceGxmContext *context, float xOffset, float xSc
     // Set viewport to enable, enable more offset and scale to set
     if (context->state.viewport.offset.x != xOffset || (context->state.viewport.offset.y != yOffset) || (context->state.viewport.offset.z != zOffset)
         || (context->state.viewport.scale.x != xScale) || (context->state.viewport.scale.y != yScale) || (context->state.viewport.scale.z != zScale)) {
+        if (xOffset == 0 && xScale == 0 && yOffset == 0 && yScale == 0 && zOffset == 0 && zScale == 0)
+            return;
```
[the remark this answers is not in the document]
bookmist: Thanks for catch. It's a hack for some not working game [PCSF00216] I will delete it.

bookmist:
@Macdu I fixed problems you found in new commit, to make it easier for review. I'll squash it before merge. (And yes, now this PR is  ready to review)

vita3k/gui/src/user_management.cpp
```diff
@@ -686,8 +686,9 @@ void draw_user_management(GuiState &gui, EmuEnvState &emuenv) {
         ImGui::TextColored(GUI_COLOR_TEXT, "%s", lang["name"].c_str());
         ImGui::SetCursorPos(INPUT_NAME_POS);
         ImGui::PushItemWidth(INPUT_NAME_SIZE);
-        if (ImGui::InputText("##user_name", temp.name.data(), SCE_NP_ONLINEID_MAX_LENGTH))
-            temp.name = temp.name.data(); // Hack
+        char user_name_buf[SCE_NP_ONLINEID_MAX_LENGTH + 1]{};
```
Zangetsu38, line 689: Should not works att all here...
This code request global value, look more code
If char is set from nothing any time nothing works, my code works fine

Zangetsu38: I have tested, and like exepted, no works att all 😕
Just create char like that is wrong
Temp.name is already temp value and is set in one function, for set default name, or current in edit
Like i said, is more’complicated you thinks, is better noctouch it, code perfectly works here

vita3k/gui/src/user_management.cpp
```diff
@@ -689,8 +686,9 @@ void draw_user_management(GuiState &gui, EmuEnvState &emuenv) {
         ImGui::TextColored(GUI_COLOR_TEXT, "%s", lang["name"].c_str());
         ImGui::SetCursorPos(INPUT_NAME_POS);
         ImGui::PushItemWidth(INPUT_NAME_SIZE);
-        if (ImGui::InputText("##user_name", temp.name.data(), SCE_NP_ONLINEID_MAX_LENGTH))
-            temp.name = temp.name.data();
+        char user_name_buf[SCE_NP_ONLINEID_MAX_LENGTH + 1]{};
+        if (ImGui::InputText("##user_name", user_name_buf, sizeof(user_name_buf)))
+            temp.name = user_name_buf;
```
Zangetsu38, line 691:
```suggestion
        char user_name_buf[SCE_NP_ONLINEID_MAX_LENGTH + 1]{};
        strncpy(user_name_buf, temp.name.c_str(), SCE_NP_ONLINEID_MAX_LENGTH);
        if (ImGui::InputText("##user_name", user_name_buf, SCE_NP_ONLINEID_MAX_LENGTH))
            temp.name = user_name_buf;
```
iIf you really want using your code, change your code not worksq by this, tested and works fine 👌

vita3k/gui/src/user_management.cpp
```diff
@@ -687,7 +687,7 @@ void draw_user_management(GuiState &gui, EmuEnvState &emuenv) {
         ImGui::SetCursorPos(INPUT_NAME_POS);
         ImGui::PushItemWidth(INPUT_NAME_SIZE);
         if (ImGui::InputText("##user_name", temp.name.data(), SCE_NP_ONLINEID_MAX_LENGTH))
-            temp.name = temp.name.data(); // Hack
+            temp.name = temp.name.data();
```
Macdu, line 690: At least add a comment saying this works because of small string optimisations.

Macdu (approved):
lgtm, a lot of small nice improvements

PR #3227, opened 2024-02-24
external: Import CppCommon to detect CPU and OS

nishinji:
I implemented this based on information from bookmist.
It works on my PC. But Android and Arm are not supported.
Macdu said that is not very useful, but it is useful for state_report (that's why I'm implementing this).

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -63,6 +63,9 @@ target_include_directories(stb INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}/stb")
 add_library(ddspp INTERFACE)
 target_include_directories(ddspp INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}/ddspp")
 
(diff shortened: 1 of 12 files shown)


vita3k/util/src/cpu_detect.cpp
```diff
@@ -0,0 +1,234 @@
...
+        return "<unknown>";
+
+    // Remove trailing 0x20 space from pBuffer
+    while (!std::isgraph(pBuffer[dwBufferSize - 1]) && dwBufferSize > 0) {
+        pBuffer[--dwBufferSize] = '\0';
+    }
+
+    return std::string(pBuffer);
+#else
+#error Unsupported platform
```
bookmist, line 91: I'm not sure if android is linux, but maybe android also should be supported. At least don't crash on android.

Macdu: Android is indeed linux (and both __linux and __linux__ are defined) so it's fine

Macdu (approved):

PR #3234, opened 2024-03-03
gui/live area: Fix transparency of Info bar.

Zangetsu38:
- gui/live area: Fix transparency of Info bar.
small reworks for separe window to info bar.


Macdu (approved):
nice

PR #3235, opened 2024-03-03
app, config, ctrl, motion: CemuHookUDP support (SteamDeckGyroDSU support on Steam Deck)

popsUlfr:
Hi,

This pull requests adds basic CemuHookUDP protocol support to enable gyro input and is supposed to be used in conjunction with https://github.com/kmicki/SteamDeckGyroDSU on the Steam Deck.
The source files `udp_client.h`, `udp_client.cpp`, `udp_protocol.h`, `udp_protocol.cpp` were taken from the yuzu project and modified and simplified to better integrate with the Vita3K codebase.

Here's a video of the working gyro in Gravity Rush, I raised the default sensitivity a bit since filming this:

https://photos.app.goo.gl/iSKQyuFCZ9CYM9h1A

There are two new configuration options:

* `cemuhookudp-address`: expects an address in the form of `127.0.0.1:26760`
* `cemuhookudp-gyro-scale-divider`: modifies the gyro sensitivity. The calculation is done like this: `gyro.x/y/z * (1.0 / cemuhookudp-gyro-scale-divider)`. So the higher the value the lower the sensitivity and the lower the. value the higher the sensitivity. The default is `39`. A value of 1 maps gyro input 1:1.

They are not configurable through the interface yet but exist in the configuration file.

So in effect to get Vita3K with working gyro on the Steam Deck:

1. Install **SteamDeckGyroDSU**: https://github.com/kmicki/SteamDeckGyroDSU (EmuDeck has an option to install it in its interface if you are already using it)
2. Apply this pull request and compile Vita3K.
3. In Steam Input set `Touchscreen Native Support` to always on for multi-touch support.

In the controller interface if the SteamDeckGyroDSU server has replied to the CemuHookUDP client, motion support will be marked as such:

Screenshot_20240303_123937


popsUlfr:
I hopefully fixed the CI building for windows and macos builds.

vita3k/ctrl/src/udp_client.cpp
```diff
@@ -0,0 +1,315 @@
...
+    if (!last_pad_data)
+        return;
+    LOG_TRACE("UDP PadData: gyro.pitch={}, gyro.roll={}, gyro.yaw={}, accel.x={}, accel.y={}, accel.z={}, motion_timestamp={}",
+        last_pad_data->gyro.pitch,
+        last_pad_data->gyro.roll,
+        last_pad_data->gyro.yaw,
+        last_pad_data->accel.x,
+        last_pad_data->accel.y,
+        last_pad_data->accel.z,
+        last_pad_data->motion_timestamp);
```
Macdu, line 302: Don't log something that will happen multiple times every second

Macdu (commented):
Thank you very much for this PR, I believe if/when we migrate to SDL3 it won't be needed anymore.

popsUlfr:
> Thank you very much for this PR, I believe if/when we migrate to SDL3 it won't be needed anymore.

Ah, so SDL3 can expose the Steam Deck's gyro directly ? Looking at the discussion here https://github.com/libsdl-org/SDL/issues/9148 it seems that one would still need to manually query the gyro values.

vita3k/ctrl/src/udp_client.cpp
popsUlfr: Even if it's a `LOG_TRACE` ? I thought it was supposed to be for this kind of logging since it would only show on the lowest log level. If that's the case then the `LOG_TRACE("PadData packet received");` from the `OnPadData()` method is actually the one that will happen even more times than this one when looking at my logs.
Should I remove both ?

nn9dev:
> Thank you very much for this PR, I believe if/when we migrate to SDL3 it won't be needed anymore.

DSU is helpful for a lot of non-SDL gyro stuff, not just Steam Deck. Wiimotes, JoyCons, and I believe even some VR Controllers. It's definitely very nice to have as an option. I'm not sure if SDL3 plans to support any or all of these in a verbose manner, but again, very nice to have.

bookmist:
If now more boost modules are used, then this modules should be added into `BOOST_MODULES_TO_FIND` in main CMakeLists.txt

popsUlfr:
> If now more boost modules are used, then this modules should be added into `BOOST_MODULES_TO_FIND` in main CMakeLists.txt It used for windows compilation without vcpkg

Ok, thank you, I'll add them there.

vita3k/ctrl/src/udp_client.cpp
Macdu: Yes please, even if it's the trace level, if a log is called more than once every second it will just bloat the log file and hide other useful logs.

popsUlfr: Understood, thank you. I'll remove them.

Macdu:
> > If now more boost modules are used, then this modules should be added into `BOOST_MODULES_TO_FIND` in main CMakeLists.txt It used for windows compilation without vcpkg
> 
> Ok, thank you, I'll add them there.
> 
> EDIT: I think `asio` and `crc` are already part of the `system` cmake boost component but please correct me if I'm wrong. @bookmist

Cmake is missing some boost components. That's not really an issue (if you don't have the asio or crc component you'll always get an error at some point). However, the fallback repository https://github.com/Vita3K/ext-boost does not have boost crc. It needs to be updated before this PR is merged or Vita3K will break for people not having an existing boost library availble (@Zangetsu38 )

popsUlfr:
@Macdu Should I try to submit a pull request to the https://github.com/Vita3K/ext-boost repo to get the crc module in ?

Macdu:
Yes you can (except if @Zangetsu38 wants to do it first), that's the only thing preventing this PR from being merged.

Zangetsu38:
oh, i have not see sory, i can yes :ok:

scribam:
@popsUlfr FYI, the CRC module is now available with ext-boost and Vita3K

nn9dev:
@Macdu / @Zangetsu38 can we get this merged finally? I don't know if it needs any rebasing but I checked out and built this PR and it works terrific.

Off-topic, but it's probably the best way to play Gravity Rush to-date.

Zangetsu38:
Maybe need rebase before

dima-xd:
image
image

german77:
I rebased this PR to latest. I will appreciate if you can do it as well so this PR can move forward. https://github.com/german77/Vita3K/tree/cemuhookrebase

vita3k/motion/src/motion.cpp
```diff
@@ -99,8 +99,12 @@ void refresh_motion(MotionState &state, CtrlState &ctrl_state) {
...
     }
 
-    if (!found_accel && !found_gyro)
-        return;
+    if (!found_accel && !found_gyro) {
+        if (ctrl_state.udp_client.IsConnected())
+            ctrl_state.udp_client.GetGyroAccel(gyro, gyro_timestamp, accel, accel_timestamp);
+        else
+            return;
+    }
```
german77, line 107:
```suggestion
    if (!found_accel && !found_gyro) {
-        if (ctrl_state.udp_client.IsConnected())
-            ctrl_state.udp_client.GetGyroAccel(gyro, gyro_timestamp, accel, accel_timestamp);
-        else
+        if (!ctrl_state.udp_client.IsConnected()) {
             return;
+        }
+        ctrl_state.udp_client.GetGyroAccel(gyro, gyro_timestamp, accel, accel_timestamp);
+        // udp client already has the correct magnitude.
+        accel *= -SDL_STANDARD_GRAVITY;
     }
```

Cemu hook is based on dsu. It already provides gyro and accelerometer in the correct magnitude. The motion code divides the accelerometer by gravity. So we need to do the inverse here to keep the values the same.

EXtremeExploit:
With the merge of https://github.com/Vita3K/Vita3K/pull/4100, i dont think this is needed anymore as gyro/motion controls work natively without a server or interpreter of some kind.

Feel free to reopen if thats not the case tho

PR #3236, opened 2024-03-04
vita3k: Fix some arm64 running issues

FantasyGmm:
```
Vita3K: /media/fan/odin2/soft/Vita3K/external/VulkanMemoryAllocator-Hpp/VulkanMemoryAllocator/include/vk_mem_alloc.h:14965: VkResult VmaAllocator_T::CalcAllocationParams(VmaAllocationCreateInfo &, bool, bool): Assertion `(inoutCreateInfo.flags & (VMA_ALLOCATION_CREATE_HOST_ACCESS_SEQUENTIAL_WRITE_BIT | VMA_ALLOCATION_CREATE_HOST_ACCESS_RANDOM_BIT)) != 0 && "When using VMA_ALLOCATION_CREATE_MAPPED_BIT and usage = VMA_MEMORY_USAGE_AUTO*, you must also specify VMA_ALLOCATION_CREATE_HOST_ACCESS_SEQUENTIAL_WRITE_BIT or VMA_ALLOCATION_CREATE_HOST_ACCESS_RANDOM_BIT."' failed.
```
When I tried running it on arm64 Linux, I encountered this error. The game crashes as soon as I start it. However, after I made some modifications to the memory allocation part, I was able to successfully enter the game.

### DIFF
--- a/vita3k/renderer/src/vulkan/renderer.cpp
+++ b/vita3k/renderer/src/vulkan/renderer.cpp
@@ -806,7 +806,7 @@ bool VKState::map_memory(MemState &mem, Ptr<void> address, uint32_t size) {
         // also make sure later the mapped address is 4K aligned
         vkutil::Buffer buffer(size + KiB(4));
         constexpr vma::AllocationCreateInfo memory_mapped_alloc = {
-            .flags = vma::AllocationCreateFlagBits::eMapped,
+            .flags = vma::AllocationCreateFlagBits::eHostAccessSequentialWrite | vma::AllocationCreateFlagBits::eMapped,
             .usage = vma::MemoryUsage::eAutoPreferHost,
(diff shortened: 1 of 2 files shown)


Macdu:
Indeed thank you, although I would advise you for performance reasons not to make a debug build (these crashes are not linked to arm64 but instead to some asserts in vma which appear only in debug builds)

Macdu (approved):

FantasyGmm:
> Indeed thank you, although I would advise you for performance reasons not to make a debug build (these crashes are not linked to arm64 but instead to some asserts in vma which appear only in debug builds)

Yes, I did debug the build because I couldn't start Vita3k on the built-in screen, so I did some debugging. However, when I used the external screen, it started with these errors, so I think it may be due to these reasons

PR #3238, opened 2024-03-05
vkutil: Fix regression

Macdu:
Fix a regression causing the vulkan renderer to crash from #3236 .
Sorry, this one is entirely my bad.


nishinji:
Certainly fixed.

bookmist (approved):

PR #3247, opened 2024-03-12
shader: fixed misuse of spv::Builder::createOp with spv::OpVectorShuffle

blackbird806:
The spv::Builder::createOp function seems to only take spv::Builder generated id rather than arbitrary values, and the spv::Builder::addIdOperand throw an assert when receiving 0 as an id. So i replaced all c++ id literals by makeIntConstant, I'm really not sure if this is what should be done here but at least it prevent the assert.

### DIFF
--- a/vita3k/shader/src/spirv_recompiler.cpp
+++ b/vita3k/shader/src/spirv_recompiler.cpp
@@ -153,7 +153,7 @@ static spv::Id create_array_if_needed(spv::Builder &b, const spv::Id param_id, c
 
 static spv::Id get_type_basic(spv::Builder &b, const Input &input) {
     switch (input.type) {
-    // clang-format off
+        // clang-format off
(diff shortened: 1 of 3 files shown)


Macdu:
Thank you for this PR. However, its content is wrong, if you were to try it with a game, you would almost definitely get a crash. As written in the specification, OpVectorShuffly expects 32-bit integers for the components and not Ids:
image
If you are making a debug build, you can comment this assert (the generated spirv is right). The correct way to solve it would be to pass a vector of IdImmediate to createOp instead of a Vector of Id.

blackbird806:
Thanks for your feedbacks, I'm not familiar at all with all of this ^^' I updated my pr to make the changement you recomended.

Macdu (approved):
lgtm, thank you for your contribution

PR #3248, opened 2024-03-12
app/gui: support HiDPI on macOS

KhoraLee:
This PR will enable HiDPI on macOS. 

### **Screenshots**
| Latest Vita3K | This PR |
| ------------- | ------------- |
| Screenshot 2024-03-13 at 4 15 28 AM | Screenshot 2024-03-13 at 4 15 24 AM |
| Screenshot 2024-03-13 at 4 15 49 AM | Screenshot 2024-03-13 at 4 15 45 AM |
|  | Screenshot 2024-03-13 at 4 19 31 AM |

### DIFF
--- a/vita3k/app/src/app_init.cpp
+++ b/vita3k/app/src/app_init.cpp
@@ -319,10 +319,18 @@ bool init(EmuEnvState &state, Config &cfg, const Root &root_paths) {
         window_type |= SDL_WINDOW_ALLOW_HIGHDPI;
         state.dpi_scale = ddpi / 96;
     }
+#elif defined(__APPLE__)
+    window_type |= SDL_WINDOW_ALLOW_HIGHDPI;
+    state.dpi_scale = 2; // It is fixed value on macOS.
 #endif
     state.res_width_dpi_scale = static_cast<uint32_t>(DEFAULT_RES_WIDTH * state.dpi_scale);
     state.res_height_dpi_scale = static_cast<uint32_t>(DEFAULT_RES_HEIGHT * state.dpi_scale);
+
+#ifndef __APPLE__
     state.window = WindowPtr(SDL_CreateWindow(window_title, SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED, state.res_width_dpi_scale, state.res_height_dpi_scale, window_type | SDL_WINDOW_RESIZABLE), SDL_DestroyWindow);
+#else
+    state.window = WindowPtr(SDL_CreateWindow(window_title, SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED, DEFAULT_RES_WIDTH, DEFAULT_RES_HEIGHT, window_type | SDL_WINDOW_RESIZABLE), SDL_DestroyWindow);
+#endif
 
     if (!state.window) {
         LOG_ERROR("SDL failed to create window!");
--- a/vita3k/gui/src/imgui_impl_sdl.cpp
+++ b/vita3k/gui/src/imgui_impl_sdl.cpp
@@ -120,6 +120,13 @@ IMGUI_API void ImGui_ImplSdl_NewFrame(ImGui_State *state) {
     int display_w, display_h;
     SDL_GetWindowSize(state->window, &w, &h);
     ImGui_ImplSdl_GetDrawableSize(state, display_w, display_h);
+#if __APPLE__
+    // On macOS, if HiDPI is enabled(SDL_WINDOW_ALLOW_HIGHDPI) SDL will create a window that size is 960x544 points.
+    // But it is actually 1920x1088 pixels. Since SDL_GetWindowSize returns 960x544 points, we need to multiply by 2.
+    w *= 2;
+    h *= 2;
+#endif
+
     io.DisplaySize = ImVec2((float)w, (float)h);
     io.DisplayFramebufferScale = ImVec2(w > 0 ? ((float)display_w / w) : 0, h > 0 ? ((float)display_h / h) : 0);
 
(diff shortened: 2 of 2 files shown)


vita3k/app/src/app_init.cpp
```diff
@@ -319,10 +319,18 @@ bool init(EmuEnvState &state, Config &cfg, const Root &root_paths) {
         window_type |= SDL_WINDOW_ALLOW_HIGHDPI;
         state.dpi_scale = ddpi / 96;
     }
+#elif defined(__APPLE__)
+    window_type |= SDL_WINDOW_ALLOW_HIGHDPI;
+    state.dpi_scale = 2; // It is fixed value on macOS.
```
Macdu, line 324: Does using the windows/linux approach (calling SDL_GetDisplayDPI then dividing it by 96) not return 2?

vita3k/app/src/app_init.cpp
```diff
@@ -319,10 +319,18 @@ bool init(EmuEnvState &state, Config &cfg, const Root &root_paths) {
...
 #endif
     state.res_width_dpi_scale = static_cast<uint32_t>(DEFAULT_RES_WIDTH * state.dpi_scale);
     state.res_height_dpi_scale = static_cast<uint32_t>(DEFAULT_RES_HEIGHT * state.dpi_scale);
+
+#ifndef __APPLE__
     state.window = WindowPtr(SDL_CreateWindow(window_title, SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED, state.res_width_dpi_scale, state.res_height_dpi_scale, window_type | SDL_WINDOW_RESIZABLE), SDL_DestroyWindow);
+#else
+    state.window = WindowPtr(SDL_CreateWindow(window_title, SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED, DEFAULT_RES_WIDTH, DEFAULT_RES_HEIGHT, window_type | SDL_WINDOW_RESIZABLE), SDL_DestroyWindow);
+#endif
```
Macdu, line 333: I think calling SDL_CreateWindow with  DEFAULT_RES_WIDTH and  DEFAULT_RES_HEIGHT should be the correct behavior on all platforms

vita3k/gui/src/imgui_impl_sdl.cpp
```diff
@@ -120,6 +120,13 @@ IMGUI_API void ImGui_ImplSdl_NewFrame(ImGui_State *state) {
     int display_w, display_h;
     SDL_GetWindowSize(state->window, &w, &h);
     ImGui_ImplSdl_GetDrawableSize(state, display_w, display_h);
+#if __APPLE__
+    // On macOS, if HiDPI is enabled(SDL_WINDOW_ALLOW_HIGHDPI) SDL will create a window that size is 960x544 points.
+    // But it is actually 1920x1088 pixels. Since SDL_GetWindowSize returns 960x544 points, we need to multiply by 2.
+    w *= 2;
+    h *= 2;
+#endif
+
```
Macdu, line 129: SDL_GetWindowSize returns the screen dimensions in coordinates while ImGui_ImplSdl_GetDrawableSize returns the same screen dimension in pixel, this is done because of hidpi scaling. What you are doing is making ImGui believe the dpi scaling is 1.0 which sounds kind of counter-intuitive (and also forces you to implement workarounds like the one below for SDL_GetMouseState)

KhoraLee: But from my windows pc with 150% scale, EmuEnvState's dpi_scale is 1.5 but it's w and display_w is same(960) and h and display is also same(544) so ImGui's DisplayFramebufferScale is set to (1, 1). 

Also as screenshot shows it was rendering at real pixel. So I was thinking HiDPI is correctly working.

vita3k/app/src/app_init.cpp
KhoraLee: macOS's SDL_GetDisplayDPI returns display's real ppi (in my case around 255: 3024x1964 14.2").

Macdu:
Hmm, I'm not entirely sure, but looking at SDL high_dpi stuff. We might actually not care at all about the SDL_WINDOW_ALLOW_HIGHDPI flag. What we could instead is not use this flag and just scale everything using the dpi scale. The result would be the same but not have different behavior on mac and other devices.

What setting SDL_WINDOW_ALLOW_HIGHDPI does is make screen coordinates bigger than screen pixels, but we really don't care about that as we use our own renderer and everything. What we care about is just making everything big/sharp enough which we should be able to do only using the dpi_scale variable.

KhoraLee:
It's been a while from when I first proposed this PR. 

From my understanding, `SDL_WINDOW_ALLOW_HIGHDPI` flag is essential for enabling proper HiDPI support. Without it, regardless of the resolution we render at, SDL will always create a 960x544 window, and macOS will simply scale it up by 2x. This will results in the same state as now—without HiDPI support.

vita3k/app/src/app_init.cpp
saturnsky: This part seems to need modification. On Mac, external monitors can be connected in regular mode instead of HiDPI mode, and in that case, dpi_scale should be 1.

On displays connected in HiDPI mode, drawableWidth and drawableHeight will be twice DEFAULT_RES_WIDTH/HEIGHT, but on normal displays, drawableWidth and drawableHeight will be equal to DEFAULT_RES_WIDTH/HEIGHT.

Mac's dpi_scale should be indirectly obtained through the ratio of these values.

KhoraLee: I made a new PR with fixes because of my mistake.
Now it calculates the scale when the display changes, so it works well in environments with both LoDPI and HiDPI screens.

PR #3252, opened 2024-03-19
shader: replaced std map by std array to store shader code in convert_gxp_usse_to_spirv

blackbird806:
The std map seems unnecessary here and can safely be replaced by std array. Doing so will avoid unnecessary dynamic allocations and more expensive iterations/lookups.

### DIFF
--- a/vita3k/shader/include/shader/usse_types.h
+++ b/vita3k/shader/include/shader/usse_types.h
@@ -436,7 +436,7 @@ struct ProgramInput {
 };
 
 enum class ShaderPhase {
-    SampleRate, // Secondary phase
+    SampleRate = 0, // Secondary phase
     Pixel, // Primary phase
 
     Max,
--- a/vita3k/shader/include/shader/usse_types.h
+++ b/vita3k/shader/include/shader/usse_types.h
@@ -436,7 +436,7 @@ struct ProgramInput {
 };
 
 enum class ShaderPhase {
-    SampleRate = 0, // Secondary phase
+    SampleRate, // Secondary phase
     Pixel, // Primary phase
 
     Max,
--- a/vita3k/shader/src/usse_translator_entry.cpp
+++ b/vita3k/shader/src/usse_translator_entry.cpp
@@ -25,7 +25,6 @@
 #include <shader/usse_translator_types.h>
 #include <util/log.h>
 
-#include <map>
 #include <optional>
 
 namespace shader::usse {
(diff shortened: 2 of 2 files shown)


vita3k/shader/include/shader/usse_types.h
```diff
@@ -436,7 +436,7 @@ struct ProgramInput {
 };
 
 enum class ShaderPhase {
-    SampleRate, // Secondary phase
+    SampleRate = 0, // Secondary phase
```
Macdu, line 439:
```suggestion
    SampleRate, // Secondary phase
```
No need for the = 0, enum fields start by default at 0. Usually one sets a specific value for an enum field when it's not the expected one.

Macdu (commented):
Indeed, using a map here is completely unnecessary (although talking about unnecessary stuff, I would advise you not to look to much at the gui code).

vita3k/shader/include/shader/usse_types.h
blackbird806: It's an habit of mine to explicit at lest the first enum value when some code relies on it but ye that's purely stylistic.

vita3k/shader/src/usse_translator_entry.cpp
```diff
@@ -1078,13 +1077,13 @@ void convert_gxp_usse_to_spirv(spv::Builder &b, const SceGxmProgram &program, co
     const uint64_t *secondary_program_start = program.secondary_program_start();
     const uint64_t *secondary_program_end = program.secondary_program_end();
 
-    std::map<ShaderPhase, std::pair<const std::uint64_t *, std::uint64_t>> shader_code;
+    std::array<std::pair<const std::uint64_t *, std::uint64_t>, (size_t)ShaderPhase::Max> shader_code;
```
Macdu, line 1080:
```suggestion
    std::array<std::pair<const std::uint64_t *, std::uint64_t>, (size_t)ShaderPhase::Max> shader_code{};
```
Without this, the content of the array is undefined, which will likely lead to a crash if one of the two phases is not defined.

blackbird806: well the whole array is filled in the code below so this shouldn't matter.

Macdu: Indeed, my bad

Macdu (approved):
lgtm

PR #3254, opened 2024-03-19
shader: Improve signed integer normalization

Macdu:
There are two main ways used by GPU to normalize and un-normalize signed integers, each with its pros and cons.
This is why signed integer normalization is implemented by Vita3K's shader recompiler... in a third way that is both really inefficient performance wise and just wrong.

I corrected this to match the behavior of a real PS Vita GPU. I also fixed the range from the fx10 type (from my testing on Killzone, fx10 fixed range is actually [-2,2])


Macdu:
I usually only update the shader cache version if using old shaders with a newer version would cause the emulator to crash or big graphical glitches, which isn't the case here.

bookmist (approved):

PR #3255, opened 2024-03-21
gui/settings dialog: Fix allow using surface sync on ogl.

Zangetsu38:
# About
- fix one mistake causing imposible enable surface sync in ogl.


Macdu (approved):
indeed, my bad

PR #3259, opened 2024-03-23
.github: push builds to store repository

nishinji:
~~Vita3K can run more games and regressions are not unusual for now, it is better to retain past builds.
Looking at PCSX2, there does not appear to be a capacity limit on releases.
Release looks like this.
image
If you do not want a large number of releases created, i can change code to have a release created when a tag is created.~~

Edited: To avoid makeing dirty the main repository, changed approach to push builds to store repository(https://github.com/Vita3K/Vita3K-builds)
we can also create a release there, maybe need to give push permission for token.

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -234,3 +234,16 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Push to store repository
+        uses: cpina/github-action-push-to-another-repository@main
+        env:
+          SSH_DEPLOY_KEY: ${{ secrets.SSH_DEPLOY_KEY }}
+        with:
+          source-directory: artifacts/
+          destination-github-username: 'Vita3K'
+          destination-repository-name: 'Vita3K-builds'
+          target-branch: master
+          target-directory: Builds/
+          commit-message: "Build: ${{ env.Build_Variable }}"
+          user-name: github-actions[bot]
+          user-email: github-actions[bot]@users.noreply.github.com


Zangetsu38:
little problem, missing this for auto updated in vita3k
image

nishinji:
> little problem, missing this for auto updated in vita3k image

Build number is provided in release title.
I can't understand what you asking.

Zangetsu38:
auto updated get broken with this code
It using description of release for get build number
exactly like is on current, with looking second line and take number after Vita3K Build:

nishinji:
The action I am trying to use does not seem to support the changes of the release notes(and I don't think the current one is user friendly). 🙃
Could you please change the code for the update?
And yes, we need to change the code for bot and website.

Zangetsu38:
is not possible adding description like current ? 🤔 
commit source and etc is very better compare here, more proper

Macdu:
@nishinji @Zangetsu38 is this PR ready to be merged?

nishinji:
I think yes, but i want to prepare the Vita3K-Builds repository(update readme, create release workflow, switch new branch), could you give me write access to it?

nishinji:
OK, I think it's ready.
Let's hope it will work property.

Macdu:
Are you sure about this?
secrets.GITHUB_TOKEN only grants right for the current repository. I'm pretty sure if you want to act on another repository, you need to specify either a SSH key or an API token with more privileges.

nishinji:
> Are you sure about this? secrets.GITHUB_TOKEN only grants right for the current repository. I'm pretty sure if you want to act on another repository, you need to specify either a SSH key or an API token with more privileges.

You are right. I can't do anything about that part, please add SSH key following document (https://cpina.github.io/push-to-another-repository-docs/setup-using-ssh-deploy-keys.html#setup-ssh-deploy-keys).

Macdu (approved):
Ok, I added the ssh key, let's see if it works

PR #3263, opened 2024-03-28
compat: Fixed issue count discrepancy after update.

Zangetsu38:
# About
- Fix issue detected by @nishinji, causing issue count is all time on 0 when update it.


Macdu (approved):
Fine... although the main issue is that `load_app_compat_db` can call `update_app_compat_db` and `update_app_compat_db` can call `load_app_compat_db` which is really confusing and causes this kind of errors.

Issue #81, opened 2018-02-22
Check for unnecessary use of `shared_ptr`


petmac:
As mentioned by @VelocityRa we should prefer to use `unique_ptr` or `weak_ptr`.

alimh1214:
I'm trying to remove unnecessary shared pointers for a class project but it's kind of hard to know if it will screw up the functionality somewhere else.

PR #3275, opened 2024-04-16
removed some unnecessary shared_ptr #81

blackbird806:
For the SDL_Semaphore going back to raw C style is what seemed the most straightforward for me, using unique ptr with custom deleter feels overkill since it's only destroyed at one place.

### DIFF
--- a/vita3k/gui/src/live_area.cpp
+++ b/vita3k/gui/src/live_area.cpp
@@ -820,12 +820,12 @@ void draw_live_area_screen(GuiState &gui, EmuEnvState &emuenv) {
                 std::vector<ImVec4> str_color;
 
                 if (!str_tag.color.empty()) {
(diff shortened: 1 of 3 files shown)


Macdu:
Nice (although I'm not sure whether I would prefer to use a reference instead of a pointer for mz_zip_archive).
For the semaphore stuff, what the code is doing right now is
- Thread 1 allocates ThreadParams on its stack then starts Thread 2
- Thread 2 copies ThreadParams to it stack then notifies Thread 1 using a semaphore
- Thread now can now return, making the inital ThreadParams invalid

This is really overcomplicated for nothing, it would be a lot easier to just do:
- Thread 1 allocates ThreadParams on the heap using new, starts Thread 2 and return
- Thread 2 reads ThreadParams  then delete it using delete
No more need for a semaphore.

PR #3277, opened 2024-04-18
kernel/load_self: fix memory leak

bookmist:
New variable was introduced and immediately go out of scope. Instead of this an existent variable should be used


Macdu (approved):
Indeed, my bad

PR #3281, opened 2024-04-25
build: fix macos build

bookmist:
I'm not sure about making release.


KhoraLee:
Instead of `macos-latest-large `, try `macos-12` or `macos-13`. large runners are pay runner.
Previous runner was macOS 12. So `macos-12` shoud be okay.

This is only temporary action. I will properly fix it on #3282

PR #3282, opened 2024-04-25
ci: respond to github action's macos runner chnage

KhoraLee:
Recently github changed github action's macos-latest runner to arm64 based runner.
So this PR will make runner to create x86_64 binary on arm64 runner.

I had to patch xxHash's CMakeLists.txt because it is getting system's architecture, so it is returning arm64 while it is building x86_64 binary.


Macdu (dismissed):
Thank you for the quick fix!
It's a bit annoying that we have to patch xxhash but it looks more like xxhash's issue for not supporting cross-architecture compilation support with dispatching.

Macdu:
On a side note, we could almost disable the xxhash dispatch mode on Mac. Rosetta does not support AVX anyway, so it's only useful for old x86_64 mac.

KhoraLee:
Disabled xxHash dispatch on macOS.

Macdu (approved):

PR #3287, opened 2024-05-12
compat: Download zipped database

nishinji:
Currently the compatibility database is downloaded uncompressed and is small enough in size, but it is updated frequently.
Therefore reducing the size is a good thing.

It will be decreased from 399KB to 31.6KB


Macdu:
Do not use `path.string().c_str()` (it is not correct all the time), instead use `fs_utils::path_to_utf8(path)`

nishinji:
> Do not use `path.string().c_str()` (it is not correct all the time), instead use `fs_utils::path_to_utf8(path)`

Fixed.

vita3k/compat/src/compat.cpp
```diff
@@ -43,6 +44,42 @@ static std::string db_updated_at;
...
+        mz_zip_archive_file_stat file_stat;
+        if (!mz_zip_reader_file_stat(&zip_archive, i, &file_stat)) {
+            LOG_ERROR("Failed to get file information from ZIP archive");
+            mz_zip_reader_end(&zip_archive);
+            return false;
+        }
+
+        // Extract the file from the ZIP archive
+        fs::path output_file_path = output_path / file_stat.m_filename;
+        if (!mz_zip_reader_extract_to_file(&zip_archive, i, output_file_path.string().c_str(), 0)) {
```
Macdu, line 69: `fs_utils::path_to_utf8` also needs to be used here

nishinji: Oups 🤦‍♂️

Macdu (approved):
Nice work!

PR #3292, opened 2024-05-14
io/modules: Make modules load the case sensitive file found in fs

EXtremeExploit:
Allows Sir Eatsalot to go a bit further on linux, still nothing tho, so its now on par with windows.

Also added vs0 support for finding case sensitive paths

### DIFF
--- a/vita3k/io/src/io.cpp
+++ b/vita3k/io/src/io.cpp
@@ -164,6 +164,12 @@ bool find_case_isens_path(IOState &io, VitaIoDevice &device, const fs::path &tra
         final_path = system_path.string().substr(0, system_path.string().find(addcont_id)) + addcont_id;
         break;
     }
+    case +VitaIoDevice::vs0: {
+        // This only works if ALL the parent folders of the path are the correct case or are in a case insensitive fs
+        // Only the file's name is searched for, not the parent folders
+        final_path = system_path.string().substr(0, system_path.string().find_last_of('/'));
+        break;
+    }
     default: {
         return false;
     }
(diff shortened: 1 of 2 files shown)


vita3k/io/src/io.cpp
```diff
@@ -164,6 +164,10 @@ bool find_case_isens_path(IOState &io, VitaIoDevice &device, const fs::path &tra
         final_path = system_path.string().substr(0, system_path.string().find(addcont_id)) + addcont_id;
         break;
     }
+    case +VitaIoDevice::vs0: {
+        final_path = system_path.string().substr(0, system_path.string().find_last_of('/'));
+        break;
```
Macdu, line 169: This is not entirely accurate, this works if only the file name can be case insensitive (the folder it is in needs to have the correct casing with this approach). This is fine for me, but add a comment specifying it.

EXtremeExploit: I mainly copied what IO had and yeah, its not super great but so far only the filenames have been in the wrong case, ill add the comment on both places when i get home

EXtremeExploit: Sometimes devs put stuff in their own folders inside app0 and still have the wrong case, (`app0:resources/resource.bla`) and "resources" is case sensitive, same for addcont0 where devs decide randomly how to store their DLCs like this mess:
image

Macdu: I don't see what you are saying has to do with my comment. 
If case isensitive paths are enabled, a file app0:folder/file.txt exists and the user asks for app0:fOLder/fILe.txt, it will resolve to the former file correctly.
However, with your approach, if a file vs0:folder/file.txt exists and the user asks for vs0:fOLder/fILe.txt, it won't work (but vs0:folder/fILe.txt will resolve correctly).
Also it looks like you mixed sensitive and insensitive in your comment.

Macdu (approved):

PR #3293, opened 2024-05-15
vita3k/packages: Handle fs errors when installing pkg

ihopeitdoes:
- Fixes possible crash when installing pkg to disk with low storage space.
- I'm aware of **decrypt_fself()** in **sce_utils.cpp** which has a similar problem, which also seems to be very optimistic about it's file operations. I'll get to it eventually.
- Feel free to nitpick, I don't know if my "fix" is really the best solution, I kind of just went off what was already there.

Consideration:
- Update lang "*pkg_install/failed_install_package*" message to mention something along the lines of "check if there is enough storage space as well"


vita3k/packages/src/pkg.cpp
```diff
@@ -271,9 +271,25 @@ bool install_pkg(const fs::path &pkg_path, EmuEnvState &emuenv, std::string &p_z
         LOG_INFO(string_name);
 
         if ((byte_swap(entry.type) & 0xFF) == 4 || (byte_swap(entry.type) & 0xFF) == 18) { // Directory
-            fs::create_directories(path / string_name);
+            if (!fs::create_directories(path / string_name)) {
+                LOG_ERROR("Failed to create directory");
+                evp_cleanup();
+                return false;
```
Macdu, line 277: fs::create_directories will return false if the directory already exists, we may want to be a bit careful with this (for update I wouldn't be surprised if this happens).

ihopeitdoes: Good point. There's an overload of the function that gives an error code that can be checked instead.

Macdu:
While using the non-throwing version, you also changed the call from `fs::create_directories` to `fs::create_directory`. The latter will also return an error if the parent directory does not exist (while the former just creates it). I'm not entirely sure if this can happen, but is there a reason for it?

ihopeitdoes:
> While using the non-throwing version, you also changed the call from `fs::create_directories` to `fs::create_directory`. The latter will also return an error if the parent directory does not exist (while the former just creates it). I'm not entirely sure if this can happen, but is there a reason for it?

Typo, had it fixed locally forgot to push

vita3k/packages/src/pkg.cpp
```diff
@@ -359,4 +388,4 @@ bool install_pkg(const fs::path &pkg_path, EmuEnvState &emuenv, std::string &p_z
     create_license(emuenv, zRIF);
     progress_callback(100);
     return true;
-}
+}
```
nishinji, line 391: ?

Zangetsu38: Endline is removed here

Issue #3286, opened 2024-05-06
[BUG]: Helldivers (Where it crashes and how I got it running....for Liberty!)


nckstwrt:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Affected Build(s)

Latest git (as of 2024-05-05)

### Description of Issue

Helldivers (HELLDIVERS™ [PCSA00134] [USA) crashes straight after the title screens and straight after the screen talking about not turning off your Vita when Helldivers is saving

![](https://i.imgur.com/BqCxZmD.png)

### Expected Behavior

Not to crash and continue :)

It's crashing while doing one particular call in module_parent.cpp - call_import - this one call inside call_import can be ignored. I hack past it at present by doing a:
```
       pc = read_pc(cpu);
        if (pc == 3435973836)
            return;
````
 However after that there is a subsequent crash in scheduler.cpp - VoiceScheduler::update - it crashes just doing the "voice->inputs.reset_inputs();" and seeing as that's just filling vectors with zeros - means there's a corruption happening somewhere (going to continue to look into that). I workaround that by just returning immediately from VoiceScheduler::update

But once you work around these to items - the game seems to begin to function well!
![](https://i.imgur.com/6PyXT33.png)

With the above info - hoping that's motivation for you to fix it properly. I will continue to give it a shot though.

EDIT: Now looking through it in more detail - I think this is all a NGS issue of some sort - while processing the modules in the VoiceSheduler::update seems to be where the problem really lies.

EDIT2: On further investigation - this is seemingly caused by the Atrac9 Decoder

EDIT3: After a lot of digging it's these lines in vita3k/ngs/src/modules/atrac9.cpp that eventually causes the crash:

**EDIT4**: Seems to be caused by this same bug reported here (caused by commit: https://github.com/Vita3K/Vita3K/commit/7d542171c72a0fe5dd17e400e2c61117d91de793):
https://github.com/Vita3K/Vita3K/issues/3270 [BUG]: sceNgs issue that happens in the "The Legend of Heroes: Trails in the Sky series"

```
data.invoke_callback(kern, mem, thread_id, SCE_NGS_AT9_END_OF_DATA, 0, 0);
...
data.invoke_callback(kern, mem, thread_id, SCE_NGS_AT9_SWAPPED_BUFFER, prev_index,
    params->buffer_params[state->current_buffer].buffer.address());
```
Comment that out and no more crashing and audio still works to a degree - but clips very early.

### Reproduction Steps

[...]

PR #3299, opened 2024-05-20
modules/SceNgs: fix crash of Helldivers game

bookmist:
This PR fix https://github.com/Vita3K/Vita3K/issues/3286
I do reverse engineering too and found a small difference. 
Also I'm not sure about https://github.com/Vita3K/Vita3K/blob/76bd1ce4459363425e2ba0518b2f741da345cfba/vita3k/modules/SceNgsUser/SceNgs.cpp#L140-L141
I did not find anything like this in decompiled code, but game works with it, so I don't remove this code


Macdu:
Indeed you are right, I made a few mistakes when reverse engineering this function:
image
And you are also right about the line you're not sure about, after looking back at it, it shouldn't be here, you can remove it and I'll approve the PR.

Macdu (approved):

PR #3320, opened 2024-06-23
.github/workflows: Push build to store repository using ghr

nishinji:
Create releases directly using ghr to avoid bloat in Vita3K-builds.
Too bad there will need to re-set token. 🙏
image


Macdu (approved):
I set up the token, let's see if it works!

PR #3326, opened 2024-07-09
[gui] common_dialog: select most recent save in dialog when first opening

CeruleanSky:
Selecting the most recent save by default when first opening the dialog, especially when you have 70 saves to scroll through was useful to me so I thought this minor tweak might be useful to others as well.


Zangetsu38:
i no sur for this, psvita no using that ?
normally it juste selefct first visible

Zangetsu38:
rip

PR #3337, opened 2024-07-25
[gui] common_dialog.cpp fix save info dialog crash

CeruleanSky:
If you use the [I] button to view info for a save file in the save dialog then press the down button a few times with your controller there is an access violation on this patched line.  Tales of Innocence R where the info contains stats of 8 members of the party and requires scrolling down to view.  This fixes a crash that could cause a person to lose progress if inside the info dialog.


Macdu:
Nice catch, thank you! You need to run clang format (`./format.sh` or `./format.bat`) to fix the formatting issues.

Macdu (approved):

PR #3344, opened 2024-08-05
sdl: build compatibility with version < 2.24.0

scribam:
Works with Ubuntu 22.04 and SDL 2.0.20


Macdu (approved):
Personally I don't really see the point of supporting building with older SDL versions (during runtime it makes a bit more sense given that the OS sdl is used on Linux), but fine.

PR #3358, opened 2024-09-02
external: Update boost to 1.86.

Zangetsu38:
# About
- external: Update boost to 1.86.
Adding crc module.
- Fix Boost linkage issues in release build.
Fix path of BOOST_LIBRARYDIR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,9 +90,10 @@ endmacro(b2_build)
 macro(boost_compile)
     list(TRANSFORM BOOST_MODULES_TO_FIND PREPEND --with- OUTPUT_VARIABLE b2_cmd_line_common)
     set(b2_cmd_line_common ${b2} ${b2_cmd_line_common} -j${CPU_COUNT} --build-dir=${BOOST_INSTALLDIR} --stagedir=${BOOST_INSTALLDIR} cxxflags=${BOOST_CXX_FLAGS})
+	set(Boost_USE_DEBUG_RUNTIME "")
 	if (CMAKE_SYSTEM_NAME STREQUAL "Windows")
 		execute_process(
-			COMMAND ${b2_cmd_line_common} address-model=64 --architecture=x64 toolset=${BOOST_TOOLSET} stage
+			COMMAND ${b2_cmd_line_common} address-model=64 --architecture=x64 toolset=${BOOST_TOOLSET} variant=debug,release stage
 			WORKING_DIRECTORY ${BOOST_SOURCEDIR}
             OUTPUT_QUIET
 		)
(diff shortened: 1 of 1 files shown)


Macdu:
Doesn't the CMakeFile need to be changed if you want to fix building with custom Boost?

Zangetsu38:
there is already a pr for that

Macdu:
I mean just build the correct variant of boost using B2 depending on if the target is debug or not, it takes 4 lines which I showed you. Redpolline's PR doesn't even do that and is nowhere near close to getting accepted from my point of view.

Zangetsu38:
> I mean just build the correct variant of boost using B2 depending on if the target is debug or not, it takes 4 lines which I showed you. Redpolline's PR doesn't even do that and is nowhere near close to getting accepted from my point of view.

is see, but that change no broken debug build ? 🤔

CMakeLists.txt
```diff
@@ -127,7 +127,7 @@ macro(get_boost)
 
 		  # First, try to find Boost without any hints (system Boost)
 		  if(NOT VITA3K_FORCE_CUSTOM_BOOST)
-		      find_package(Boost 1.81 COMPONENTS ${BOOST_MODULES_TO_FIND} QUIET)
+		      find_package(Boost 1.83 COMPONENTS ${BOOST_MODULES_TO_FIND} QUIET)
```
scribam, line 130: Why 1.83 is required instead of 1.81?

Zangetsu38: i just for increment minimum requiement
but for now i try fix one things...

Zangetsu38:
Is finnaly ready, thx to @Macdu for help of fixing linking issue... stupid cmake...

CMakeLists.txt
Macdu: I do agree with scribam, the minimum version should only be increased if it is required.

Zangetsu38: Ah, i have thinks is good with see all workflow using 1.83 🥲

Zangetsu38: I have revert it is ready

CMakeLists.txt
```diff
@@ -138,8 +139,8 @@ macro(get_boost)
 		      set(BOOST_SOURCEDIR "${CMAKE_SOURCE_DIR}/external/boost")       # For internal use
 		      set(BOOST_INSTALLDIR "${CMAKE_BINARY_DIR}/external/boost")      # For internal use
 		      set(BOOST_ROOT "${BOOST_SOURCEDIR}")                            # find_package(Boost ...) hint
-		      set(BOOST_INCLUDEDIR "${BOOST_SOURCEDIR}")                      # find_package(Boost ...) hint
-		      set(BOOST_LIBRARYDIR "${BOOST_INSTALLDIR}/libs")                # find_package(Boost ...) hint
+		      set(BOOST_INCLUDE_DIRS "${BOOST_SOURCEDIR}")                      # find_package(Boost ...) hint
+		      set(BOOST_LIBRARY_DIRS "${BOOST_INSTALLDIR}/lib")                # find_package(Boost ...) hint
```
Macdu, line 143: Is there a reason for changing these?
According to [FindBoost](https://cmake.org/cmake/help/latest/module/FindBoost.html), the former are hints while the latter are variable set by find_package(boost) and thus shouldn't be set by the app cmake.

Zangetsu38: Ha yes... i have no see hint value... i have only see result and cache
But second fail is libs folder no exist, is lib

Zangetsu38: fixed.

Macdu (approved):

PR #3359, opened 2024-09-02
imgui: ImGuiChildFlags_Border migration

scribam:
Before imgui 1.90 (November 2023), the "ImGuiChildFlags child_flags = 0" parameter was "bool border = false" in BeginChild.

Although the imgui code should be backward compatible, when using a Debug build, I am hitting this static_assert https://github.com/ocornut/imgui/blob/cb16be3a3fc1f9cd146ae24d52b615f8a05fa93d/imgui.cpp#L5541 because `ImGuiWindowFlags_AlwaysAutoResize` is used with BeginChild.
> IM_ASSERT((window_flags & ImGuiWindowFlags_AlwaysAutoResize) == 0 && "Cannot specify ImGuiWindowFlags_AlwaysAutoResize for BeginChild(). Use ImGuiChildFlags_AlwaysAutoResize!");

This PR fixes it and does the migration from bool to ImGuiChildFlags


Macdu (approved):
I like their definition of "backward compatibility"

PR #3361, opened 2024-09-04
Vita3K/CI: set macOS minimum version to 11.0

KhoraLee:
Homebrew support only two major macOS version. So due to this, we did changed OSX_DEPLOYMENT_TARGET based on this.

From now, CI will build Boost and OpenSSL thats OSX_DEPLOYMENT_TARGET is 11.0 and download MoltenVK from MoltenVK repo which is also have OSX_DEPLOYMENT_TARGET 11.0.


vita3k/renderer/src/vulkan/renderer.cpp
```diff
@@ -32,7 +32,9 @@
 #include <SDL_vulkan.h>
 
 #ifdef __APPLE__
-#include <mvk_config.h>
+#include <MoltenVK/mvk_config.h>
+typedef VkResult (VKAPI_PTR *PFN_vkSetMoltenVKConfigurationMVK)(VkInstance ignored, const MVKConfiguration* pConfiguration, size_t* pConfigurationSize); // #include <MoltenVK/mvk_deprecated_api.h>
+
```
KhoraLee, line 37: KhronosGroup deprecated SetMoltenVKConfigurationMVK and changed it to use the standard Vulkan `VK_EXT_layer_settings` extension. But `VK_EXT_layer_settings` is introduced by Vulkan API revision 1.3.275. So to adapt this, we need to update Vulkan-Header and related things.

I couldn't use `mvk_deprecated_api.h` header because of conflict with our PTR class implementation and macOS's internal PTR typedef.

Macdu: I can update the vulkan headers, I'll do it

KhoraLee: Changed to use `VK_EXT_layer_settings` when setting full image view swizzle.
Ref. https://github.com/KhronosGroup/MoltenVK/blob/main/Docs/MoltenVK_Configuration_Parameters.md / https://vulkan.lunarg.com/doc/sdk/1.3.290.0/windows/layer_configuration.html

Macdu:
Lgtm, but you'll have to rebase your PR.

KhoraLee:
Rebased.

Macdu (approved):

PR #3362, opened 2024-09-04
imgui updates

scribam:
**Added:**

- New cursors from upstream

**Fixed:**

- Only the OpenGL backend currently frees the SDL cursors. By moving the code to ImGui_ImplSdl_Shutdown, it will be free for all backends.

**Breaking changes related to v1.91.1 I fixed:**

> - BeginChild(): renamed ImGuiChildFlags_Border to ImGuiChildFlags_Borders for consistency.

> - IO: moved clipboard functions from ImGuiIO to ImGuiPlatformIO:
>   - io.GetClipboardTextFn -> platform_io.Platform_GetClipboardTextFn
>   - io.SetClipboardTextFn -> platform_io.Platform_SetClipboardTextFn
>   - in function signatures, changed void* user_data to ImGuiContext* ctx for consistency with other functions. Pull your user data from platform_io.ClipboardUserData if used.


Macdu (approved):
lgtm, we never free ImGui, but that's still good to have

PR #3367, opened 2024-09-06
renderer/gl: Add support for U8U8U8 color format

saturnsky:
## Description
This PR adds support for the `SCE_GXM_COLOR_BASE_FORMAT_U8U8U8` color format in the OpenGL renderer. This format was previously unsupported, which could lead to rendering issues in games that use it.

## Changes
- Added `SCE_GXM_COLOR_BASE_FORMAT_U8U8U8` to the `translate_internal_format` function, mapping it to `GL_RGB8`
- Updated `translate_format` to return `GL_RGB` for `SCE_GXM_COLOR_BASE_FORMAT_U8U8U8`
- Modified `translate_type` to return `GL_UNSIGNED_BYTE` for `SCE_GXM_COLOR_BASE_FORMAT_U8U8U8`
- Updated `translate_swizzle` to handle `SCE_GXM_COLOR_BASE_FORMAT_U8U8U8`
- Added `SCE_GXM_COLOR_BASE_FORMAT_U8U8U8` to `bytes_per_pixel_in_gl_storage`, returning 3 bytes per pixel

## How Has This Been Tested?
- Tested with Ken Ga Kimi For V (PCSG00553), which now renders correctly with this change

### DIFF
--- a/vita3k/renderer/src/gl/color_formats.cpp
+++ b/vita3k/renderer/src/gl/color_formats.cpp
@@ -101,6 +101,9 @@ static const GLint *translate_swizzle(SceGxmColorSwizzle1Mode mode) {
 // Translate popular color base format that can be bit-casted for purposes
 GLenum translate_internal_format(SceGxmColorBaseFormat base_format) {
     switch (base_format) {
+    case SCE_GXM_COLOR_BASE_FORMAT_U8U8U8:
+        return GL_RGB8;
+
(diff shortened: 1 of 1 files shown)


Zangetsu38 (approved):

EXtremeExploit (approved):

Macdu (approved):
LGTM. However, please be aware that most modern GPU actually do not support this format (it is not used on Vulkan and instead emulated with rgba8). When using it, the OpenGL driver will itself emulate it using RGBA8.

PR #3373, opened 2024-09-08
renderer/vulkan: Increase scenes per frame in VKRenderTarget creation

saturnsky:
## Description
This PR slightly increases the number of scenes per frame in the VKRenderTarget creation process. This change aims to reduce warning messages in certain games that require more scenes per frame than previously allocated.

## Changes
- Modified the calculation of `samples_per_frame` in `VKRenderTarget` constructor:
  - Changed from `params.scenesPerFrame + 2` to `params.scenesPerFrame + 3`

## How Has This Been Tested?
- Tested with PhotoKano Kiss (PCSG00139)
  - The "Render Target is using more scenes per frame than what was planned!" warning message no longer appears
  - Note: This change does not improve the actual rendering results for this game

## Additional Notes
While this change reduces warning messages in PhotoKano Kiss, I hope that we won't need to increase this value further.

### DIFF
--- a/vita3k/renderer/src/vulkan/creation.cpp
+++ b/vita3k/renderer/src/vulkan/creation.cpp
@@ -191,7 +191,7 @@ VKRenderTarget::VKRenderTarget(VKState &state, const SceGxmRenderTargetParams &p
 
     constexpr uint16_t SCE_GXM_MAX_SCENES_PER_RENDERTARGET = 8;
     // hopefully this will always be enough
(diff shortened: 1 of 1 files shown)


Macdu:
I mean If you really want you can change the WARN to a WARN_ONCE. 

But increasing by 50% the amount of sync objects allocated in every game to reduce the amount of single time warnings generated without any graphical change in a single game is a bit too much.

saturnsky:
Understood. I thought removing Warnings would be the preferred development direction. If these messages are meant to help identify games that may encounter issues later, it’s better not to modify the code.
It also seems best not to change it to WARN_ONCE, as having the warning occur at the moment the issue arises could help in identifying problems later.

PR #3374, opened 2024-09-08
gui/interface: Fix reinstall using DnD and Vulkan

scribam:
With a vpk installed and the renderer backend set to Vulkan, trying to reinstall the same vpk using drag and drop will crash the app. Using OpenGL related functions with Vulkan is not a good idea, this PR fixes it.


Macdu (approved):
Thank you!

PR #3375, opened 2024-09-09
util/net utils: Implement file download resume.

Zangetsu38:
# About
util/net utils: Implement file download resume.
gui/vita3k update: Restore resume download new build.

### DIFF
--- a/vita3k/util/src/net_utils.cpp
+++ b/vita3k/util/src/net_utils.cpp
@@ -393,29 +393,43 @@ bool download_file(const std::string &url, const std::string &output_file_path, 
     if (!curl_download)
         return false;
 
-    CurlDownloadCallback curl_callback = +[](void *user_data, long bytes_total, long downloaded_bytes) {
+    CurlDownloadCallback curl_callback = +[](void *user_data, long total_bytes, long downloaded_bytes) {
         const auto data = (CallbackData *)user_data;
 
-        if (bytes_total == 0)
-            return 0; // Ignore if we dont have the total yet
+        if ((total_bytes == 0) || (downloaded_bytes == 0))
+            return 0; // Ignore if we dont have the total yet or if we haven't downloaded anything
 
         if (!data->second)
             return 0;
-        const auto progress_percent = static_cast<float>(downloaded_bytes) / static_cast<float>(bytes_total) * 100.0f;
 
-        const auto elapsed_time_ms = std::difftime(get_current_time_ms(), data->first);
+        // Calculate progress percentage
+        const auto total_bytes_downloaded = static_cast<double>(downloaded_bytes + data->first.bytes_already_downloaded);
+        const auto file_size = total_bytes + data->first.bytes_already_downloaded;
+        const auto progress_percent = static_cast<float>(total_bytes_downloaded) / static_cast<float>(file_size) * 100.0f;
+
+        // Calculate elapsed time
+        const auto current_time = get_current_time_ms();
+        const auto elapsed_time_ms = std::difftime(current_time, data->first.time);
 
         // Calculate remaining time in seconds
-        const auto remaining_bytes = static_cast<double>(bytes_total - downloaded_bytes);
+        const auto remaining_bytes = static_cast<double>(total_bytes - downloaded_bytes);
         const auto remaining_time = static_cast<uint64_t>((remaining_bytes / downloaded_bytes) * elapsed_time_ms) / 1000;
 
         ProgressState *callback_result = data->second(progress_percent, remaining_time);
 
         std::unique_lock<std::mutex> lock(callback_result->mutex);
 
+        // Store the current pause state
+        const auto pause = callback_result->pause;
+
+        // Wait until the pause state becomes false (unpaused)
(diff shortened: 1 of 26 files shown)


vita3k/util/src/net_utils.cpp
```diff
@@ -393,47 +393,58 @@ bool download_file(const std::string &url, const std::string &output_file_path,
...
 
         ProgressState *callback_result = data->second(progress_percent, remaining_time);
 
         std::unique_lock<std::mutex> lock(callback_result->mutex);
 
-        callback_result->cv.wait(lock, [&]() {
-            return !callback_result->pause;
-        });
+        if (callback_result->pause) {
+            callback_result->cv.wait(lock);
```
Macdu, line 423: Don't wait without a predicate, c++ condition variables are allowed to have a spurious wakeup

vita3k/util/src/net_utils.cpp
```diff
@@ -393,47 +393,58 @@ bool download_file(const std::string &url, const std::string &output_file_path,
...
+
+            // Add the time spent paused to the total elapsed time
```
Macdu, line 425: Isn't it the opposing: "Remove to time spent paused..." instead (the code is right, just the comment means the opposite)

Zangetsu38: no commentaire is fine, i no remove, i adding time elipsed in pause

vita3k/util/src/net_utils.cpp
Zangetsu38: yes, i have fixed it withlittle reworks this part

Zangetsu38:
Pr is ready

vita3k/util/src/net_utils.cpp
Macdu: You are increasing `data->first.time` which is effectively the starting time. The elapsed time is `now - data->first.time` which as a consequence decreases.
 What the line under does is "remove the time spent paused from the total elapsed time" which is indeed what you want to do for the remaining time approximation to be accurate.

Zangetsu38: Current comment is little diferent but

The goal here is to keep track of the actual elapsed time excluding any periods where the download was paused. The data->first.time variable represents the moment the download started. When we pause the download, the time essentially 'stops,' so when we resume, we need to account for that gap in time to ensure the total elapsed time reflects only the periods when the download was actively in progress. By adding the difference between the current time and the time right before we paused (current_time), we're compensating for the pause duration. This way, the elapsed time will continue to reflect the correct duration from start to resume, excluding any pauses.

Macdu: I mean, the fact that you are giving a chatgpt-generated answer (or at least something clearly not by you) shows you don't understand why this comment is or is not misleading. I can do the same thing:
image

Zangetsu38: I just used chatgpt to better explain to you what I wanted to say in English, he just helped me explain it correctly, he is also the one who confirmed to me that the comment posted perfectly explained what the code was doing, (the default comment was automatically written by copilot), and that you would have misunderstood the code.

I find it funny that they tell you that from your side when he literally explains the opposite to me.

But so to use my words, what the code does is when exiting the pause, it adds the time that has elapsed to the start time, so the comments explain this perfectly, the value of time increases, it does not decrease precisely

Zangetsu38: For mor explain why i no can't using your version
Screenshot_20240919_223502_ChatGPT

Macdu: Are you really saying that given that the comment was written by copilot and chatgpt said it was correct, it is correct? ....

Also the AI generated response in your last message makes absolutely no sense.

Zangetsu38: I'm not sure I understand what the problem is with the comment 🤔 
It's literally written what the code does, it adds to the starting time the time that elapsed during the break. 
Precisely to not take the break into account and to calculate the remaining time.

For copilot, he just made a suggestion once I had written the beginning "// Add" but then I rewrote this comment myself to be clearer.

Macdu: Yes, this exactly what you said:
- It adds to the starting time the time "elapsed during the break", i.e the time spent paused

Then why is your comment not mentioning the starting time and instead saying that the time spent paused is added to the total elapsed time? The total elapsed time can be computed as "now() - starting_time", so this comment is describing the opposite of what the code below is doing.

Zangetsu38: Oh yeah, sorry...

antoniodesousa: I think Macdu wants the comment to be changed to `Add the time spent paused to the starting time`, which makes sense.

Zangetsu38: > I think Macdu wants the comment to be changed to `Add the time spent paused to the starting time`, which makes sense.

yes no worry, i have undersstund, and here i have no idea why i have confonded start time to elapsed time...

and i have juste removed elapsed time copilot have added alone start time... anyway, is fixed

Macdu (approved):

PR #3376, opened 2024-09-10
external: Fix OpenSSL path installation for the new 3.3.2 version. 

Zangetsu38:
# About
- vita3k: fix openssl path install of new 3.3.2 version.
Set and use OPENSSL_ROOT_DIR as a cache variable to copy the bin files.

Fix generate project on windows with using pre-build version

### DIFF
--- a/vita3k/CMakeLists.txt
+++ b/vita3k/CMakeLists.txt
@@ -269,12 +269,12 @@ elseif(WIN32)
 		POST_BUILD
 		COMMAND ${CMAKE_COMMAND} -E copy_if_different "${CMAKE_BINARY_DIR}/external/discord_game_sdk/lib/x86_64/discord_game_sdk.dll" "$<TARGET_FILE_DIR:vita3k>")
 	endif()
-	if(EXISTS "${CMAKE_BINARY_DIR}/external/openssl")
+	if(EXISTS "${OPENSSL_ROOT_DIR}")
 		add_custom_command(
 		TARGET vita3k
 		POST_BUILD
-		COMMAND ${CMAKE_COMMAND} -E copy_if_different "${CMAKE_BINARY_DIR}/external/openssl/openssl-3/x64/bin/libssl-3-x64.dll" "$<TARGET_FILE_DIR:vita3k>"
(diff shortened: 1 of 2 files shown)


vita3k/CMakeLists.txt
```diff
@@ -275,12 +275,12 @@ elseif(WIN32)
...
 	endif()
-	if(EXISTS "${CMAKE_BINARY_DIR}/external/openssl")
+	if(EXISTS "${OPENSSL_ROOT_DIR}")
 		add_custom_command(
 		TARGET vita3k
 		POST_BUILD
-		COMMAND ${CMAKE_COMMAND} -E copy_if_different "${CMAKE_BINARY_DIR}/external/openssl/openssl-3/x64/bin/libssl-3-x64.dll" "$<TARGET_FILE_DIR:vita3k>"
-		COMMAND ${CMAKE_COMMAND} -E copy_if_different "${CMAKE_BINARY_DIR}/external/openssl/openssl-3/x64/bin/libcrypto-3-x64.dll" "$<TARGET_FILE_DIR:vita3k>")
+		COMMAND ${CMAKE_COMMAND} -E copy_if_different "${OPENSSL_ROOT_DIR}/bin/libssl-3-x64.dll" "$<TARGET_FILE_DIR:vita3k>"
+		COMMAND ${CMAKE_COMMAND} -E copy_if_different "${OPENSSL_ROOT_DIR}/bin/libcrypto-3-x64.dll" "$<TARGET_FILE_DIR:vita3k>")
```
Macdu, line 283: I don't think this is the best idea, this will cause cmake to try to copy openssl dlls even when using static openssl.

Zangetsu38: Ah ? That is code of red, i have thinks is good idea but 🤔

Zangetsu38: i have search, is not problem, if you look my code, OPENSSL_ROOT_DIR is set inside cache of cmake only in case it using msvc and download prebuild version, so dynamic version.
i have thinks using it, (OPENSSL_USE_STATIC_LIBS) but it is set only on mac and no inside cache

Macdu (dismissed):

Zangetsu38:
🗿 ok wait lol

Zangetsu38:
Done

Macdu (approved):

PR #3379, opened 2024-09-15
vita3k: More cleanup

scribam:


vita3k/host/dialog/include/host/dialog/filesystem.hpp
```diff
@@ -105,8 +103,4 @@ Result pick_folder(std::filesystem::path &resulting_path, const std::filesystem:
  */
 std::string get_error();
 
-} // namespace filesystem
-
-} // namespace dialog
-
-} // namespace host
+} // namespace host::dialog::filesystem
```
Macdu, line 106: Please keep newlines at the end of files

scribam: Fixed! For all project files

I tried to add [InsertNewlineAtEOF](https://releases.llvm.org/16.0.0/tools/clang/docs/ClangFormatStyleOptions.html#insertnewlineateof) in the .clang-format file to enforce this rule but it is only supported by clang-format 16+ (CI runner ubuntu-22.04 has version 15)

.github/format-check.sh
```diff
@@ -4,7 +4,7 @@ set -ex
 
 if [ "$GITHUB_EVENT_NAME" == "pull_request" ]; then 
     git fetch origin $GITHUB_BASE_REF:refs/remotes/origin/$GITHUB_BASE_REF
-    for f in $(git diff --diff-filter=AM --name-only origin/$GITHUB_BASE_REF 'vita3k/**.cpp' 'vita3k/**.h' 'tools/**.cpp' 'tools/**.h'); do
+    for f in $(git diff --diff-filter=AM --name-only origin/$GITHUB_BASE_REF 'vita3k/**.cpp' 'vita3k/**.h' 'vita3k/**.hpp' 'tools/**.cpp' 'tools/**.h' 'tools/**.hpp'); do
```
Macdu, line 7: I think for consistency, it is better to rename the only .hpp in the project to .h (there is no reason to have both .hpp and .h files)

scribam: Done

Macdu (approved):

PR #3380, opened 2024-09-17
modules/SceJpeg: Improve JPEG decoding accuracy

saturnsky:
## Description
This PR enhances the SceJpeg module to more accurately emulate the PS Vita's hardware JPEG decoder, including its specific limitations. The improvements were developed by creating and comparing with a real HomeBrew application running on PS Vita hardware, aiming to replicate its behavior and constraints as closely as possible.

## Changes
- Refined pitch calculation, which greatly improves emulation accuracy for Visual Novels
- Implemented BGRA decoding
- Added downscale decoding functionality
- Emulated PS Vita hardware limitations
- Implemented sceJpegCsc function for color space conversion from YUV444 to RGBA
- Added support for iFrameWidth argument in sceJpegMJpegCsc function

## Hardware Emulation Details
- Accurately emulated PS Vita's decoding size limits, ensuring that oversized images are handled the same way as on actual hardware
- Replicated other hardware-specific quirks and limitations discovered during the development of the HomeBrew test application

## Limitations
- Due to FFmpeg limitations, some features are not fully equivalent to PS Vita hardware:
  - YUV441 color space is not supported in the current implementation (though it's unlikely any software uses this)
  - Certain uncommon MCU configurations that would cause errors on PS Vita hardware are processed without issues in this implementation

## How Has This Been Tested?
The changes have been tested with the following games:
- Kiss Ato (PCSG00742)
- end sleep (PCSG00963)
- Ken ga Kimi for V (PCSG00553)
- Natsumegu (PCSG00650)

## Additional Notes
While BGRA decoding and downscale decoding have been implemented, it's unclear if any games actually utilize these features.
The emulation of hardware limitations is a crucial aspect of this PR, as it ensures that the emulator behaves as closely as possible to the actual PS Vita hardware, including its constraints and edge cases.

### DIFF
--- a/vita3k/codec/include/codec/state.h
+++ b/vita3k/codec/include/codec/state.h
@@ -34,8 +34,6 @@ union DecoderSize {
     struct {
         uint32_t width;
         uint32_t height;
-        uint32_t pitch_width;
-        uint32_t pitch_height;
     };
     struct {
         uint32_t samples;
@@ -46,8 +44,11 @@ enum DecoderColorSpace {
     COLORSPACE_UNKNOWN,
     COLORSPACE_GRAYSCALE,
     COLORSPACE_YUV444P,
+    COLORSPACE_YUV440P,
+    COLORSPACE_YUV441P,
     COLORSPACE_YUV422P,
     COLORSPACE_YUV420P,
+    COLORSPACE_YUV411P,
 };
 
 // glGet sort of API to use virtual stuff
@@ -121,14 +122,28 @@ struct H264DecoderState : public DecoderState {
     ~H264DecoderState() override;
 };
 
+struct MJpegPitch {
+    uint32_t x;
+    uint32_t y;
+};
+
+struct MJpegDecoderOptions {
+    bool use_standard_decoder;
+    int downscale_ratio;
+};
+
 struct MjpegDecoderState : public DecoderState {
-    uint32_t mcu_width;
-    uint32_t mcu_height;
+    bool use_standard_decoder = false;
+    int downscale_ratio = 1;
+
+    MJpegPitch pitch[4];
 
     DecoderColorSpace color_space_out;
 
     bool send(const uint8_t *data, uint32_t size) override;
     bool receive(uint8_t *data, DecoderSize *size) override;
+    void configure(void *options);
+    void get_pitch_info(MJpegPitch pitch[4]);
     DecoderColorSpace get_color_space();
 
     MjpegDecoderState();
@@ -265,8 +280,9 @@ struct PlayerState {
     ~PlayerState();
 };
 
-void convert_rgb_to_yuv(const uint8_t *rgba, uint8_t *yuv, uint32_t width, uint32_t height, const DecoderColorSpace color_space, int32_t inPitch);
-void convert_yuv_to_rgb(const uint8_t *yuv, uint8_t *rgba, uint32_t width, uint32_t height, const DecoderColorSpace color_space);
+void convert_rgb_to_yuv(const uint8_t *rgba, uint8_t *yuv, uint32_t width, uint32_t height, const DecoderColorSpace color_space, int32_t in_pitch);
+void convert_yuv_to_rgb(const uint8_t *yuv, uint8_t *rgba, uint32_t frame_width, const DecoderColorSpace color_space, bool is_bgra, MJpegPitch pitch[4]);
 int convert_yuv_to_jpeg(const uint8_t *yuv, uint8_t *jpeg, uint32_t width, uint32_t height, uint32_t max_size, const DecoderColorSpace color_space, int32_t compress_ratio);
 void copy_yuv_data_from_frame(AVFrame *frame, uint8_t *dest, const uint32_t width, const uint32_t height, bool is_p3);
+void calculate_pitch_info(uint32_t width, uint32_t height, int downscale_ratio, DecoderColorSpace color_space, bool use_standard_decoder, MJpegPitch output_pitch[4]);
 std::string codec_error_name(int error);
--- a/vita3k/codec/src/mjpeg.cpp
+++ b/vita3k/codec/src/mjpeg.cpp
@@ -30,85 +30,11 @@ extern "C" {
 
 #include <cassert>
 
-namespace {
-
-struct JpegInfo {
-    int mcu_width;
-    int mcu_height;
-    int width;
-    int height;
-};
-
-struct ChannelInfo {
-    uint32_t width_divisor;
-    uint32_t height_divisor;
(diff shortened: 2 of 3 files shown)


vita3k/codec/include/codec/state.h
```diff
@@ -121,14 +122,28 @@ struct H264DecoderState : public DecoderState {
...
+    int downscale_ratio;
+};
+
 struct MjpegDecoderState : public DecoderState {
-    uint32_t mcu_width;
-    uint32_t mcu_height;
+    bool use_standard_decoder = false;
+    int downscale_ratio = 1;
+
+    MJpegPitch pitch[4];
```
Macdu, line 139: Is there a reason for pitch to be an array of size 4? All the formats considered have at most 3 planes (I know the pitch field of SceJpegOutputInfo has size 4, but afaik the last one is never used).

vita3k/modules/SceJpeg/SceJpegUser.cpp
```diff
@@ -131,41 +181,46 @@ EXPORT(int, sceJpegDecodeMJpeg, const unsigned char *pJpeg, SceSize isize, uint8
...
 
-    convert_yuv_to_rgb(temporary.data(), pRGBA, size.pitch_width, size.pitch_height, state->decoder->get_color_space());
+    SceJpegPitch yuv_pitch[4];
+    state->decoder->get_pitch_info(yuv_pitch);
+
+    if (isUnsupportedImageSize(yuv_pitch[0].x, yuv_pitch[0].y)) {
+        return RET_ERROR(SCE_JPEG_ERROR_UNSUPPORT_IMAGE_SIZE);
+    }
+
+    convert_yuv_to_rgb(temporary.data(), pRGBA, size.width, size.height, state->decoder->get_color_space(), false, yuv_pitch);
```
Macdu, line 201: size contains the pre-downscaled size, while I believe we want the post-downscaled dimension here, right?

vita3k/codec/src/mjpeg.cpp
```diff
@@ -344,76 +365,136 @@ bool MjpegDecoderState::receive(uint8_t *data, DecoderSize *size) {
...
+                std::memcpy(dst_row, last_row, original_pitch[plane].x);
+            }
+
+            // Move to the next plane
+            original_data += original_pitch[plane].x * original_pitch[plane].y;
+        }
+
+        if (this->downscale_ratio != 1) {
+            // The image downscaling on PS Vita is not based on width and height scaling, but rather on pitch size.
+            // When attempting to emulate this using scs_scale, the code complexity increases significantly. Therefore, the implementation has been simplified.
```
Macdu, line 459:
```suggestion
            // When attempting to emulate this using sws_scale, the code complexity increases significantly. Therefore, the implementation has been simplified.
```

vita3k/codec/src/mjpeg.cpp
```diff
@@ -344,76 +365,136 @@ bool MjpegDecoderState::receive(uint8_t *data, DecoderSize *size) {
...
+                uint8_t last_pixel = src_row[src_width - 1];
+                std::fill(dst_row + src_width, dst_row + original_pitch[plane].x, last_pixel);
+            }
+
+            // Fill remaining rows with the last row
+            const uint8_t *last_row = original_data + (src_height - 1) * original_pitch[plane].x;
+            for (int y = src_height; y < original_pitch[plane].y; y++) {
+                uint8_t *dst_row = original_data + y * original_pitch[plane].x;
+                std::memcpy(dst_row, last_row, original_pitch[plane].x);
+            }
```
Macdu, line 451: I think it is better to zero-fill instead (except if it is exactly the behavior of the PS Vita)

vita3k/codec/include/codec/state.h
saturnsky: Theoretically, the YCCK type of JPEG format can have 4 color planes, and the PS Vita header seems to be designed with this situation in mind. Also, the PS Vita's default gallery app supports YCCK type JPEGs.
However, SceJpeg doesn't support YCCK type images. So I can't know the implementation details of how the PS Vita's gallery app is implemented. For example, it's possible that SceJpegArm supports YCCK type images. Or Sony might have implemented a dedicated decoder in their own apps.
The reason for setting the pitch to 4 in the current implementation is because there might be software on the PS Vita that uses 4 planes, and I didn't have enough confidence to reduce it to 3 at the cost of losing consistency with the PS Vita. It would be possible to implement it with 3 first, and then expand it if situations using all 4 are discovered later. Which approach do you think is better?

vita3k/modules/SceJpeg/SceJpegUser.cpp
saturnsky: That's correct. I missed it because there were no issues during testing, but the post-downscaled size should be used there.
It seems that due to internal implementation characteristics of FFmpeg, it coincidentally worked without problems. (It probably allocated a larger workspace and then processed only the space actually occupied by the image.) It's right to correct this.

vita3k/codec/src/mjpeg.cpp
saturnsky: That part is a loose emulation of the PS Vita's behavior.
In a typical JPEG decoding process, the JPEG is decoded in MCU units and then values outside the image's width/height range are discarded. However, the PS Vita doesn't discard values outside the pixel range in the decoded JPEG. Therefore, even if it's outside the image's width/height range, values are recorded in the decoding result if they're within the MCU block. Since it's difficult to exactly mimic this at a low level, I simulated the implementation through copying.
I question whether there are games where these values in this area have meaning (and there shouldn't be such games), but console games often make strange assumptions about hardware. So I guessed that the current approach is less likely to cause problems than zero-filling.

saturnsky:
I implemented sceJpegCsc additionally to fix the regression that could have occurred due to the PR.

vita3k/codec/include/codec/state.h
Macdu: Well, I think it's better to write the code assuming only at most 3 planes, but the way it is right now is fine.

Macdu:
Btw, we usually use a snake case convention for function names (even though it may not be properly enforced everywhere), could you change your function names to match this convention?

Also, you forgot to remove the ChatGPT response at the beginning of the PR description.

saturnsky:
Okay. I'll modify the function names overall according to the convention and push again.

As I'm not proficient in English writing, I've been getting help from an LLM for PR writing, but I accidentally left some text. It's been corrected now.​​​​​​​​​​​​​​​​

saturnsky:
@Macdu I have pushed again after modifying the function names.

Macdu (approved):

PR #3382, opened 2024-09-21
renderer: Improve depth buffer accuracy

Macdu:
Thanks to the amazing work of Saturnsky, it looks like after viewport transformations, z values on the PS Vita have a non standard behavior:
- If the z value is strictly greater than 1 it gets clipped
- If it is less than 0 it gets clamped

Emulate this by clamping in the shader and telling the renderer to clip primitives.
This is more accurate, fixes the text in the amazing spiderman and might fix some other clipping geometry too.


bookmist (approved):

imWatchful:
Note from Macdu: "...not completely accurate (but still better than the current implementation)"

Issue #3244, opened 2024-03-11
[Feature Request]: controller swapping


drp00per:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What feature are you suggesting?

Vita3k forces the first controller connected to it to be priority and controller 1. The ability to switch controller priority and switch between controllers.

### Why would this feature be useful?

Most devices, simply disconnecting your controller and switching them would be fine, but with the ever growing list of Android handheld devices, the inability to switch away from the installed controller means we cannot get full usability from the device or the emulator.

### Other emulators or software with a similar feature

Pretty much all of them

### Platform (if relevant)

None

Zangetsu38:
Vita3k force nothing, is managed by sdl directly, but is totaly possible change order, i have tested it

drp00per:
Would you mind sharing how to change order?  Using a retroid pocket 4 pro with a built in controller.  I've tried everything and looked at all the options and can't find anything regarding controller swapping.  Thanks

DisasterIncarnate:
this used to work perfectly fine on the steamdeck i could use the decks own controller or use a ps5 controller when the deck is connected to my tv, now my ps5 controller works fine in the grid menu, i can move all over and select the game i want, but once a game is actually launched only the steam decks own controller works, the ps5 controller becomes non functional.

drp00per:
That's the same issue I have, Can't use an external controller with any type of gaming handheld.  What I originally saw was that vita3k reads the first controller attached as controller 1 and since there is a controller physically attached its automatically controller 1.  But Zengetsu said that isn't true, but didn't tell how to fix the problem.  It's very confusing and disappointing.

karthikeyanat:
I do have same problem in rog ally. Vita3k show both internal pad and external controller but no way to change the order.

theRealSlimHenry:
I'm having the same issue. If it's possible to change the order in SDL, why is it not implemented as an option within the program? If it cannot be implemented, can we at least get the steps on how to fix the issue ourselves?

maurozb:
[...]

PR #3383, opened 2024-09-22
gui/controllers dialog: Implement swap for controller ports.

Zangetsu38:
# About
- gui/controllers dialog: Implement swap for controller ports.

Implement featur request by https://github.com/Vita3K/Vita3K/issues/3244

# Result
image

### DIFF
--- a/vita3k/gui/src/controllers_dialog.cpp
+++ b/vita3k/gui/src/controllers_dialog.cpp
@@ -173,6 +173,37 @@ static void add_bind_to_table(GuiState &gui, EmuEnvState &emuenv, const SDL_Game
     ImGui::PopID();
 }
 
+static void swap_controller_ports(CtrlState &state, int source_port, int dest_port) {
+    // Check if the ports are valid
+    if ((source_port < 0) || (source_port >= SCE_CTRL_MAX_WIRELESS_NUM) || (dest_port < 0) || (dest_port >= SCE_CTRL_MAX_WIRELESS_NUM)) {
+        LOG_ERROR("Ports are not valid.");
+        return;
+    }
+
+    // Find the controllers corresponding to the source and destination ports
+    auto source_controller_it = std::find_if(state.controllers.begin(), state.controllers.end(),
+        [&](const auto &pair) { return pair.second.port == (source_port + 1); });
+    auto dest_controller_it = std::find_if(state.controllers.begin(), state.controllers.end(),
+        [&](const auto &pair) { return pair.second.port == (dest_port + 1); });
+
+    // Check that both controllers exist
+    if ((source_controller_it == state.controllers.end()) || (dest_controller_it == state.controllers.end())) {
+        LOG_ERROR("Unable to find one or more controllers on the specified ports.");
+        return;
+    }
+
+    LOG_INFO("Controller on source port {} {} swapped with controller on destination port {} {}", source_port, state.controllers_name[source_port], dest_port, state.controllers_name[dest_port]);
+
+    // Swap controllers port and player index
+    SDL_GameControllerSetPlayerIndex(source_controller_it->second.controller.get(), dest_port);
+    SDL_GameControllerSetPlayerIndex(dest_controller_it->second.controller.get(), source_port);
+    std::swap(source_controller_it->second.port, dest_controller_it->second.port);
+
+    // Swap controller names and motion support
+    std::swap(state.controllers_name[source_port], state.controllers_name[dest_port]);
+    std::swap(state.controllers_has_motion_support[source_port], state.controllers_has_motion_support[dest_port]);
+}
+
 void draw_controllers_dialog(GuiState &gui, EmuEnvState &emuenv) {
     const ImVec2 VIEWPORT_POS(emuenv.viewport_pos.x, emuenv.viewport_pos.y);
     const ImVec2 VIEWPORT_SIZE(emuenv.viewport_size.x, emuenv.viewport_size.y);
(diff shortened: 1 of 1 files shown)


vita3k/gui/src/controllers_dialog.cpp
```diff
@@ -173,6 +173,39 @@ static void add_bind_to_table(GuiState &gui, EmuEnvState &emuenv, const SDL_Game
...
+    std::swap(source_controller_it->second.port, dest_controller_it->second.port);
+    std::swap(source_controller_it->second.has_accel, dest_controller_it->second.has_accel);
+    std::swap(source_controller_it->second.has_gyro, dest_controller_it->second.has_gyro);
+    std::swap(source_controller_it->second.has_led, dest_controller_it->second.has_led);
+
+    // Swap controller names
+    std::swap(state.controllers_name[source_port - 1], state.controllers_name[dest_port - 1]);
+    std::swap(state.controllers_has_motion_support[source_port - 1], state.controllers_has_motion_support[dest_port - 1]);
+}
+
```
Macdu, line 208: There are like 10 different issues with this code. Can't you just swap the port value and be done?

Zangetsu38: Hmmm 🤔 i check

Zangetsu38: Done, sorry, i have not see this error, for’other is obligatory for controller name keep same
If now switch name it keep dualsense when is ds4 after swap

vita3k/gui/src/controllers_dialog.cpp
```diff
@@ -173,6 +173,36 @@ static void add_bind_to_table(GuiState &gui, EmuEnvState &emuenv, const SDL_Game
...
+    if (source_port < 1 || source_port > SCE_CTRL_MAX_WIRELESS_NUM || dest_port < 1 || dest_port > SCE_CTRL_MAX_WIRELESS_NUM) {
+        LOG_ERROR("Ports are not valid.");
+        return;
+    }
+
+    // Find the controllers corresponding to the source and destination ports
+    auto source_controller_it = std::find_if(state.controllers.begin(), state.controllers.end(),
+        [source_port](const auto &pair) { return pair.second.port == source_port; });
+    auto dest_controller_it = std::find_if(state.controllers.begin(), state.controllers.end(),
+        [dest_port](const auto &pair) { return pair.second.port == dest_port; });
```
Macdu, line 187:
```suggestion
    auto source_controller_it = std::find_if(state.controllers.begin(), state.controllers.end(),
        [&](const auto &pair) { return pair.second.port == source_port; });
    auto dest_controller_it = std::find_if(state.controllers.begin(), state.controllers.end(),
        [&](const auto &pair) { return pair.second.port == dest_port; });
```

vita3k/gui/src/controllers_dialog.cpp
```diff
@@ -214,10 +242,17 @@ void draw_controllers_dialog(GuiState &gui, EmuEnvState &emuenv) {
...
+            static int selected_port = -1;
             for (auto i = 0; i < ctrl.controllers_num; i++) {
                 ImGui::TableNextRow();
                 ImGui::TableSetColumnIndex(0);
-                ImGui::Text("%-8d", i);
+                selected_port = i;
+                ImGui::PushID(i);
+                ImGui::SetNextItemWidth(50.f * emuenv.dpi_scale);
+                if (ImGui::Combo("##swap_port", &selected_port, port_names, ctrl.controllers_num))
+                    swap_controller_ports(ctrl, i + 1, selected_port + 1);
```
Macdu, line 254:
```suggestion
                    swap_controller_ports(ctrl, i, selected_port);
```
In `swap_controller_ports`, at 6 different times, you need to decrement the port, while you need the 1-indexed version only twice.

vita3k/gui/src/controllers_dialog.cpp
```diff
@@ -173,6 +173,36 @@ static void add_bind_to_table(GuiState &gui, EmuEnvState &emuenv, const SDL_Game
...
+    // Check that both controllers exist
+    if (source_controller_it == state.controllers.end() || dest_controller_it == state.controllers.end()) {
+        LOG_ERROR("Unable to find one or more controllers on the specified ports.");
+        return;
+    }
+
+    LOG_INFO("Controller on source port {} {} swapped with controller on destination port {} {}", source_port, state.controllers_name[source_port - 1], dest_port, state.controllers_name[dest_port - 1]);
+
+    // Swap controllers port and player index
+    SDL_GameControllerSetPlayerIndex(source_controller_it->second.controller.get(), dest_port - 1);
```
Macdu, line 198: Shouldn't `SDL_GameControllerSetPlayerIndex` also be called for dest_controller?

vita3k/gui/src/controllers_dialog.cpp
Zangetsu38: hmm i see yes

vita3k/gui/src/controllers_dialog.cpp
Zangetsu38: is swaped alone when you moving one player index

Zangetsu38: so if you have player 1 2, swap done 2 1 alone

but if have more 2 controller i no know

Macdu: It should work because internally SDL re-assigns the first available index to the other controller, however this relies on the internal behavior of SDL and not on the public API documentation, so I think adding a
`SDL_GameControllerSetPlayerIndex(dest_controller_it ->second.controller.get(), source_port ); `
doesn't hurt

Zangetsu38: ok, done, in first version i have added it, but removed after see is not needed
player index tecnicaly is only important for Playstation controller with it indicate position of player

vita3k/gui/src/controllers_dialog.cpp
```diff
@@ -214,10 +243,17 @@ void draw_controllers_dialog(GuiState &gui, EmuEnvState &emuenv) {
             ImGui::Spacing();
             ImGui::TableSetColumnIndex(2);
             ImGui::TextColored(GUI_COLOR_TEXT_TITLE, "%s", "Motion Support");
+            const char *port_names[] = { "1", "2", "3", "4" };
+            static int selected_port = -1;
             for (auto i = 0; i < ctrl.controllers_num; i++) {
                 ImGui::TableNextRow();
                 ImGui::TableSetColumnIndex(0);
-                ImGui::Text("%-8d", i);
+                selected_port = i;
```
Macdu, line 251:
```suggestion
                int selected_port = i;
```
Last small thing, no need to make a static variable if you always reset it.

Zangetsu38: done

Macdu (approved):

Zangetsu38:
after this all change i have re tested with my two contreoller for verify all 👌

PR #3389, opened 2024-09-24
gui/information bar: Fix iterator invalidation bug in notice list processing

saturnsky:
## Description
This PR addresses a critical iterator invalidation bug in the notice list processing logic within the information bar component. The bug was caused by erasing elements from a container while iterating over it, which can lead to undefined behavior and potential crashes.
Actual issues in the existing code would have primarily occurred only when manually deleting multiple files, so most users would not have experienced problems in the majority of usage scenarios. 

## Changes
- Replaced index-based erase with safe iterator-based erase


Macdu (approved):
Thank you for the PR. Also the GUI code is completely plagued with issues like this...

PR #3390, opened 2024-09-24
ffmpeg: update submodule

oltolm:
see title


Macdu:
You can put this commit in your other PR #3360, the only thing the ffmpeg update does is add mingw binaries, to it makes sense to be bundled with your mingw support pr.

oltolm:
done

PR #3392, opened 2024-09-26
make unicorn optional and disable per default

oltolm:
Disabling Unicorn makes building Vita3k a bit simpler.


Macdu:
I don't think "Making the compile time 30 seconds faster for the first time" is only in itself a good enough reason to remove Unicorn.

oltolm:
Should I close the PR or enable Unicorn per default?

Macdu:
> Should I close the PR or enable Unicorn per default?

The option to disable unicorn is already available in the Android version (and is almost exactly the same as the content of your PR), which will be merged to master at some point, so you can close this PR, except if you have a good reason for wanting to be able to disable Unicorn right now.

PR #3395, opened 2024-10-02
.github/workflows: Detail the name of the artifacts

nishinji:
Detail the name of the artifact for store repository. This is useful for identifying when you are downloading a lot from a store repository.
image
Also, 7z compression is used to reduce the size. (Reduction of more than 30%.)

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -31,7 +31,6 @@ jobs:
           - os: ubuntu-latest
             version: zip
             cache_path: ~/.ccache
-            extra_cmake_args: 
             cmake_preset: linux-ninja-clang15
           - os: windows-latest
             version: zip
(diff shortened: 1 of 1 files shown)


.github/workflows/c-cpp.yml
```diff
@@ -181,9 +180,14 @@ jobs:
...
     outputs:
       BuildTag: ${{ env.build_variable }}
+      ShortSHA: ${{ env.git_short_sha }}
 
   create-release:
     needs: [build]
+    strategy:
+      fail-fast: false
+      matrix:
+        repo: [master, store]
```
Macdu, line 190: Are you sure about disabling fail-fast on the release build? Won't that mess up the auto-updater for an OS if the build for this OS fails?

nishinji: This is so that if one job fails, the parallel workflow is not aborted. (Here is master and store) This means that if one release fails, although I don't think that will happen, it will run until the other completes.
Also, since the update script only references the master repo, I don't think it will cause any problems.

Macdu (approved):

PR #3396, opened 2024-10-04
modules/SceLibc: Implement various module variables

saturnsky:
## Description

This PR implements some variables in the SceLibc module. I proceeded with the development after seeing the 0xdeadbeef message occurring in several games. However, after proceeding, it appears that libc doesn't actually use SceLibc.cpp, so it seems to have no practical significance.
Additionally, the starting number for file descriptors has been modified. This modification was made because if fd starts at 0 as before, there's a possibility of having fds like _Stdout(0x401), which I thought could potentially cause issues. It's unclear whether this would actually be problematic.
Nevertheless, since I've already done the work, I thought it would be better to submit the PR rather than not submitting it. Whether to accept this PR seems like a policy issue, and if it's deemed meaningless, I think it's fine to reject it according to policy.


Macdu:
Something to note is that `alloc` has a page granularity, so every alloc you do (even if it only allocates 4 bytes) allocates a whole 4kb page. So this PR adds a 80KB memory increase for everyone (even when not using libc).
Also it should be noted that being able to HLE libc is more of a legacy feature, I don't think there is a reason for anyone to HLE libc instead of LLE it.

Finally, the commit name hides the major change: changing the fd start offset (which is not related in any way to libc) and might cause improvements/regressions on some games (although I doubt it).

saturnsky:
I included the fd modification in the same commit because I thought it was part of the implementation for _Stdio and related functions. However, you're right about the impact of the commit.

I wasn't aware of the memory penalty, and considering that, the PR for HLE seems even more unnecessary. I'll close this PR and either keep only the fd-related part or improve it, then submit it as a separate PR later.

PR #3397, opened 2024-10-04
modules/SceKernel: Implement sceKernelGetEventPattern

saturnsky:
## Description
This PR implements the `sceKernelGetEventPattern` function in the SceLibKernel module, enhancing the emulation of the PS Vita's event handling system.

## Changes
- Implemented `sceKernelGetEventPattern` function in SceLibKernel module
- Implemented `_sceKernelGetEventPattern` function in SceKernelThreadMgr module
- Updated function signatures and declarations in header files

## Emulation Details
- The implementation follows the expected behavior of the PS Vita's kernel, retrieving the event pattern for a given event ID
- While this implementation covers the basic functionality, there might be edge cases or specific hardware behaviors that are not yet fully emulated

## Additional Notes
- The implementation follows the code convention used in other `event_id` related functions for consistency

### DIFF
--- a/vita3k/modules/SceKernelThreadMgr/SceThreadmgr.cpp
+++ b/vita3k/modules/SceKernelThreadMgr/SceThreadmgr.cpp
@@ -249,9 +249,16 @@ EXPORT(int, _sceKernelGetEventInfo) {
     return UNIMPLEMENTED();
 }
 
-EXPORT(int, _sceKernelGetEventPattern) {
-    TRACY_FUNC(_sceKernelGetEventPattern);
-    return UNIMPLEMENTED();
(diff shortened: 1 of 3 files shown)


bookmist:
Good improvement. Thank you.
`Ptr<SceUInt32>` in function headers can be changed to `SceUInt32*` . System that calls HLE functions converts guest pointers to host pointers automatically. It's not a problem at all to use Ptr<> here but I think that simple pointers are simpler and shorter.

saturnsky:
Thank you for your input on the internal structure. I agree that the suggested approach would be better, not only for conciseness but also considering the implementation of functions using different event_ids. I have modified the related code accordingly.

bookmist (approved):

Macdu (approved):

Issue #3217, opened 2024-02-16
[BUG]: Regression after improvement of virtual memory


dima-xd:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Affected Build(s)

3571

### Description of Issue

After #3043 the game https://github.com/Vita3K/compatibility/issues/943 (and probably others too) doesn't boot anymore.
I looked a little bit at it, and found out that the game boots only if the start page is 0 in <a href="https://github.com/Vita3K/Vita3K/pull/3043/files#diff-190b4f60f0116bccabf48fdb04073c304ab3de0101e52a1db64485b596f8c891R200">alloc_aligned</a>.

### Expected Behavior

The game boots

### Reproduction Steps

Open the game

### Log File

[vita3k.log](https://github.com/Vita3K/Vita3K/files/14309756/vita3k.log)

### System Configuration

CPU: AMD Ryzen 5 5600H
GPU/Driver: NVIDIA GeForce RTX 3600
RAM: 16 GB
OS: Windows 11

EXtremeExploit:
Fixed by #3398

PR #3398, opened 2024-10-05
modules/SceSysmem: Set correct start_address for USER_RW_UNCACHE

saturnsky:
## Description
This PR addresses a regression introduced after Build 3463 ([#3043](https://github.com/Vita3K/Vita3K/pull/3043)) that affected some games ([#3217](https://github.com/Vita3K/Vita3K/issues/3217)). The issue was caused by games incorrectly interpreting the address returned by `sceKernelGetMemBlockBase` as a signed value and treating negative values as errors.

## Changes
- Updated the `sceKernelAllocMemBlock` function in the SceSysmem module
- Set the correct `start_address` for `SCE_KERNEL_MEMBLOCK_TYPE_USER_RW_UNCACHE`

## Emulation Details
- Verified that `SCE_KERNEL_MEMBLOCK_TYPE_USER_RW_UNCACHE` uses the same memory region as `SCE_KERNEL_MEMBLOCK_TYPE_USER_MAIN_PHYCONT_NC_RW`
- Both memory block types now use the 0x70000000 address range

## Impact
This fix should resolve the regression issues reported in #3217, improving compatibility for affected games.

## Testing
- Tested with games that were previously affected by the regression
- Confirmed that memory allocation now works as expected for these cases

### DIFF
--- a/vita3k/modules/SceSysmem/SceSysmem.cpp
+++ b/vita3k/modules/SceSysmem/SceSysmem.cpp
@@ -125,6 +125,7 @@ EXPORT(SceUID, sceKernelAllocMemBlock, const char *pName, SceKernelMemBlockType 
     // https://wiki.henkaku.xyz/vita/SceSysmem_Types#memtype_bit_value
     Address start_address;
     switch (type) {
+    case SCE_KERNEL_MEMBLOCK_TYPE_USER_RW_UNCACHE:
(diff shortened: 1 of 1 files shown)


bookmist (approved):

Macdu (approved):
Indeed, nice catch, the 0x70000000-0x80000000 virtual address range is reserved for uncached user memory.

PR #3399, opened 2024-10-05
gui: Fix load_image pointer validity check

scribam:
Also set the correct type for `size`, `stbi_load_from_memory` uses an `int`
```
STBIDEF stbi_uc *stbi_load_from_memory(stbi_uc const *buffer, int len, int *x, int *y, int *comp, int req_comp)
```


Macdu:
If you want to make the types match, might as well make them match completely (i.e use an `int` instead of a `std::int32_t`)

scribam:
> If you want to make the types match, might as well make them match completely (i.e use an `int` instead of a `std::int32_t`)

Done 🙂

Macdu (approved):

PR #3400, opened 2024-10-06
modules/SceGxm: Prevent viewport updates outside of the scene

saturnsky:
This PR implements improvements and refactoring for the viewport update logic. The main changes are as follows:

1. Modified the logic to prevent viewport updates outside of the Scene.
2. Integrated duplicate viewport update code into a single function, enhancing code readability and maintainability.

These modifications have resolved an issue where certain games would crash due to attempting viewport updates before Scene initialization. (Reference: [Vita3K/compatibility#212](https://github.com/Vita3K/compatibility/issues/212))

### DIFF
--- a/vita3k/modules/SceGxm/SceGxm.cpp
+++ b/vita3k/modules/SceGxm/SceGxm.cpp
@@ -1316,6 +1316,16 @@ EXPORT(int, sceGxmAddRazorGpuCaptureBuffer) {
     return UNIMPLEMENTED();
 }
 
+static void update_viewport(renderer::State &state, SceGxmContext *context) {
+    if (context->state.viewport.enable == SCE_GXM_VIEWPORT_ENABLED) {
+        renderer::set_viewport_real(state, context->renderer.get(), context->state.viewport.offset.x,
+            context->state.viewport.offset.y, context->state.viewport.offset.z, context->state.viewport.scale.x,
+            context->state.viewport.scale.y, context->state.viewport.scale.z);
+    } else {
+        renderer::set_viewport_flat(state, context->renderer.get());
+    }
+}
+
 EXPORT(void, sceGxmSetDefaultRegionClipAndViewport, SceGxmContext *context, uint32_t xMax, uint32_t yMax) {
     TRACY_FUNC(sceGxmSetDefaultRegionClipAndViewport, context, xMax, yMax);
     const std::uint32_t xMin = 0;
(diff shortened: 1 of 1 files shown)


vita3k/modules/SceGxm/SceGxm.cpp
```diff
@@ -4191,6 +4191,9 @@ EXPORT(void, sceGxmSetViewportEnable, SceGxmContext *context, SceGxmViewportMode
     // Set viewport to enable/disable, no additional offset and scale to set.
     if (context->state.viewport.enable != enable) {
         context->state.viewport.enable = enable;
+        if (!context->renderer->current_render_target) {
```
Macdu, line 4194: This should be checked [here](https://github.com/Vita3K/Vita3K/blob/f43ac8765ea1a28bb0270092dadab9c900f991de/vita3k/renderer/src/state_set.cpp#L150) instead.

Moreover, this does not check that this happens within a scene (current_render_target never gets set back to null after being set). This will cause the first scene in each game to completly ignore the viewport (and even the scenes after could depending on race conditions).

saturnsky: I modified it to check context->state.active instead of current_render_target to verify if we are within a Scene when executing the command, as it seems more appropriate.
Apart from that, as I understand it, sceGxmBeginScene calls sceGxmSetDefaultRegionClipAndViewport during initialization, and sceGxmSetDefaultRegionClipAndViewport calls set_viewport_real and set_viewport_flat depending on whether the viewport is enabled, so it shouldn't be problematic.
This is because when a Scene is initialized, the Viewport is reset based on its Enable status.

saturnsky:
I rewrote the commit message and source code of a PR that fixes the same issue, including corrections to parts I had misunderstood.
The text has changed significantly, but the problem I'm trying to fix remains the same.

Macdu (dismissed):

vita3k/modules/SceGxm/SceGxm.cpp
```diff
@@ -1316,6 +1316,16 @@ EXPORT(int, sceGxmAddRazorGpuCaptureBuffer) {
     return UNIMPLEMENTED();
 }
 
+void update_viewport(renderer::State &state, SceGxmContext *context) {
```
Zangetsu38, line 1319: no better using static void ? 🤔

saturnsky: You're right. There's no reason for that function not to be static.

Macdu (approved):

PR #3408, opened 2024-10-17
compat: change time style to yyyy-mm-dd

nishinji:
Follow ISO 8601.
`yyyy-mm-dd` is defined in ISO. The compat-repo must be changed at the same time.

### DIFF
--- a/vita3k/compat/src/compat.cpp
+++ b/vita3k/compat/src/compat.cpp
@@ -106,7 +106,7 @@ bool load_app_compat_db(GuiState &gui, EmuEnvState &emuenv) {
 
     // Check if compatibility database is up to date in first load
     if (db_updated_at.empty()) {
-        db_updated_at = compatibility.attribute("db_updated_at").as_string();
+        db_updated_at = compatibility.attribute("iso_db_updated_at").as_string();
         if (update_app_compat_db(gui, emuenv))
             return true;
     }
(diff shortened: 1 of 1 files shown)


Zangetsu38 (approved):

Macdu (approved):

Macdu:
@nishinji Tell me when I can merge so you can update the compat repo at the same time.

nishinji:
> @nishinji Tell me when I can merge so you can update the compat repo at the same time.

Whenever. I have already opened a PR in compat repo https://github.com/Vita3K/compatibility/pull/2688.
We just need to merge them at the same time(more precisely, merge the PR of the compat repo as soon as a new release is made ).

Macdu:
Just one thought, won't that mess up with people using an older version (and in particular the android version)?

Zangetsu38:
> Just one thought, won't that mess up with people using an older version (and in particular the android version)?

For user using old yes same for android, but only android build is probleme here, for other can just update 🤔

nishinji:
Workaround added.
Builds prior to this commit will download compatibility every time they are launched, but this should not be a major issue.

Zangetsu38:
> Workaround added. Builds prior to this commit will download compatibility every time they are launched, but this should not be a major issue.

Your workaround change nothing here 🤔
Is not inside XML or on build release
You juste create variable no used 🤔

vita3k/compat/src/compat.cpp
```diff
@@ -173,7 +173,7 @@ bool update_app_compat_db(GuiState &gui, EmuEnvState &emuenv) {
     auto &lang = gui.lang.compat_db;
 
     // Get current date of last compat database updated at
-    const auto updated_at = net_utils::get_web_regex_result(latest_link, std::regex(R"(Updated at: (\d{2}-\d{2}-\d{4} \d{2}:\d{2}:\d{2}))"));
+    const auto updated_at = net_utils::get_web_regex_result(latest_link, std::regex(R"(Updated at: (\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}Z))"));
```
Zangetsu38, line 176:
```suggestion
    const auto updated_at = net_utils::get_web_regex_result(latest_link, std::regex(R"(ISO Updated at: (\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}Z))"));
```

Zangetsu38 (approved):
lgtm

Macdu (approved):

Issue #3403, opened 2024-10-09
[BUG]: Freeze occurring in non-standard multi-buffering games


saturnsky:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Affected Build(s)

v0.2.0 3664

### Description of Issue

In general, when PS Vita performs multi-buffering, arguments are passed to sceGxmDisplayQueueAddEntry in the following order:
sceGxmDisplayQueueAddEntry(FrontBufferSyncObject, BackBufferSyncObject, BackBufferDisplayData)

However, in a specific game, I confirmed that the arguments are passed in the order of:
sceGxmDisplayQueueAddEntry(BackBufferSyncObject, FrontBufferSyncObject, BackBufferDisplayData)
This means that the next buffer is assigned to oldBuffer and the current buffer to newBuffer.

The sceGxmDisplayQueueAddEntry function in Vita3K is developed assuming that the arguments will always come in the order of frontBuffer - backBuffer, causing operational issues in games like this.

On the PS Vita, it seems that sceGxmDisplayQueueAddEntry puts both oldBufferSyncObject and newBufferSyncObject into the queue, resulting in the same operation regardless of the order of arguments.

Modifying this part will fix the freezing in games where multi-buffering is implemented in a non-standard way.

### Expected Behavior

On the PS Vita, it is presumed that sceGxmDisplayQueueAddEntry performs the same operation on both oldBuffer and newBuffer. (At the very least, entering them in reverse order does not cause issues with the game's operation.)
In Vita3K, the game operates differently depending on the order of these two arguments.

### Reproduction Steps

I can confirm the issue immediately after launching Furmins.
Currently, Furmins can only call sceGxmDisplayQueueAddEntry once.
For Furmins, I recorded the arguments of sceGxmColorSurfaceInit, sceGxmBeginScene, sceGxmEndScene, and sceGxmDisplayQueueAddEntry to examine how the arguments are passed.
Additionally, I confirmed that if the arguments of sceGxmDisplayQueueAddEntry are swapped at the beginning of the function to be recognized in reverse order, the game does not freeze.

### Log File

[10:37:13.196] |I| [config::init_config]: backend-renderer: Vulkan
[10:37:13.196] |I| [config::init_config]: log-level: Trace
[10:37:13.686] |I| [SDL_main]: Vita3K v0.2.0 3664-93759d9c-Saturnsky/SceGxm_Viewport
[...]

PR #3410, opened 2024-10-21
modules/SceGxm: Improve sync object handling and simplify display queue logic

saturnsky:
This PR is intended to address issue #3403.

The logic of `sceGxmDisplayQueueAddEntry` has been modified according to the implementation of the PS Vita, so that it now waits for both `oldBufferSync` and `newBufferSync`. After this change, workaround codes that seemed unnecessary have also been removed. Additionally, the `handle_wait_sync_object` command has been removed, as it no longer appears logically necessary after these changes. Now, blocks related to `SyncObject` exist only within the logic of `sceGxmDisplayQueueAddEntry` (specifically in `display_entry_thread`).

This is a very aggressive modification, so there is a possibility of errors. However, based on my understanding, this change does not seem to introduce any logical issues.

The part related to frame prediction has been left unchanged as I couldn't fully assess the complexity. However, it may also be necessary to modify the frame prediction to target both the old and new buffers. This aspect is outside the scope of this PR.

Although I did not observe any regressions in my game library, I cannot be entirely certain that there are no issues due to the large scale of the changes. Additionally, there is a possibility that I misunderstood the existing lock logic and proceeded with development based on incorrect assumptions. Therefore, I am approaching this PR with caution.

### DIFF
--- a/vita3k/modules/SceGxm/SceGxm.cpp
+++ b/vita3k/modules/SceGxm/SceGxm.cpp
@@ -901,7 +901,6 @@ static void display_entry_thread(EmuEnvState &emuenv) {
         LOG_CRITICAL("display_thread not found. thid:{}", emuenv.gxm.display_queue_thread);
         return;
     }
-    Ptr<SceGxmSyncObject> previous_sync = Ptr<SceGxmSyncObject>();
 
     while (true) {
         auto display_callback = display_queue.top();
@@ -911,7 +910,10 @@ static void display_entry_thread(EmuEnvState &emuenv) {
         SceGxmSyncObject *old_sync = display_callback->old_sync.get(emuenv.mem);
         SceGxmSyncObject *new_sync = display_callback->new_sync.get(emuenv.mem);
 
-        renderer::wishlist(new_sync, display_callback->new_sync_timestamp);
+        // sceGxmDisplayQueueAddEntry waits for both buffers to complete
+        renderer::wishlist(old_sync, display_callback->old_sync_timestamp);
+        if (old_sync != new_sync)
+            renderer::wishlist(new_sync, display_callback->new_sync_timestamp);
 
         // now we can remove the thread from the display queue
         display_queue.pop();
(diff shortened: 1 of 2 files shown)


Macdu:
You are removing the sync object waits on scenes and transfer gxm functions. This definitely loosk wrong: a scene/copy after a displayQueueAddEntry must wait for the sync object before starting .One of the goal of the sync object is not to draw to the surface being rendered (although this wouldn't happen anyway on Vita3K), also it is possible this can introduce race conditions for the rendering buffers.

saturnsky:
I understood it. I had removed the WaitSyncObject because I thought the handling would be explicitly done in sceGxmDisplayQueueAddEntry on the PS Vita, but it seems that change won't be beneficial. I reverted the changes related to WaitSyncObject and kept only the modifications in other parts.
I tested it within the same scope as when I previously submitted the PR, and within the range I tested, the new PR performed equivalently to the previous one. Additionally, the possibility of the side effects you were concerned about should also be reduced.

saturnsky:
My previous PR did not ensure that `sceGxmDisplayQueueAddEntry` would return only after the sync through `SceGxmSyncObject` had been completed. The current code more accurately emulates the actual behavior of `sceGxmDisplayQueueAddEntry` compared to before.

In fact, this revised code addresses a regression in [RecoLove: Blue Ocean](https://github.com/Vita3K/compatibility/issues/820).

However, it appears that the original Vita3K code aimed to minimize blocking on the main thread, while the revised PR introduces blocking on the main thread. Although I believe this PR is more accurate, I'm uncertain if it’s acceptable, as I lack sufficient historical information to make that determination.

Could you provide your thoughts on this aspect?

saturnsky:
Sorry for the confusion. I made a poor decision while developing in a fatigued state. It’s natural that making the logic synchronous would fix rendering issues in some games. 

However, it's correct for `sceGxmDisplayQueueAddEntry` not to block, and blocking should be handled within `display_entry_thread`. I should have remembered that just because more games function properly doesn't mean the issue was correctly identified. The previous PR was correct, and the cause of that game's malfunction likely lies elsewhere.

The PR has been reverted to its original state.

Macdu:
Indeed, the only time sceGxmDisplayQueueAddEntry should block is when the display queue is full. With the PR in its current state, it can still lead to race conditions (scenes began after sceGxmDisplayQueueAddEntry with the matching sync object should wait for the entry flip to be done).
What needs to be done for both the old and new sync objects is to:
- Wishlist at the timestamp when the entry was added
- Display the entry
- Call subject_done with the next timestamp

Also if the old and new sync objects are the same (this happens on some games), this should only be done once.

saturnsky:
Thank you for your continued advice.
I modified it to increment timestamp_ahead for both SyncObjects in sceGxmDisplayQueueAddEntry and to perform subject_done when display_thread->run_guest_function is called.
Come to think of it, with these changes, WaitSyncObject now has a meaningful purpose.

vita3k/modules/SceGxm/SceGxm.cpp
```diff
@@ -911,6 +910,8 @@ static void display_entry_thread(EmuEnvState &emuenv) {
         SceGxmSyncObject *old_sync = display_callback->old_sync.get(emuenv.mem);
         SceGxmSyncObject *new_sync = display_callback->new_sync.get(emuenv.mem);
 
+        // sceGxmDisplayQueueAddEntry waits for both buffers to complete
+        renderer::wishlist(old_sync, display_callback->old_sync_timestamp);
         renderer::wishlist(new_sync, display_callback->new_sync_timestamp);
```
Macdu, line 915: You can add a check to perform the second wishlist only if old_sync != new_sync

saturnsky: Okay. I didn’t add a comparative statement since I believed there was no functional difference between the two, however, considering it involves access to a lock, there may be a slight performance difference.

Macdu (approved):
lgtm, If possible, some testing would be nice to make sure this does not cause any regression.

nishinji:
No problems with games over 30.

PR #3411, opened 2024-10-23
[WIP] gui/themes: Add support of BGM.

Zangetsu38:


Macdu:
I mean, from my point of view, is it really useful to be able to play background music in the launcher?

Zangetsu38:
> I mean, from my point of view, is it really useful to be able to play background music in the launcher?

That just added possibility for that like on psvita

> I mean, from my point of view, is it really useful to be able to play background music in the launcher?

it's just a feature that is like on the psvita, I would have done it from the beginning of the implementation of the themes if I had known how to do it, and if the person doesn't like it, they just have to put the volume to 0, but personally I already love it 🗿

Macdu:
I guess sure but then:
- Make it disabled by default
- Do not use cubeb directly, use the api from audio/state.h which is an abstraction of cubeb/sdl audio.

Zangetsu38:
> I guess sure but then:
> 
>     * Make it disabled by default
> 
>     * Do not use cubeb directly, use the api from audio/state.h which is an abstraction of cubeb/sdl audio.

This is unfortunately not possible, just because of the way it works and the difference in processing compared to the sound of games.

Which is why I did that, the sound already starts at the very moment the emulator displays the applications home screen, the current sound is initialized after launching the game too, and I must necessarily process the entire sound, its reading state and volume completely separately from the sound of the game.

Also the audio processing by a game is completely different, it does it frame by frame in streaming, for me the audio file is decoded entirely at once and then read.

Macdu:
As I said, we have an audio abstraction layer for a reason. If there is something not possible right now in the abstraction layer, implement it there with support for both SDL and Cubeb.
Also decode audio only when needed instead of keeping the full decoded audio in RAM (including ingame) at all times.

Zangetsu38:
> As I said, we have an audio abstraction layer for a reason. If there is something not possible right now in the abstraction layer, implement it there with support for both SDL and Cubeb. Also decode audio only when needed instead of keeping the full decoded audio in RAM (including ingame) at all times.

ok but so as I said, it is currently not possible, and I don't see how to manage the decoding in stream, nor how to manage the pause and the volume separately from the sound of the game?

the volume and the pause status are managed for cubeb by the stream, and on the current code there is only one, if I made a code isolated from the rest it is not for nothing.

Is same for loop, if you looking code you can see it loop when music finish, current code no can doing that, so exept juste when is finished restart again all decoding and stream

vita3k/gui/src/themes.cpp
```diff
@@ -249,6 +255,10 @@ bool init_theme(GuiState &gui, EmuEnvState &emuenv, const std::string &content_i
                 if (!home_property.child("m_hostCollabo").child("m_iconFilePath").text().empty())
                     theme_icon_name["NPXS10026"] = home_property.child("m_hostCollabo").child("m_iconFilePath").text().as_string();
 
+                // Bgm theme
+                if (!home_property.child("m_bgmFilePath").text().empty())
+                    path_bgm = { "ux0", (fs::path("theme") / content_id / home_property.child("m_bgmFilePath").text().as_string()).string() };
```
bookmist, line 260:
```suggestion
                    path_bgm = { "ux0", fmt::format("theme/{}/{}",content_id,home_property.child("m_bgmFilePath").text().as_string()) };
```
No need to do double conversion string->path->concatenation->string

Macdu:
As I said, the reason we have an abstraction layer for audio output is for it to be used. Everything you want to do can be done with minimal modification to the audio abstraction layer.

Zangetsu38:
> As I said, the reason we have an abstraction layer for audio output is for it to be used. Everything you want to do can be done with minimal modification to the audio abstraction layer.

I repeat but therefore, problem, how can you manage the control of the pause and the volumes with also the loop of the bgm in a way totally separate from the sound of the game...

both need to have separate and totally isolated and separated control...

Macdu:
> > As I said, the reason we have an abstraction layer for audio output is for it to be used. Everything you want to do can be done with minimal modification to the audio abstraction layer.
> 
> I repeat but therefore, problem, how can you manage the control of the pause and the volumes with also the loop of the bgm in a way totally separate from the sound of the game...
> 
> both need to have separate and totally isolated and separated control...

The volume is controlled per stream and pausing can also be controlled per stream, I don't see a problem. 

Also the hardware itself (meaning the audio device) might not at all like being opened twice. It is specified for SDL2 that opening an audio device twice is not supported on all platforms.

Zangetsu38:
> > > As I said, the reason we have an abstraction layer for audio output is for it to be used. Everything you want to do can be done with minimal modification to the audio abstraction layer.
> > 
> > 
> > I repeat but therefore, problem, how can you manage the control of the pause and the volumes with also the loop of the bgm in a way totally separate from the sound of the game...
> > both need to have separate and totally isolated and separated control...
> 
> The volume is controlled per stream and pausing can also be controlled per stream, I don't see a problem.
> 
> Also the hardware itself (meaning the audio device) might not at all like being opened twice. It is specified for SDL2 that opening an audio device twice is not supported on all platforms.

I know that the volume and pause are controlled by the stream, that's what I said 🤔 that's literally why I say it's not possible, that the bgm code must be completely isolated from the game sound code with another stream used, that's exactly what I did, to have management that is independent

For double init, i not know 🤔 i used cubeb and no have problem

Issue #1756, opened 2022-05-24
Game patch support 


illusion0001:
Feature request for patching games and apps.

Basic features:
- Generate hash from loaded elf file
- Apply patch before elf is booted
- Write data to specified location in elf file. I.e 1, 2, 4 and 8 byte. `[ word, 0x810068b2, 0x02e0 ]`

RokkuDayo:
Crazy how this hasn't been implemented yet, it would open the gates for a ton of 60 FPS patches and other enhancements.

PR #3419, opened 2024-10-29
vita3k: introduce game patch files

SpikeHD:
Should address #1756 

Introduces a new folder, `patch`, that should contain text files ending in `.txt`. These `.txt` files should be named the corresponding titleID that the patch(es) apply to (for example, a patch file for "Dragon Quest Builders" may be called `PCSB00981.txt`). They can also be named things like `TITLEID_60FPS.txt`, to better organize patches, or `TITLEID1_TITLEID2.txt` if a patch applies to more than one game.

The patches themselves are pretty rudimentary byte-level replacements. The format was inspired by [Electry/VitaGrafix](https://github.com/Electry/VitaGrafix/) but does not come with the same creature comforts, lol. The format is as follows:
```
<seg>:<offset> <byte1> <byte2> <...byteN>
```

For example, to test, I have converted the "Hatsune Miku: Project Diva F 2nd" 60FPS patch from [VitaGrafix](https://github.com/Electry/VitaGrafixPatchlist/blob/bb3148070924b608fc18aee63ead91547907cbcb/patchlist.txt#L417) to be the following:

```
# 60 FPS
0:0xA994 0x01 0x20
0:0xAAF2 0xF7 0xEE 0x00 0x0A
```

...and it seems to work as expected! (excuse the sped up animations, I believe that's a problem with the original patch. There is a [pull request in the VitaGrafixPatchlist repo](https://github.com/Electry/VitaGrafixPatchlist/pull/136/files) that may address it, but I have not translated it over to test)

https://github.com/user-attachments/assets/0d96ded1-42e7-4388-9560-cbe96dcd3d0b

Some notes:
* This could definitely have more helpers, like VitaGrafix does. For example, being able to write `0:0xA994 t1_mov(0, 1)` instead of `0:0xA994 0x01 0x20`. I wanted to submit this PR first to ensure there was interest in it, before doing more complicated stuff for it.
* I'm a little rusty when it comes to C++ because I've been spending the last couple years writing Rust (lol). Please forgive any silly oversights.

Thank you to all that have built Vita3K up thus far ❤️


vita3k/kernel/src/load_self.cpp
```diff
@@ -633,6 +633,18 @@ SceUID load_self(KernelState &kernel, MemState &mem, const void *self, const std
...
 
+                for (auto &patch : patches) {
+                    if (seg_index == patch.seg) {
+                        for (int i = 0; i < patch.values.size(); i++) {
+                            auto value = patch.values[i];
+
+                            LOG_INFO("Applying patch at offset 0x{:08X} with value 0x{:08X}", patch.offset + (sizeof(value) * i), patch.values[i]);
+
+                            *(uint8_t *)(seg_ptr.get(mem) + patch.offset + (sizeof(value) * i)) = patch.values[i];
+                        }
```
bookmist, line 644: Why not to use `memcpy` instead of whole loop?

SpikeHD: Good catch, thanks!

SpikeHD:
Tested Project Diva F 2nd again, by converting the patch from the PR I mentioned:
```
# 60 FPS
0:0xA994 0x01 0x20
0:0xAAF2 0xF7 0xEE 0x00 0x0A
0:0x2896F6 0xDC 0xF8 0x80 0x83
0:0x2896FE 0xDC 0xF8 0x84 0x93
0:0x2849b4 0x20 0x22
```
...and results look good!

https://github.com/user-attachments/assets/f93e3a23-9ab8-465c-8601-abb44c02016d

vita3k/app/src/app_init.cpp
```diff
@@ -157,6 +158,7 @@ void init_paths(Root &root_paths) {
         root_paths.set_config_path(base_path);
         root_paths.set_shared_path(base_path);
         root_paths.set_cache_path(base_path / "cache" / "");
+        root_paths.set_patch_path(base_path / "patch" / "");
 
 #if defined(__linux__) && !defined(__ANDROID__) && !defined(__APPLE__)
         // XDG Data Dirs.
```
EXtremeExploit, line 164: There should also be a set_patch_path for linux (preferably right before the endif) that looks something like:
```cpp
root_paths.set_patch_path(shared_path / "patch" / "");
```
Linux's base_path isnt always writable

SpikeHD: Good call 👍

vita3k/patch/src/patch.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+    while ((pos = values.find(' ')) != std::string::npos) {
+        values_vec.push_back(strtoull(values.substr(0, pos).c_str(), nullptr, 16));
+        values.erase(0, pos + 1);
+    }
+
+    values_vec.push_back(strtoull(values.c_str(), nullptr, 16));
+
+    return Patch{ seg, offset, values_vec };
+}
```
Zangetsu38, line 81: Not forget add new line in end file

SpikeHD: My IDE is supposed to do that for me, but I guess it forgot 🙄

vita3k/patch/src/patch.cpp
```diff
@@ -0,0 +1,81 @@
...
+std::vector<Patch> get_patches(fs::path &path, const std::string &titleid) {
+    // Find a file in the path with the titleid
+    std::vector<Patch> patches;
+
+    LOG_INFO("Looking for patches for titleid {}", titleid);
+
+    for (auto &entry : fs::directory_iterator(path)) {
+        auto filename = entry.path().filename().string();
+        // Just in case users decide to use lowercase filenames
+        auto upper_filename = std::transform(filename.begin(), filename.end(), filename.begin(), ::toupper);
```
EXtremeExploit, line 31: IMO, the filenames of patches should be case sensitive just in case theres 2 files with different casing but for the same titleid and the user is dumb and cant figure out why their patches arent working. Also just a bit for simplification and make every patch's titleid be uppercase c:

I want to know what the others think of this tho

SpikeHD: My concern is that there is no way for the user to know that their patch isn't being read simply because it is in lower case. I think it's better to assume the user might be dumb and will used lower/mixed case names because there is no real reason for them to be case-sensitive anyways.

I can already imagine an issue being submitted with the title "Patches not applying", with the answer to their problem being "make the filename uppercase", which some might annoying. Or imagine someone creates a lowercase patch file, tries it, sees that it doesn't work, and then just goes "ugh, I guess the patch system doesn't work". This kind of thing is easily preventable by keeping this line, but I'd also be happy to hear what others think :)

saturnsky: I think case-insensitive implementation would be better.

1. It's difficult for end-users using Windows systems to understand that software behavior can differ due to filename case differences.
2. Even when multiple patches with only case differences exist on Linux systems, all those patches are applied regardless of case.

Among these two reasons, reason 1 is more important. I'm also curious about others' opinions on this issue.

antoniodesousa: I also agree that it should be case-insensitive. I don't know of any case where this could be a problem.

vita3k/patch/src/patch.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+        if (filename.find(titleid) != std::string::npos && filename.ends_with(".VPATCH")) {
```
nishinji, line 33: why `.VPATCH`?
Is this format used anywhere?
In VitaGrafix it is txt, so I think that would be better.

vita3k/patch/src/patch.cpp
EXtremeExploit: > 1. It's difficult for end-users using Windows systems to understand that software behavior can differ due to filename case differences.

Windows wouldnt be affected by the change as the filesystem is case-insesitive already, it would only change for linux/android users, but i see your point, it seems that i underestimated how utterly dumb the user can be especially android

vita3k/patch/src/patch.cpp
SpikeHD: I wanted it to be distinct because, while VitaGrafix uses `.txt`, this system isn't compatible with VitaGrafix patches, so it might be misleading (ie users may think they can just put the VitaGrafix `patchlist.txt` into the `patch` folder, which would not work).

If this isn't really a concern though, I am totally fine with just making them `.txt` 👍

SpikeHD: Changed to `.txt` since it probably isn't much of a concern

vita3k/modules/module_parent.cpp
```diff
@@ -220,7 +221,10 @@ SceUID load_module(EmuEnvState &emuenv, const std::string &module_path) {
         LOG_ERROR("Failed to read module file {}", module_path);
         return SCE_ERROR_ERRNO_ENOENT;
     }
-    SceUID module_id = load_self(emuenv.kernel, emuenv.mem, module_buffer.data(), module_path, emuenv.log_path);
+
+    const std::vector<Patch> patches = get_patches(emuenv.patch_path, emuenv.io.title_id);
```
Macdu, line 225: You are re-loading the patches from their file every time a module is being loaded. You should at least cache them or only load them when loading the eboot module considering it's the only one which can be patched right now.

vita3k/patch/src/patch.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+    return patches;
+}
+
+Patch parse_patch(const std::string &patch) {
+    // FORMAT: <seg>:<offset> <values>
+    // Example, equivalent to `t1_mov(0, 1)`:
+    // 0:0xA994 0x01 0x20
+    // Keep in mind that we are in little endian
+    uint8_t seg = atoi(patch.substr(0, patch.find(':')).c_str());
```
Macdu, line 61:
```suggestion
    uint8_t seg = std::stoi(patch.substr(0, patch.find(':'));
```

vita3k/patch/src/patch.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+    // Everything after the first colon, and before the first space, is the offset
+    uint32_t offset = strtoull(patch.substr(patch.find(':') + 1, patch.find(' ') - patch.find(':') - 1).c_str(), nullptr, 16);
```
Macdu, line 64:
```suggestion
    uint32_t offset = std::stoull(patch.substr(patch.find(':') + 1, patch.find(' ') - patch.find(':') - 1), nullptr, 16);
```

vita3k/patch/src/patch.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+    // All following values (separated by spaces) are the values to be written
+    std::string values = patch.substr(patch.find(' ') + 1);
+    std::vector<uint8_t> values_vec;
+
+    // Get all additional values separated by spaces
+    size_t pos = 0;
+
+    while ((pos = values.find(' ')) != std::string::npos) {
+        values_vec.push_back(strtoull(values.substr(0, pos).c_str(), nullptr, 16));
```
Macdu, line 74:
```suggestion
        values_vec.push_back(std::strtoull(values.substr(0, pos), nullptr, 16));
```

vita3k/patch/src/patch.cpp
```diff
@@ -0,0 +1,81 @@
...
+        values.erase(0, pos + 1);
+    }
+
+    values_vec.push_back(strtoull(values.c_str(), nullptr, 16));
```
Macdu, line 78:
```suggestion
    values_vec.push_back(std::stoull(values, nullptr, 16));
```

SpikeHD:
Hopefully everything has been addressed 👍

Macdu (approved):
lgtm

PR #3424, opened 2024-11-01
renderer: Fix incorrect timestamp offset in wait_cmd function

saturnsky:
The wait_cmd function was reading the timestamp value from an incorrect memory offset, which could lead to synchronization issues.


Macdu (approved):
thank you

PR #3428, opened 2024-11-02
kernel: more correct thread local storage emulation

bookmist:
also stub ThreadVfpException functions
used mostly for lle libkernel


Macdu (approved):
lgtm

PR #3432, opened 2024-11-05
modules/SceVshBridge: Stub vshDisplayRegisterFrameBufCallback .

Zangetsu38:
# About
- modules/SceVshBridge: Stub vshDisplayRegisterFrameBufCallback.

### DIFF
--- a/vita3k/modules/SceVshBridge/SceVshBridge.cpp
+++ b/vita3k/modules/SceVshBridge/SceVshBridge.cpp
@@ -17,6 +17,8 @@
 
 #include <module/module.h>
 
+#include "../SceDisplay/SceDisplay.h"
+
 EXPORT(int, _vshAppMgrAcInstGetAcdirParam) {
     return UNIMPLEMENTED();
 }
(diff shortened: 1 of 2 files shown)


vita3k/modules/SceVshBridge/SceVshBridge.cpp
```diff
@@ -746,7 +748,8 @@ EXPORT(int, vshSblSsIsDevelopmentMode) {
 }
 
 EXPORT(int, vshSblUtMgrHasComTestFlag) {
-    return UNIMPLEMENTED();
+    STUBBED("");
+    return 1;
```
Macdu, line 752: I don't remember, is this function expected to return 1 on retail firmware?

Zangetsu38: > I don't remember, is this function expected to return 1 on retail firmware?

Without it vsh no works, so i no know, pri cess have speaked about it too

Macdu: For example, looking at the disassembly, this function will return an error if the process executing it is not a system app.

Zangetsu38: I see, i need check it

Zangetsu38: > For example, looking at the disassembly, this function will return an error if the process executing it is not a system app.

ok, yes it return -0x7ffeffa8 if is not system app

and after it call kernel version of function and change value to 1 if is not on 0 🤔

Zangetsu38: My current pr opf decrypt fself in a fly get help for correct implement this pr

bookmist (approved):

Issue #3438, opened 2024-11-10
AddressSanitizer: heap-use-after-free


oltolm:
I get this error when click on "Update" on the game starting screen.

```
00000044	22:39:53	[19892] =================================================================	
00000045	22:39:53	[19892] ==19892==ERROR: AddressSanitizer: heap-use-after-free on address 0x11e0e5961af0 at pc 0x7ff6edbee2c9 bp 0x005c297e92c0 sp 0x005c297e9308	
00000046	22:39:53	[19892] READ of size 8 at 0x11e0e5961af0 thread T0	
00000047	22:39:53	[19892]     #0 0x7ff6edbee2c8 in TextureUpdateDescriptorSet(ImGui_VulkanState&, TextureState*) C:/src/Vita3K/vita3k/gui/src/imgui_impl_sdl_vulkan.cpp:200:18	
00000048	22:39:53	[19892]     #1 0x7ff6edbe6c19 in ImGui_ImplSdlVulkan_RenderDrawData(ImGui_VulkanState&) C:/src/Vita3K/vita3k/gui/src/imgui_impl_sdl_vulkan.cpp:559:17	
00000049	22:39:53	[19892]     #2 0x7ff6ed27023c in ImGui_ImplSdl_RenderDrawData(ImGui_State*) C:/src/Vita3K/vita3k/gui/src/imgui_impl_sdl.cpp:181:16	
00000050	22:39:53	[19892]     #3 0x7ff6ed28ffbf in gui::draw_end(GuiState&) C:/src/Vita3K/vita3k/gui/src/gui.cpp:751:5	
00000051	22:39:53	[19892]     #4 0x7ff6ece9c2a1 in SDL_main C:/src/Vita3K/vita3k/main.cpp:458:9	
00000052	22:39:53	[19892]     #5 0x7ff6f0b13150 in main_getcmdline (E:\build-vita3k-clang-debug\bin\Vita3K.exe+0x143c83150)	
00000053	22:39:53	[19892]     #6 0x7ff6eda24627 in main C:/M/B/src/mingw-w64/mingw-w64-crt/crt/crtexewin.c:67:10	
00000054	22:39:53	[19892]     #7 0x7ff6ece91312 in __tmainCRTStartup C:/M/B/src/mingw-w64/mingw-w64-crt/crt/crtexe.c:259:15	
[...]

PR #3439, opened 2024-11-10
gui: improve settings dialog

oltolm:
I fixed some problems with the settings dialogs.

1. As a user I don't know what the "Close" button does. Does it save & close or simply close? I renamed it to "Cancel" to make clear that is cancels the dialog.
2. "Save" does not close the dialog which means that I have to click "Close" afterwards. I changed it to close the the dialog. One click less.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ CMakeUserPresets.json
 compile_commands.json
 # AppImage builder
 appimage/linuxdeploy.AppImage
+/.gdbinit
--- a/vita3k/gui/src/settings_dialog.cpp
+++ b/vita3k/gui/src/settings_dialog.cpp
@@ -1216,10 +1216,11 @@ void draw_settings_dialog(GuiState &gui, EmuEnvState &emuenv) {
     ImGui::SameLine(0, 20.f * SCALE.x);
     const auto is_apply = !emuenv.io.app_path.empty() && (!is_custom_config || (emuenv.app_path == emuenv.io.app_path));
     const auto is_reboot = (emuenv.renderer->current_backend != emuenv.backend_renderer) || (config.resolution_multiplier != emuenv.cfg.current_config.resolution_multiplier);
-    if (ImGui::Button(is_apply ? (is_reboot ? lang.main_window["save_reboot"].c_str() : lang.main_window["save_apply"].c_str()) : common.save_data.save["title"].c_str(), BUTTON_SIZE)) {
+    if (ImGui::Button(is_apply ? (is_reboot ? lang.main_window["save_reboot"].c_str() : lang.main_window["save_apply"].c_str()) : lang.main_window["save_close"].c_str(), BUTTON_SIZE)) {
         save_config(gui, emuenv);
         if (is_apply)
             set_config(gui, emuenv, emuenv.io.app_path);
+        show_settings_dialog = false;
     }
     SetTooltipEx(lang.main_window["keep_changes"].c_str());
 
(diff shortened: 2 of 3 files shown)


nn9dev:
While you're at it, you should change the "Update" button to "Refresh" in LiveArea. I keep meaning to do it, but I've simply never gotten around to it. I'm fairly certain that the button is supposed to mimic the button with the arrows in a circle on a real Vita's LiveArea (which technically does ""update"" the content displayed, it's supposed to be for like re-checking announcements and other external data the LiveArea presents). Currently it serves as a "crash emulator" button, as well as confuses users basically daily (who think it's supposed to update the game's files/content).

If possible it would be nice to have this button linked to the same operation that the "Refresh" button on the Home Screen does, but I don't know what it's current behavior is supposed to be doing (since it just crashes).

Before posting this comment, I checked the [LiveArea manual](https://manuals.playstation.net/document/en/psvita/basic/screens.html) and I do see there IS an "Update" button that should display, but considering the object I'm talking about is almost always there (and there's no code to actually check Sony's servers for game/app updates), I don't believe these two to be the same button.

I apologize for hijacking. If you can't be bothered, I'll get around to it Eventually™

Zangetsu38:
> While you're at it, you should change the "Update" button to "Refresh" in LiveArea. I keep meaning to do it, but I've simply never gotten around to it. I'm fairly certain that the button is supposed to mimic the button with the arrows in a circle on a real Vita's LiveArea (which technically does ""update"" the content displayed, it's supposed to be for like re-checking announcements and other external data the LiveArea presents). Currently it serves as a "crash emulator" button, as well as confuses users basically daily (who think it's supposed to update the game's files/content).
> 
> If possible it would be nice to have this button linked to the same operation that the "Refresh" button on the Home Screen does, but I don't know what it's current behavior is supposed to be doing (since it just crashes).
> 
> Before posting this comment, I checked the [LiveArea manual](https://manuals.playstation.net/document/en/psvita/basic/screens.html) and I do see there IS an "Update" button that should display, but considering the object I'm talking about is almost always there (and there's no code to actually check Sony's servers for game/app updates), I don't believe these two to be the same button.
> 
> I apologize for hijacking. If you can't be bothered, I'll get around to it Eventually™

I already know that, is old things relate refresh content on live area screen, exemple after update game or install license, and for info, i already have all code for check update of game in sony server...
And crash is caused by vulkan, ogl no crash

oltolm:
BTW I posted the ASAN report for the update button crash #3438.

nishinji:
I agree with this change, but I think it needs to be changed from ‘save’ to ‘save and close’

oltolm:
> I agree with this change, but I think it needs to be changed from ‘save’ to ‘save and close’

I am OK with it, but then we can just as well call it "OK" like everyone else.

oltolm:
I changed it to "Save & Close".

scribam:
@oltolm Do you think you can improve this PR? I would like to have the use case below fixed with your changes:

**Preconditions**
Backend renderer is OpenGL

**Steps to reproduce**

1. Open the settings dialog
2. Change the backend renderer (OpenGL to Vulkan)
3. Use the cancel button
4. Open the settings dialog again

**Expected result**
The backend renderer is OpenGL

**Actual result**
The backend renderer is Vulkan

oltolm:
> @oltolm Do you think you can improve this PR? I would like to have the use case below fixed with your changes:

The graphics backend is not the only option that is applied immediately. Same is true for example for the audio backend and debug settings. There are probably others. I don't want to change it, it's too much work. We could mark them in GUI with an "*" that says something like "This option is applied immediately".

oltolm:
Can someone review my PR?

Zangetsu38:
Keep close, no have really sense said cancel here, you only close dialog ^^

oltolm:
I changed it back to "close".

bookmist (approved):

oltolm:
Can someone merge it?

PR #3441, opened 2024-11-12
modules/SceMotion: Improve motion support

saturnsky:
This PR aims to make sceMotion behavior more similar to PS Vita and implement some unimplemented functions.

I made the following modifications:

1. Basic Orientation is calculated according to gravity, identical to PS Vita.
2. `sceMotionGetState` function updates basicOrientation information when called.
3. Modified gyro-related behavior to match PS Vita values. This involved modifying the `UpdateRotation` function and `get_orientation` function.

I implemented the following functions:

1. `sceMotionReset` function updates quat values to their initial values. This function has been implemented.
2. `sceMotionRotateYaw` function does exactly what its name suggests.
3. Added some stubbed functions.

Testing across multiple games is required for merging this PR.

Since I tried to solve this by modifying only variable assignments while minimizing changes to the `UpdateOrientation` function's code, there might be some parts where code readability is slightly compromised. However, since the existing implementation was already swapping rad_gyro coordinates, this should be within acceptable bounds.

Also, this PR was written assuming equivalence between a DualShock4 controller lying face-up on a desk and a PS Vita device lying face-up on a desk. This assumption was made because in the existing implementation, these scenarios resulted in identical Accelerometer and Gyroscope values between the two devices. This aspect can be modified according to policy.

### DIFF
--- a/vita3k/modules/SceDriverUser/SceMotion.cpp
+++ b/vita3k/modules/SceDriverUser/SceMotion.cpp
@@ -24,9 +24,9 @@
 #include <util/tracy.h>
 TRACY_MODULE_NAME(SceMotion);
 
-EXPORT(int, sceMotionGetAngleThreshold) {
+EXPORT(SceFloat, sceMotionGetAngleThreshold) {
     TRACY_FUNC(sceMotionGetAngleThreshold);
-    return UNIMPLEMENTED();
+    return get_angle_threshold(emuenv.motion);
 }
 
 EXPORT(int, sceMotionGetBasicOrientation, SceFVector3 *basicOrientation) {
@@ -211,9 +211,10 @@ EXPORT(int, sceMotionRotateYaw, const float radians) {
     return 0;
 }
 
-EXPORT(int, sceMotionSetAngleThreshold, const float angle) {
+EXPORT(int, sceMotionSetAngleThreshold, SceFloat angle) {
     TRACY_FUNC(sceMotionSetAngleThreshold, angle);
-    return UNIMPLEMENTED();
+    set_angle_threshold(emuenv.motion, angle);
+    return 0;
 }
 
 EXPORT(int, sceMotionSetDeadband, SceBool setValue) {
--- a/vita3k/modules/SceDriverUser/SceMotion.cpp
+++ b/vita3k/modules/SceDriverUser/SceMotion.cpp
@@ -35,18 +35,18 @@ EXPORT(int, sceMotionGetBasicOrientation, SceFVector3 *basicOrientation) {
         return RET_ERROR(SCE_MOTION_ERROR_NULL_PARAMETER);
 
     std::lock_guard<std::mutex> guard(emuenv.motion.mutex);
-    Util::Quaternion quat = get_orientation(emuenv.motion);
+    SceFVector3 accelerometer = get_acceleration(emuenv.motion);
 
     *basicOrientation = { 0.f, 0.f, 0.f };
     // get the basic orientation, only one component is not zero and will be 1 or -1
-    // TODO: this is probably wrong
-    float max_val = std::max({ std::abs(quat.xyz.x), std::abs(quat.xyz.y), std::abs(quat.xyz.z) });
-    if (max_val == std::abs(quat.xyz.x)) {
-        basicOrientation->x = quat.xyz.x > 0.0f ? 1.0f : -1.0f;
-    } else if (max_val == std::abs(quat.xyz.y)) {
-        basicOrientation->y = quat.xyz.y > 0.0f ? 1.0f : -1.0f;
+    // The basic orientation is determined by gravity, so it uses the Accelerometer value.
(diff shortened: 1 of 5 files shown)


vita3k/modules/SceDriverUser/SceMotion.cpp
```diff
@@ -35,26 +35,26 @@ EXPORT(int, sceMotionGetBasicOrientation, SceFVector3 *basicOrientation) {
...
-        basicOrientation->y = quat.xyz.y > 0.0f ? 1.0f : -1.0f;
+    // The basic orientation is determined by gravity, so it uses the Accelerometer value.
+    float max_val = std::max({ std::abs(accelerometer.x), std::abs(accelerometer.y), std::abs(accelerometer.z) });
+    if (max_val == std::abs(accelerometer.x)) {
+        basicOrientation->x = accelerometer.x > 0.0f ? -1.0f : 1.0f;
+    } else if (max_val == std::abs(accelerometer.y)) {
+        basicOrientation->y = accelerometer.y > 0.0f ? -1.0f : 1.0f;
     } else {
-        basicOrientation->z = quat.xyz.z > 0.0f ? 1.0f : -1.0f;
+        basicOrientation->z = accelerometer.z > 0.0f ? -1.0f : 1.0f;
```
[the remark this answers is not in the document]
saturnsky: Since your point seems logically correct, I tested the suggested scenario and found that the current implementation is correct. While I can't understand Sony's implementation intent, the basic orientation depends solely on the accelerometer. (I hadn't tested the scenario you pointed out, and since your observation was very reasonable, I thought my implementation was wrong until I actually tested it)
When testing like that, the values indeed alternate between +x and -x (and +z during low acceleration periods).

saturnsky:
This PR includes deadband-related implementations worked on by @german77 and modifications to reflect the angle threshold in the basic orientation calculation process.
The current implementation is entirely based on real device testing.

The implementation included in the previous PR was a special implementation for when the angle threshold was 45 degrees, and according to subsequent additional testing, the results were not accurate for some numerical ranges. That has also been fixed.
Currently, when moving the PS Vita at slow speeds, except for issues with significant figures in trigonometric functions, the actual and calculated results are almost identical across all tested angle threshold ranges.

Considering the occasional errors that occur at high speeds, there must be some influence from the gravity vector, and this aspect is likely missing from the current implementation.

However, I suspect that either the quat values are not the input source, or they have a much smaller impact compared to the accelerometer values. If quat values had a significant impact on the results, calling sceMotionReset every frame should noticeably change the calculations since quat values would be constantly reset, but I couldn't observe any such effects.

There's also a possibility that it maintains a separate gravity vector calculation for basic orientation, but I haven't been able to verify these aspects yet.

I believe the current state is better than before, even with the errors that occur during high-speed movement. Since further research would significantly increase in difficulty and success is uncertain, I think it's appropriate to consolidate at the current stage.

vita3k/motion/src/motion_input.cpp
```diff
@@ -17,23 +22,34 @@ void MotionInput::SetPID(SceFloat new_kp, SceFloat new_ki, SceFloat new_kd) {
...
         gyro_bias = (gyro_bias * 0.9999f) + (gyroscope * 0.0001f);
     }
 
-    if (gyro.Length2() < gyro_threshold) {
-        gyro = {};
-    } else {
+    // Enable gyro compensation if gyro is active
+    if (gyro.Length() > gyro_deadband) {
         only_accelerometer = false;
     }
```
Macdu, line 54: only_accelerometer is to detect if there is a gyro at all (it is never set back to true), is it correct to compare it against gyro_deadband, which can be set by the game?

vita3k/motion/src/motion_input.cpp
```diff
@@ -17,23 +22,34 @@ void MotionInput::SetPID(SceFloat new_kp, SceFloat new_ki, SceFloat new_kd) {
...
+    gyro = gyroscope;
+
     if (bias_enabled) {
-        gyro = gyroscope - gyro_bias;
-    } else {
-        gyro = gyroscope;
+        gyro -= gyro_bias;
+    }
+
+    if (deadband_enabled && gyro.Length() < gyro_deadband) {
```
Macdu, line 38: I think it would be better to store gyro_deadband2 and use gyro.Length2 and gyro.Length (it is best not to use square roots if we don't need it)

vita3k/motion/src/motion_input.cpp
german77: When only accelerometer is enabled gyro will be emulated from accelerometer values. This is a counter measure when the motion source doesn't have a gyro sensor.

This flag is permanently disabled if any gyro input is detected. Returning to normal behavior. It can be changed to `if (gyro.Length2() > 0) `

saturnsky:
Based on Macdu's comments and german77's response, I modified the code and additionally improved the UpdateBasicOrientation function so that unit_xy is not calculated when unnecessary.

Macdu (approved):

PR #3445, opened 2024-11-18
app: Ignore XDG_DATA_DIRS for shared path

EXtremeExploit:
XDG_DATA_DIRS is used mainly to find default configuration files or system-wide configuration for apps, like a window manager having a default/system configuration file in /usr/share/WM/foo.
The idea for it is that it checks that path only if there is no user configuration, but because Vita3K doesnt have a default/system-wide configuration file AND the configuration is generated on first start, it really isnt needed + shared path shouldnt be system-wide as it needs to be read-write by the user

### DIFF
--- a/vita3k/app/src/app_init.cpp
+++ b/vita3k/app/src/app_init.cpp
@@ -212,15 +212,7 @@ void init_paths(Root &root_paths) {
         if (env_home != NULL)
             root_paths.set_shared_path(fs::path(env_home) / ".local/share" / app_name / "");
 
-        if (XDG_DATA_DIRS != NULL) {
-            auto env_paths = string_utils::split_string(XDG_DATA_DIRS, ':');
-            for (auto &i : env_paths) {
-                if (fs::exists(fs::path(i) / app_name)) {
-                    root_paths.set_shared_path(fs::path(i) / app_name / "");
-                    break;
-                }
-            }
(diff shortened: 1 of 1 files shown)


Macdu:
I guess this is going to change the shared path of some users without warnings, is this fine?

EXtremeExploit:
Like i said XDG_DATA_DIRS mainly points to directories like `/usr/share/` or `/usr/local/share`, they are used for specifying which directories do programs go to look like for default files, and i HIGHLY doubt someone would manually set XDG_DATA_DIRS to a writable directory.

Its somewhat common for people to set XDG_DATA_HOME and the others, but XDG_DATA_DIRS is the exception as the default files are provided by package managers, i dont think anyone actually sets data dirs manually to a different directory

In the super rare case in which an user gets affected the only affected directories would be custom languages, patches and texture dumps/replacements

XargonWan:
Greetings, yes, I am the culprit, the issue came from me :)
In a Flatpak or read-only OS/filesystem environment (such as RetroDECK) the emulator crashes as it's unable to create the patch folder.

```
terminate called after throwing an instance of 'boost::filesystem::filesystem_error'
  what():  boost::filesystem::create_directories: Read-only file system [system:30]: "/app/share/Vita3K/patch", "/app/share/Vita3K/patch"
 ```

Macdu (approved):

PR #3447, opened 2024-11-21
gui/controllers: align controller assignment behavior with Vita TV

saturnsky:
Feature Implementation:
I modified the controller assignment functionality to match the Vita TV.
On Vita TV, newly connected controllers are always sequentially assigned starting from port 1 to the next available number, and users can explicitly reassign controllers to any number from 1-4. No special handling is implemented when a disconnected controller is reconnected.

Bug Fix:
I fixed an issue where incorrect behavior occurred when a connected controller was disconnected. Previously, when a disconnected controller created a gap in port numbers, only iterated through the number of connected controllers, causing issues like being unable to check controllers with higher port numbers.

### DIFF
--- a/vita3k/ctrl/include/ctrl/state.h
+++ b/vita3k/ctrl/include/ctrl/state.h
@@ -35,7 +35,7 @@ typedef std::shared_ptr<_SDL_Haptic> HapticPtr;
 
 struct Controller {
     GameControllerPtr controller;
-    int port; // 1-based
+    int port; // SDL_GameController index
     bool has_accel;
     bool has_gyro;
     bool has_led;
@@ -65,5 +65,5 @@ struct CtrlState {
     SceCtrlPadInputMode input_mode_ext = SCE_CTRL_MODE_DIGITAL;
 
     // last vsync the data was read
-    uint64_t last_vcount[5] = {};
+    uint64_t last_vcount[5] = {}; // sceCtrl ports.
 };
--- a/vita3k/ctrl/include/ctrl/state.h
+++ b/vita3k/ctrl/include/ctrl/state.h
@@ -35,10 +35,11 @@ typedef std::shared_ptr<_SDL_Haptic> HapticPtr;
 
 struct Controller {
     GameControllerPtr controller;
-    int port;
+    int port; // 1-based
     bool has_accel;
     bool has_gyro;
     bool has_led;
+    const char *name;
 };
 
 struct ControllerBinding {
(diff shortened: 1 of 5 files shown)


vita3k/gui/src/controllers_dialog.cpp
```diff
@@ -186,22 +186,30 @@ static void swap_controller_ports(CtrlState &state, int source_port, int dest_po
...
-    // Swap controller names and motion support
-    std::swap(state.controllers_name[source_port], state.controllers_name[dest_port]);
-    std::swap(state.controllers_has_motion_support[source_port], state.controllers_has_motion_support[dest_port]);
+        // Swap controllers port and player index
+        SDL_GameControllerSetPlayerIndex(source_controller_it->second.controller.get(), dest_port);
+        SDL_GameControllerSetPlayerIndex(dest_controller_it->second.controller.get(), source_port);
+        std::swap(source_controller_it->second.port, dest_controller_it->second.port);
+
+        // Swap controller names and motion support
+        std::swap(state.free_ports[source_port], state.free_ports[dest_port]);
```
Macdu, line 211: This just swap two booleans, it has nothing to do with swapping controller names and motion support

vita3k/ctrl/include/ctrl/state.h
```diff
@@ -35,10 +35,11 @@ typedef std::shared_ptr<_SDL_Haptic> HapticPtr;
 
 struct Controller {
     GameControllerPtr controller;
-    int port;
+    int port; // 1-based
```
Macdu, line 38: Are you sure making it 1-base is the best? for allmost all uses you need to add/remove 1 to it.

vita3k/gui/src/controllers_dialog.cpp
saturnsky: I didn't update the existing comments while modifying the logic.

And looking at it again, since the free_ports values on both sides are identical when entering that part of the conditional statement, that code is unnecessary.
I think those lines remained because they were written before the conditional branching. I'll remove that part. Thank you.

vita3k/ctrl/include/ctrl/state.h
saturnsky: That part is legacy, and I left only the comments to prevent confusion for future developers after experiencing confusion myself with port numbers switching between 1-based and 0-based during development. The original 1-based port numbering seems to have been done to match it with sceCtrl.
However, as you mentioned, there's a lot of code getting complicated just to maintain that part. I think refactoring all parts including the existing code to be 0-based would help prevent future mistakes. I'll proceed with additional development to address this as well.

Macdu:
I haven't looked at the reason, but when trying your build and having the controller window opened, plugging in a controller always causes a nice segfault (null pointer dereference)

saturnsky:
Can I know in which environment the issue occurs?
I tried multiple connection/reconnection attempts with a DualShock 4 controller + Xbox 360 controller in:

* Windows 11 (10.0.26100.2605), Visual Studio 2022 (17.11.6) environment and
* Ubuntu 22.04.5 LTS, clang 14.0.0 environment

Since I couldn't reproduce the issue through these attempts, I'd like to confirm the specific environment.

Macdu:
Ok, after more testing, it turns out the crash was caused by SDL_GameControllerOpen returning NULL (I guess it can), it caused the controller name to also be null which when displayed caused the segfault.
This issue is not coming from your PR but I guess it would be nice to fix this at the same time.

saturnsky:
Theoretically, if a joystick is disconnected after calling SDL_NumJoysticks but before calling SDL_GameControllerOpen or SDL_GameControllerNameForIndex, it seems to could return NULL.
I failed to reproduce this issue, but I don't see any other functions that could cause problems. I have fixed this issue for now.

Macdu (approved):
Thank you!

PR #3449, opened 2024-11-26
touch: Validate all input params and simplify code

german77:
This is a quick pass to the current touch code. I don't expect any games be affected by these changes.

- Ensures all input parameters are properly validated to return the same error code as the actual console
- Removes a bit of duplication code in `sceTouchGetPanelInfo` to simplify reading.
- ~~Partially implements `sceTouchGetPixelDensity` with `GetHardwareInfo` calls~~. In reality these values are stored in a global set at module init but vita3k doesn't seem to have any proper initialization for modules.

### DIFF
--- a/vita3k/modules/SceTouch/SceTouch.cpp
+++ b/vita3k/modules/SceTouch/SceTouch.cpp
@@ -51,7 +51,7 @@ EXPORT(int, sceTouchDisableTouchForce, SceUInt32 port) {
         return RET_ERROR(SCE_TOUCH_ERROR_INVALID_ARG);
     }
     touch_set_force_mode(port, false);
-    return 0;
+    return SCE_TOUCH_OK;
 }
 
 EXPORT(int, sceTouchDisableTouchForceExt) {
@@ -70,7 +70,7 @@ EXPORT(int, sceTouchEnableTouchForce, SceUInt32 port) {
         return RET_ERROR(SCE_TOUCH_ERROR_INVALID_ARG);
     }
     touch_set_force_mode(port, true);
-    return 0;
+    return SCE_TOUCH_OK;
 }
 
 EXPORT(int, sceTouchEnableTouchForceExt) {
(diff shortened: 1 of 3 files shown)


vita3k/modules/SceVshBridge/SceVshBridge.cpp
```diff
@@ -314,7 +314,8 @@ EXPORT(int, _vshSysconCtrlManualChargeMode) {
 }
 
 EXPORT(int, _vshSysconGetHardwareInfo) {
```
Macdu, line 316: Looking at the disassembly for _vshSysconGetHardwareInfo, it has signature `int _vshSysconGetHardwareInfo(char info[4])` and returns 0 in case of success (the hardware info is stored in info). Also I'm not really sure if its worth implementing this function if the hardware info returned is completely opaque.

german77: `GetHardwareInfo` didn't had any argument in this instance. Although this is the kernel call.
```
      uVar2 = SceSysconForDriver::GetHardwareInfo_CBD6D8BC();
      if (((uVar2 & 0xff0000) == 0x300000) || ((uVar2 & 0xff0000) == 0x310000)) {
        PixelDensity1 = 22.0;
        FLOAT_810073e4 = 22.0;
        PixelDensity2 = 22.0;
        FLOAT_810074d8 = 22.0;
      }
```

I don't have any necessity to implement it at the moment. So I'm going to just revert it in the meantime.

Macdu: Ah ok, the kernel space call (`ksceSysconGetHardwareInfo`) directly returns the hardware info while the user one `vshSysconGetHardwareInfo` calls the former but returns an error code and stores the hardware info in an address given in parameters.

Macdu (approved):

PR #3450, opened 2024-11-30
gui: add keyboard navigation support for menus

saturnsky:
This is a PR for implementing keyboard-based menu navigation. It's still in WIP status as there are several concerns to address. I plan to implement joystick-based menu control after this work. The main implementation goal is to eliminate the need for touch menu control on devices like the SteamDeck. I'd like to hear others' thoughts on these issues and the current implementation direction.

Current implementation:
Pressing Alt moves focus to the menu, after which the menu can be controlled via keyboard. While the menu has focus, home_screen control is blocked. Menu state is shared between mouse and keyboard menu access. When moving between menu items left and right, I set having the menu expanded as the default state.

Issue 1: I couldn't find a way to achieve the goal without using imgui_internal. This may result in higher version dependency on imgui, which I'm not happy about.
Issue 2: The is_any_dialog_open function... is there really no other way? I couldn't find an alternative in imgui.

Better implementation might be possible by directly accessing ImGuiContext, but I wanted to avoid this at all costs. While both depend on imgui_internal, I judged that function access versus direct member variable access would have different maintenance costs going forward.
While the functionality works as intended, I'm not satisfied with the current implementation for the reasons mentioned above. However, if there seems to be no other implementation method available, this appears to be unavoidable.

Points to consider:

Using Alt for menu access was arbitrarily decided based on typical Windows programs, but the relevant key will differ by OS, and there could be conflicts with user-set keymaps. This needs more consideration.
For joystick control, I'm thinking of using the PS button (Xbox button) in the home menu to move focus to the menu. This should guarantee no conflicts. (@Zangetsu38 Did you know could there be any conflict cases?)

Tasks to proceed with in this PR later:

1. ~~Game controls are currently not blocked. Game controls should also be blocked when the menu has focus.~~ Updated
2. ~~Joystick control has not been implemented~~

### DIFF
--- a/vita3k/gui/src/gui.cpp
+++ b/vita3k/gui/src/gui.cpp
@@ -711,6 +711,8 @@ void pre_init(GuiState &gui, EmuEnvState &emuenv) {
     init_font(gui, emuenv);
     lang::init_lang(gui.lang, emuenv);
 
+    auto &io = ImGui::GetIO();
+
     bool result = ImGui_ImplSdl_CreateDeviceObjects(gui.imgui_state.get());
     assert(result);
 }
--- a/vita3k/interface.cpp
+++ b/vita3k/interface.cpp
@@ -568,6 +568,7 @@ static void take_screenshot(EmuEnvState &emuenv) {
 bool handle_events(EmuEnvState &emuenv, GuiState &gui) {
     refresh_controllers(emuenv.ctrl, emuenv);
     const auto allow_switch_state = !emuenv.io.title_id.empty() && !gui.vita_area.app_close && !gui.vita_area.home_screen && !gui.vita_area.user_management && !gui.configuration_menu.custom_settings_dialog && !gui.configuration_menu.settings_dialog && !gui.controls_menu.controls_dialog && gui::get_sys_apps_state(gui);
+    auto &io = ImGui::GetIO();
 
     const auto ui_navigation = [&emuenv, &gui, allow_switch_state](const uint32_t sce_ctrl_btn) {
         if (gui.vita_area.app_close) {
@@ -647,6 +648,9 @@ bool handle_events(EmuEnvState &emuenv, GuiState &gui) {
         default: break;
         }
     };
+
+    for (int i = 0; i < IM_ARRAYSIZE(io.NavInputs); i++)
+        io.NavInputs[i] = 0.0f;
 
     SDL_Event event;
     while (SDL_PollEvent(&event)) {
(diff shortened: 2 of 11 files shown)


Zangetsu38:
I had already tried in the past with this same function (ImGuiConfigFlags_NavEnableKeyboard), but elsewhere, and activated all the time, both for the keyboard and for the controller, I gave up because it was seriously crappy...
And it had nothing to do with the control on psvita, it also takes the focus on the window... it's just atrocious

and yes, you just can't use the PS key, it's already allocated to quit/return to games, like on the vita

And really, I hate their implementation of how the interface is controlled, it's a horror...

it's for dca that much later I did it myself manually to have a control As I wanted, as close as possible to the psvita itself

saturnsky:
I agree. After modifying that option myself, I understood why you didn't use ImGuiConfigFlags_NavEnableKeyboard in the previous implementation.
I also think the default behavior of ImGuiConfigFlags_NavEnableKeyboard quite difficult to accept, so I made it conditionally turn On/Off and overrode some of its detailed operations. In this process, I had to use imgui_internal. I don't prefer using such internal functions, but I couldn't find an alternative.
Thank you for letting me know about the scenario of using the PS key. I think I can consider Options key and others as alternative candidates. I'll think about the details of this part after finishing the keyboard-based work first.

saturnsky:
I've implemented controller support to better evaluate the actual experience.
In the live area, access the menu with the Start key. During gameplay, there's no hotkey for accessing the menu (though it's still possible to navigate with the controller once the menu is focused). This was an intentional choice to prevent gameplay interruption.
While it feels decent to me at the moment, I need to think more about whether this key layout is truly the right approach.

saturnsky:
I added a feature to prevent returning to the home screen while navigating the menus.

vita3k/gui/src/gui.cpp
```diff
@@ -711,6 +711,8 @@ void pre_init(GuiState &gui, EmuEnvState &emuenv) {
     init_font(gui, emuenv);
     lang::init_lang(gui.lang, emuenv);
 
+    auto &io = ImGui::GetIO();
+
```
Macdu, line 715: Is there a reason for this (or if GetIO initialises some stuff, there is no reason to assign its output to a variable)

vita3k/interface.cpp
```diff
@@ -648,6 +649,9 @@ bool handle_events(EmuEnvState &emuenv, GuiState &gui) {
         }
     };
 
+    for (int i = 0; i < IM_ARRAYSIZE(io.NavInputs); i++)
+        io.NavInputs[i] = 0.0f;
```
Macdu, line 653: NavInput has the following comment:
[LEGACY] Since 1.88, NavInputs[] was removed. Backends from 1.60 to 1.86 won't build. Feed gamepad inputs via io.AddKeyEvent() and ImGuiKey_GamepadXXX enums.

Are you sure about this? I am fine with using internal ImGui functions but we should at least not use deprecated/about to be removed ones.

vita3k/gui/src/gui.cpp
saturnsky: This is my mistake. That part remained when I was removing unnecessary code after adding different code to that section and confirming it was not needed.

saturnsky:
Due to compatibility issues in the current version, io.AddKeyEvent cannot be used, but the point that the current implementation will become obsolete according to ImGui's policy is correct.
Following @Macdu's comment, I will rewrite a PR with equivalent functionality based on PR #3475 to align with ImGui's policy. Therefore, I will close the current PR for now.

PR #3455, opened 2024-12-02
ngs/player: ngs/player: Prevent crash by rounding playback rate up.

Zangetsu38:
# About
 - ngs/player: Prevent crash by rounding playback rate up.
Some Unity games set very low playback frequencies (e.g., 0.xx), causing FFmpeg to crash.
- Improve validation for playback_scalar and playback_frequency.
Replaced redundant checks for playback_scalar and playback_frequency with dedicated lambda functions for better readability and maintainability.
- Adjusted the validation logic to properly handle edge cases:
Allowed playback_scalar values between 0 and 10.
Allowed playback_frequency values between 0 and 192000.

fix crash of Looney Tunes Galactic Sports, Spongebob, Stick it to the man, etc....
credit to @bookmist for help

### DIFF
--- a/vita3k/ngs/src/modules/player.cpp
+++ b/vita3k/ngs/src/modules/player.cpp
@@ -51,19 +51,24 @@ void PlayerModule::on_param_change(const MemState &mem, ModuleData &data) {
     const SceNgsPlayerParams *old_params = reinterpret_cast<SceNgsPlayerParams *>(data.last_info.data());
     SceNgsPlayerParams *new_params = static_cast<SceNgsPlayerParams *>(data.info.data.get(mem));
 
-    if (isnan(new_params->playback_scalar) || new_params->playback_scalar <= 0) {
+    // check for invalid playback values
+    const auto is_invalid_playback_value = [](const float playback_value, const float max_value) {
+        return isnan(playback_value) || (playback_value < 0.f) || (playback_value > max_value);
+    };
+
+    if (is_invalid_playback_value(new_params->playback_scalar, 10.f)) {
         new_params->playback_scalar = old_params->playback_scalar;
         LOG_ERROR_ONCE("Invalid playback rate scaling.");
-        if (isnan(new_params->playback_scalar) || new_params->playback_scalar <= 0) {
+        if (is_invalid_playback_value(new_params->playback_scalar, 10.f)) {
             new_params->playback_scalar = 1.0;
         }
     }
 
-    if (isnan(new_params->playback_frequency) || new_params->playback_frequency <= 0) {
+    if (is_invalid_playback_value(new_params->playback_frequency, 192000.f)) {
         new_params->playback_frequency = old_params->playback_frequency;
         LOG_ERROR_ONCE("Invalid playback frequency.");
-        if (isnan(new_params->playback_frequency) || new_params->playback_frequency <= 0) {
-            new_params->playback_frequency = 48000.0;
+        if (is_invalid_playback_value(new_params->playback_frequency, 192000.f)) {
+            new_params->playback_frequency = 48000.f;
         }
     }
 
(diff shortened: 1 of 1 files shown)


Macdu:
I'm a bit concerned about fixes like this. A game never "willingly" set a playback frequency to something < 1.0 (or to a NaN like the previous check was doing), what this means is that some part of the code is not properly initializing (or overwriting) some structure related to this, and this is just a way to try to minimize the consequences instead of fixing the cause.

Zangetsu38:
> I'm a bit concerned about fixes like this. A game never "willingly" set a playback frequency to something < 1.0 (or to a NaN like the previous check was doing), what this means is that some part of the code is not properly initializing (or overwriting) some structure related to this, and this is just a way to try to minimize the consequences instead of fixing the cause.

is same of Loony toones and spongebob, and laybe lore unity game, it freeze durring loading ingame render, and get spammed value like that, but value is all time same
image

Zangetsu38:
stick it to the man have same, b t durring menu, spam value like that, but sound works
my pr fix very lot game of unity
image

Macdu:
Just make the change in on_param_change, like the nan check workaround, there is no reason for the player to have to check any of this, it should be able to assume its input is coherent.

Macdu:
> > I'm a bit concerned about fixes like this. A game never "willingly" set a playback frequency to something < 1.0 (or to a NaN like the previous check was doing), what this means is that some part of the code is not properly initializing (or overwriting) some structure related to this, and this is just a way to try to minimize the consequences instead of fixing the cause.
> 
> is same of Loony toones and spongebob, and maybe more unity game, it freeze durring loading ingame render, and get spammed value like that, but value is all time same image

Yes, as I said, no game is ever going to ask some audio to be processed at 0.04 hertz, it means it is either reading not properly set stack data or at some point the wrong value was retrieved for some multiplication/addition.

Zangetsu38:
> Just make the change in on_param_change, like the nan check workaround, there is no reason for the player to have to check any of this, it should be able to assume its input is coherent.

I no really like set it here, this fonction check if value is invalide, and like bookmist have explain me, 0 value is valid for ngs, problem here is not this value, problem is it want resampling because src_sample_rate and sample_rate is diferent so it try resemple stream...
And ffmeg said NOP with 0 value lol
But scaling is on 1.f

So my code is not wrong, if for ngs 0 valid is valide value, is normal condition onoy for ngs no supoort 0 value for that

Macdu:
I can see ngs having a special case  when specifying a frequency of 0 Hertz. However a game specifying a frequency a frequency of 0.03 is definitely not expected (or if it is feel free to show me a homebrew with ngs reacting correctly when specifying such a frequency) and indicates an issue with the game prior to calling this function. Therefore please add the check when parsing the frequency (check which shouldn't be needed, same as checking for a NaN frequency)

Zangetsu38:
> I can see ngs having a special case when specifying a frequency of 0 Hertz. However a game specifying a frequency a frequency of 0.03 is definitely not expected (or if it is feel free to show me a homebrew with ngs reacting correctly when specifying such a frequency) and indicates an issue with the game prior to calling this function. Therefore please add the check when parsing the frequency (check which shouldn't be needed, same as checking for a NaN frequency)

Is frequency used by some unity game, i no sur any hb using it
A’d again for ngs 0 frequancy is full valid, confirmed by bookmist, my code is fine, problem is only relate vita3k because it try resempling it 😐, psvita don't

Macdu:
> > I can see ngs having a special case when specifying a frequency of 0 Hertz. However a game specifying a frequency a frequency of 0.03 is definitely not expected (or if it is feel free to show me a homebrew with ngs reacting correctly when specifying such a frequency) and indicates an issue with the game prior to calling this function. Therefore please add the check when parsing the frequency (check which shouldn't be needed, same as checking for a NaN frequency)
> 
> Is frequency used by some unity game, i no sur any hb using it A’d again for ngs 0 frequancy is full valid, confirmed by bookmist, my code is fine, problem is only relate vita3k because it try resempling it 😐, psvita don't

I have no issue with a sample rate of 0 Hertz being handled as a special case, looks like it can happen. I have an issue about handling specially a sample rate of 0.05 Hertz (or NaN hertz) because it means we have an issue upfront, these should be checked as soon as possible (so in the on_param_change), with a comment saying these checks are a fix and not something really expected.

Zangetsu38:
code changed pushed, request review

bookmist (approved):

PR #3456, opened 2024-12-04
modules/SceAppMgr: fix Sfo loading and improve AppParamGetString.

Zangetsu38:
# About
- modules/SceAppMgr: fix Sfo loading and improve AppParamGetString.
- packages/sfo/get data by id: Set key corresponding id.
- interface: load sfo in boot app.

Fix call of getstring used by adhoc mode in app and on vsh too

### DIFF
--- a/vita3k/modules/SceAppMgr/SceAppMgr.h
+++ b/vita3k/modules/SceAppMgr/SceAppMgr.h
@@ -20,7 +20,8 @@
 #include <module/module.h>
 
 enum SceAppMgrErrorCode : uint32_t {
-    SCE_APPMGR_ERROR_INVALID_PARAMETER = 0x80800001, //!< Invalid parameter
+    SCE_APPMGR_ERROR_INVALID_PARAMETER = 0x80800001, //!< Invalid parameter (NULL)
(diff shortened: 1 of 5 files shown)


vita3k/modules/SceAppMgr/SceAppMgr.h
```diff
@@ -53,8 +54,15 @@ typedef struct SceAppMgrLoadExecOptParam {
     int reserved[256 / 4]; //!< Reserved area
 } SceAppMgrLoadExecOptParam;
 
+typedef struct sceAppMgrAppParamGetStringOptParam { // size is 0x10 on 3.60
```
Macdu, line 57: This comment is a bit misleading, it makes you think you are talking about the size field of sceAppMgrAppParamGetStringOptParam and not the size of the structure.

Zangetsu38: Is @bookmist  comment

Macdu: Don't use size with a field named size right below, you can use anything else (like this structure is expected to be 0x10 bytes on 3.60 or whatever)

Zangetsu38: Adcorded with bookmist, i have just removed comment.

bookmist (approved):

PR #3461, opened 2024-12-10
motion: Return proper error codes

german77:
Makes sure we return the same error codes as the console

### DIFF
--- a/vita3k/modules/SceDriverUser/SceMotion.cpp
+++ b/vita3k/modules/SceDriverUser/SceMotion.cpp
@@ -31,15 +31,22 @@ EXPORT(SceFloat, sceMotionGetAngleThreshold) {
 
 EXPORT(int, sceMotionGetBasicOrientation, SceFVector3 *basicOrientation) {
     TRACY_FUNC(sceMotionGetBasicOrientation, basicOrientation);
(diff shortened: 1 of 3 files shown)


Macdu:
As far as I understand, it's not really possible for the motion library to be uninitialized, does it make sense to check for it and return errors then?

german77:
This module is meant to be loaded and unloaded as needed. But this emulator doesn't have proper initialization for modules so it's a hardcoded to always initialized. They can be removed as it doesn't serve a purpose at the moment.

lohyft:
下次更新是几时？

Macdu:
> This module is meant to be loaded and unloaded as needed. But this emulator doesn't have proper initialization for modules so it's a hardcoded to always initialized. They can be removed as it doesn't serve a purpose at the moment.

The motion module is part of the modules that are preloaded before the application starts (it doesn't even have a module ID). As such it will always be loaded. NOT_INIT kind of errors happen when a module has a init function, which if there is an internal one is called when the module is loaded.

Also Vita3K supports module loading (and unloading) perfectly fine. But if you were to call a function from an unloaded module, you would get a nice crash on real hardware.

Macdu (approved):

PR #3464, opened 2024-12-14
gui: Remove High DPI support on Linux

saturnsky:
Currently, there are various HighDPI-related issues in Linux.
I want to disable High DPI with this PR until High DPI is properly implemented in Linux.
This will solve the problem where Vita3K couldn't be used in some Linux desktop environments.

Here are the detailed issues:

1. In Linux environments, the SDL_GetDisplayDPI function returns the monitor's physical DPI.
   * On some displays where the monitor's length is not provided, this value becomes 0. In such cases, Vita3K's screen does not display properly.
   * On some HighDPI displays, the physical DPI is too high, causing Vita3K's window to exceed 100% of the screen. In this case, it becomes impossible to use Vita3K normally.
2. Even when native DPI scaling is supported in the desktop environment, there are cases where Vita3K's screen doesn't render properly.
   * In Wayland environments, depending on the settings, Windows and other elements can be drawn at sizes larger than the given parameters. In this case, the window size is rendered larger, and text within the window can also be expanded according to the DPI multiplier, causing the rendering results to overflow on the screen.
   * In other words, rendering issues always occur when the detected dpi_scale value is not 1.

I considered only stopping the operation in certain situations, thinking there might be environments where High DPI implementation could be used properly, but testing showed that such environments barely exist.

### DIFF
--- a/vita3k/app/src/app_init.cpp
+++ b/vita3k/app/src/app_init.cpp
@@ -297,27 +297,11 @@ bool init(EmuEnvState &state, Config &cfg, const Root &root_paths) {
         state.display.fullscreen = true;
         window_type |= SDL_WINDOW_FULLSCREEN_DESKTOP;
     }
-#if defined(_WIN32) || defined(__linux__)
-    const auto isSteamDeck = []() {
-#ifdef __linux__
-        std::ifstream file("/etc/os-release");
-        if (file.is_open()) {
-            std::string line;
-            while (std::getline(file, line)) {
-                if (line.find("VARIANT_ID=steamdeck") != std::string::npos)
-                    return true;
-            }
-        }
-#endif
-        return false;
-    };
-
-    if (!isSteamDeck()) {
-        float ddpi, hdpi, vdpi;
-        SDL_GetDisplayDPI(0, &ddpi, &hdpi, &vdpi);
-        window_type |= SDL_WINDOW_ALLOW_HIGHDPI;
-        state.dpi_scale = ddpi / 96;
-    }
+#if defined(_WIN32)
+    float ddpi, hdpi, vdpi;
+    SDL_GetDisplayDPI(0, &ddpi, &hdpi, &vdpi);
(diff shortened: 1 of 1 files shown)


Zangetsu38:
no sur is good idea, whjen i have set it correctly, lot user have very apreciate it

saturnsky:
Currently, the DPI scaling feature only works as intended when the following conditions are met:

1. In Linux desktop environments with insufficient OS-level DPI scaling (For Linux desktops that properly provide DPI scaling, the Display scale must be set to 100% for the feature to work as currently implemented)
2. Using a High DPI display
3. When that display correctly returns the physical size of the display

If any of these conditions are not met, it causes various issues.
Furthermore, there aren't many user environments that satisfy all these conditions, and more significantly, it's practically difficult to determine whether a user's environment is capable of properly functioning with the current DPI scaling implementation. For starters, just obtaining the Display Scale settings from the user's system is not a simple task due to the diversity of Linux desktop environments.

Therefore, with this PR, I aim to proactively resolve issues for people who are experiencing various problems, such as being unable to run Vita3K due to the current implementation. I understand that this will result in a degraded user experience for some people for a while. However, I believe this step is necessary.

Don't worry. I am working on implementing DPI scaling properly, and High DPI-related features will be reimplemented with compatibility in mind.
And if users don't want to experience a decline in their playing experience during this period, they can continue using the existing version for a while. However, I believe that users who find it difficult to play games normally due to the current implementation need a faster patch before the High DPI reimplementation is completed.

Zangetsu38:
hmm no sur all user apreciate removing it here :/
maybe keep user choice ?

Macdu:
I agree with Zangetsu here that just removing the High DPI scaling without any alternative might not be the best idea, the base font size looks really bad on a high DPI screen.

saturnsky:
I have reimplemented the entire HighDPI and created PR #3473. This PR is no longer necessary.

PR #3473, opened 2024-12-20
gui/highdpi: Overall Reimplementation of HighDPI Functionality

saturnsky:
Due to incomplete implementation of the existing HighDPI feature, the implementation has been thoroughly revised. This implementation affects all OS platforms.

1. When SDL supports system-level DPI scale implementation (=Mac, Windows, Wayland, and other unverified systems), DPI scaling is implemented using this feature.
2. Otherwise, if manual_dpi_scale value is provided, Vita3K will emulate HighDPI internally (operates similarly to Windows support before this PR).
   1. Currently, x11 is implemented for Linux support.
   2. If needed later, additional implementations can be easily added using the same method as x11 support.
3. The actual renderable area and OS-handled window size are managed separately.
4. Font sizes are prepared in 1x, 1.5x, and 2x sizes, and fonts are dynamically changed according to changes in the rendering area size.
   1. This improvement helps enhance rendering quality not only for HighDPI systems but also when window size is increased on regular systems.
5. When font texture preparation exceeds system GPU texture size limits, the maximum font size is reduced to ensure proper operation in environments with low-end graphics cards.

Additional Trial and Error Notes
1. Using a single large font size and downscaling it isn't viable because ImGUI's font downscaling algorithm is not excellent, resulting in poor font quality in low-resolution environments.
2. Attempted implementation through MipMap to improve this issue - while font rendering quality was excellent, the MipMap generation process caused white outlines around text due to interference from other character areas. Creating sufficient margins between characters to solve this would result in impractically large texture sizes.
3. While Wayland environment supports system-level HighDPI implementation, the Wayland-ImGUI-SDL-Vulkan environment currently has issues, making the software work properly only with OpenGL. This is not my bug.
4. When SDL improves to properly support x11 display scale in the future, manual_dpi_scale supply should be discontinued. Otherwise, users will experience DPI configuration at the square of system settings (400% for 200% settings).

Additionally, this patch fixes an issue where Vita3K wouldn't run on Linux environments that incorrectly report display sizes. This problem occurred because SDL_GetDisplayDPI function is implemented differently across OS platforms - while Windows returns system settings, Linux returns hardware configuration values. Some faulty displays return 0 inches for diagonal size.
Fortunately, this issue could be reproduced when running Vita3K in virtual machines like VMWare. Otherwise, identifying the cause would have been more difficult.

### DIFF
--- a/vita3k/app/src/app_init.cpp
+++ b/vita3k/app/src/app_init.cpp
@@ -55,6 +55,10 @@ void update_viewport(EmuEnvState &state) {
     int w = 0;
     int h = 0;
 
+    SDL_GetWindowSize(state.window.get(), &w, &h);
+    state.window_size.x = w;
+    state.window_size.y = h;
+
     switch (state.renderer->current_backend) {
     case renderer::Backend::OpenGL:
         SDL_GL_GetDrawableSize(state.window.get(), &w, &h);
@@ -71,6 +75,11 @@ void update_viewport(EmuEnvState &state) {
 
     state.drawable_size.x = w;
     state.drawable_size.y = h;
+
+    state.dpi_scale = static_cast<float>(state.drawable_size.x) / state.window_size.x;
+
+    state.res_width_dpi_scale = static_cast<uint32_t>(DEFAULT_RES_WIDTH * state.dpi_scale);
+    state.res_height_dpi_scale = static_cast<uint32_t>(DEFAULT_RES_HEIGHT * state.dpi_scale);
 
     if (h > 0) {
         const float window_aspect = static_cast<float>(w) / h;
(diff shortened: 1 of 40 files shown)


Zangetsu38:
you can squash all commit in one, all relate same.

saturnsky:
I intentionally organized the commit units as they are now. (There were about 16 commits before organizing)
Since there were too many modified files, and many of them resulted from renaming, combining these commits into one would make it very cumbersome for future developers to review the changes. Therefore, I determined it was necessary to divide them into meaningful units.

vita3k/app/src/app_init.cpp
```diff
@@ -229,6 +279,53 @@ void init_paths(Root &root_paths) {
     fs::create_directories(root_paths.get_patch_path());
 }
 
+#ifdef __LINUX__
+static float fetch_x11_display_dpi() {
```
scribam, line 283: :thinking: SDL3 does more than just retrieve a value from "Xft.dpi", you can see there are other potential values to check: https://github.com/libsdl-org/SDL/blob/c6b5c21aaf66dc8fb4214e7bebb1258084abfae6/src/video/x11/SDL_x11modes.c#L49-L119
Do you think if it will be worth it to do the same logic?

saturnsky: Thank you. Since there's no guarantee that we can obtain the Xft.dpi value from the system, having more fallback options is better. The fallbacks included in SDL3's implementation should be quite reliable. I will add fallbacks with reference to the cited code.

saturnsky: After review, getting values from XSETTINGS seems too costly compared to its benefits. We would need to either add a dependency on libxsettings-client or implement direct value access, but the benefits don't seem worth adding these for this purpose.
While GDK_SCALE is also not widely used currently, it's not a bad option since it has minimal dependencies. Therefore, I only added GDK_SCALE as a fallback.

saturnsky:
I have rebased to resolve the conflicts, making it ready for merge.

bookmist (approved):

PR #3475, opened 2024-12-21
gui/imgui: Backport imgi_impl_sdl from vita3k-android

saturnsky:
Backported imgui_impl_sdl.cpp from the vita3k-android project to the current project to avoid using features that will become obsolete in ImGui during subsequent development.


Macdu:
Could you add me as a co-author and remove the android part (ifdef Android and ImGui_ImplSDL2_HandleTouch) ?

saturnsky:
I should have definitely added you as a co-author. That was my mistake.
I intentionally left the Android section considering the possibility of code base integration, but going forward, if there are similar tasks, I will proceed by removing such sections.

scribam (approved):

PR #3486, opened 2025-01-01
shader/translator/texture: Ignore lod mode bias unimplemented.

Zangetsu38:
# About 
- shader/translator/texture: Ignore lod mode bias unimplemented.
Fix texture missing on one game

Problem on this game founded by @saturnsky with game using lod_mod 1, and current code literally skip using texture of shader only because lod mode bias is not impelented, no have sense.
Lod detail level relate distance is not very importante in computer compare psvita
But so yes, need implement it on days...

# Result
- Before
image
- After
image

### DIFF
--- a/vita3k/shader/src/translator/texture.cpp
+++ b/vita3k/shader/src/translator/texture.cpp
@@ -201,12 +201,6 @@ bool USSETranslatorVisitor::smp(
     Imm7 src0_n,
     Imm7 src1_n,
     Imm7 src2_n) {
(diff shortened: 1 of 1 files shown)


Macdu:
I mean, might as well implement LOD bias if we have a game using it. I'll implement it tomorrow.

Zangetsu38:
> I mean, might as well implement LOD bias if we have a game using it. I'll implement it tomorrow.

I see

Zangetsu38:
Closed with get implemented in #3488

PR #3488, opened 2025-01-02
shader: Add sample lod bias support

Macdu:
Add support for the lod bias sample instruction.
Supersedes #3486.

Note that I haven't tested it, please do so before approving.


Zangetsu38:
works fine
image

PR #3489, opened 2025-01-02
motion: Return motion sample history when requested

german77:
Motion is actually stored in a ring buffer and sceMotionGetSensorState is different from sceMotionGetState. The important part here is that sceMotionGetSensorState returns unfiltered values. This means you can access raw data from the motion sensors, this is probably done so the game is able to use it's own motion filter.


Zangetsu38 (approved):

Zangetsu38:
@german77 why close ?

PR #3497, opened 2025-01-08
Fix GDB remote connection

FirebirdTA01:
The change in thread.cpp allows the thread to be resumed by the GDB stub correctly by registering the thread as suspended. This also reduces CPU usage as the thread isn't just running when it's supposed to be suspended.

GDB command matching now won't find a match with just any command that contains the first letter if the command is shorter than what is passed. This was a problem when just sending 'q' and it would find a match with longer commands that started with 'q'.  Lastly, arm-vita-eabi-gdb would complain that unrecognized commands were received and then hang. This was due to some double parsing being caused by calling 
`LOG_GDB("GDB Server Unrecognized Command. {}", std::string(command.content_start, command.content_length));`
even after finding a matching command.

### DIFF
--- a/vita3k/gdbstub/src/gdb.cpp
+++ b/vita3k/gdbstub/src/gdb.cpp
@@ -696,7 +696,8 @@ constexpr bool cmp_less(T t, U u) noexcept {
 }
 
 static bool command_begins_with(PacketCommand &command, const std::string_view small_str) {
-    if (!cmp_less(small_str.size(), command.content_length))
+    //If the command's content is shorter than small_str, it can't match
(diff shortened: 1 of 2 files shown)


bookmist:
Please squash the commits into one and name it according to the project conventions. Something like `gdbstub: Fix GDB command matching and don't reply with "Unrecognized Command"`

VelocityRa:
@bookmist You can do this yourself when merging if you choose the Squash option in the dropdown.

VelocityRa (approved):
lgtm!

PR #3500, opened 2025-01-11
main: restore serialize_config removed by #2932

nn9dev:
Without this line, a fresh install of the emulator will never boot to the main app select screen. Tested on Linux.


bookmist (approved):

Zangetsu38:
Is not needed, condig is already saved in finish setup
And code works same for all os

nn9dev:
Zangetsu and I have found the root cause, which will hopefully be solved in a different upcoming pr :)

PR #3504, opened 2025-01-17
modules/SceAudiodec: Limit the value of pcm_size_given to pass assert.

Dicot0721:


bookmist:
I suggest splitting this into three separate commits, as it addresses three distinct issues. The fixes in the modules and packages look good to me, and I am ready to approve them. However, I am not very familiar with the renderer, so I would prefer to wait for Macdu's review on those changes. If the renderer fixes are moved to a separate PR, I will be happy to approve the current changes.

bookmist (approved):

bookmist:
I reviewed the source code more thoroughly, and I don't think suppressing this assertion is a good idea. Both pcm_size_given and pcm_size_max are calculated parameters within the emulator code, and any discrepancy between them likely indicates an error. Could you please specify where you encountered the assertion failure?

PR #3509, opened 2025-01-22
ngs: Fix attempts to allocate huge amounts of memory causing the process to freeze.

Dicot0721:
- `uint32_t decoded_size` may overflow to a very large number by subtracting a number larger than itself, which causes `data.extra_storage.resize(curr_pos + decoded_size * sizeof(float) * 2)` to try to allocate huge amounts of memory (and the process freezes as a result).
- I'm not actually sure if I wrote it correctly this way, I just referenced the previous line of code, but judging by the results, it works fine.

### DIFF
--- a/vita3k/ngs/src/modules/atrac9.cpp
+++ b/vita3k/ngs/src/modules/atrac9.cpp
@@ -185,7 +185,7 @@ bool Atrac9Module::decode_more_data(KernelState &kern, const MemState &mem, cons
         const uint32_t samples_left_after = (frame_bytes_gotten / superframe_size - 1) * samples_per_superframe;
         if (bufparam.samples_discard_end_off > samples_left_after) {
             // last chunk
-            decoded_size -= bufparam.samples_discard_end_off;
+            decoded_size -= std::min(decoded_size, bufparam.samples_discard_end_off - samples_left_after);
         }
     }
 


vita3k/ngs/src/modules/atrac9.cpp
```diff
@@ -182,10 +182,10 @@ bool Atrac9Module::decode_more_data(KernelState &kern, const MemState &mem, cons
             decoded_size -= skipped_samples;
         }
 
-        const uint32_t samples_left_after = (frame_bytes_gotten / superframe_size - 1) * samples_per_superframe;
+        const uint32_t samples_left_after = ((frame_bytes_gotten / superframe_size) - 1) * samples_per_superframe;
         if (bufparam.samples_discard_end_off > samples_left_after) {
             // last chunk
-            decoded_size -= bufparam.samples_discard_end_off;
+            decoded_size -= std::min(samples_per_superframe, bufparam.samples_discard_end_off - samples_left_after);
```
bookmist, line 188: If the previous `if` statement is executed, the bug will still occur.
```suggestion
            decoded_size -= std::min(decoded_size, bufparam.samples_discard_end_off - samples_left_after);
```

vita3k/ngs/src/modules/atrac9.cpp
```diff
@@ -182,10 +182,10 @@ bool Atrac9Module::decode_more_data(KernelState &kern, const MemState &mem, cons
             decoded_size -= skipped_samples;
         }
 
-        const uint32_t samples_left_after = (frame_bytes_gotten / superframe_size - 1) * samples_per_superframe;
+        const uint32_t samples_left_after = ((frame_bytes_gotten / superframe_size) - 1) * samples_per_superframe;
```
bookmist, line 185: Revert the unnecessary brackets

vita3k/ngs/src/modules/atrac9.cpp
```diff
@@ -246,7 +246,7 @@ bool Atrac9Module::decode_more_data(KernelState &kern, const MemState &mem, cons
     }
 
     const int32_t sample_rate = data.parent->rack->system->sample_rate;
-    if (params->playback_scalar != 1 || static_cast<int>(round(params->playback_frequency)) != sample_rate) {
+    if (params->playback_scalar != 1.0f || static_cast<int>(round(params->playback_frequency)) != sample_rate) {
```
bookmist, line 249: I would also revert this change as it is not related to the bug. However, it is not as critical.

vita3k/ngs/src/modules/atrac9.cpp
Zangetsu38: I like it personaly get more clear view ^^

bookmist (approved):

PR #3517, opened 2025-02-02
Updated Swedish translation

yeager:

PR #3519, opened 2025-02-02
lang: revert some wrong string of sv.

Zangetsu38:
# About
- lang: revert some wrong string of sv.
using official translation for it.

- fix get broken by last commit #3517


Zangetsu38:
@scribam never merge any lang pr without me or @Croden1999 checking

bookmist (approved):

PR #3530, opened 2025-02-26
ngs/packages: align classes for 64bit and refactor constexpr for static int

GermanAizek:
SceFiber: 72 -> 64 bytes
SceHeader: 48 -> 40 bytes
SceNgsPlayerStates: 88 -> 80 bytes

### DIFF
--- a/vita3k/modules/SceFiber/SceFiber.cpp
+++ b/vita3k/modules/SceFiber/SceFiber.cpp
@@ -56,11 +56,11 @@ enum class FiberStatus {
 };
 
 typedef struct SceFiber {
+    char name[32];
+    CPUContext *cpu;
     Ptr<SceFiberEntry> entry;
     Address addrContext;
     SceSize sizeContext;
-    char name[32];
-    CPUContext *cpu;
     SceUInt32 argOnInitialize;
     Ptr<uint32_t> argOnRun;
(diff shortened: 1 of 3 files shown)


vita3k/modules/SceFiber/SceFiber.cpp
```diff
@@ -56,11 +56,11 @@ enum class FiberStatus {
 };
 
 typedef struct SceFiber {
+    char name[32];
+    CPUContext *cpu;
     Ptr<SceFiberEntry> entry;
     Address addrContext;
     SceSize sizeContext;
-    char name[32];
-    CPUContext *cpu;
```
Zangetsu38, line 63: Why move variable ?

GermanAizek:
> The members are most likely declared in a certain order (e.g. [vitasdk](https://docs.vitasdk.org/)) and shouldn't be changed, and the application doesn't lack this little bit of memory.

in other words, do I need to align in vitasdk headers first?

Zangetsu38:
> > The members are most likely declared in a certain order (e.g. [vitasdk](https://docs.vitasdk.org/)) and shouldn't be changed, and the application doesn't lack this little bit of memory.
> 
> in other words, do I need to align in vitasdk headers first?

Don't touch it 🧍

PR #3532, opened 2025-02-27
shader/renderer/module/io: using reserve for reduce inserts time

GermanAizek:


vita3k/gui/src/vita3k_update.cpp
```diff
@@ -79,6 +79,7 @@ bool init_vita3k_update(GuiState &gui) {
         std::thread get_commit_desc([dif_from_current]() {
             // Calculate Page and Per Page Get
             std::vector<std::pair<uint32_t, uint32_t>> page_count;
+            page_count.reserve(dif_from_current / 100);
```
Zangetsu38, line 82: if you can revert it please, i have pr one totaly change code, and this code is wrong here

vita3k/shader/include/shader/usse_translator.h
```diff
@@ -96,6 +96,7 @@ class USSETranslatorVisitor final {
 
             std::vector<spv::Id> consts;
 
+            consts.reserve((i + 1) - 1);
```
EXtremeExploit, line 99: isnt this the same as just `i`?

vita3k/interface.cpp
```diff
@@ -436,12 +443,12 @@ static ExitCode load_app_impl(SceUID &main_module_id, EmuEnvState &emuenv) {
...
     init_savedata_app_path(emuenv.io, emuenv.pref_path);
 
-    // Load param.sfo
-    vfs::FileBuffer param_sfo;
-    if (vfs::read_app_file(param_sfo, emuenv.pref_path, emuenv.io.app_path, "sce_sys/param.sfo"))
-        sfo::load(emuenv.sfo_handle, param_sfo);
-
-    init_exported_vars(emuenv);
+    // todo: VAR_NID(__sce_libcparam, 0xDF084DFA) is loaded wrong
+    emuenv.kernel.export_nids.reserve(get_var_exports().size());
```
EXtremeExploit, line 447: Might be better to store the get_var_exports return into a variable instead of calling it twice

vita3k/io/src/io.cpp
```diff
@@ -173,6 +178,7 @@ bool find_case_isens_path(IOState &io, VitaIoDevice &device, const fs::path &tra
     if (!fs::exists(final_path))
         return false;
 
+    io.cachemap.reserve(std::distance(fs::directory_iterator{final_path}, fs::directory_iterator{}));
```
EXtremeExploit, line 181: Im not sure this works correctly

PR #3533, opened 2025-02-27
shader: optimize make_vert_finalize using reserve with iterable vector

GermanAizek:
Affects the function call `load_shader_generic()`.

std::vector is better suited as an iterator for-range-loop (C++17) than std::list, and it also requires less memory.

If std::map and std::unordered_map are replaced, use .at() operator is negligible.

It becomes possible to use reserve(), which significantly reduces insertion time in lambda function `add_vertex_output_info`.

### DIFF
--- a/vita3k/shader/src/spirv_recompiler.cpp
+++ b/vita3k/shader/src/spirv_recompiler.cpp
@@ -121,7 +121,7 @@ struct VertexProgramOutputProperties {
     std::uint32_t component_count{};
     std::uint32_t location{};
 };
-using VertexProgramOutputPropertiesMap = std::map<SceGxmVertexProgramOutputs, VertexProgramOutputProperties>;
+using VertexProgramOutputPropertiesMap = std::unordered_map<SceGxmVertexProgramOutputs, VertexProgramOutputProperties>;
 
 // ******************************
 // * Functions (implementation) *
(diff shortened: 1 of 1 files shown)


vita3k/shader/src/spirv_recompiler.cpp
```diff
@@ -1532,11 +1532,13 @@ static spv::Function *make_vert_finalize_function(spv::Builder &b, const SpirvSh
 
     VertexProgramOutputPropertiesMap vertex_properties_map;
     // list is used here to gurantee the vertex outputs are written in right order
-    std::list<SceGxmVertexProgramOutputs> vertex_outputs_list;
+    std::vector<SceGxmVertexProgramOutputs> vertex_outputs_list;
```
bookmist, line 1535: Do you read the comment?

EXtremeExploit: vector should still be ordered the same way list is, plus it makes everything contiguous in memory so its better for cache and memory footprint a little bit

EXtremeExploit (approved):

vita3k/shader/src/spirv_recompiler.cpp
```diff
@@ -1569,10 +1574,10 @@ static spv::Function *make_vert_finalize_function(spv::Builder &b, const SpirvSh
     for (const auto vo : vertex_outputs_list) {
```
Dicot0721, line 1574: Is `vertex_outputs_list` necessary? Don't we just need `vertex_properties_map`?
```suggestion
    for (const auto &[vo, properties] : vertex_properties_map) {
```
If we remove `vertex_outputs_list`, we won't even need `add_vertex_output_info`.

PR #3534, opened 2025-02-27
renderer: code refactor const refs, keywords, constexpr C++17

GermanAizek:
Just the usual code refactoring for newer C++ standarts.

### DIFF
--- a/vita3k/renderer/src/texture/pvrt-dec.cpp
+++ b/vita3k/renderer/src/texture/pvrt-dec.cpp
@@ -55,19 +55,19 @@ struct Pixel32 {
 struct Pixel128S {
     int32_t red, green, blue, alpha;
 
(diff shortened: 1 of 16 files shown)


vita3k/renderer/src/texture/pvrt-dec.cpp
```diff
@@ -55,19 +55,19 @@ struct Pixel32 {
 struct Pixel128S {
     int32_t red, green, blue, alpha;
 
-    Pixel128S operator+(const int32_t uiFactor) {
+    Pixel128S operator+(const int32_t uiFactor) const {
```
bookmist, line 58: This file is taken from PowerVR SDK, so it would be better to send your fixes to upstream

xerpi:
@GermanAizek, can you rebase this and, for now, drop the pvrt change as @bookmist said, unless it's already upstream?

PR #3537, opened 2025-02-27
gui/renderer: Replace sscanf when not parsing a complex string

Dicot0721:
- The use of `sscanf` should be reduced, to improve efficiency and get rid of the cumbersome format specifiers.

### DIFF
--- a/vita3k/gui/src/live_area.cpp
+++ b/vita3k/gui/src/live_area.cpp
@@ -820,12 +820,12 @@ void draw_live_area_screen(GuiState &gui, EmuEnvState &emuenv) {
                 std::vector<ImVec4> str_color;
 
                 if (!str_tag.color.empty()) {
-                    unsigned int color = 0xFFFFFFFF;
+                    uint32_t color = 0xFF'FF'FF'FF;
 
                     if (frame.autoflip)
-                        sscanf(str[app_path][frame.id][current_item[app_path][frame.id]].color.c_str(), "#%x", &color);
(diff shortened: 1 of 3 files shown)


vita3k/gui/src/themes.cpp
```diff
@@ -106,8 +106,7 @@ static ImU32 convert_hex_color(const std::string &src_color) {
     std::string result = src_color.substr(src_color.length() - 6, 6);
     result.insert(0, "ff");
 
-    unsigned int color;
-    sscanf(result.c_str(), "%x", &color);
+    uint32_t color = std::stoul(result, nullptr, 16);
```
bookmist, line 109: What if `result` value is invalid? (see util/string_utils:stol_def)

vita3k/gui/src/live_area.cpp
```diff
@@ -820,12 +820,12 @@ void draw_live_area_screen(GuiState &gui, EmuEnvState &emuenv) {
                 std::vector<ImVec4> str_color;
 
                 if (!str_tag.color.empty()) {
-                    unsigned int color = 0xFFFFFFFF;
+                    uint32_t color = std::numeric_limits<uint32_t>::max();
```
bookmist, line 823: It's not uint32_t max value, it's ARGB hex color (white non-transparent). I think it's better to let colors be in hex values as they are used in web. Or use some constant (I don't remember if there is a constant for white color)

bookmist (approved):

PR #3459, opened 2024-12-09
package: Implement decrypt fself in a fly.

Zangetsu38:
# About
- package: Implement decrypt fself in a fly.

Request by @Princess-of-Sleeping for obligatory things with vsh support
Should help for improve it https://github.com/Vita3K/Vita3K/pull/3432

PR #3545, opened 2025-03-13
Add option to use old vita3k install app and firmware method

ikhoeyZX:
Add option to install app / firmware in unencrypted like old version vita3k, so app can run in old version vita3k (before this pr : https://github.com/Vita3K/Vita3K/pull/3459 ) without reinstall again

Screenshot_20250313_231440_Vita3K ZX

### DIFF
--- a/vita3k/config/include/config/config.h
+++ b/vita3k/config/include/config/config.h
@@ -148,6 +148,7 @@ enum ScreenshotFormat {
     code(bool, "display-info-message", false, display_info_message)                                     \
     code(bool, "show-welcome", true, show_welcome)                                                      \
     code(bool, "check-for-updates", true, check_for_updates)                                            \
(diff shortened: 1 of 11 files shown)


Zangetsu38:
Ouch no ?
For why ?
My pr have specialy done stuf for devrypt in a fly, so no revert this 🫠

vita3k/config/include/config/config.h
```diff
@@ -148,6 +148,7 @@ enum ScreenshotFormat {
     code(bool, "display-info-message", false, display_info_message)                                     \
     code(bool, "show-welcome", true, show_welcome)                                                      \
     code(bool, "check-for-updates", true, check_for_updates)                                            \
+    code(bool, "dencrypt-install", true, dencrypt_install)                                              \
```
Zangetsu38, line 151: Why true ?
Is not normal situation decrypt selfie in install, you get broken self after

PR #3551, opened 2025-03-31
ngs/player: Prevent infinite loop by checking empty buffer.

Zangetsu38:
# About
- ngs/player: Prevent infinite loop by checking empty buffer.


Zangetsu38:
Ready for review

bookmist (approved):

PR #3505, opened 2025-01-20
gui: Fix Chinese characters are displayed as Japanese glyphs

Dicot0721:
- Fix CJK characters are still displayed as Japanese glyphs when the system language is Chinese.

PR #3553, opened 2025-04-01
gui: Fix missing Chinese fonts in initial setup screen.

Dicot0721:
- This PR is separated from #3505.
- Before this PR, Chinese fonts cannot be loaded when we launch the emulator for the first time, because the default system language is English (US).

> We can see a lot of question marks in the image, these are the characters that are missing from the font.

image
image

### DIFF
--- a/vita3k/gui/src/gui.cpp
+++ b/vita3k/gui/src/gui.cpp
@@ -271,7 +271,7 @@ static void init_font(GuiState &gui, EmuEnvState &emuenv) {
                     io.Fonts->AddFontFromFileTTF(fs_utils::path_to_utf8(default_font_path / "mplus-1mn-bold.ttf").c_str(), font_config.SizePixels, &font_config, japanese_and_extra_ranges.Data);
 
                     const auto sys_lang = static_cast<SceSystemParamLang>(emuenv.cfg.sys_lang);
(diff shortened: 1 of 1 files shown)


vita3k/gui/src/gui.cpp
```diff
@@ -271,7 +271,7 @@ static void init_font(GuiState &gui, EmuEnvState &emuenv) {
                     io.Fonts->AddFontFromFileTTF(fs_utils::path_to_utf8(default_font_path / "mplus-1mn-bold.ttf").c_str(), font_config.SizePixels, &font_config, japanese_and_extra_ranges.Data);
 
                     const auto sys_lang = static_cast<SceSystemParamLang>(emuenv.cfg.sys_lang);
-                    if (sys_lang == SCE_SYSTEM_PARAM_LANG_CHINESE_S)
+                    if (sys_lang == SCE_SYSTEM_PARAM_LANG_CHINESE_S || sys_lang == SCE_SYSTEM_PARAM_LANG_ENGLISH_US) // System language may not have been selected
```
Zangetsu38, line 274: change it by using config for check if init is already completed
is more precise

Zangetsu38 (approved):

bookmist (approved):

PR #3554, opened 2025-04-01
gui/lang: Add one translatable string

Dicot0721:


Zangetsu38:
> > I no likenit 😔
> > Problème relate button size, is fix, and some languague can have big size of string
> 
> Strings in other languages are not necessarily larger than in English, even if the string length is slightly longer than the button length, it's better than just displaying it in English.

Not really button size is based on english Word size
And here i using shirt Word version Preinst is complicated to have same short version in other languague

Zangetsu38:
But adding already one langue so, no need create it if add 0 languague translate ^^

Zangetsu38 (approved):

Zangetsu38 (approved):

bookmist (approved):

PR #3557, opened 2025-04-03
Add the run status to sceKernelStartThread immediately after calling it

RyleFury:
Add a small delay after sceKernelStartThread. If this delay is not put, sceKernelWaitThreadEnd and sceKernelDeleteThread functions which are too close work randomly and many glitches can appear. Tested with Ciel Nosurge Offline, fixes all the random freezes and graphical glitches with 3D models. 1µs is enough with tests I made. However, more tests could be performed to make sure it works properly everywhere since it's a very crucial function.


Zangetsu38:
Is little hacky here 🤔

RyleFury:
Changed to a much better solution for the same result ^^

Zangetsu38:
Is see, @bookmist is for you 👌

bookmist (approved):

PR #3560, opened 2025-04-07
external: Update dynarmic

EXtremeExploit:
The current version of dynarmic points to a commit that doesnt exists and fails to checkout to it in some cases, so just make the dynarmic point to the actual latest commit in our fork


1whatleytay (approved):
Seems sane

Zangetsu38:
Waiting for merge, master is keep draft
🧍

PR #3562, opened 2025-04-09
Set a pixel perfect ratio for 16:9 resolutions in fullscreen mode

RyleFury:
Since the resolution of the PS Vita is very close from a 16:9 one (just 4 extra height pixels), I propose this change. It allows to take all the screen width and crop 2 pixels in the top and the bottom in fullscreen mode when the resolution of the screen is 16:9 only. It's especially interesting for 1080p (pixels x 4) and 4K (pixels x 16) resolutions which are integer multiples of 960x540 to get the best possible rendering. The same method is used by the plugin Sharpscale on PSTV with the Integer mode in 1080i.


Zangetsu38:
You need learn how using git, and set only one commit with good name, here is invalid for this

PR #3563, opened 2025-04-10
Add the "Fullscreen HD res pixel perfect" checkbox in the GUI section

RyleFury:
Since the resolution of the PS Vita is very close from a 16:9 one (just 4 extra height pixels), I propose this change. It allows to take all the screen width and crop 2 pixels in the top and the bottom in fullscreen mode when the resolution of the screen is 16:9 only. It's especially interesting for 1080p (pixels x 4) and 4K (pixels x 16) resolutions which are integer multiples of 960x540 to get the best possible rendering. The same method is used by the plugin Sharpscale on PSTV with the Integer mode in 1080i.

If you prefer, we could also add a combo box with three modes to replace the "Stretch to Display Area" one:

- Fit (Standard mode)
- Stretch to full screen
- Stretch to full width (keep aspect ratio)

This time, I think I did things properly with the pull request ^^

### DIFF
--- a/vita3k/gui/src/settings_dialog.cpp
+++ b/vita3k/gui/src/settings_dialog.cpp
@@ -272,6 +272,7 @@ void init_config(GuiState &gui, EmuEnvState &emuenv, const std::string &app_path
 
     get_modules_list(gui, emuenv);
     config.stretch_the_display_area = emuenv.cfg.stretch_the_display_area;
+    config.fullscreen_hd_res_pixel_perfect = emuenv.cfg.fullscreen_hd_res_pixel_perfect;
     config_cpu_backend = set_cpu_backend(config.cpu_backend);
     current_aniso_filter_log = static_cast<int>(log2f(static_cast<float>(config.anisotropic_filtering)));
     max_aniso_filter_log = static_cast<int>(log2f(static_cast<float>(emuenv.renderer->get_max_anisotropic_filtering())));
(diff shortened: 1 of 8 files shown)


Zangetsu38:
I have no said close and ré open pr, just learn how squash, édit last commit, force push etc...
Here commit name is keep invalide, check all othe for look how write it.
And so here using edit last commit for change name

Thx

RyleFury:
OK, I'll remember it. I changed the commit name with the amend option. Please be patient, I'm not used to Pull Requests or your community and it's not easy at first. Tell me if you still see something wrong.

RyleFury:
Got it, done!

Zangetsu38:
fine, i left other review :)

Zangetsu38:
ah, for your ask yes is better if you can create option for this ^^
i personally more like respect aspect ratio and no lost some pixel

RyleFury:
Fullscreen_HD_res_pixel_perfect

OK, I changed my strategy.
I added a checkbox "Fullscreen HD res pixel perfect" in the GUI tab.

This checkbox will only affect the fullscreen mode and the HD resolutions which are multiple integers of 960x540 (standard ones being 1080p and 4K).

The point is that pixels are multiplied by an integer, making a very sharp rendering. The game screen will take all the display area's width and crop 2 native pixels in the top and the bottom, which is barely visible.

The "Fullscreen HD res pixel perfect" checkbox has the priority over the "Stretch The Display Area" one, since requirements for the former are pretty strict. In any case, this checkbox is safe.

I know pretty a lot of people who would be interested by this feature, especially since we can attune the upscaling of the internal resolution consequently, which is a real plus. Of course, even without this feature, the "Nearest" Screen Filter does almost the same, but at the price of lost pixels which make the animations strange, so it's frustrating.

Now, I'll just wait for your review :)

vita3k/gui/src/settings_dialog.cpp
```diff
@@ -375,11 +376,21 @@ static void save_config(GuiState &gui, EmuEnvState &emuenv) {
         emuenv.cfg.psn_signed_in = config.psn_signed_in;
     }
 
+    bool update_viewport_en = 0;
```
Zangetsu38, line 379: bool 0 ? 
```suggestion
    bool update_viewport_en = false;
```

bookmist:
Looks good to merge. @RyleFury please, resolve conflict.

bookmist (approved):

PR #3565, opened 2025-04-15
gui/ime: Allow moving the cursor under numeric keypad

Dicot0721:


Zangetsu38:
You can just check if your change respect comportement of real ime keyboard on real psvita ?

Zangetsu38:
> > no have ime in my paf branch, or in my home branch, IME is only used by app, in vsh you only view ime dialog, is diferent things how it works
> > you need real vita with using real game for this for now
> > you can look with nishinji for test this
> 
> I can use ime when changing folder names in the home screen of the psv system, I tested it here

is not IME, is ime dialog, is very diferent

Zangetsu38 (approved):

PR #3575, opened 2025-04-23
renderer/vulkan/renderer: Align address and size to 4K in no page table.

Zangetsu38:
# About:
- renderer/vulkan/renderer: Align address and size to 4K in no page table.
Ensure memory address is aligned to 4K and size is adjusted accordingly.
Handle special cases where no page table mapping is used.

# Result
- Fix this game to go ingame again in vulkan render.
image


vita3k/modules/SceGxm/SceGxm.cpp
```diff
@@ -2694,25 +2694,33 @@ EXPORT(int, sceGxmMapMemory, Ptr<void> base, uint32_t size, uint32_t attribs) {
...
     }
 
+    // Add 4KiB alignment to the base address
+    const uint64_t base_ptr_val = std::bit_cast<uint64_t>(base.get(emuenv.mem));
+    const uint64_t base_offset = align(base_ptr_val, KiB(4)) - base_ptr_val;
+    Address base_aligned{ static_cast<uint32_t>(base_ptr_val + base_offset) };
+
+    // Add 4KiB alignment to the size
+    size = align(base_offset + size, KiB(4));
+
```
Macdu, line 2704:
```suggestion
     if (base.address() % KiB(4) != 0 || size % KiB(4) != 0) {
           LOG_WARN_ONCE("Mapping unaligned GPU memory");
     }
    // Make sure the base address and size are 4KiB-aligned
    Address base_ptr = align_down(base.address(), KiB(4));
    uint32_t size_aligned = align(base.address() + size, KiB(4)) - base_ptr;

```
- There is no reason to retrieve the host (64-bit) address from the guest address for alignment, it can be done on the 32-bit guest address directly
- The base address should be aligned down, not up 
- You are converting back a host adress to a guest one by truncating the upper 32-bit. The only reason this doesn't fail horribly is because usually the guest start location is 32-bit aligned (but this is not a guarantee).

Zangetsu38: done

Macdu (approved):

bookmist (approved):

PR #3584, opened 2025-04-28
mem/ptr.h: Remove template for 'Ptr(U *, const MemState &)'

Dicot0721:
- `Ptr<T>::Ptr<U>(U *, const MemState &)` has redundant template, which allows code like `Ptr<int>::Ptr<char>(char *, const MemState &)` to pass compilation.

### DIFF
--- a/vita3k/modules/SceLibKernel/SceLibKernel.cpp
+++ b/vita3k/modules/SceLibKernel/SceLibKernel.cpp
@@ -343,7 +343,7 @@ EXPORT(int, sceClibStrcatChk) {
 
 EXPORT(Ptr<char>, sceClibStrchr, const char *str, int c) {
     TRACY_FUNC(sceClibStrchr, str, c);
-    const char *res = strchr(str, c);
+    char *res = const_cast<char *>(strchr(str, c));
     return Ptr<char>(res, emuenv.mem);
 }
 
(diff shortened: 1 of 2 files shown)


bookmist (dismissed):

vita3k/modules/SceLibKernel/SceLibKernel.cpp
```diff
@@ -343,7 +343,7 @@ EXPORT(int, sceClibStrcatChk) {
 
 EXPORT(Ptr<char>, sceClibStrchr, const char *str, int c) {
     TRACY_FUNC(sceClibStrchr, str, c);
-    const char *res = strchr(str, c);
+    char *res = const_cast<char *>(strchr(str, c));
```
bookmist, line 346: I think it's better to add or remove const from function header then use const_cast. Like `EXPORT(Ptr<char>, sceClibStrchr, char *str, int c)` or `EXPORT(Ptr<const char>, sceClibStrchr, const char *str, int c)`
Same for sceClibStrrchr

bookmist (approved):

vita3k/modules/SceLibKernel/SceLibKernel.cpp
bookmist: You are right. I look on C++ definition of `strchr` and here is С function

PR #3591, opened 2025-05-04
modules/SceSysmem: Handle memory using SFO and pool info.

Zangetsu38:
# About
- modules/SceSysmem: Handle memory using SFO and pool info.
Fix crash and prevent memory over-commitment in sceKernelGetFreeMemorySize.
Support +29MiB, +77MiB, and +109MiB memory expansion modes via ATTRIBUTE2 in SFO.
Adjust max usable memory (max_user) dynamically based on selected expansion mode.
Account for all memory pools: normal, CDRAM, and phycont, for accurate memory tracking.
Ensure proper scaling and alignment of memory size when memory is limited.

thx to @bookmist for lot information and explain.

# Result
- Get playable for first time
image

- Move from nothing to Intro for first time
image

### DIFF
--- a/vita3k/modules/SceSysmem/SceSysmem.cpp
+++ b/vita3k/modules/SceSysmem/SceSysmem.cpp
@@ -20,7 +20,10 @@
 #include <kernel/state.h>
 #include <kernel/types.h>
 
+#include <packages/sfo.h>
+
 #include <util/align.h>
+#include <util/string_utils.h>
 
 #include <util/tracy.h>
 TRACY_MODULE_NAME(SceSysmem);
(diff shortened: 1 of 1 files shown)


Macdu:
I"m not sure this is a good idea, the reason being that Vita3K likely allocates a lot more guest memory than actual hardware. It may break other games. If you go with this approach it may be better to track the different types of memory when allocating it rather than in sceKernelGetFreeMemorySize, as it may be called a lot and the way you implemented it right now is really inefficient.

Also what was the issue with the game to begin with? Was it because this function was reporting a lot more cdram/phycont memory than what was expected? Because it was not decreasing as expected after an allocation? Because of some overflow (wouldn't be surprised)?

Zangetsu38:
@Macdu is done, have changed pr, more precise on user

PR #3592, opened 2025-05-05
ngs/player: fix small typo

nishinji:
I investigated and found that Power Pro Baseball 2013 is no longer playable.
This will allow to play Power Pro Baseball 2013 again.

### DIFF
--- a/vita3k/ngs/src/modules/player.cpp
+++ b/vita3k/ngs/src/modules/player.cpp
@@ -210,8 +210,8 @@ bool PlayerModule::process(KernelState &kern, const MemState &mem, const SceUID 
                 decoder->receive(nullptr, &samples_count);
 
                 // Playback rate scaling
-                float src_sample_rate = std::ceil(params->playback_frequency);
-                if ((params->playback_scalar != 1.f) || (static_cast<int>(src_sample_rate) != sample_rate)) {
+                float src_sample_rate = params->playback_frequency;
+                if ((src_sample_rate >= 1.f) && ((params->playback_scalar != 1.f) || (static_cast<int>(src_sample_rate) != sample_rate))) {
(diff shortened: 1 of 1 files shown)


vita3k/ngs/src/modules/player.cpp
```diff
@@ -53,14 +53,14 @@ void PlayerModule::on_param_change(const MemState &mem, ModuleData &data) {
 
     // check for invalid playback values
     const auto is_invalid_playback_value = [](const float playback_value, const float max_value) {
-        return isnan(playback_value) || (playback_value < 0.f) || (playback_value > max_value);
+        return isnan(playback_value) || (playback_value <= 0.f) || (playback_value > max_value);
```
Zangetsu38, line 56: ?? 0 value is totaly valid for ngs

Zangetsu38: ah after test it i understund issue... 
relate scaling, it try re scaling with 0 value, and like unity game, it do'nt like that

Zangetsu38:
For fix it, done just my old fix...
for ,in line 213 and 214 and copy past that
>float src_sample_rate = params->playback_frequency;
if ((params->playback_scalar != 1.f) || (static_cast<int>(src_sample_rate) != sample_rate)) {

bookmist:
Documentation said that Playback Frequency and Playback Scalar can be zero. So I think fix may be somewhere else.

Zangetsu38:
> Documentation said that Playback Frequency and Playback Scalar can be zero. So I think fix may be somewhere else.

Real fix is just ignore rescal sample if sample freq is < 0 , is what i have suggested like change

Zangetsu38 (approved):

Zangetsu38:
is more if is < 1 here ^^

bookmist (approved):

PR #3594, opened 2025-05-11
module/load module: Add HTTPS loading with RUDP module and include it in auto-load list

Zangetsu38:
# About 
- module/load module: Add HTTPS loading with RUDP module.
Prevent games from calling RUDP without loading HTTPS.  
Add RUDP to the automatic modules list and reorder the list to match the enum order.

### DIFF
--- a/vita3k/module/src/load_module.cpp
+++ b/vita3k/module/src/load_module.cpp
@@ -59,7 +59,7 @@ static SysmodulePaths init_sysmodule_paths() {
     p[SCE_SYSMODULE_NP_SCORE_RANKING] = { "np_ranking" };
     p[SCE_SYSMODULE_SQLITE] = { "libSceSqlite" };
     p[SCE_SYSMODULE_TRIGGER_UTIL] = { "trigger_util" };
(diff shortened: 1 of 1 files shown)


bookmist:
librudp does not use libhttp and libssl. It's ok for me to include it in auto-load list, but not Add HTTPS loading with RUDP module

Zangetsu38:
> librudp does not use libhttp and libssl. It's ok for me to include it in auto-load list, but not Add HTTPS loading with RUDP module

How explain so i have fixed issue with this ?
It call rudp module and after directly using http and ssl function...

bookmist:
> > librudp does not use libhttp and libssl. It's ok for me to include it in auto-load list, but not Add HTTPS loading with RUDP module
> 
> How explain so i have fixed issue with this ? It call rudp module and after directly using http and ssl function...

What game? and when it happens?

Zangetsu38:
Drive girl in boot, it load some module and just after load rudp, call instant http ans ssl function  but no have call module, and is impossible... so want said is loaded by other

PR #3601, opened 2025-05-18
util/exec.h: Fix include

Dicot0721:
- `"exec.h"` lacks the necessary header file, and although it does not report an error at compile time (because there is no file that references it), opening the file in IDE displays with error.


bookmist:
So why not to delete it completely, if this file is not used?

EXtremeExploit (approved):

PR #3611, opened 2025-05-25
shader/usse utilities: implement FX10 pack function for C10 format.

Zangetsu38:
# About
- shader/usse utilities: implement FX10 pack function for C10 format.


Macdu:
A few comments:
- Is there actually a game using this?
- Nice AI generated code. However it has like 4 different distinct issues

PR #3615, opened 2025-05-29
modules/SceSysmodule: Avoid reloading already loaded module.

Zangetsu38:
# About
- modules/SceSysmodule: Avoid reloading already loaded module.
- module/load_module: Use contains() instead of find() in is_module_loaded.

fix regression on sao move for ingame to bootable, with no can using fios module already loaded.

# Result
- get back to ingame again
image

### DIFF
--- a/vita3k/modules/SceSysmodule/SceSysmodule.cpp
+++ b/vita3k/modules/SceSysmodule/SceSysmodule.cpp
@@ -198,6 +198,8 @@ EXPORT(int, sceSysmoduleLoadModule, SceSysmoduleModuleId module_id) {
         return RET_ERROR(SCE_SYSMODULE_ERROR_INVALID_VALUE);
 
     LOG_INFO("Loading module ID: {}", to_debug_str(emuenv.mem, module_id));
+    if (is_module_loaded(emuenv.kernel, module_id))
+        return SCE_SYSMODULE_LOADED;
     if (is_module_enabled(emuenv, module_id)) {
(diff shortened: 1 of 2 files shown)


vita3k/modules/SceSysmodule/SceSysmodule.cpp
```diff
@@ -198,6 +198,11 @@ EXPORT(int, sceSysmoduleLoadModule, SceSysmoduleModuleId module_id) {
         return RET_ERROR(SCE_SYSMODULE_ERROR_INVALID_VALUE);
 
     LOG_INFO("Loading module ID: {}", to_debug_str(emuenv.mem, module_id));
+    {
+        std::lock_guard<std::mutex> guard(emuenv.kernel.mutex);
+        if (emuenv.kernel.loaded_sysmodules.contains(module_id))
+            return SCE_SYSMODULE_LOADED;
+    }
```
Macdu: There is a is_module_loaded function which does exactly this

Zangetsu38: I know... but problem here is this game load fios in boot with so it already loaded in pre boot, and loading it fail with return parametter error

Disable it in pre-load get same issue

Zangetsu38: > There is a is_module_loaded function which does exactly this

Yes, thx, is done

PR #3616, opened 2025-05-30
gui/live area: Add display last frame in gate.

Zangetsu38:
# About
- gui/live area: Add display last frame in gate.

# Result
image


Zangetsu38:
Rebase guy... oh is me 🗿🤣

PR #3617, opened 2025-05-31
ngs: Fix freezing for unknown reasons

Dicot0721:
- In the debug build, when wrapping `ngs::Voice::voice_mutex` with `std::unique_ptr`, the process freezes at `ngs::Voice::remove_patch` (the game is _Catherine: Full Body_ [PCSG01179]).
- Even without this error, `std::mutex` should be declared directly when used as a member variable, rather than wrapped with `std::unique_ptr`.


bookmist (approved):

Macdu:
The reason the mutex is wrapped in a unique_ptr is because on windows a mutex is really big (I believe 80 bytes). The voice is saved on guest memory so it must not exceed the expected voice size on the guest system. This is why I put it in a unique_ptr (to decrease its size to 8 bytes).

PR #3630, opened 2025-07-04
gxm: add support for SCE_GXM_COLOR_FORMAT_U8_A.

Zangetsu38:
# About
- gxm: add support for SCE_GXM_COLOR_FORMAT_U8_A.
- renderer/vulkan/gxm to Vulkan: Add Support for SCE_GXM_COLOR_SWIZZLE1_A.

Used by toukiden 2


Macdu:
SCE_GXM_COLOR_FORMAT_U8_A is not a color format that can be implemented this way (except if you are lucky and the shader writes the same value to the red and alpha channels). The shader itself must be modified to properly handle it.

Zangetsu38:
I see... 🤔

PR #3631, opened 2025-07-07
modules/SceNgs: Fix softlock when no callback is provided.

Zangetsu38:
# About
- modules/SceNgs: Fix softlock when no callback is provided.
ngs: Add sanity check on allocator block count to avoid false double-free detection.

Relates to an old issue with Jetpack Joyride.

This PR addresses an issue where some games call sceNgsRackRelease without providing a callback, expecting the rack to be released synchronously.
However, the rack might also be released later by the engine (e.g., in VoiceScheduler::update), leading to a double destruction and potential crash or undefined behavior.

# Fix implementation
- init_rack now stores the initial allocator block count after allocation.
- release_rack compares the current block count to the stored value:
If it matches → the rack is assumed to be valid, and destruction proceeds.
If it differs → the rack is assumed to have already been released (e.g., by game-side memory reuse), so destruction is skipped.

This is a non-invasive workaround to avoid softlocks and crashes, especially in games that reuse memory aggressively or rely on single-threaded update logic.


Zangetsu38:
stupid game with strange issue... need more search...

PR #3653, opened 2025-08-15
renderer: Remove VK_KHR_portability_enumeration on macOS

KhoraLee:
Before migrating to SDL3, VK_KHR_portability_enumeration was added to optional_instance_extensions.
Since we are not using the Vulkan Loader, this extension did not appear in vk::enumerateInstanceExtensionProperties and was simply ignored.

However, the current SDL3 implementation adds VK_KHR_portability_enumeration through SDL_Vulkan_GetInstanceExtensions.
This caused the extension to be included in the instance creation process, leading to a crash.

As we have no plans to use the Vulkan Loader, VK_KHR_portability_enumeration should be manually removed from instance_extensions.

This should fix issue #3636.


squidbus:
Removing it from optional extensions will break anyone who *does* have the Vulkan loader installed, as the dynamic loader will try it first. Tested on my system and:
```
[05:49:11.927] |C| [operator()]: Unhandled C++ exception. vk::createInstance: ErrorIncompatibleDriver
```

I see why you're removing it from the required extensions but is there a reason to remove it from optional?

KhoraLee:
@squidbus Since we are embedding MoltenVK into our binary, I thought all users would not use the Vulkan Loader. I’ve reverted the optional part so that anyone who does have the Vulkan Loader installed will have no problem as well

PR #3657, opened 2025-09-03
modules/SceNpTus: Prevent mis-leading games into thinking TUS/TSS functionality is available by returning abort error codes

nadia1461:
This is the second fix following up from https://github.com/Vita3K/Vita3K/pull/3655
This resolves all blocking issues with [Lumines Electronic Symphony [PCSE00009]](https://github.com/Vita3K/compatibility/issues/693#top), which should bring the game from `InGame+` to `Playable`

Lumines has online scoreboard functionality which it uses the TSS/TUS libraries for.
When booting the game, it pulls back data from `sceNpTssGetSmallStorage` via this flow:

- Creates a TUS request with `sceNpTusCreateRequest`
- Attempts to pull two datetime strings from TSS via `sceNpTssGetSmallStorage`
- `sceNpTssGetSmallStorage` returns `UNIMPLEMENTED()` which is interpreted as a success code (0)
- Error checks have passed and the game attempts to parse several datetime values out of the TSS data buffer
- Before parsing the datetimes, an attempt is made to put a null terminator on the end of the TSS data buffer - `data[contentLength] = '\0'`

`sceNpTssGetSmallStorage` never actually initialized the provided `void *data` or `SceSize *contentLength`.
Because of that, this operation results in the following buffer-overwrite, due to a garbage, uninitialized `contentLength` value

`[ArmDynarmicCallback::MemoryWrite]: Invalid write of uint8_t at addr: 0xcd9c28f2, val = 0x0
PC: 0x8100d49e`

The game has error handling on all `sceNpTus*` / `sceNpTss*` calls.
If these calls return an error code, the game continues without issue and gracefully handles it.
The only impact is the lack of the online scoreboard, which is not critical to the game play.

I've aimed to have minimal potential impact on other games, so most of the TUS/TSS library has not been touched.
The intent here is only to communicate that there is an issue with the TUS/TSS initialization calls.

It may be desired to have all TUS/TSS calls return error codes until the networking side has been implemented, but I'm inexperienced in this codebase so I wanted to keep my impacts small.

I have pulled in error codes from the PS Vita SDK following the same format as other modules.
I have also added a small comment describing the intent of the current TSS/TUS implementation to help provide context for anyone looking to extend the functionality in the future.

Let me know if the comments should be removed, or anything stylistic changed.


Zangetsu38:
Suggestion, remove comment inside code, squash and rename commit in respect us style, look other commit

nadia1461:
Done. I've updated the title of the PR + commit to be in-line with each other, removed the comment inside the code, consolidated to a single commit, and updated the title to be less ambiguous

Zangetsu38:
@nadia1461 you have try this pr with other game using it ?
you can check with knyt undergrund

nishinji:
List of Games Using SceNpTus: https://github.com/search?q=repo%3Anishinji%2FVita3K_Logs+%22sceNpTus%22&type=code&p=1

Zangetsu38:
> List of Games Using SceNpTus: https://github.com/search?q=repo%3Anishinji%2FVita3K_Logs+%22sceNpTus%22&type=code&p=1

I have same in my repo, but little old xd

nadia1461:
I did run into an issue with double-checking Lumines just now. In the first revision of the new approach, I mistakenly returned error codes for the TSS Small Storage calls. I patched that out today, but only just now re-tested Lumines. It appears Lumines doesn't like the Empty String approach being used, so I'll need to poke around a bit further.

It will still work if an error code is returned for the TSS Small Storage calls. That is still a significantly smaller surface area for potential regressions, and I can test out the above listed games again with the `sceNpTssGetSmallStorage*` call returning an error code.

nadia1461:
I pushed a new build. This commit limits the scope of potentially affected games to 4 titles (discovered via [Vita3K_Logs](https://github.com/nishinji/Vita3K_Logs/tree/393029415e421074cb69e81c68c01cc30a855b44)/[compatibility](https://github.com/nishinji/Vita3K_Logs/tree/393029415e421074cb69e81c68c01cc30a855b44/compatibility))

Only these titles use any `sceNpTssGetSmallStorage*` API.
Other titles use `sceNpTus*`, but none of those calls have been modified.

The approach provides an empty, null-terminated string buffer back as `pBuffer`, with an error code.
This ensures the `char* pBuffer` pointer is not left potentially uninitialized as it previously was.
The error code picked does not particularly matter, but `SCE_NP_COMMUNITY_ERROR_NO_RESOURCE` was used as it seemed closest to the actual issue
This allows Lumines to skip the download of the TSS payload and continue to gameplay.

The error codes provided as enums are from the PSVita SDK Reference, and are the proper enum names + codes.
This follows the same pattern done in other modules like `SceAudiodecUser`

As can be seen by the testing performed below, this commit does not cause new regressions.
- Ecolibrium suffered a regression **before** this PR
- One Piece: Pirate Warriors 2 has a graphical issue on my PC, which was present **before** this PR. Gameplay is unaffected

## Motorstorm RC (PCSC00014) (Intro)
  - `sceNpTssGetSmallStorage` is used during the intro loading screen
  - Game does not progress beyond intro (before PR)
  - Game does not progress beyond intro. TSS call is observed to occur correctly (after PR)

Screenshot 2025-10-31 202455
Screenshot 2025-10-31 202511

## Ecolibrium (PCSA00087) (Ingame-)
  - Game does not currently load past intro screen (before PR)
  - Game does not currently load past intro screen (after PR)

Screenshot 2025-10-31 203227
Screenshot 2025-10-31 203247

## One Piece: Pirate Warriors 2 (PCSG00142) (Playable)
- `sceNpTssGetSmallStorage` is used during the initial loading screens where PSN connectivity occurs
- Game gets into game play without issue, but has some graphical issues where the UI is smaller than it should be (before PR)
- Game get into game play without issue, but has the same graphical issues as before (after PR)
Screenshot 2025-10-31 203737
Screenshot 2025-10-31 203748
Screenshot 2025-10-31 203924

## Lumines Electronic Symphony (PCSE00009) (Ingame+)
- Game freezes at intro loading screen at black screen with spinner (before PR)
- Game loads without freezing and a user can play the game without issue (after PR)
Screenshot 2025-10-31 204445
Screenshot 2025-10-31 204457
Screenshot 2025-10-31 204508

PR #3664, opened 2025-09-21
modules/SceCommonDialog: Fix 'check_save_file' crash on Linux

Dicot0721:
- When `iconPath` is empty string, `vfs::read_file` will try to open dictory `ux0` as a binary file to read, which have no effect on Windows but crash on *nix (Linux/Android/...).
- `std::ifstream` can open a directory as file on *nix (but can not on Windows), and after that `std::ifstream::tellg()` return the max value that the return type can represent (strangely not fail), which leads to throwing an exception when buffer is being resized.


EXtremeExploit (dismissed):

Zangetsu38:
> > is on what game ?
> 
> Known:
> 
>     * Sly Cooper: Thieves in Time™ [PCSA00068]
> 
>     * 仮面ライダー バトライド・ウォー 創生 [PCSG00675]

i see yes, sly cooper for empty save get len 0 but not null hmm
but sad, change it no fix empty icon, i need look here is strange

EXtremeExploit (approved):

EXtremeExploit (approved):

Issue #2892, opened 2023-09-01
Analog Calibration


Panderner:
Analog calibration is need for some games that has issues for most controllers with diagonal deadzone.

For example: Muramasa Rebirth, Evasion I is required to pass the tutorial. 
![Screenshot (163)](https://github.com/Vita3K/Vita3K/assets/37503397/09f7fcd7-9b6e-42af-8d70-03ce29bb311f)

Unable to perform the action for DualSense controller but it works for Keyboard.

GoldenRodArt:
> Analog calibration is need for some games that has issues for most controllers with diagonal deadzone.
> 
> For example: Muramasa Rebirth, Evasion I is required to pass the tutorial. ![Screenshot (163)](https://user-images.githubusercontent.com/37503397/265092420-09f7fcd7-9b6e-42af-8d70-03ce29bb311f.png)
> 
> Unable to perform the action for DualSense controller but it works for Keyboard.

I have the same issue :(

cobra838:
> I ran into this exact problem

Yes, the sticks on the gamepad should draw a square, not a circle, then everything works correctly.
It works:
![image](https://github.com/user-attachments/assets/65d4f760-f259-4ae6-8348-d428eed8d0b7)

It doesn't work:
![image](https://github.com/user-attachments/assets/7e6cacbe-26f3-44a0-bdf9-a0dbae434b85)

https://hardwaretester.com/gamepad

To solve this, you can use ds4windows:
![image](https://github.com/user-attachments/assets/87b007ec-899e-4fba-b090-e3641a1abfbe)

PR #3666, opened 2025-09-25
ctrl: add analog stick multiplier

nishinji:
The Vita apparently has a slightly larger analog input range compared to the DualShock, which causes issues (fe. P4G, the characters walk slowly diagonally).

relate to #2892


vita3k/ctrl/src/ctrl.cpp
```diff
@@ -240,6 +217,26 @@ std::array<ControllerBinding, 15> get_controller_bindings_ext(EmuEnvState &emuen
     } };
 }
 
+static float axis_to_axis(int16_t axis, EmuEnvState &emuenv) {
```
EXtremeExploit: No need to pass the entirety of emuenv here, just a float is enough

EXtremeExploit (changes requested):

PR #3668, opened 2025-09-26
renderer/vulkan: rebuild swapchain when surface size differs from window

### LINKED ISSUE #3632: [BUG]: Linux: Full screen cuts off UI

### Affected Build(s)

3806

### Description of Issue

In the latest nightly build, whenever you enter fullscreen mode or resize the window, the System dialogs are rendered larger than the actual screen thus making it unusable (this is particularly noticable when saving, see image) 

Image

### Expected Behavior

When entering full screen mode the UI should be completely visible in the full screen window

Image

### Reproduction Steps

1st: Enter full screen mode

### Log File

[vita3k.log](https://github.com/user-attachments/files/21236173/vita3k.log)

### System Configuration

CPU: AMD Ryzen 7 7735HS
GPU/Driver: Nvidia RTX 4060 laptop / 575.64.03
RAM: 16GB
OS: archlinux, linux 5.12.38

ekisu:
When a window is resized on Wayland, `eSuboptimalKHR` is not raised (unlike other platforms, apparently - see https://github.com/ocornut/imgui/issues/7671), which meant the surface was never resized.

To fix this, I've added a specific check to see if the surface still matches the current window size; and if it doesn't it rebuilds the swapchain.

Fixes #3632


Zangetsu38:
need rebase and fix conflict

ekisu:
@Zangetsu38 Updated it, fix still works. Not quite sure if the logic around `if (acquire_result == vk::Result::eErrorSurfaceLostKHR)` is 100% correct, but I'm not really sure how to test this scenario anyway :thinking:

PR #3670, opened 2025-10-03
modules/SceCommonDialog: Clamp slot_list_size in SaveDataDialog.

Zangetsu38:
# About
- modules/SceCommonDialog: Clamp slot_list_size in SaveDataDialog.
Clamp slot_list_size in Continue to not exceed the value set by Init, preventing vector overflow crash.
Clamp slot_list_size in Init to SLOTLIST_MAXSIZE (256) to respect SDK maximum.

- Fix crash in Sly Cooper: Thieves in Time when returning to Save List.
When selecting a save slot and cancelling the dialog, Continue would try to display the list
with a slotListSize of 32, exceeding the 31 slots set in Init, causing a vector overflow.
This clamps slot_list_size to prevent the crash.

### DIFF
--- a/vita3k/modules/SceCommonDialog/SceCommonDialog.cpp
+++ b/vita3k/modules/SceCommonDialog/SceCommonDialog.cpp
@@ -1106,9 +1106,9 @@ EXPORT(int, sceSaveDataDialogContinue, const SceSaveDataDialogParam *p) {
         emuenv.common_dialog.savedata.mode_to_display = SCE_SAVEDATA_DIALOG_MODE_LIST;
         list_param = p->listParam.get(emuenv.mem);
         if (list_param->slotListSize > 0)
-            emuenv.common_dialog.savedata.slot_list_size = list_param->slotListSize;
-        slot_list.resize(list_param->slotListSize);
-        for (std::uint32_t i = 0; i < list_param->slotListSize; i++) {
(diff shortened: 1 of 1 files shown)


vita3k/modules/SceCommonDialog/SceCommonDialog.cpp
```diff
@@ -1106,9 +1106,9 @@ EXPORT(int, sceSaveDataDialogContinue, const SceSaveDataDialogParam *p) {
         emuenv.common_dialog.savedata.mode_to_display = SCE_SAVEDATA_DIALOG_MODE_LIST;
         list_param = p->listParam.get(emuenv.mem);
         if (list_param->slotListSize > 0)
-            emuenv.common_dialog.savedata.slot_list_size = list_param->slotListSize;
```
Dicot0721, line 1109: If I understand correctly, this might be clearer (●◡●)
```suggestion
        if ((0 < list_param->slotListSize) && (list_param->slotListSize < emuenv.common_dialog.savedata.slot_list_size))
            emuenv.common_dialog.savedata.slot_list_size = list_param->slotListSize;
```

Zangetsu38: Put 0 in first have killed my eye lol
And so thx, but i more like using std min haha

EXtremeExploit (approved):

PR #3673, opened 2025-10-16
external: update fmt to v12

oltolm:
This is necessary to build with Clang 21. See https://github.com/fmtlib/fmt/issues/4543

### DIFF
--- a/vita3k/interface.cpp
+++ b/vita3k/interface.cpp
@@ -21,6 +21,7 @@
 
 #include <app/functions.h>
 #include <config/state.h>
+#include <ctime>
 #include <ctrl/functions.h>
 #include <ctrl/state.h>
 #include <dialog/state.h>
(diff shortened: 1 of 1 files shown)


vita3k/interface.cpp
```diff
@@ -536,8 +537,16 @@ static void take_screenshot(EmuEnvState &emuenv) {
...
+    struct tm localtime;
+#ifdef _WIN32
+    localtime_s(&localtime, &t);
+#else
+    localtime_r(&t, &localtime);
+#endif
+
     const auto img_format = emuenv.cfg.screenshot_format == JPEG ? ".jpg" : ".png";
-    const fs::path save_file = save_folder / fmt::format("{}_{:%Y-%m-%d-%H%M%OS}{}", string_utils::remove_special_chars(emuenv.current_app_title), fmt::localtime(std::time(nullptr)), img_format);
+    const fs::path save_file = save_folder / fmt::format("{}_{:%Y-%m-%d-%H%M%OS}{}", string_utils::remove_special_chars(emuenv.current_app_title), localtime, img_format);
```
Dicot0721, line 549: Using `std::chrono::system_clock::now()` might be better?

scribam (approved):

PR #3680, opened 2025-11-02
Support vertex streams with stride set to 0 in OpenGL backend

ZenoArrows:
If a stride of 0 is specified in OpenGL then the array is assumed to be tightly packed. This doesn't match GXM where a stride of 0 simply means to repeat the first element.

The equivalent in OpenGL is to disable the vertex attribute array and set the vertex attribute directly through `glVertexAttrib*`. Unfortunately the API isn't quite as flexible as the `glVertexAttribPointer`, so a switch case is required to select the function variant that matches the data type in the shader. To avoid making this code even longer all attributes are padded to 4 elements so we can always call the 4-element `glVertexAttrib4*` variant.

The values of attributes set through `glVertexAttrib*` don't refer to the array buffer, so technically we could avoid allocating memory in the storage buffer for them but it would require more drastic changes to this function.

Vulkan already supports this case as it allows for an arbitrary stride including 0, so no additional code is required for that backend.

### DIFF
--- a/vita3k/renderer/src/gl/sync_state.cpp
+++ b/vita3k/renderer/src/gl/sync_state.cpp
@@ -26,6 +26,7 @@
 #include <gxm/functions.h>
 #include <gxm/types.h>
 #include <util/align.h>
+#include <util/float_to_half.h>
 #include <util/log.h>
 #include <util/vector_utils.h>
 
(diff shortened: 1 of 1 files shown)


vita3k/renderer/src/gl/sync_state.cpp
```diff
@@ -518,14 +522,82 @@ void sync_vertex_streams_and_attributes(GLContext &context, GxmRecordState &stat
...
+                case GL_FLOAT:
+                    {
+                        GLfloat v[4] = { 0.0f, 0.0f, 0.0f, 1.0f };
+                        memcpy(v, data, component_count * sizeof(v[0]));
+                        glVertexAttrib4fv(attrib_location + i, v);
+                    }
+                    break;
+                case GL_INT:
+                    {
+                        GLint v[4] = { 0, 0, 0, 0 };
```
Zangetsu38, line 592: format is wrong for {}
```suggestion
                case GL_INT: {
                        GLint v[4] = { 0, 0, 0, 0 };
```

PR #3689, opened 2025-11-16
.github: update actions

nishinji:
Not important changes.

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -4,10 +4,10 @@ on: [push, pull_request]
 
 jobs:
   format-check:
-    runs-on: ubuntu-22.04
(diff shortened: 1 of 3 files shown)


bookmist:
Is it ok to increase ubuntu version? As far as I remember it was intentionally downgraded due to too new libc version. Vita3K did not run on old linux if compiled with ubuntu-latest or something like.

Dicot0721:
> Is it ok to increase ubuntu version? As far as I remember it was intentionally downgraded due to too new libc version. Vita3K did not run on old linux if compiled with ubuntu-latest or something like.

ninshinji only modified the runtime environment for CodeQL, and the action passed.
In fact, I suspect the "libc version" issue might be the same as #3645.

bookmist (approved):

PR #3706, opened 2025-12-21
build: Add Android version.

Zangetsu38:
# About 
- build: Add Android version.
PS button support added to the overlay controller, with improved default button positioning


vita3k/mem/src/mem.cpp
```diff
@@ -116,8 +116,8 @@ bool init(MemState &state, const bool use_page_table) {
     const BOOL ret = VirtualProtect(state.memory.get(), state.page_size, PAGE_NOACCESS, &old_protect);
     LOG_CRITICAL_IF(!ret, "VirtualAlloc failed: {}", get_error_msg());
 #else
-    const int ret = mprotect(state.memory.get(), state.page_size, PROT_NONE);
-    LOG_CRITICAL_IF(ret == -1, "mprotect failed: {}", get_error_msg());
+    // const int ret = mprotect(state.memory.get(), state.page_size, PROT_NONE);
```
EXtremeExploit: Instead of disabling mprotect for everything except windows, it should be disabled for android only here

vita3k/renderer/src/batch.cpp
```diff
@@ -128,12 +131,18 @@ void process_batches(renderer::State &state, const FeatureState &features, MemSt
                 return;
 
             // keep the old behavior for opengl with vsync as it looks like the new one causes some issues
-            if (state.current_backend == Backend::OpenGL && config.current_config.v_sync)
-                return;
+            // if (state.current_backend == Backend::OpenGL && config.current_config.v_sync)
```
EXtremeExploit: Does this still allow for vsync to work on opengl?

EXtremeExploit (changes requested):

vita3k/mem/src/mem.cpp
Zangetsu38: Ah yes sorry...

vita3k/renderer/src/batch.cpp
Zangetsu38: For this i have no idea... 🤔

EXtremeExploit (approved):
I think its fine, i'd like some external review just in case, this is huge

PR #3708, opened 2025-12-22
gui&config: Add support of custom config of backend-render.

Zangetsu38:
# About
- gui&config: Add support of custom config of backend-render.
Allow using custom config par app for backend render.
set variable ready using for custom driver.


EXtremeExploit (approved):
i think its fine

PR #3600, opened 2025-05-17
Migration to SDL3

bookmist:
Draft PR with migration to SDL3

A lot of renaming, some changes in imgui_impl_sdl, total rewriting of sdl_audio
I don't check controllers/gamepads. Buttons (ABXY) can be incorrect.
Also there may be problems with HiDPI

PR #3694, opened 2025-12-03
modules/SceCamera: SDL3 camera implementation

bookmist:
This PR implement camera in Vita3k
Camera can be actual web camera, static image or solid color. It can be configured in settings.

It's first time I made architectural decisions, so any help/notes/comments are welcome. Maybe, the code can be structured better.

Camera support is not totally correct, but it works in tested games.

Note. On Windows some web cameras can be inaccessible due to SDL limitations

Edit: Now ready for review.

PR #3709, opened 2025-12-23
motion: Add device motion support and fix elapsed timestamp handling.

Zangetsu38:
# About
 - motion: Add device motion support and fix elapsed timestamp handling.
Move Update(Orientation/Rotation) to refresh_motion to update once per refresh instead of per sensor event.
Fix elapsed time using sensor timestamps (ns -> us), with fallback when timestamp is 0.
Add support for device motion (accelerometer and gyroscope) and handle all native device rotations.

Fix regression caused by moving to sdl3 with using event #3600

### DIFF
--- a/vita3k/motion/include/motion/state.h
+++ b/vita3k/motion/include/motion/state.h
@@ -28,5 +28,10 @@ struct MotionState {
     uint64_t last_gyro_timestamp = 0;
     uint64_t last_accel_timestamp = 0;
 
+    bool has_device_motion_support = false;
     bool is_sampling = false;
+
+    void init();
+    void stop_sensor_sampling();
+    void start_sensor_sampling();
 };
--- a/vita3k/motion/src/motion.cpp
+++ b/vita3k/motion/src/motion.cpp
@@ -21,8 +21,151 @@
 
 #include <ctrl/state.h>
 
+#include <util/log.h>
+
 #include <SDL3/SDL_gamepad.h>
+#include <SDL3/SDL_sensor.h>
+
 #include <numbers>
+
+enum DeviceRotation : int32_t {
+    ROTATION_UNKNOWN = -1,
+    ROTATION_0,
+    ROTATION_90,
+    ROTATION_180,
+    ROTATION_270
+};
+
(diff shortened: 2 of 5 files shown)


vita3k/motion/src/motion.cpp
```diff
@@ -22,8 +22,113 @@
...
+#else
+constexpr bool is_device_landscape = true;
+#endif
+
+static void detect_device_motion_support(MotionState &state) {
+    state.device_accel_id = 0;
+    state.device_gyro_id = 0;
+
+    int num_sensors;
+    auto sensors_id = SDL_GetSensors(&num_sensors);
```
bookmist, line 72: sensors_id should be freed with [SDL_free](https://wiki.libsdl.org/SDL3/SDL_free)() when it is no longer needed.

vita3k/motion/include/motion/state.h
```diff
@@ -19,14 +19,26 @@
...
     std::mutex mutex;
     MotionInput motion_data;
     uint32_t last_counter = 0;
-    uint64_t last_gyro_timestamp = 0;
-    uint64_t last_accel_timestamp = 0;
+    uint64_t last_refresh_motion_timestamp = 0;
+
+    // gyroscope and accelerometer on the device itself
+    SDL_Sensor *device_gyro = nullptr;
+    SDL_Sensor *device_accel = nullptr;
```
bookmist: This can be changed to unique_ptr with custom destructor (SDL_CloseSensor) like this https://github.com/Vita3K/Vita3K/pull/3694/files#diff-8943f98aa2f4259c79bba5083cc50cf38d71d963e5ce72146807d339bc2ecd93R33-R47

vita3k/motion/src/motion.cpp
Zangetsu38: Ah i see

vita3k/motion/include/motion/state.h
Zangetsu38: is done

bookmist (approved):

PR #3735, opened 2026-01-01
host/dialog/filesystem android: Cleanup file picker path handling.

Zangetsu38:
# About
- host/dialog/filesystem android: Cleanup file picker path handling.
- Resolve real filesystem paths directly from the Android file picker.
- Remove all SAF fallback logic, temporary file copies, and FD-based handling.
- Drop file-descriptor mapping entirely.
- Unify file opening with desktop by using FOPEN().
- Simplify JNI: filedialogReturn now returns only a UTF-8 path.
- Add permission request on Android 11+ to ensure picker paths remain accessible.

### DIFF
--- a/vita3k/gui/src/pkg_install_dialog.cpp
+++ b/vita3k/gui/src/pkg_install_dialog.cpp
@@ -59,21 +59,27 @@ void draw_pkg_install_dialog(GuiState &gui, EmuEnvState &emuenv) {
         result = host::dialog::filesystem::open_file(pkg_path, { { "PlayStation Store Downloaded Package", { "pkg" } } });
         draw_file_dialog = false;
         if (result == host::dialog::filesystem::Result::SUCCESS) {
-            FILE *infile = host::dialog::filesystem::resolve_host_handle(pkg_path);
-            PkgHeader pkg_header{};
-            fread(&pkg_header, sizeof(PkgHeader), 1, infile);
-            fclose(infile);
-            std::string title_id_str(pkg_header.content_id);
-            std::string title_id = title_id_str.substr(7, 9);
-            const auto work_path{ emuenv.pref_path / fmt::format("ux0/license/{}/{}.rif", title_id, pkg_header.content_id) };
-            if (fs::exists(work_path)) {
-                LOG_INFO("Found license file: {}", work_path);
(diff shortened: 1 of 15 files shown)


vita3k/util/src/fs_utils.cpp
```diff
@@ -57,6 +57,17 @@ void dump_data(const fs::path &path, const void *data, const std::streamsize siz
...
 }
+
+FILE *open_file_handle_from_path(const fs::path &path) {
+    FILE *result;
+#ifdef _WIN32
+    _wfopen_s(&result, path.c_str(), L"rb");
+#else
+    result = fopen(path.c_str(), "rb");
+#endif
+    return result;
```
bookmist:
```suggestion
return FOPEN(path.c_str(), "rb");
```

Zangetsu38: Ha 🗿 ok 👌  thx 😅
Baka me lol

Zangetsu38: Is done 👌

vita3k/gui/src/pkg_install_dialog.cpp
```diff
@@ -59,21 +59,27 @@ void draw_pkg_install_dialog(GuiState &gui, EmuEnvState &emuenv) {
...
+                    fs::ifstream binfile(work_path, std::ios::in | std::ios::binary | std::ios::ate);
+                    zRIF = rif2zrif(binfile);
+                    ImGui::OpenPopup("install");
+                    state = State::INSTALL;
+                } else {
+                    ImGui::OpenPopup("install");
+                }
             } else {
-                ImGui::OpenPopup("install");
+                LOG_CRITICAL("Failed to load pkg file in path: {}", pkg_path.generic_path().string());
```
bookmist: Do not use `.string()`. For logs nothing is ok
```suggestion
                LOG_CRITICAL("Failed to load pkg file in path: {}", pkg_path.generic_path());
```

Zangetsu38: I have just copy past other line for log, look another open file, all using this like that 🤔

bookmist (approved):

PR #3739, opened 2026-01-03
Ci linux arm64/aarch64 zip appimage support 

mrcmunir:
Hello Second attempt at CI support, updated to the current branch, taking advantage of the fact that the Android part has already been merged.

I set clang-15 as the default on Linux because it doesn't compile on arm64 with the default compiler that comes with Ubuntu 22.04.
Linux scripts are multiarch friendly, so if more platforms are supported in the future, it's much easier to maintain them now.

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -25,11 +25,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-22.04, windows-latest, macos-latest, android]
+        os: [ubuntu-22.04, ubuntu-22.04-arm, windows-latest, macos-latest, android]
         config: [Release]
         version: [zip, appimage, apk]
         include:
           - os: ubuntu-22.04
+            arch: x86_64
             version: appimage
             cache_path: ~/.ccache
             extra_cmake_args: -DLINUXDEPLOY_COMMAND=/usr/local/bin/linuxdeploy-x86_64.AppImage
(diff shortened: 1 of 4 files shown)


.github/workflows/c-cpp.yml
```diff
@@ -85,12 +96,23 @@ jobs:
...
+        if: startsWith(matrix.os, 'ubuntu-22.04')
         run: |
+          # easy update clang linux os from external repo for specific clang version
+          wget https://apt.llvm.org/llvm.sh
+          chmod +x ./llvm.sh
+          sudo ./llvm.sh 15
+          sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-15 100
+          sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-15 100
+          sudo update-alternatives --set clang /usr/bin/clang-15
+          sudo update-alternatives --set clang++ /usr/bin/clang++-15
```
Dicot0721: It should be separated into a step specific to Linux ARM64.

.github/workflows/c-cpp.yml
```diff
@@ -147,6 +169,15 @@ jobs:
...
         if: matrix.os == 'ubuntu-22.04' && matrix.version == 'appimage'
 
+      - name: Set up linuxdeploy Appimage (ubuntu-22.04 ARM64)
+        run: |
+          if [[ ! -e linuxdeploy-aarch64.AppImage ]]; then
+            curl -sLO https://github.com/linuxdeploy/linuxdeploy/releases/latest/download/linuxdeploy-aarch64.AppImage
+          fi
+          sudo cp -f linuxdeploy-aarch64.AppImage /usr/local/bin/
+          sudo chmod +x /usr/local/bin/linuxdeploy-aarch64.AppImage
+        if: matrix.os == 'ubuntu-22.04-arm' && matrix.version == 'appimage'
```
Dicot0721: You could set the value of the variable `ARCH` to `x86_64` or `aarch64` based on the value of `matrix.os`, then use `$ARCH`/`${ARCH}`. The extra step is unnecessary.

mrcmunir: Done

.github/workflows/c-cpp.yml
mrcmunir: Will be done.

mrcmunir:
It's already updated with the suggestions with push force

Dicot0721 (approved):

Issue #3604, opened 2025-05-20
[Feature Request]: Add the ability to reset game "Last Time Played" via right-click menu


MontyJack2020:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What feature are you suggesting?

I would like to be able to reset a game's "Last Time Played" counter by being able to right-click on the title via the menu and have an option to reset it there.

### Why would this feature be useful?

It would let me set all games on the menu to a clean state after play testing them so I know which ones to return to and also avoid me from setting up the emulator from scratch in order to clear out this data.

### Other emulators or software with a similar feature

PCSX2

### Platform (if relevant)

Windows

PR #3745, opened 2026-01-04
Add the ability to reset game "Last Time Played" via right-click menu

iWumboUWumbo2:
Adding the following feature https://github.com/Vita3K/Vita3K/issues/3604

### DIFF
--- a/vita3k/regmgr/src/regmgr.cpp
+++ b/vita3k/regmgr/src/regmgr.cpp
@@ -18,6 +18,7 @@
 #include <array>
 #include <regex>
 #include <regmgr/functions.h>
+#include <sstream>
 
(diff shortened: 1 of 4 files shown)


vita3k/regmgr/src/regmgr.cpp
```diff
@@ -18,6 +18,7 @@
 #include <array>
 #include <regex>
 #include <regmgr/functions.h>
+#include <sstream>
```
Dicot0721, line 21: What is the purpose of this change?

iWumboUWumbo2: My compiler was complaining because stringstream was being used without the proper include.

PR #3752, opened 2026-01-07
cmake: symlink compile_commands.json to project root

eagleflo:
`clangd` doesn't find the `compile_commands.json` already being generated inside `build` by default.

Symlinking (or copying on Windows) it to project root works out of the box.


EXtremeExploit:
This isn't really needed anymore

PR #3753, opened 2026-01-08
renderer/vulkan: skip surface cache for old entries

### LINKED ISSUE #2733: Trails in the Sky FC Evolution [PCSG00488] red screen when entering/leaving a battle/building

eagleflo:
While investigating #2733, I noticed that Vita3K would render an _ancient_ frame from a previous scene transition (with a red hue) whenever a new scene transition occurred.

I started investigating the texture & surface caching code and realized that if you just skip the whole cache when there is a big delay between frames it makes the issue go away by falling back to the texture cache and guest memory instead.

I tried to mirror the existing style and found another place where 60 frames was used as the threshold for an old frame.

@Macdu, you're the author of this code -- does this sound at all reasonable to you, or is there some much better way to fix this? Will this break other games?

Fixes #2733.

### DIFF
--- a/vita3k/renderer/src/vulkan/surface_cache.cpp
+++ b/vita3k/renderer/src/vulkan/surface_cache.cpp
@@ -328,6 +328,14 @@ std::optional<TextureLookupResult> VKSurfaceCache::retrieve_color_surface_as_tex
     if (!overlap)
         return std::nullopt;
 
+    // If the surface hasn't been rendered to for a long time, don't return stale GPU data.
+    // Let the texture cache read from guest memory instead.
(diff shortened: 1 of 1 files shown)


eagleflo:
While this mostly fixes the linked issue, I did find a place in the game where you can run through two doors to trigger two scene transitions in less than 60 frames and the dreaded red screen flash from surface cache reappears. So the time-based heuristic might not be good enough.

I'm now trying to figure out if there is a better way to do cache invalidation within the surface cache by working out what _exactly_ the game does during a scene transition.

PR #3758, opened 2026-01-09
renderer/vulkan: Add dirty flag to surface cache

### LINKED ISSUE #2733: Trails in the Sky FC Evolution [PCSG00488] red screen when entering/leaving a battle/building

eagleflo:
While investigating #2733, I noticed that Vita3K would render an _ancient_ frame from a previous scene transition (with a red hue) whenever a large scene transition occurred.

This happens because `surface_cache` includes a stale texture from the previous transition. By adding `dirty` flag to the cache entries and trapping writes to texture memory we can skip the `surface_cache` for stale textures and pass through to `texture_cache` instead.

I also noticed that `can_mprotect_mapped_memory` seems to just work on Steam Deck? I'd be very curious what kind of problems there have been before.

Fixes #2733.

### DIFF
--- a/vita3k/mem/include/mem/functions.h
+++ b/vita3k/mem/include/mem/functions.h
@@ -37,6 +37,16 @@ enum struct MemPerm {
     ReadWrite = ReadOnly | WriteOnly
 };
 
+constexpr MemPerm most_restrictive_perm(MemPerm a, MemPerm b) {
+    if (a == MemPerm::None || b == MemPerm::None)
+        return MemPerm::None;
+    if (a == MemPerm::ReadOnly || b == MemPerm::ReadOnly)
+        return MemPerm::ReadOnly;
+    if (a == MemPerm::WriteOnly || b == MemPerm::WriteOnly)
+        return MemPerm::WriteOnly;
+    return MemPerm::ReadWrite;
+}
+
 bool init(MemState &state, const bool use_page_table);
 Address alloc(MemState &state, uint32_t size, const char *name, Address start_addr = user_main_memory_start);
 Address alloc_aligned(MemState &state, uint32_t size, const char *name, unsigned int alignment, Address start_addr = user_main_memory_start);
--- a/vita3k/mem/src/mem.cpp
+++ b/vita3k/mem/src/mem.cpp
@@ -331,6 +331,7 @@ bool add_protect(MemState &state, Address addr, const uint32_t size, const MemPe
         protect.size = std::max(it->first + it->second.size, addr + protect.size) - start;
         addr = start;
         protect.blocks.merge(it->second.blocks); // transfer blocks to the new protect
+        protect.perm = most_restrictive_perm(protect.perm, it->second.perm);
 
         if (it == state.protect_tree.begin()) {
             state.protect_tree.erase(it);
@@ -341,7 +342,7 @@ bool add_protect(MemState &state, Address addr, const uint32_t size, const MemPe
         state.protect_tree.erase(it--);
     }
 
-    protect_inner(state, addr, protect.size, perm);
+    protect_inner(state, addr, protect.size, protect.perm);
 
     state.protect_tree.emplace(addr, std::move(protect));
     return true;


vita3k/renderer/src/vulkan/surface_cache.cpp
```diff
@@ -59,6 +60,32 @@ static bool format_need_additional_memory(SceGxmColorBaseFormat format) {
 
 namespace renderer::vulkan {
 
+static void protect_surface(MemState &mem, ColorSurfaceCacheInfo &info) {
+    const bool supports_surface_sync = (info.tiling == SurfaceTiling::Linear
+        && format_support_surface_sync(info.format));
+
+    uint32_t addr_start = align_down(info.data.address(), mem.page_size);
+    uint32_t addr_end = align(info.data.address() + info.total_bytes, mem.page_size);
```
Macdu: The range is already page-aligned inside the add_protect call.
Also, the protect has page granularity, meaning that it will trigger for any write in a page overlapping the surface range. This could cause regressions, in particular for small surfaces which can be marked as dirty without being modified. The way to work around this would be to only apply it to big surfaces and only to their interior pages.

vita3k/renderer/src/vulkan/surface_cache.cpp
```diff
@@ -59,6 +60,32 @@ static bool format_need_additional_memory(SceGxmColorBaseFormat format) {
...
+
+    // Use MemPerm::None to trap both reads and writes for surfaces that support sync,
+    // MemPerm::ReadOnly to trap only writes for other surfaces
+    MemPerm perm = supports_surface_sync ? MemPerm::None : MemPerm::ReadOnly;
+    MemPerm existing_perm;
+    if (is_protecting(mem, addr_start, &existing_perm))
+        perm = most_restrictive_perm(perm, existing_perm);
```
Macdu: The memory protecting system is quite basic, it will apply the new permission to all protected ranges overlapping the one being queried. Also the most_restrictive_perm function should be applied inside of add_protect instead of here.

vita3k/renderer/src/vulkan/surface_cache.cpp
```diff
@@ -188,6 +215,11 @@ SurfaceRetrieveResult VKSurfaceCache::retrieve_color_surface_for_framebuffer(Mem
             color_surface_queue.set_as_lru(&info);
         } else {
             color_surface_queue.set_as_mru(&info);
+            *info.dirty = false;
+
+            if (info.data)
+                protect_surface(mem, info);
```
Macdu: protect_surface is an expensive call and should only be called if the surface was dirty before

vita3k/renderer/src/vulkan/surface_cache.cpp
eagleflo: Reverted this logic to what was done for surface sync before, but not quite sure if that addresses what you bring up.

vita3k/renderer/src/vulkan/surface_cache.cpp
eagleflo: I moved this inside `add_protect`. I'm as of yet unsure what to do with the overlapping ranges.

vita3k/renderer/src/vulkan/surface_cache.cpp
eagleflo: Thanks for catching this, dumb error on my part.

vita3k/renderer/src/vulkan/surface_cache.cpp
```diff
@@ -275,27 +312,11 @@ SurfaceRetrieveResult VKSurfaceCache::retrieve_color_surface_for_framebuffer(Mem
...
-            addr_end = align(info_added.data.address() + info_added.total_bytes, KiB(4));
-        }
-        add_protect(mem, addr_start, addr_end - addr_start, MemPerm::None, [need_sync = info_added.need_surface_sync](Address addr, bool) {
-            *need_sync = true;
-            return true;
-        });
-    }
+    info_added.need_surface_sync = std::make_shared<bool>(false);
+    info_added.dirty.reset();
+    info_added.dirty = std::make_shared<bool>(false);
```
Macdu:
```suggestion
    info_added.dirty = std::make_shared<bool>(false);
```
No need to reset if you assign right after

vita3k/renderer/src/vulkan/surface_cache.cpp
```diff
@@ -59,6 +59,38 @@ static bool format_need_additional_memory(SceGxmColorBaseFormat format) {
 
 namespace renderer::vulkan {
 
+static void protect_surface(MemState &mem, ColorSurfaceCacheInfo &info) {
+    const bool supports_surface_sync = (info.tiling == SurfaceTiling::Linear
+        && format_support_surface_sync(info.format));
```
Macdu:
```suggestion
    const bool trap_reads = (info.tiling == SurfaceTiling::Linear
        && format_support_surface_sync(info.format));
```
Because we look at both reads and writes, the name support_suface_sync is a bit ambiguous (it's not whether we support but whether we apply it here)

vita3k/renderer/src/vulkan/surface_cache.cpp
```diff
@@ -59,6 +59,38 @@ static bool format_need_additional_memory(SceGxmColorBaseFormat format) {
...
+    uint32_t addr_end = align_down(info.data.address() + info.total_bytes, KiB(4));
+    if (addr_start >= addr_end) {
+        // we still need to protect something, even if it's not completely accurate
+        if (!supports_surface_sync) {
+            // Small surface with no inner pages - skip dirty tracking to avoid
+            // false positives from unrelated writes to the same page
+            return;
+        }
+        addr_start = align_down(info.data.address(), KiB(4));
+        addr_end = align(info.data.address() + info.total_bytes, KiB(4));
```
Macdu: Small surfaces are not dirty tracked only if we are also trapping reads. They should be not dirty tracked in any case.

Macdu (commented):
The PR is mostly fine from my point of view. However, as I said before, trapping memory does not work with mesa drivers on Linux (it causes a crash) when using the external memory method (I believe this is also the case for page table). 

Did you check that this is no longer the case? If not, the can_mprotect_mapped_memory check should not be removed (but you can additionally check which mapping method is used when setting it).

vita3k/renderer/src/vulkan/surface_cache.cpp
eagleflo: I was wondering why we did this for `need_surface_sync` above... you're right, it shouldn't be necessary. I'll remove.

vita3k/renderer/src/vulkan/surface_cache.cpp
eagleflo: You're right, `trap_reads` is much better. Will fix.

PR #3765, opened 2026-01-10
patch: slightly modified

nishinji:
Attempt to get closer to vitagrafix
The following file, which slightly modifies vitagrafix_patchlist, works.
https://github.com/nishinji/vita3K_patches/blob/main/patchlist.txt

### DIFF
--- a/vita3k/patch/src/patch.cpp
+++ b/vita3k/patch/src/patch.cpp
@@ -21,21 +21,20 @@
 
 #include <util/fs.h>
 #include <util/log.h>
+#include <util/string_utils.h>
 
 std::vector<Patch> get_patches(fs::path &path, const std::string &titleid, const std::string &bin) {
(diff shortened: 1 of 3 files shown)


vita3k/patch/src/patch.cpp
```diff
@@ -85,6 +86,10 @@ Patch parse_patch(const std::string &patch) {
 
     // All following values (separated by spaces) are the values to be written
     std::string values = patch.substr(patch.find(' ') + 1);
+    // set vblank to 1(60Hz) for now
+    string_utils::replace(values, "4 - vblank", "3");
```
SpikeHD, line 90: Idk how I feel about hard-coding these vblank operations but other than that this seems good to me!

nishinji: you fix it later 🙃

PR #3770, opened 2026-01-11
ci: Add Windows ARM64 build

Dicot0721:
- Update ffmpeg to last commit.
- Simplify vcpkg installation with `--triplet=<triplet>`.
- Remove useless `if` in the job `Compute git short SHA`.
- Rename macOS/Windows x86_64 artifacts.
- Disable Discord Rich Presence on Windows ARM64.
- Add flags to disable inclusion of timestamps in PCH files with clang-cl. (See <https://reviews.llvm.org/D20867>)
- Enable Visual Studio for ARM64 in CMake presets.
- Fix the Windows update script still using '0.2.0' when parsing version.

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -43,12 +43,24 @@ jobs:
             cmake_preset: linux-ninja-clang-appimage
 
           - os: windows-latest
+            arch: x86_64
             cache_path: |
-                C:\vcpkg\installed
-                C:\vcpkg\packages
-                C:\Users\runneradmin\AppData\Local\ccache
+              C:\vcpkg\installed
+              C:\vcpkg\packages
+              C:\Users\runneradmin\AppData\Local\ccache
+            vcpkg_triplet: x64-windows-static-md
             extra_cmake_args: -DCMAKE_TOOLCHAIN_FILE=C:\vcpkg\scripts\buildsystems\vcpkg.cmake -DVCPKG_TARGET_TRIPLET=x64-windows-static-md
             cmake_preset: windows-ninja
+
+          - os: windows-11-arm
+            arch: arm64
+            cache_path: |
+              C:\vcpkg\installed
+              C:\vcpkg\packages
+              C:\Users\runneradmin\AppData\Local\ccache
+            vcpkg_triplet: arm64-windows-static-md
+            extra_cmake_args: -DCMAKE_TOOLCHAIN_FILE=C:\vcpkg\scripts\buildsystems\vcpkg.cmake -DVCPKG_TARGET_TRIPLET=arm64-windows-static-md
+            cmake_preset: windows-ninja-clang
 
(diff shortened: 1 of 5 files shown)


vita3k/script/update-windows.bat
```diff
@@ -9,38 +9,50 @@ echo ============================================================
 
 for /f "delims=" %%g in ('powershell "((Invoke-RestMethod https://api.github.com/repos/Vita3K/Vita3K/releases/latest -timeout 2).body.Split("\"`n"\") | Select-String -Pattern 'Vita3K Build:') -replace  'Vita3K Build: '"') do @set git_version=%%g
 if exist Vita3K.exe (
-   for /f "delims=" %%v in ('powershell "((Get-Item Vita3K.exe).VersionInfo.FileVersion) -replace '0.2.0.'"') do @set version=%%v
+    for /f "delims=" %%v in ('powershell "((Get-Item Vita3K.exe).VersionInfo.FileVersion) -replace '0.2.1.'"') do @set version=%%v
```
Zangetsu38: why adding space ? 🤔 
and good catch, i have forget it lol

Dicot0721: To match the indentation of 4 spaces lol

Zangetsu38: ah yes why i have set only 3 lol

.github/workflows/c-cpp.yml
```diff
@@ -43,13 +43,23 @@ jobs:
             cmake_preset: linux-ninja-clang-appimage
 
           - os: windows-latest
+            arch: x64
```
nishinji: use x86_64 for consistency

Dicot0721: It's not just a name but be used in build

nishinji: Then let's add an environment variable named `name`.

Dicot0721: How? And WHY? Architecture descriptions vary across different build systems and carry distinct meanings—for instance, `x86_64` and `arm64-v8a` on Android. Different OS use different build workflows, which should not cause confusion.

nishinji: How: add `name: x64`
Why:  As already explained, to maintain consistency,
is there any reason to call the same architecture by a different name?
It is more clear to literally add the `name` when the OS names architectures.

Dicot0721:
> I'm guessing it's not quite ready yet? Just tried latest build and got this error on launch on my Windows ARM Snapdragon X Elite device:

Well, it seems I need to create a Windows on ARM VM to debug 💀

Dicot0721:
> Gets further now, still crashes, but gets far enough to create a log file. Log file attached. [vita3k.log](https://github.com/user-attachments/files/24587673/vita3k.log)

Edit `config.yml` and change `backend-renderer` to `OpenGL`

PR #3773, opened 2026-01-12
Build: Add ad-hoc codesigning for macOS builds

### LINKED ISSUE #2184: [Feature Request] Codesigning on macOS
Since Apple is kind enough with `macOS on ARM requires applications to be signed.` I wonder, do you have any plans for codesigning the app on macOS?

shinra-electric:
### Issue:

Arm builds have recently been added to the CI, but they are currently not signed. 

The Arm builds will launch if the extended attributes are cleared manually, running `xattr -cr ./Vita3K.app` or by removing the quarantine attribute with`xattr -d com.apple.quarantine ./Vita3K.app` from the terminal, but this shouldn't be a required step for the average user. 

On macOS, x86 apps do not need to be signed. But it is a requirement for Arm apps, which is why it hasn't been an issue until now. 

### Solution: 
This PR adds an extra ad-hoc code-signing step to macOS builds. 
It has been added at the end, as any subsequent changes to the app bundle will invalidate the existing signature. 
This allows the Arm build to launch, and shouldn't make any difference to the x86 build. 

*The user will still have to authorise it by going through the security settings or by relaxing Gatekeeper controls.*

Closes #2184

### Testing:
- Ensure that the CI Arm build can launch without running `xattr`
- Ensure the CI x86 build still works on both Intel and Arm Macs 

### Note about codesigning:
There is a lot of confusion about codesigning on macOS apps, so just to clear this up: 
- This is ***not*** notarising. That requires a paid Apple Developer account and is only required when distributing through the App Store. Notarised apps can also bypass the annoying security approval step. 
- This is not codesigning with a developer account, which also requires a payment. 
- This is an ad-hoc signature which is good enough to allow the app to launch.


shinra-electric:
In addition to this, I see that the update script also runs `xattr` on the downloaded update. This should probably also be a codesigning step, but I can't test it locally

Dicot0721 (approved):

shinra-electric:
> Hope to resolve #3785 🙏

Unlikely, though fingers crossed. 

Autoupdating  has been broken on macOS for quite a while. I looked into it, but couldn’t identify the problem.

Edit:
I didn’t look at that issue properly, yeah this might fix it

gh0sti:
> > Hope to resolve #3785 🙏
> 
> 
> 
> Unlikely, though fingers crossed. 
> 
> 
> 
> Autoupdating  has been broken on macOS for quite a while. I looked into it, but couldn’t identify the problem.
> 
> 
> 
> Edit:
> 
> I didn’t look at that issue properly, yeah this might fix it

How are you getting around needing a dev account to get signing privileges from Apple?

gh0sti:
> Hope to resolve #3785 🙏

Why implement this commit instead of trying to fix the code signing? https://github.com/Vita3K/Vita3K/commit/855c34befa434b7186644fdc2243a8b8eef4c58e

Dicot0721:
> Why implement this commit instead of trying to fix the code signing? https://github.com/Vita3K/Vita3K/commit/855c34befa434b7186644fdc2243a8b8eef4c58e

This commit is intended to allow time for the fixes, since I'm not sure if this is the only bug, and whether this PR is effective.

shinra-electric:
> ...since I'm not sure if this is the only bug, and whether this PR is effective.

Just to clarify, this is what the PR does. 
- Both examples below are Arm builds from Github Actions
- The signature details are easily viewable with a QuickLook plugin called [Apparancy](https://www.mothersruin.com/software/Apparency/)

### Master: 
Screenshot 2026-01-18 at 14 04 40
Screenshot 2026-01-18 at 14 05 02

The ad-hoc signature from the Linker is invalidated by the time the app bundle is packaged, so the OS reports it as "broken" and won't launch it. This can be worked around by running `xattr -cr` on the app bundle through terminal.

### This PR: 

Screenshot 2026-01-18 at 14 05 28
Screenshot 2026-01-18 at 14 05 37

I have Gatekeeper set to [allow apps from anywhere](https://macreports.com/how-to-disable-gatekeeper-on-mac-and-enable-the-anywhere-option-for-installing-any-software/), which is why I get the option to open it. Regular users will have to authorise from the security settings.

Dicot0721:
Thanks for the explanation! Btw, is the signature generated the same for every build? Does macOS check for signature consistency?

shinra-electric:
> Thanks for the explanation! Btw, is the signature generated the same for every build? Does macOS check for signature consistency?

It is different for every build. As far as I'm aware an ad-hoc signature is just a basic hash of the whole app bundle (including executable, dylibs, resources, info.plist etc). When the OS tries to launch an app for the first time, it compares a newly calculated hash to the existing signature. If it matches, it's safe to launch. If it doesn't match, it means something has changed since it was signed and shouldn't be trusted. 

Running `xattr -cr` clears the "first run" attribute and skips the hash check, which is why it works around the issue. (Not sure which exact attribute it is, maybe the "downloaded from internet" attribute? Because I know the issue doesn't occur when building locally)

gh0sti:
> > Why implement this commit instead of trying to fix the code signing? [855c34b](https://github.com/Vita3K/Vita3K/commit/855c34befa434b7186644fdc2243a8b8eef4c58e)
> 
> This commit is intended to allow time for the fixes, since I'm not sure if this is the only bug, and whether this PR is effective.

shouldn't you disable macOS arm64 builds? I still see them as available to download and newer versions of vita3k are building with arm64.

shinra-electric:
@Dicot0721 would it be possible to get another reviewer for this?

nn9dev (approved):
Image

nn9dev:
@Dicot0721 in the issue above, someone not involved with development says it [works on their machine](https://github.com/Vita3K/Vita3K/issues/3908#issuecomment-4347599294). I think that's enough to merge this PR :)

Dicot0721 (approved):

imWatchful (approved):

PR #3777, opened 2026-01-14
 renderer/vulkan: Automatically select supported deep stencil 

ikhoeyZX:
most potato phones doesn't support D32S8 Deep stencil, that's why most old phone like PowerVR GE and older Mali driver can't run this emulator.


ikhoeyZX:
image
didn't work?

Yukkid:
Hi, could you optimize Persona 4 Golden for phones with 4GB of RAM and a Snapdragon 600+ processor? I really want to play it. =)

ikhoeyZX:
nah i give up, i will open new one

PR #3779, opened 2026-01-15
script & gui/vita3k_update: Revert & bugfixes

Dicot0721:
- Revert the macOS update link to x64 build because the ARM64 build is unstable.
- Fix using `.zip` suffix for macOS download.
- Use bash to fix error `read: arg count` in `update-linux.sh` (<https://unix.stackexchange.com/questions/581409>).
- Fix the zip name in `update-linux.sh` differing from that in GUI.


Dicot0721:
I'm not sure if i should change the link of ARM64 build in vita3k_update.cpp 🤔

Dicot0721:
@KhoraLee What do you think?

Zangetsu38 (approved):

PR #3782, opened 2026-01-16
gui/initial setup: refactor grant file access permissions.

Zangetsu38:
# About
- gui/initial setup: refactor grant file access permissions.
Rework storage permission handling during initial setup.
Block setup progression until storage access is granted.
Avoid invalid states and SAF-related issues on some Android devices.


Croden1999:
👍🏻 it is good for us, nice sir

Issue #3767, opened 2026-01-10
[BUG]: Vita3K deadlocks on GUI exit after playing a game


eagleflo:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Affected Build(s)

v0.2.1 3888

### Description of Issue

When you close the Vita3K window from window manager (not File -> Quit), it deadlocks instead of cleanly exiting. Window manager steps in and asks the user to terminate, generating a coredump.

I am surprised there is not an existing issue for this. I saw some people discussing this in https://github.com/Vita3K/Vita3K/issues/3278, but that is a separate issue.

### Expected Behavior

Clean exit, no coredump generated.

### Reproduction Steps

1. Start Vita3K and open a game
2. Exit by closing the window (rather than File -> Quit)
3. Vita3K deadlocks and needs to be terminated by the WM, creating a coredump

### Log File

Last lines in log:

[15:53:22.659] |I| [save_pipeline_cache]: Saving pipeline cache...
[15:53:22.659] |I| [save_pipeline_cache]: Pipeline cache saved
zsh: killed     ./build/linux-ninja-clang/bin/Release/Vita3K

### System Configuration

CPU: AMD Ryzen 7 5800X3D 8-Core Processor | 16 Threads | 4.157 GHz
GPU/Driver: NVIDIA GeForce RTX 4070 Ti SUPER / 590.48.1.0
RAM: Available ram memory: 32012 MiB
OS: Arch Linux

PR #3795, opened 2026-01-24
Further exit cleanup

eagleflo:
- Pause audio before destroying ThreadState
- Don't run thread end callbacks during teardown
- Wake up waiting threads in `exit_delete`

With these fixes イースVIII -Lacrimosa of DANA- (PCSG00881) now exits cleanly via `SDL_EVENT_QUIT`.

Additional improvements for #3767, although this was a crash instead of a deadlock.

### DIFF
--- a/vita3k/kernel/src/kernel.cpp
+++ b/vita3k/kernel/src/kernel.cpp
@@ -168,9 +168,9 @@ Ptr<Ptr<void>> KernelState::get_thread_tls_addr(MemState &mem, SceUID thread_id,
 
 void KernelState::exit_delete_all_threads() {
     const std::lock_guard<std::mutex> lock(mutex);
-    for (auto &[_, thread] : threads) {
-        thread->exit_delete();
-    }
+    for (auto &[_, thread] : threads)
+        // Skip end callbacks; running guest code can access torn-down state
+        thread->exit_delete(false);
 }
 
 void KernelState::pause_threads() {
--- a/vita3k/kernel/src/thread.cpp
+++ b/vita3k/kernel/src/thread.cpp
@@ -182,6 +182,13 @@ void ThreadState::exit_delete(bool exit) {
     } else {
         stop(*cpu);
     }
+
+    // Wake if thread waiting on status_cond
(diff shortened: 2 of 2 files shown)


vita3k/kernel/src/thread.cpp
```diff
@@ -175,6 +175,9 @@ void ThreadState::exit_delete(bool exit) {
 
     run_end_callback = exit;
 
+    if (status == ThreadStatus::wait)
+        update_status(ThreadStatus::run);
+
```
bookmist: I'm not sure about this. what about normal thread ending, for example from sceKernelExitDeleteThread

eagleflo: The point of this is to wake up threads that are blocked on `status_cond` by sync primitives (and not on `something_to_do`). Previously these threads would survive `exit_delete_all_threads`.

`sceKernelExitDeleteThread` is used by a thread to delete itself, so it would be already running and not be affected by this.

The other callsite is `sceKernelDeleteThread` which can be used for other threads, but right now that already checks that the other thread has `status == ThreadStatus::dormant` before it does anything.

I actually found out that there's a third way for a thread to get blocked, via `sceKernelWaitSignal`. I'll add handling for that and comments.

vita3k/kernel/src/kernel.cpp
```diff
@@ -168,9 +168,9 @@ Ptr<Ptr<void>> KernelState::get_thread_tls_addr(MemState &mem, SceUID thread_id,
 
 void KernelState::exit_delete_all_threads() {
     const std::lock_guard<std::mutex> lock(mutex);
-    for (auto &[_, thread] : threads) {
-        thread->exit_delete();
-    }
+    for (auto &[_, thread] : threads)
+        // Skip end callbacks; running guest code can access torn-down state
+        thread->exit_delete(false);
```
eagleflo, line 173: Not running callbacks in this case is a behavior change, but based on my reading `exit_delete_all_threads` is meant for teardown.

Please correct me if I'm mistaken.

PR #3799, opened 2026-01-26
Fix crackling audio with SDL3 audio backend in certain games

### LINKED ISSUE #3798: [BUG]: Crackling audio with SDL backend in certain games

### Affected Build(s)

v0.2.1 3911

### Description of Issue

Using the SDL audio backend, there is *massive* crackling and log spam with certain games.

### Expected Behavior

Audio should sound normal as it does with the `cubeb` backend.

### Reproduction Steps

Open 零の軌跡 Evolution [PCSG00042] (Trails from Zero) or 碧の軌跡 Evolution [PCSG00246] (Trails to Azure) and listen to the background music in main menu.

Similar crackle persists inside game as well.

### Log File

```
[02:42:01.954] |E| [audio_output]: SDL audio error: Parameter 'buf' is invalid
[02:42:01.965] |E| [audio_output]: SDL audio error: Parameter 'buf' is invalid
[02:42:01.976] |E| [audio_output]: SDL audio error: Parameter 'buf' is invalid
[02:42:01.986] |E| [audio_output]: SDL audio error: Parameter 'buf' is invalid
[02:42:01.997] |E| [audio_output]: SDL audio error: Parameter 'buf' is invalid
```

### System Configuration

CPU: AMD Ryzen 7 5800X3D 8-Core Processor | 16 Threads | 3.554 GHz
GPU/Driver: NVIDIA GeForce RTX 4070 Ti SUPER
RAM: 32012 MiB
OS: Arch Linux

eagleflo:
First commit helps with massive log spam, probably enough to slow the game down.

Second commit skips pacing sleep in audio thread for SDL3 audio backend.

Fixes #3798.

### DIFF
--- a/vita3k/audio/src/audio.cpp
+++ b/vita3k/audio/src/audio.cpp
@@ -69,6 +69,8 @@ AudioOutPortPtr AudioState::open_port(int nb_channels, int freq, int nb_sample) 
 }
 
 void AudioState::audio_output(ThreadState &thread, AudioOutPort &out_port, const void *buffer) {
+    if (!buffer)
+        return;
     adapter->audio_output(thread, out_port, buffer);
 
     uint64_t now = std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::system_clock::now().time_since_epoch()).count();


bookmist:
Why you calculate sleep threshold this way?  Maybe (adapter->device_buffer_samples) would be better. On linux device buffer can be much bigger then on windows.
Also problem may be in very rough sleeping via std::this_thread::sleep_for
Try to use SDL_DelayNS or SDL_DelayPrecise, maybe it helps. (I cant check this bug myself)

>It sounds like the "output all remaining sound" case with null buf remains valid for cubeb?

It's valid for Vita, so both backends should do the same in this case.

eagleflo:
(Just adding as a note here that I also did #3804 while researching this to bring the audio adapters closer to each other, but that's a big change. I'll keep this PR open to investigate a smaller, easier to review fix.)

I added some custom logging and this is the pattern that the two games I'm trying to fix are doing:

```
[23:54:27.602] |D| [audio_output]: Audio output called with buffer size 2048
[23:54:27.608] |D| [audio_output]: Audio output called with null buffer
[23:54:27.613] |D| [audio_output]: Audio output called with buffer size 2048
[23:54:27.618] |D| [audio_output]: Audio output called with null buffer
[23:54:27.624] |D| [audio_output]: Audio output called with buffer size 2048
[23:54:27.629] |D| [audio_output]: Audio output called with null buffer
[...]
```

I'm pretty sure this behavior is what's causing the crackle with the emulator. Other more normal games seem to never call `audio_output` with a null buffer, although I saw that some use a smaller buffer.

To explore this further, I quickly tried adding a `wait_for_drain` for both audio backends (using their cond vars) and calling that from `AudioState::audio_output` if we get a null buffer, but this extra blocking absolutely destroys the case above -- worse crackling than without.

Next I tried just doing an early return from `AudioState::audio_output` on null buffer and voilà! That fixes the crackling issue on its own, without any adverse effects on other games I have. https://github.com/eagleflo/Vita3K/commit/71a49467d12d82c1a7b06a169f6be7652947f4f8 if you want to test.

Now, I don't have a very large catalogue, but do you know of any games that *rely* on `sceAudioOutOutput` blocking when called with null buffer? I'm now somewhat doubtful that the real Vita blocks on null buffer at all, given that these two games still work flawlessly on original hardware.

PR #3804, opened 2026-01-29
audio: Rewrite SDL backend to use ring buffer & callbacks

eagleflo:
Model the SDL audio backend after cubeb: push audio to ring buffer, block when full. Callback pulls from ring buffer and wakes the producer when a slot is freed.

I basically just copied what @Macdu did in the cubeb backend here verbatim. This way both audio backends resemble each other. In my testing after this change both audio backends keep the ring buffers full, gracefully blocking the thread until audio is consumed. No underruns either.

This is arguably a better fix for #3798, but it is way more work to test and review than #3799. Works for me on Linux & pipewire, but I have limited ability to test on other platforms...

### DIFF
--- a/vita3k/audio/src/impl/sdl_audio.cpp
+++ b/vita3k/audio/src/impl/sdl_audio.cpp
@@ -20,6 +20,7 @@
 #include "kernel/thread/thread_state.h"
 
 #include <SDL3/SDL_audio.h>
+#include <SDL3/SDL_hints.h>
 
 #include "util/log.h"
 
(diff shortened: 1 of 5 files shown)


vita3k/audio/src/impl/sdl_audio.cpp
```diff
@@ -85,25 +103,46 @@ AudioOutPortPtr SDLAudioAdapter::open_port(int nb_channels, int freq, int nb_sam
...
     auto port = std::make_shared<SDLAudioOutPort>(stream, *this);
-    SDL_CHECK(SDL_SetAudioStreamGetCallback(stream.get(), SDLAudioAdapter::thread_wakeup_callback, port.get()));
+    SDL_CHECK(SDL_SetAudioStreamGetCallback(stream.get(), SDLAudioAdapter::audio_callback, port.get()));
     port->channels = nb_channels;
     port->len_microseconds = (nb_sample * 1'000'000ULL) / freq;
     port->len_bytes = nb_sample * nb_channels * sizeof(int16_t);
+
+    // SDL3 has no equivalent to cubeb_get_min_latency(), so we calculate buffer count
+    // based on a target latency. 25ms matches cubeb's PulseAudio safe minimum.
+    const int target_latency_ms = 25;
```
eagleflo, line 113: https://github.com/mozilla/cubeb/blob/484857522c73318c06f18ba0a3e17525fa98c608/src/cubeb_pulse.c#L764-766

eagleflo: https://github.com/PCSX2/pcsx2/blob/204829865dc91721a1d946b9ad222563c24edc0d/pcsx2/Host/AudioStreamTypes.h#L55

bookmist: on windows wasapi device_buffer_samples (in frames) is the same as cubeb_get_min_latency (in ms) and on my system it equals to 10 ms

eagleflo: Yeah, we explicitly ask the SDL to give us 512 frames above, by default it would give 1024. But SDL3 is very explicit that it might not honor the request: https://github.com/libsdl-org/SDL/blob/main/include/SDL3/SDL_hints.h#L337-L339. If we *do* get 512 I think 10 ms is then around the right value. Maybe this should be tied to `device_buffer_samples` -- at least we should not be using anything less.

`cubeb_get_min_latency` results in 1200 frames on my PC, maybe too conservative default now that we have PipeWire on Linux. I tested setting this to 10 ms and it works on my machine; it shrinks the ring buffer size to 2 compared to 4 on cubeb. It's low latency, but also the slightest jitter could easily cause underruns / crackling... didn't hear any in my limited testing, though.

It's a tradeoff between low latency and safety, and feels very hardware and OS dependent. Some emulators (used to?) have specific setting for "low-latency audio" which would match that 10 ms. I'm not sure if adding a new setting for this would make sense... maybe it would help people on potatoes that can't get as low as 10 ms.

PR #3820, opened 2026-02-10
vita3k/config: Changed the BGM disabled by default

nishinji:
As [macdu said](https://github.com/Vita3K/Vita3K/pull/3411#issuecomment-2439723570), this should clearly be disabled by default.
Also, there's a bug where it clacking sound if the at9 file path is invalid, which is extremely frustrating.

### DIFF
--- a/vita3k/gui/include/gui/state.h
+++ b/vita3k/gui/include/gui/state.h
@@ -186,7 +186,7 @@ struct User {
     std::string avatar = "default";
     gui::SortType sort_apps_type = gui::TITLE;
     gui::SortState sort_apps_state = gui::ASCENDANT;
-    bool system_music = true;
+    bool system_music = false;
     std::string theme_id = "default";
     bool use_theme_bg = true;
     std::string start_type = "default";
--- a/vita3k/main.cpp
+++ b/vita3k/main.cpp
@@ -328,9 +328,9 @@ int main(int argc, char *argv[]) {
         gui::pre_init(gui, emuenv);
         gui::init_bgm_player(emuenv.cfg.bgm_volume);
         if (!emuenv.cfg.initial_setup) {
-            emuenv.cfg.system_music.emplace(true);
+            emuenv.cfg.system_music.emplace(false);
             if (gui::init_bgm(gui, emuenv))
-                gui::switch_bgm_state(false);
+                gui::switch_bgm_state(true);
             while (!emuenv.cfg.initial_setup) {
                 wait_for_frame_done();
                 if (handle_events(emuenv, gui)) {


Zangetsu38:
you can close ,it, is fixed now 🙆‍♂️

Zangetsu38:
No, now is done, you can disable it quickely if you want, and issue with audio broken if file missing is fixed
Disable by default is no sense...
And set volume to 0 + disable in same time is totaly stupid 😐

vita3k/gui/include/gui/state.h
```diff
@@ -186,7 +186,7 @@ struct User {
     std::string avatar = "default";
     gui::SortType sort_apps_type = gui::TITLE;
     gui::SortState sort_apps_state = gui::ASCENDANT;
-    bool system_music = true;
+    bool system_music = false;
```
Zangetsu38, line 189: Ok, accorded to default setting of real psvita, and now sound issue glitch if file missing is fixed, revert this and and set bgm volume to 65 % 🙆‍♂️, oin vita is 66.67 % (20/30) but MEH... 
like that new user can see bgm is here, and get no big sound, and can disable when it want easy 🙆‍♂️

Zangetsu38 (changes requested):

Macdu (approved):

Zangetsu38:
Not request anymore, can now disable it in first boot if user no want

nishinji:
Why was it closed?
The goal was to change the default behavior, and the merged PR does not fully replace this PR.
Your usual stance of ignoring the overwhelming majority’s opinion is truly admirable.

Zangetsu38:
> Why was it closed? The goal was to change the default behavior, and the merged PR does not fully replace this PR. Your usual stance of ignoring the overwhelming majority’s opinion is truly admirable.

I didn't ignore anything; my PR was precisely there to provide a real solution to give users a choice at first startup and, above all, to inform them that this feature exists...

The majority opinion you're talking about, which was only about ten people, was based solely on a survey done at the time when there was a bug that caused a terrible sound that could be heard if there was no audio file present.

Except that this bug was fixed shortly afterwards, and therefore, disabling it by default no longer made any sense and was just stupid...

When users discovered this feature, I received a lot of feedback from people who, having heard it for the first time, reported liking it and saying it was even closer to the PS Vita version.

So once the bug was fixed, and the option added so that each user could disable and enable it as they saw fit, this PR was no longer necessary.
And only user i have see request disable it by default is some dev..., not real user...

In my opinion, wanting disable this feature by default, which doesn't correspond to the console's actual behavior, means you really must dislike the vibe or the console itself...

I provided a real solution, which corresponds to the actual behavior of the console, rather than a hidden forcing, which would have even hidden the very existence of this feature from all users; it's simply common sense.

nishinji:
> > Why was it closed? The goal was to change the default behavior, and the merged PR does not fully replace this PR. Your usual stance of ignoring the overwhelming majority’s opinion is truly admirable.
> 
> I didn't ignore anything; my PR was precisely there to provide a real solution to give users a choice at first startup and, above all, to inform them that this feature exists...
> 
> The majority opinion you're talking about, which was only about ten people, was based solely on a survey done at the time when there was a bug that caused a terrible sound that could be heard if there was no audio file present.
> 
> Except that this bug was fixed shortly afterwards, and therefore, disabling it by default no longer made any sense and was just stupid...
> 
> When users discovered this feature, I received a lot of feedback from people who, having heard it for the first time, reported liking it and saying it was even closer to the PS Vita version.
> 
> So once the bug was fixed, and the option added so that each user could disable and enable it as they saw fit, this PR was no longer necessary. And only user i have see request disable it by default is some dev..., not real user...
> 
> In my opinion, wanting disable this feature by default, which doesn't correspond to the console's actual behavior, means you really must dislike the vibe or the console itself...
> 
> I provided a real solution, which corresponds to the actual behavior of the console, rather than a hidden forcing, which would have even hidden the very existence of this feature from all users; it's simply common sense.

image
Saying, “It’s a great feature, so we’ve forced it to be enabled!” and activating unnecessary features is exactly the same approach Apple and Microsoft take.
The background music playback feature is completely unnecessary for users on PCs or smartphones rather than consoles. Since I always use apps other than the emulator to play music, having music start playing just because I launch the GUI for debugging purposes is utter nonsense and annoying.
The developers all agreed with this change because they understood my point of view, and many users on Discord agree as well. As an option is provided, users will likely notice this feature, and there is no problem with disabling it by default, as many users have requested.
By the way, I always had background music playing on my console.

Zangetsu38:
> image Saying, “It’s a great feature, so we’ve forced it to be enabled!” and activating unnecessary features is exactly the same approach Apple and Microsoft take. The background music playback feature is completely unnecessary for users on PCs or smartphones rather than consoles. Since I always use apps other than the emulator to play music, having music start playing just because I launch the GUI for debugging purposes is utter nonsense and annoying. The developers all agreed with this change because they understood my point of view, and many users on Discord agree as well. As an option is provided, users will likely notice this feature, and there is no problem with disabling it by default, as many users have requested. By the way, I always had background music playing on my console.

You just said it yourself, this is only for developers here...
Not for end users, who don't launch the program with music playing, it doesn't make sense, and who therefore appreciate this music which gives them a similar vibe to the console.

And again, you conducted the survey when the bug was present and on a very small sample of users...
So right there, the survey is biased and in no way represents the users' wishes...

And no, the option is in the app's system settings, so if you don't look, you won't see it.

But I plan to fix this by doing something when each user is created.
With v3kn things...

imWatchful:
Developers are also primary end users, and as Macdu has approved this PR there is no need to discuss further. Will merge after rebase.

Zangetsu38:
> Developers are also primary end users, and as Macdu has approved this PR there is no need to discuss further. Will merge after rebase.


I don't understand this obsession with wanting to disable or destroy everything I do to improve the user experience...
And no, developers are absolutely not end users...
Except if someone finishes the game on it :) Developers just code, test, and fix bugs; they rarely play normally like normal users...

And i don't see the connection with Macdu here?
Who, moreover, accepted it simply because they didn't like the background music, just like you, nothing more...
And furthermore, it was based on a survey with few users at the time the background music had a problem when no audio files were present...
So the survey is no longer even valid...

This is therefore anything but objective...

If I say that I provided a solution to correct what was intended, then indeed, there's no longer any need to discuss it, and this request is therefore no longer required...

vita3k/main.cpp
```diff
@@ -328,9 +328,9 @@ int main(int argc, char *argv[]) {
         gui::pre_init(gui, emuenv);
         gui::init_bgm_player(emuenv.cfg.bgm_volume);
         if (!emuenv.cfg.initial_setup) {
-            emuenv.cfg.system_music.emplace(true);
+            emuenv.cfg.system_music.emplace(false);
             if (gui::init_bgm(gui, emuenv))
-                gui::switch_bgm_state(false);
+                gui::switch_bgm_state(true);
```
Zangetsu38, line 333: Why have changed this too ?
code no have sense anymore now 💀 

init bgm no start playing... so here you set pause on things not started, is stupid
Is crazy like you hate bgm 💀 

All is just changed like noob here..., don't touch my code if you no undestund it...

Zangetsu38:
And you have merged with keep this false information
>Also, there's a bug where it clacking sound if the at9 file path is invalid, which is extremely frustrating.

Bug what you said is already fixed from long time ago now...

PR #3824, opened 2026-02-13
gui/app_context_menu: Disable the report creation function for ANDROID

nishinji:
For my mental health :pleading_face:

### DIFF
--- a/vita3k/gui/src/app_context_menu.cpp
+++ b/vita3k/gui/src/app_context_menu.cpp
@@ -572,7 +572,7 @@ static void update_last_time_app_used_helper(GuiState &gui, EmuEnvState &emuenv,
 }
 
 void reset_last_time_app_used(GuiState &gui, EmuEnvState &emuenv, const std::string &app) {
-    update_last_time_app_used_helper(gui, emuenv, app, (time_t)0);
+    update_last_time_app_used_helper(gui, emuenv, app, static_cast<time_t>(0));
 }
 
 void update_last_time_app_used(GuiState &gui, EmuEnvState &emuenv, const std::string &app) {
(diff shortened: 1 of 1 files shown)


vita3k/gui/src/app_context_menu.cpp
```diff
@@ -739,10 +739,10 @@ void draw_app_context_menu(GuiState &gui, EmuEnvState &emuenv, const std::string
                         if (ImGui::MenuItem(lang.compat["copy_vita3k_summary"].c_str()))
                             copy_vita3k_summary();
                         if (ImGui::MenuItem(lang.compat["open_state_report"].c_str())) {
-                            copy_vita3k_summary();
```
Zangetsu38, line 742: keep it, is not error, i have set it for good reason :)

nishinji: I understand this was added intentionally, but it should be removed.
Who would expect the clipboard to be overwritten when pressing a button labeled “Open Link”?
At the very least, a button to copy information to the clipboard exists directly above it, so usability isn't compromised.

vita3k/gui/src/app_context_menu.cpp
```diff
@@ -739,10 +739,10 @@ void draw_app_context_menu(GuiState &gui, EmuEnvState &emuenv, const std::string
...
                             open_path(fmt::format("{}/{}", ISSUES_URL, gui.compat.app_compat_db[title_id].issue_id));
                         }
                     } else {
+#ifndef ANDROID
```
Zangetsu38: using good version here:
```suggestion
#ifndef __ANDROID__
```

PR #3835, opened 2026-02-23
Revert "renderer/vulkan: Change the automatic detection order of GPUs"

Dicot0721:
- Reverts Vita3K/Vita3K#3776.
- #3776 may break on some Windows device when using D3D.

### DIFF
--- a/vita3k/renderer/src/vulkan/renderer.cpp
+++ b/vita3k/renderer/src/vulkan/renderer.cpp
@@ -35,8 +35,6 @@
 
 #include <SDL3/SDL_vulkan.h>
 
-#include <ranges>
-
 #ifdef __APPLE__
 #include <MoltenVK/mvk_vulkan.h>
 #endif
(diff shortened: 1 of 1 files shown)


vita3k/renderer/src/vulkan/renderer.cpp
```diff
@@ -536,13 +534,14 @@ bool VKState::create(SDL_Window *window, std::unique_ptr<renderer::State> &state
...
-                if (device.getProperties().deviceType == vk::PhysicalDeviceType::eDiscreteGpu)
-                    break;
+            for (const auto &device : physical_devices) {
+                if (device_is_compatible(device)) {
+                    physical_device = device;
+
+                    // if it is an integrated gpu, try to find a discrete one
+                    if (device.getProperties().deviceType == vk::PhysicalDeviceType::eDiscreteGpu)
+                        break;
+                }
```
Zangetsu38, line 544:
```suggestion
            vk::PhysicalDevice fallback_device;
            for (auto it = physical_devices.rbegin(); it != physical_devices.rend(); ++it) {
                if (!device_is_compatible(*it))
                    continue;

                const auto props = it->getProperties();
                if (!fallback_device)
                    fallback_device = *it;

                const bool is_microsoft_software = props.vendorID == 0x1414
                    || std::string_view(props.deviceName).find("Microsoft") != std::string_view::npos;
                if (is_microsoft_software)
                    continue;

                physical_device = *it;

                // if it is a discrete gpu, stop here
                if (props.deviceType == vk::PhysicalDeviceType::eDiscreteGpu)
                    break;
            }

            if (!physical_device && fallback_device)
                physical_device = fallback_device;
        }
```

should be fine like that for what you want too

PR #3836, opened 2026-02-24
renderer/vulkan: Change GPU priority

Dicot0721:
- Old: discrete > other.
- New: discrete > integrated > other.


mrcmunir (approved):
Tested under linux arm64 helped and fixes priority GPU returned now 1 position instead llvmpipe  if are present in the system.

PR #3840, opened 2026-02-25
renderer/vulkan: Drain VK wait thread queue in sceGxmFinish

olistarri:
`renderer::finish()` only waited for the renderer command thread (via Nop), but the VK wait thread that writes notification values after GPU fences runs asynchronously. This let `sceGxmFinish` return before notifications were fully written, so late writes could overwrite values the game already reset, causing deadlocks on scene transitions.

Fixes Sly Cooper: Thieves in Time (PCSF00270 / PCSA00097) black screen on Vulkan during scene transitions.

Tested on Linux x86, plays through intro and into Episode 1 with no issues.

### DIFF
--- a/vita3k/renderer/src/sync.cpp
+++ b/vita3k/renderer/src/sync.cpp
@@ -25,6 +25,7 @@
 #include <renderer/gl/functions.h>
 #include <renderer/vulkan/functions.h>
 #include <renderer/vulkan/state.h>
+#include <renderer/vulkan/types.h>
 
 #include <renderer/functions.h>
 #include <util/tracy.h>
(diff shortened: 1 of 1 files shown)


vita3k/renderer/src/sync.cpp
```diff
@@ -93,6 +94,12 @@ COMMAND(new_frame) {
 void finish(State &state, Context *context) {
     // Add NOP then wait for it
     renderer::send_single_command(state, context, renderer::CommandOpcode::Nop, true, 1);
+
+    // Wait for the VK wait thread to drain pending notification writes.
+    if (state.current_backend == Backend::Vulkan) {
+        auto &vk_state = reinterpret_cast<vulkan::VKState &>(state);
+        vk_state.request_queue.wait_empty();
+    }
```
Macdu, line 106: Nice catch, although this is not enough (you wait until the request queue is empty, but the last request that was on the queue might still be getting processed).
Pushing a dummy request (a CallbackRequest with an empty callback) before calling wait_empty should fix it.

Dicot0721:
> Nitpick: `reinterpret_cast` isn't the correct cast here. Use `static_cast`.

`reinterpret_cast` to a derived class from its base at zero offset behaves the same as `static_cast`. 🤔

vita3k/renderer/src/sync.cpp
olistarri: Yeah, makes sense. On further testing I was getting some inconsistent behavior, fixed and pushed.

okt04175:
After this change, SOUL SACRIFICE DELTA™ [PCSD00079] black screen with Vulkan mode.

PR #3841, opened 2026-02-27
renderer: Improve transfer operation flow

Zangetsu38:
# About
- renderer: Improve transfer operation flow
Fix crash when app runs transfer finish during an operation.

# Result
- fix one game crashing when it start transferdownscaling and call transfer finish just after, causing crash on ptr get unmaped

### DIFF
--- a/vita3k/modules/SceGxm/SceGxm.cpp
+++ b/vita3k/modules/SceGxm/SceGxm.cpp
@@ -5435,6 +5435,7 @@ EXPORT(int, sceGxmTransferFill, uint32_t fillColor, SceGxmTransferFormat destFor
 EXPORT(int, sceGxmTransferFinish) {
     TRACY_FUNC(sceGxmTransferFinish);
     // same as sceGxmFinish
+    renderer::wait_for_transfer_ops(*emuenv.renderer);
     renderer::finish(*emuenv.renderer, nullptr);
 
     return 0;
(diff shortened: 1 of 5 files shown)


Zangetsu38:
> #3840 should already fix this issue in a much nicer way.

Are you sure? The problem with this game has nothing to do with Vulkan; it just executes the transferfinish immediately after the downscell call and ends up doing a gxmUnmap, so the PTR gets canceled mid-operation. It's a general problem.

Like you see here in log and in ghidra





[16:28:12.426] |D| [export_sceGxmTransferDownscale]: Downscaling image from 60000000 to 80418000, size 960x544
[16:28:12.426] |D| [export_sceGxmTransferFinish]: Finishing transfer operations
[16:28:12.427] |D| [export_sceGxmUnmapMemory]: Unmapping GPU memory at address 80418000
[16:28:12.427] |D| [renderer::cmd_handle_transfer_downscale::<lambda_1>::operator ()]: Downscale operation from 0x60000 to 0x60000
[16:28:12.430] |W| [renderer::cmd_handle_transfer_downscale::<lambda_1>::operator ()]: Downscale destination address range invalid for format 0x60000 (dst=0x80418000-0x8049FFFF), skipping writeback

Macdu:
Calling renderer::finish should wait for all gxm operations (including transfer operations) to be done. 

The issue until now was that on vulkan, the transfer operations were done on the request thread which renderer:finish did not wait for.  The PR mentioned above fixes this.

PR #3845, opened 2026-03-06
Added support for alternate key bindings for keyboard controls

Kumiko7:
This PR implements support for secondary key bindings within the keyboard control settings. Users can now assign two different physical keys to a single emulated Vita button, or config hotkey.
Changes
UI Enhancements: Updated controls_dialog.cpp to include a second column for "Alternate Bindings."
Input Logic: Refactored the input mapping system to check for multiple key assignments per action.
Configuration: Updated the config saving/loading logic to ensure secondary bindings persist across sessions.

This improves accessibility and flexibility. For example, a user can bind "Cross" to both Space and Enter, or set up a secondary layout for specific gameplay styles (like switching between WASD and Arrow keys) without reconfiguring the entire menu.


vita3k/ctrl/src/ctrl.cpp
```diff
@@ -420,4 +420,4 @@ int ctrl_get(const SceUID thread_id, EmuEnvState &emuenv, int port, SceCtrlData2
     }
 
     return nb_returned_data;
-}
+}
```
Zangetsu38: ??

Kumiko7: somehow trailing newlines got removed, fixed

PR #3860, opened 2026-04-03
mem: Improve memory handling on 16k page devices

nn9dev:
Intended to supersede #2843

Currently, arm64 MacOS crashes after launching Gravity Rush, with Dynarmic throwing memory write errors, which were not properly being caught by the exception handler, perhaps due to memory being incorrectly counted for and protected.

Hardcoded 4KiB values were causing incorrect page table indexing and  `mprotect` calls, and incorrect page-size stride math in the file read ""intentional fault"" loop.

In this PR, these values are replaced with the actual page size of the device, and things now seem to work as intended.

I am not sure if any of this is proper, but it doesn't crash after load anymore... :)

### DIFF
--- a/vita3k/io/src/state_functions.cpp
+++ b/vita3k/io/src/state_functions.cpp
@@ -37,8 +37,18 @@ SceOff FileStats::read(void *input_data, const int element_size, const SceSize e
     // that's because host io does not work well with memory trapping and read-only buffer
     // so set 1 byte to 0 in all pages to trigger all possible pagefaults in this range
     // todo: call a mem function to check this instead
+    
+#ifdef _WIN32
+    SYSTEM_INFO system_info = {};
+    GetSystemInfo(&system_info);
+    uint32_t page_size = system_info.dwPageSize;
+#else
+    uint32_t page_size = static_cast<int>(sysconf(_SC_PAGESIZE));
+#endif
+    //LOG_WARN("Page Size: {}", page_size);
+    
     volatile uint8_t *input_addr = reinterpret_cast<volatile uint8_t *>(input_data);
-    for (int i = 0; i < element_size * element_count; i += 0x1000)
+    for (int i = 0; i < element_size * element_count; i += page_size)
         input_addr[i] = 0;
     input_addr[element_size * element_count - 1] = 0;
 
--- a/vita3k/mem/include/mem/state.h
+++ b/vita3k/mem/include/mem/state.h
@@ -25,6 +25,11 @@
 #include <memory>
 #include <mutex>
 
+#if defined(__APPLE__) && defined(__aarch64__)
+#define PAGE_NOT_4KIB
+typedef std::map<Address, std::pair<uint32_t, uint32_t>, std::greater<Address>> ForceAllocPage;
+#endif
+
 struct AllocMemPage {
     uint32_t allocated : 4;
     uint32_t size : 28;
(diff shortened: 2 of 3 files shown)


vita3k/io/src/state_functions.cpp
```diff
@@ -37,8 +37,18 @@ SceOff FileStats::read(void *input_data, const int element_size, const SceSize e
...
     // todo: call a mem function to check this instead
+    
+#ifdef _WIN32
+    SYSTEM_INFO system_info = {};
+    GetSystemInfo(&system_info);
+    uint32_t page_size = system_info.dwPageSize;
+#else
+    uint32_t page_size = static_cast<int>(sysconf(_SC_PAGESIZE));
+#endif
+    //LOG_WARN("Page Size: {}", page_size);
```
Zangetsu38, line 48: ?

nn9dev: Pulled from [mem/src/mem.cpp:66](https://github.com/Vita3K/Vita3K/blob/f780c68abbd51419d4012d3d6cd4ae9aed955154/vita3k/mem/src/mem.cpp#L66)

If you mean the log comment I can definitely remove that :P
MemState or emuenv isn't accessible from within io (I don't think) so I couldn't access MemState.page_size. 
Though now that I think about it, maybe it's not good to call that page size get on every read..?

Open to suggestions.

bookmist: Constant 4k is ok too. Based on comment before it should be not more then real page_size, so if page size can't be less then 4k, then 4k is ok.

vita3k/io/src/state_functions.cpp
```diff
@@ -37,8 +37,18 @@ SceOff FileStats::read(void *input_data, const int element_size, const SceSize e
     // that's because host io does not work well with memory trapping and read-only buffer
     // so set 1 byte to 0 in all pages to trigger all possible pagefaults in this range
     // todo: call a mem function to check this instead
+    
+#ifdef _WIN32
+    SYSTEM_INFO system_info = {};
+    GetSystemInfo(&system_info);
+    uint32_t page_size = system_info.dwPageSize;
```
Dicot0721, line 44: Refer to <https://devblogs.microsoft.com/oldnewthing/20210510-00/?p=105200>, Windows always uses the 4 KB page sizes on common architectures.

vita3k/io/src/state_functions.cpp
```diff
@@ -37,8 +37,18 @@ SceOff FileStats::read(void *input_data, const int element_size, const SceSize e
...
+#else
+    uint32_t page_size = static_cast<int>(sysconf(_SC_PAGESIZE));
```
Dicot0721, line 46:
```suggestion
    static const uint32_t page_size = static_cast<uint32_t>(sysconf(_SC_PAGESIZE));
```

vita3k/mem/include/mem/state.h
```diff
@@ -25,6 +25,11 @@
 #include <memory>
 #include <mutex>
 
+#if defined(__APPLE__) && defined(__aarch64__)
+#define PAGE_NOT_4KIB
```
Dicot0721, line 29: Android also supports 16 KB page sizes, and the page sizes on Android are flexible, thus using conditional inclusion may not be a good idea.

vita3k/io/src/state_functions.cpp
nn9dev: this code was pulled from [mem/src/mem.cpp:66](https://github.com/Vita3K/Vita3K/blob/f780c68abbd51419d4012d3d6cd4ae9aed955154/vita3k/mem/src/mem.cpp#L66), but I still think it's good to actually read the page size from the host instead of making an assumption.

nn9dev:
Superseded by #3862 :)

PR #3862, opened 2026-04-04
mem: split host/guest page sizes to support 16kib page devices

imWatchful:
In regards to #2843 and #3860. 

Align OS calls to respect the host's page size instead of hardcoding 4KiB, everything else that needs to work with guest page size keeps working as it did before.


EXtremeExploit (approved):
lgtm

PR #3868, opened 2026-04-08
audio/sdl: increase audio_output prebuffer threshold on Android.

Zangetsu38:
# About
- audio/sdl: increase audio_output prebuffer threshold on Android.
Fix audio quality on sdl.


bookmist (approved):
I don't like it, but its better when nothing

PR #3869, opened 2026-04-09
audio: fix stutter on SDL audio on android

bookmist:


vita3k/audio/src/impl/sdl_audio.cpp
```diff
@@ -75,6 +75,10 @@ AudioOutPortPtr SDLAudioAdapter::open_port(int nb_channels, int freq, int nb_sam
         .freq = freq
     };
     SDL_CHECK(SDL_GetAudioDeviceFormat(device_id, &dst_spec, &device_buffer_samples));
+#ifdef __ANDROID__
+    // Android's audio latency is higher than other platforms, so increase prebuffering to reduce underruns.
+    device_buffer_samples *= 2;
```
Zangetsu38, line 80: Here is wrong, i have changed my pr for totaly fix issue, both function, callback and output request x4 for android, so for can works, you need set x2 also inside audio_output here.
(For both done x4)

bookmist: It's same idea, what is in your pr, but much shorter code. here *2 and in `thread_wakeup_callback` again *2
No *2 in `SDLAudioAdapter::audio_output` is important and as far as we have no problems on PC, I think it's ok.

Zangetsu38: audio_output request also x4 on android for fixing audio, like before
look again my pr

bookmist: I checked for some games, this PR works.

Zangetsu38: > I checked for some games, this PR works.

yes, is what i have thinks too in begin..., for that i have reworked my pr after user and me have view vita os keep get soud issue on sdl, fixed by using also x4 in audio output, like before so

Zangetsu38:
First solution get same as before is merged now 👌

PR #3872, opened 2026-04-11
cpu: Move exclusive monitor entirely into Dynarmic backend

xerpi:
The exclusive monitor is a Dynarmic implementation detail for emulating ARM LDREX/STREX in software, it shouldn't be exposed to the kernel or protocol layers.

The shared monitor is now a static member of DynarmicCPU. All DynarmicCPU instances reference it via their processor_id slot.

### DIFF
--- a/vita3k/cpu/include/cpu/impl/dynarmic_cpu.h
+++ b/vita3k/cpu/include/cpu/impl/dynarmic_cpu.h
@@ -35,7 +35,6 @@ class DynarmicCPU : public CPUInterface {
     std::unique_ptr<Dynarmic::A32::Jit> jit;
     std::unique_ptr<ArmDynarmicCallback> cb;
     std::shared_ptr<ArmDynarmicCP15> cp15;
-    Dynarmic::ExclusiveMonitor *monitor;
 
     std::size_t core_id = 0;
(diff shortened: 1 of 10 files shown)


vita3k/cpu/include/cpu/impl/dynarmic_cpu.h
```diff
@@ -91,6 +90,10 @@ class DynarmicCPU : public CPUInterface {
     bool get_log_code() override;
     bool get_log_mem() override;
 
+    void clear_exclusive() override;
     std::size_t processor_id() const override;
     void invalidate_jit_cache(Address start, size_t length) override;
+
+    static constexpr std::size_t MAX_CORE_COUNT = 150;
```
xerpi: Hmm `KernelState` also defines `MAX_CORE_COUNT`, maybe we should grab it from there?

xerpi: Moved the `MAX_CORE_COUNT` definition to `cpu/common.h` so both can use it.

imWatchful (approved):

Macdu (approved):
lgtm

PR #3875, opened 2026-04-11
kernel/thread: Fix data race regression in run_loop inner loop

xerpi:
The previous fix (a93e009a) changed the inner do-while loop into a
while-at-top loop, checking exit conditions before the first
execution. But those conditions (res, hit_breakpoint, etc.) only
make sense after execution — checking them first caused the loop
to break immediately without executing any guest code (0 fps).

Restore do-while semantics: always execute at least once, then
check the loop condition under the mutex. The to_do read at the
top of each iteration (step vs run decision) and the exit
condition at the bottom are both protected by the mutex.

Also fix suspend() to hold the mutex before modifying to_do,
matching what resume() and exit_delete() already do.

### DIFF
--- a/vita3k/kernel/src/thread.cpp
+++ b/vita3k/kernel/src/thread.cpp
@@ -257,13 +257,9 @@ bool ThreadState::run_loop() {
                 }
             }
 
-            // Run the cpu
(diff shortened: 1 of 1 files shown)


imWatchful (approved):
👍

PR #3876, opened 2026-04-11
modules/SceLibKernel: Implement sceClibMemchr.

Rinnegatamante:


Macdu (commented):
It is not possible to turn a host pointer into a guest pointer, this will return an arbitrary guest pointer when using page tables (the Ptr(host_ptr, emuenv.mem) constructor should be completely removed, I think I mentioned it in another pr).
Instead you should take the guest pointer as an argument to the function and update it with the offset between the two host pointers before returning it.

bookmist:
Example of this pointer arithmetic can be found https://github.com/Vita3K/Vita3K/blob/de3df2dd45cad0f5a613f641d91ecd33c8811ba6/vita3k/modules/SceGxm/SceGxm.cpp#L3173

Rinnegatamante:
I was using sceClibStrstr as reference. I guess that one needs to be changed as well?

vita3k/modules/SceLibKernel/SceLibKernel.cpp
```diff
@@ -342,9 +346,12 @@ EXPORT(int, sceClibStrcatChk) {
 }
 
 EXPORT(Ptr<char>, sceClibStrchr, const char *str, int c) {
-    TRACY_FUNC(sceClibStrchr, str, c);
-    char *res = const_cast<char *>(strchr(str, c));
-    return Ptr<char>(res, emuenv.mem);
+    TRACY_FUNC(sceClibStrchr, str, ch);
+    const char *s = str.get(emuenv.mem);
+    char *res = const_cast<char *>(strchr(str, ch));
```
imWatchful, line 351: I think you meant to do `char *res = const_cast<char *>(strchr(s, ch));` here?

vita3k/modules/SceLibKernel/SceLibKernel.cpp
```diff
@@ -420,16 +427,22 @@ EXPORT(uint32_t, sceClibStrnlen, const char *s1, SceSize maxlen) {
     return static_cast<uint32_t>(strnlen(s1, maxlen));
 }
 
-EXPORT(Ptr<char>, sceClibStrrchr, const char *src, int ch) {
+EXPORT(Ptr<char>, sceClibStrrchr, Ptr<const char> src, int ch) {
     TRACY_FUNC(sceClibStrrchr, src, ch);
+    const char *s = src.get(emuenv.mem);
     char *res = const_cast<char *>(strrchr(src, ch));
```
imWatchful, line 433: same here?

vita3k/modules/SceLibKernel/SceLibKernel.cpp
```diff
@@ -420,16 +427,22 @@ EXPORT(uint32_t, sceClibStrnlen, const char *s1, SceSize maxlen) {
...
+    return Ptr<char>(nullptr);
 }
 
-EXPORT(Ptr<char>, sceClibStrstr, const char *s1, const char *s2) {
+EXPORT(Ptr<char>, sceClibStrstr, Ptr<const char> s1, const char *s2) {
     TRACY_FUNC(sceClibStrstr, s1, s2);
-    char *res = const_cast<char *>(strstr(s1, s2));
-    return Ptr<char>(res, emuenv.mem);
+    const char *src = s1.get(emuenv.mem);
+    char *res = const_cast<char *>(strstr(src, s2));
```
imWatchful, line 442: s2 would need the same treatment as s1 here I believe

vita3k/modules/SceLibKernel/SceLibKernel.cpp
```diff
@@ -420,16 +427,22 @@ EXPORT(uint32_t, sceClibStrnlen, const char *s1, SceSize maxlen) {
...
+    if (res)
+        return Ptr<char>(s1.address() + int(size_t(res) - size_t(src));
```
imWatchful, line 444: Missing the closing parenthesis

imWatchful (changes requested):

Issue #3808, opened 2026-02-01
[BUG]: Several regressions since commit b4c1335 was merged to master


antoniodesousa:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Affected Build(s)

v0.2.1 3856

### Description of Issue

Several issues since the commit [b4c1335](https://github.com/Vita3K/Vita3K/commit/b4c13354225f6e1c4e25ab29799740595bcbfabe) was merged into master:

- Several games show vertex explosions now
- Some games have severe performance issues
- All slide bars in the UI are broken when dragging using a mouse

### v0.2.0 3855

![Image](https://github.com/user-attachments/assets/6613ed84-7d7c-4fb3-8323-e051d78e9477)
![Image](https://github.com/user-attachments/assets/92e37045-98b3-405b-81cd-13d8ec0c4487)

### v0.2.1 3856

![Image](https://github.com/user-attachments/assets/19703826-2023-4c3d-b19b-2e0d91300323)
![Image](https://github.com/user-attachments/assets/177dc7c1-e53b-4464-b59b-eb9a2e1fd594)

### Expected Behavior

- No vertex explosions or same behavior as before this commit
- Same performance as before this commit
- All slide bars in the UI work when dragging using a mouse

### Reproduction Steps

Just run FIFA 15 and LittleBigPlanet.

### Log File

[vita3k_3855.log](https://github.com/user-attachments/files/24992120/vita3k_3855.log)
[vita3k_3856.log](https://github.com/user-attachments/files/24992119/vita3k_3856.log)

### System Configuration

CPU: Ryzen 9 6900HX
GPU/Driver: Radeon RX 6600M / Adrenalin 26.1.1
RAM: 64 GB
OS: Windows 11 25H2 Pro

Zangetsu38:
for lbp i have learn yes, but for fifa not yet 🤔 
and for slid scroll is complicated to fixing it

antoniodesousa:
> for fifa 15 is good with Using External host, for lbp is... hmm white screen lol

I can confirm that using external host does fix the vertex explosions in FIFA 15. Sadly, there's no luck with LBP :(

antoniodesousa:
Before this commit we didn't have any option to change the memory mapping method on PC. So I wouldn't know about it. Anyways, the regression in LittleBigPlanet still stands though.

BeezBumba:
external host on my mali g610 linux device does not work for me. Do yall have support for that GPU?

PR #3877, opened 2026-04-11
gui: Guard ScrollWhenDragging to Android only

ccmdi:
Fixes "All slide bars in the UI are broken when dragging using a mouse" for #3808


Zangetsu38:
You fix nothing here, problem is also on android, i have already fixed it, i pr soon, sorry

PR #3879, opened 2026-04-12
gdbstub: Make GDB usable

### LINKED ISSUE #1903: Built-in Guest Debugger
Ability to step through guest code, see registers set code and memory breakpoints would be useful.

### LINKED ISSUE #1984: cannot debug with gdb-stub
7(6`IP O52 $83}6NDWZ(B9

FirebirdTA01:
Can now connect with GDB, set breakpoints, pause execution, step through instructions, better control of multiple threads, bug fixes and larger GDB packet buffer. Also changed GDB logging to use existing configurable LOG_INFO/WARN/ERROR system.

Closes #1984
Closes #1903


FirebirdTA01:
Commit b51d6627 stripped bit 0 of entry_point and wrote CPSR.T=1
  for every thread start, not just under gdbstub. This corrupted
  CPSR for most games (Sorry, I was using a simpler app to test and didn't catch the regression at first).

vita3k/gdbstub/src/gdb.cpp
```diff
@@ -601,20 +891,24 @@ static std::string cmd_remove_breakpoint(EmuEnvState &state, PacketCommand &comm
     const uint32_t address = parse_hex(content.substr(first + 1, second - 1 - first));
     const uint32_t kind = static_cast<uint32_t>(std::stol(content.substr(second + 1, content.size() - second - 1)));
 
-    LOG_GDB("GDB Server Removed Breakpoint at {} ({}, {}).", log_hex(address), type, kind);
+    // Match cmd_add_breakpoint: only Z0 is handled.
+    if (type != 0)
+        return "";
+
+    LOG_INFO("GDB Server Removed Breakpoint at {} ({}, {}).", log_hex(address), type, kind);
     state.kernel.debugger.remove_breakpoint(state.mem, address);
```
Macdu: This is still broken for Thumb instructions. 
The address value given by gdb for add_breakpoint and remove_breakpoint has its lowest bit set if it targets thumb instructions.
You should pass align_down(address, 2) to `debugger.add_breakpoint` and `debugger.remove_breakpoint`

Macdu (changes requested):
Thanks a lot for this PR which is a huge improvement and sorry for taking so long to review.
While trying this PR, I wasn't able to properly use breakpoints with thumb instructions, I explain why below. I will try again after this is fixed.

vita3k/gdbstub/src/gdb.cpp
```diff
@@ -274,13 +448,31 @@ static std::string cmd_read_registers(EmuEnvState &state, PacketCommand &command
...
-
-    for (uint32_t a = 0; a < 16; a++) {
-        str += be_hex(fetch_reg(cpu, a));
+    if (state.gdb.client_has_xml) {
+        // target.xml layout: r0-r15 (64B), cpsr (4B), d0-d31 (256B), fpscr (4B) = 328B = 656 hex
+        std::string str;
+        str.reserve(656);
+        for (uint32_t a = 0; a < 16; a++)
+            str += be_hex(fetch_core_reg(cpu, a));
+        str += be_hex(read_cpsr(cpu) | 0x20); // force T bit — Vita is always Thumb
```
Macdu: That's not true, Vita can be either Thumb or regular 32-bit arm instructions

vita3k/gdbstub/src/gdb.cpp
```diff
@@ -432,28 +659,33 @@ static std::string cmd_continue(EmuEnvState &state, PacketCommand &command) {
...
-                    // TODO if that thread waits for sync primitive, dead lock.
-                    thread->status_cond.wait(thread_lock, [&]() { return thread->status == ThreadStatus::suspend; });
+                thread->resume(true);
+                {
+                    auto thread_lock = std::unique_lock(thread->mutex);
+                    thread->status_cond.wait(thread_lock, [&]() {
+                        return thread->status == ThreadStatus::suspend
+                            || thread->status == ThreadStatus::wait;
+                    });
                 }
```
Macdu: This is not enough, thread->resume sets the todo value of the thread, not its status, meaning this wait will return immediatly. (technically it is a race condition, but the other thread is not likely to win)

Macdu (commented):
When debugging thumb instructions, I still wasn't able to use the debugger.
It looks like gdb does not recognize that breaks are caused by its own breakpoints,  causing it to set back the breakpoint before continuing and the thread to immediatly break again.

vita3k/gdbstub/src/gdb.cpp
FirebirdTA01: Yeah, I had an AI draft the fix for me super late last night when I was booted into a different environment on my machine which didn't have Vita3k or my VitaSDK toolchain. I was lazy and the easiest way for me to get it into the other environment for testing was to push it through github. I hadn't tested it or even looked at it but had intended to before having you check on it again (thought I had more time, haha).  I'll fix it properly soon.

FirebirdTA01:
Added a readme which also contains a note about the windows packaged arm-vita-eabi-gdb not being functional because of missing dependencies

FirebirdTA01:
There were a few upstream changes which broke the automatic merge. I'm making updates to the PR now

vita3k/gdbstub/src/gdb.cpp
FirebirdTA01: Okay, should be ready for re-review. The watchpoint stuff isn't completely ready yet, but it was in the working tree of my other environment and they don't break anything either so I went ahead and pushed those commits up also.

imWatchful:
Hi, we recently merged a PR that changes some of the behavior in thread/kernel and this PR would need to be updated as a result. Additionally, the current `app::destroy` call closes the gdb server entirely, you probably want to remove that behavior in this PR :D

FirebirdTA01:
> Hi, we recently merged a PR that changes some of the behavior in thread/kernel and this PR would need to be updated as a result. Additionally, the current `app::destroy` call closes the gdb server entirely, you probably want to remove that behavior in this PR :D

Updated :)

devnoname120:
@FirebirdTA01 FYI gdbstub software breakpoints are still broken for Thumb-2

**Edit**: but IMO this shouldn’t be a blocker for merging. This PR already MASSIVELY improves the current state of things so it would already be very good to have it on main. Otherwise there is the risk that the reviews drag on forever and we never get it merged. Thumb-2 and other improvements can be done in follow-up PRs.

vita3k/app/src/app_init.cpp
```diff
@@ -748,8 +746,10 @@ SettingsCommitResult delete_custom_settings(EmuEnvState &emuenv, const std::stri
 
 // need to remove this function completely, waiting for gdb refactor
 void destroy(EmuEnvState &emuenv) {
```
imWatchful: I would remove this function completely and all references to it.

FirebirdTA01:
> @FirebirdTA01 FYI gdbstub software breakpoints are still broken for Thumb-2
> 
> **Edit**: but IMO this shouldn’t be a blocker for merging. This PR already MASSIVELY improves the current state of things so it would already be very good to have it on main. Otherwise there is the risk that the reviews drag on forever and we never get it merged. Thumb-2 and other improvements can be done in follow-up PRs.

I might be misunderstanding, but I thought they are working correctly. I'm testing using my Vita Native Renderer app on my github. 

3-BP test, vitasdk client:
  STOP A: main+0 Thumb-2 stmdb.w -> hit, stepi PC +4 (correct?)
  STOP B: main+12 16-bit Thumb sub -> hit, stepi PC +2 (correct?)
  STOP C: main+44 Thumb-2 mov.w -> hit, stepi PC +4 (correct?)
  
  Can you correct my understanding or provide steps to reproduce the broken behavior?

PR #3881, opened 2026-04-14
module/bridge: HLE compile-time dispatch, avoid extra pointer indirection

xerpi:
The HLE dispatching bridge is now a stateless lambda that the compiler generates per-export via a NTTP, with the ARM calling convention layout computed at compile-time time so all register and stack argument reads are resolved at compile time with no runtime dispatch.
Also replace std::function with a plain function pointer for ImportFn to avoid possible heap-allocation and indirection.

Example: `_sceKernelCancelEventFlag(SceUID event_id, SceUInt pattern, SceUInt32 *num_wait_thread)`
Before: two functions, runtime dispatch, indirect call through captured function pointer.
```
; std::_Function_handler::_M_invoke -- unwraps the std::function heap object
_ZNSt17_Function_handlerI...bridge<int,(int,uint,uint*)>...M_invoke...:
        pushq   %rax
        movq    (%rdi), %rdx            ; load captured state from heap
        movq    (%rdx), %rdi
        movq    8(%rdx), %rax           ; load captured export_fn ptr
        leaq    16(%rdx), %rcx          ; load captured args_layout ptr
        callq   _Z4call<int,(int,uint,uint*)>...
        retq

; call<> -- dispatches each arg through a runtime switch on arg.location
_Z4call<int,(int,uint,uint*)>...:
        movl    (%rdx), %eax
        cmpl    $1, %eax                ; arg[0]: gpr or stack?
        je      .stack_path_0
        testl   %eax, %eax
        jne     .done_0
        movq    8(%r14), %rsi
        callq   read_reg
        movl    16(%r14), %eax
        cmpl    $1, %eax                ; arg[1]: gpr or stack?
        je      .stack_path_1
        testl   %eax, %eax
        jne     .done_1
        movq    24(%r14), %rsi
        callq   read_reg
        movl    32(%r14), %eax
        cmpl    $1, %eax                ; arg[2]: gpr or stack?
        je      .stack_path_2
        testl   %eax, %eax
        jne     .done_2
        movq    40(%r14), %rsi
        callq   read_reg
        callq   *%r10                   ; indirect call through captured fn ptr
        jmp     write_return_value
```

After: single function with register indices hardcoded (generated at compile-time), direct call.

```
; make_bridge<&export__sceKernelCancelEventFlag>(...).__invoke
_ZZ11make_bridge<&export__sceKernelCancelEventFlag>...NUl...8__invoke...:
        xorl    %esi, %esi
        callq   read_reg                ; r0 -> SceUID (index 0, hardcoded)
        movl    $1, %esi
        callq   read_reg                ; r1 -> SceUInt (index 1, hardcoded)
        movl    $2, %esi
        callq   read_reg                ; r2 -> Ptr<SceUInt32> (index 2, hardcoded)
        callq   eventflag_cancel        ; direct call, resolved at compile time
        jmp     write_return_value
```


vita3k/module/include/module/lay_out_args.h
```diff
@@ -27,93 +27,42 @@ namespace module {
...
-    const std::size_t next_float_used = state.float_used + 1;
-
-    // Lay out in registers.
-    const ArgLayout layout = { ArgLocation::fp, float_index };
-    const LayoutArgsState next_state = { state.gpr_used, state.stack_used, next_float_used };
-
-    return { layout, next_state };
-}
-
+// Kept constexpr (not consteval) because vargs::next() calls this at runtime.
```
imWatchful: Small nit but this comment is probably unnecessary

vita3k/module/include/module/lay_out_args.h
```diff
@@ -27,93 +27,42 @@ namespace module {
...
+            const std::size_t stack_required = sizeof(Arg); // TODO Should this be aligned up?
+            const std::size_t stack_offset = align(state.stack_used, stack_alignment);
+            return { { ArgLocation::stack, stack_offset }, { state.gpr_used, stack_offset + stack_required, state.float_used } };
         }
+
+        return { { ArgLocation::gpr, gpr_index }, { gpr_index + gpr_required, state.stack_used, state.float_used } };
     }
 }
 
+// Layout is always computed at compile time.
```
imWatchful: Same here

imWatchful (dismissed):
LGTM 👍 up to you if you want to remove those comments or not

vita3k/module/include/module/lay_out_args.h
xerpi: Yeah I'll remove them

imWatchful (approved):

Macdu:
Please let me review this PR before merging (will do it today or tomorrow)

xerpi:
> Please let me review this PR before merging (will do it today or tomorrow)

Of course, please take your time!
Also, if you can find a way to simplify all the syntactic sugar, it would be really appreciated!
Imho this code is quite difficult to understand...

vita3k/module/include/module/bridge.h
```diff
@@ -21,41 +21,48 @@
...
 #ifdef TRACY_ENABLE
-        ZoneNamedC(___tracy_scoped_zone, 0xFFF34C, emuenv.cfg.tracy_primitive_impl); // Tracy - Track function scope and set color to yellow
-        ZoneNameV(___tracy_scoped_zone, export_name, strlen(export_name)); // Tracy - Edit scope name based on export_name
+        ZoneNamedC(___tracy_scoped_zone, 0xFFF34C, emuenv.cfg.tracy_primitive_impl);
+        ZoneNameV(___tracy_scoped_zone, export_name, strlen(export_name));
 #endif
-
-        using Indices = std::index_sequence_for<Args...>;
-        call(export_fn, export_name, std::get<0>(args_layout), std::get<1>(args_layout), Indices(), thread_id, cpu, emuenv);
+        constexpr auto arr = std::get<0>(layout);
```
imWatchful, line 58: You could use structured bindings here xD that's about it for syntactic improvement (that I can think of at least)

xerpi: Oh yeah, please feel free to create a new PR based on this one with simpler code!

imWatchful: i forgot that constexpr structured bindings aren't a thing until c++26

Macdu:
Small mistake I found (which already existed before your PR) about the way we lay out args: the following signature
`fn(int a, int b, int c, int64_t c, int d)` 
puts d in r3 while it should be spilled (as soon as you start spilling, you don't stop).
I guess we are lucky we did not have this kind of signature in practice (8-byte arguments are rare anyway).
Could you fix this as well in your PR?

xerpi:
> Small mistake I found (which already existed before your PR) about the way we lay out args: the following signature `fn(int a, int b, int c, int64_t c, int d)` puts d in r3 while it should be spilled (as soon as you start spilling, you don't stop). I guess we are lucky we did not have this kind of signature in practice (8-byte arguments are rare anyway). Could you fix this as well in your PR?

Ohh good catch!
I'll try to fix it.

xerpi:
> Small mistake I found (which already existed before your PR) about the way we lay out args: the following signature `fn(int a, int b, int c, int64_t c, int d)` puts d in r3 while it should be spilled (as soon as you start spilling, you don't stop). I guess we are lucky we did not have this kind of signature in practice (8-byte arguments are rare anyway). Could you fix this as well in your PR?

Added an extra commit fixing that + a test.

Macdu (approved):
lgtm, thank you for the PR!
(the code is still hard to follow, but that's just how template-oriented c++ code is)

PR #3886, opened 2026-04-18
audio: Increase SDL audio buffer size

Macdu:
Audio depends a lot on the host hardware and driver being used. 
Increasing the audio buffer fixes the stutter issues I have on all games when using pipewire (which is the default sdl driver on linux)


imWatchful (approved):
ci shit itself for some reason, had to rerun half of them

Zangetsu38 (commented):
Sorry for that... i have thinks only android request more buffer... and x2 worked fine for windows so have wanted keep good latancy for it

PR #3892, opened 2026-04-22
cpu: move halt instruction block to KernelState, shared across all threads

xerpi:
Each thread was allocating its own identical 4-byte NOP+WFI block. Since all threads point LR at the same sentinel, one allocation in KernelState is sufficient.

Note that this is a Dynarmic-specific "hack" to make it stop execution (`jit->Run()`) after the guest code returns (jumps back LR):
- Prior JIT execution, we prepare LR to point to this NOP+WFI block.
- Run guest code (`jit->Run()`)
- When the guest function returns, it will jump into this NOP+WFI block and `Dynarmic::A32::Exception::WaitForInterrupt` will be called. There, we call `cpu->jit->HaltExecution()` which will make `jit->Run()` return

### DIFF
--- a/vita3k/kernel/src/kernel.cpp
+++ b/vita3k/kernel/src/kernel.cpp
@@ -21,6 +21,7 @@
 
 #include <cpu/common.h>
 #include <kernel/state.h>
+#include <mem/functions.h>
 
 #include <kernel/thread/thread_state.h>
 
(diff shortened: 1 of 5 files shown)


vita3k/kernel/src/kernel.cpp
```diff
@@ -90,6 +91,13 @@ bool KernelState::init(MemState &mem, const CallImportFunc &call_import, bool cp
     cpu_protocol = std::make_unique<CPUProtocol>(*this, mem, call_import);
     this->cpu_opt = cpu_opt;
 
+    // Generate halt instruction (NOP + WFI)
+    halt_instruction = alloc_block(mem, 4, "halt_instruction");
+    const auto halt_ptr = halt_instruction.get_ptr<uint16_t>();
+    *halt_ptr.get(mem) = 0xBF00; // NOP
+    *(halt_ptr.get(mem) + 1) = 0xBF30; // WFI
```
bookmist:
```suggestion
    const auto halt_ptr = halt_instruction.get_ptr<uint16_t>().get(mem);
    halt_ptr[0] = 0xBF00; // NOP
    halt_ptr[1]= 0xBF30; // WFI
```

bookmist (approved):

PR #3899, opened 2026-04-24
android: Add x86_64 ABI

Dicot0721:
- Disable adrenotools in Android x86_64 build to compile.
- Fix incorrect type for parameter of JNI function.

### DIFF
--- a/vita3k/renderer/src/vulkan/renderer.cpp
+++ b/vita3k/renderer/src/vulkan/renderer.cpp
@@ -46,11 +46,14 @@
 
 #ifdef __ANDROID__
 #include <SDL3/SDL_system.h>
-#include <adrenotools/bcenabler.h>
-#include <adrenotools/driver.h>
 #include <dlfcn.h>
 #include <sys/mman.h>
 #include <util/float_to_half.h>
+
+#ifdef USE_ADRENO_TOOLS
+#include <adrenotools/bcenabler.h>
+#include <adrenotools/driver.h>
+#endif
 
 #include <android/hardware_buffer.h>
 
@@ -153,7 +156,7 @@ const static std::vector<const char *> required_device_extensions = {
 
 namespace renderer::vulkan {
 
-#ifdef __ANDROID__
+#if defined(__ANDROID__) && defined(USE_ADRENO_TOOLS)
 // need to avoid patching bcn per custom driver more than once
 static bool patch_bcn_once(void *function_to_patch) {
     static std::unordered_set<void *> patched_functions;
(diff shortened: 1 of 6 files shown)


vita3k/renderer/src/vulkan/renderer.cpp
```diff
@@ -1784,10 +1787,12 @@ bool VKState::support_custom_drivers() {
 }
 
 void VKState::set_turbo_mode(bool set) {
+#ifdef __aarch64__
```
EXtremeExploit: i think it would be better to have a define flag for supporting custom drivers rather than it being aarch64 specific, in CMake file around the same lines you edited

vita3k/util/src/android_driver.cpp
```diff
@@ -408,6 +411,7 @@ bool is_custom_driver_loaded(const std::string &driver_name, uint32_t vendor_id,
 
 } // namespace android_driver
 
+#ifdef __aarch64__
```
EXtremeExploit: same here

vita3k/util/src/android_driver.cpp
```diff
@@ -466,12 +470,12 @@ void *open_custom_vulkan_driver(const std::string &driver_name) {
...
     return vulkan_handle;
 }
+#endif
 
 namespace android_driver {
 
 PFN_vkGetInstanceProcAddr resolve_vk_get_instance_proc_addr(const std::string &driver_name) {
-    const PFN_vkGetInstanceProcAddr system_vk_get_instance_proc_addr = load_system_vk_get_instance_proc_addr();
-
+#ifdef __aarch64__
```
EXtremeExploit: and here

EXtremeExploit (changes requested):
i think its good but overall some small changes

vita3k/renderer/src/vulkan/renderer.cpp
Dicot0721: [adrenotools](https://github.com/bylaws/libadrenotools/blob/8fae8ce254dfc1344527e05301e43f37dea2df80/CMakeLists.txt#L3-L5) enables only aarch64 in their CMake file
```
if(NOT ${CMAKE_ANDROID_ARCH_ABI} STREQUAL arm64-v8a)
	message(FATAL_ERROR "Unsupported target architecture: ${CMAKE_ANDROID_ARCH_ABI}. Please make an issue on the repo!")
endif()
```

EXtremeExploit: Im talking about adding a custom define for custom drivers stuff, i dont think it should be on the arbitrary ifdef aarch64 as other arm binaries exist, you can add a custom define for it in the same cmake lines you edited

Dicot0721: Ah I see

Dicot0721: Done

EXtremeExploit (approved):

PR #3900, opened 2026-04-25
modules/SceFiber: default to LLE

xerpi:
SceFiber depends on TLS's `TLS_SP_TOP`/`TLS_SP_BOTTOM` entries. Since they have been added and populated properly at thread creation, LLE SceFiber works (or should) out of the box without needing the current HLE implementation, which has some hacks and heap memory allocations that shouldn't be there.

This also removes all non-trivial manual CPU context/register manipulation from HLE modules, reducing the surface that would need special handling for native exec.

### DIFF
--- a/vita3k/module/src/load_module.cpp
+++ b/vita3k/module/src/load_module.cpp
@@ -184,7 +184,7 @@ bool is_lle_module(SceSysmoduleModuleId module_id, EmuEnvState &emuenv) {
 }
 
 static std::vector<std::string> init_auto_lle_module_names() {
-    std::vector<std::string> auto_lle_module_names = { "libc", "libSceFt2", "libpvf" };
+    std::vector<std::string> auto_lle_module_names = { "libc", "libSceFt2", "libpvf", "libfiber" };
     for (const auto module_id : auto_lle_modules) {
         for (const auto module : sysmodule_paths[module_id]) {
             auto_lle_module_names.emplace_back(module);


Zangetsu38:
Thx but 🤔
Don't remove current implementation 🤔
Keep it is better for testing

xerpi:
> Thx but 🤔 Don't remove current implementation 🤔 Keep it is better for testing

Fair enough.
I'll just add it to `auto_lle_module_names` for now.

Zangetsu38:
Fiber module is not loaded by app by scesysmodule func ? 🤔
Pre load module is rarely

xerpi:
> Fiber module is not loaded by app by scesysmodule func ? 🤔 Pre load module is rarely

Good point. So it seems SceFiber is already being loaded as LLE when the app loads it?

Zangetsu38:
> > Fiber module is not loaded by app by scesysmodule func ? 🤔 Pre load module is rarely
> 
> Good point. So it seems SceFiber is already being loaded as LLE when the app loads it?

No sur 🤔 need look App using it for looking if app call func for load it

xerpi:
I mean when an app calls `sceSysmoduleLoadModule(SCE_SYSMODULE_FIBER);`, will it load the HLE version or the LLE version (libfiber.suprx) by default?
I think `is_lle_module()` will return true.

Zangetsu38:
> I mean when an app calls `sceSysmoduleLoadModule(SCE_SYSMODULE_FIBER);`, will it load the HLE or the LLE by default? I think `is_lle_module()` will return true.

If the application calls this function in this way, it loads the module. For this, I ask the application to use Fiber, for example.

bookmist:
Libfiber should be loaded with sceSysmoduleLoadModule, but Libfiber is used by libult, but libult does not load libfiber.
Maybe library loader load libfiber with libult.

Zangetsu38:
> Libfiber should be loaded with sceSysmoduleLoadModule, but Libfiber is used by libult, but libult does not load libfiber. Maybe library loader load libfiber with libult.

complicated xD, for tjhat i ask for app exemple using fiber, for checking

bookmist:
Wait some time with this PR. On the one side, code should be different, on the other side, I'm thinking about LLE os0:kd/sysmodule.skprx (it's very small and mostly contains conversion of module id to file name) and some rewriting of the list of automatic modules (from module ids to file names)

frangarcj (approved):

Issue #3893, opened 2026-04-23
[BUG]: [Android] System bars not hidden when launched via external intent (e.g. ES-DE frontend)


xxpc9:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Affected Build(s)

v0.2.1 3959-caf39d89

### Description of Issue

When launching Vita3K directly from its own UI, the navigation bar and status bar are correctly hidden during gameplay. However, when launched via an external frontend such as ES-DE, both bars remain visible, causing the game screen to be cut off at the top and bottom.

Image of game launched from ES-DE:
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/366966b4-9046-4db2-b836-c6e12cd38203" />

### Expected Behavior

The navigation bar and status bar should be hidden during gameplay regardless of how Vita3K was launched, whether directly or via an external intent from a frontend.

Image of game launched directly from Vita3k:

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/f3f8b043-d040-47b9-a9e2-e3ca7f28f72f" />

### Reproduction Steps

Install ES-DE frontend
Configure it to launch a Vita3K game
Launch a game through the frontend
Navigation bar and status bar are visible and game screen is clipped

### Log File

Using "adb shell dumpsys window windows" while a game is running shows the difference clearly:

When launched directly:
Requested visibilities: ITYPE_STATUS_BAR: invisible, ITYPE_NAVIGATION_BAR: invisible

When launched externally:
Requested visibilities: ITYPE_STATUS_BAR: visible, ITYPE_NAVIGATION_BAR: visible

### System Configuration

CPU: Qualcomm Snapdragon 8 Gen 2
GPU/Driver: Adreno 740
RAM: 12 GB
OS: Android 13

EXtremeExploit:
Does this till happen in the latest version?

PR #3902, opened 2026-04-26
replace SYSTEM_UI_FLAG_* with WindowInsetsControllerCompat as the for…

xxpc9:
Replace SYSTEM_UI_FLAG_* with WindowInsetsControllerCompat as the former is deprecated and ignored in certain UI race conditions.

Fix for this issue #3893

When Vita3k is launched externally via launchers like ES-DE on Android 11+ it could encounter certain race condition where SYSTEM_UI_FLAG_* is ignored in favor of the newer Android's default UI API, causing navigation bars to not hide itself. SYSTEM_UI_FLAG_* were deprecated since Android 11, and WindowInsetsControllerCompat is backward compatible which should work all the way back to [API 21 (Android 5+).](https://developer.android.com/reference/androidx/core/view/WindowInsetsControllerCompat)

### DIFF
--- a/android/src/main/java/org/vita3k/emulator/Emulator.java
+++ b/android/src/main/java/org/vita3k/emulator/Emulator.java
@@ -17,6 +17,9 @@ import android.view.Surface;
 import android.view.ViewGroup;
 import android.view.View;
 import android.widget.Toast;
+import androidx.core.view.WindowCompat;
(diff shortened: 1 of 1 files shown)


imWatchful:
Thanks a lot for opening this, I actually had these changes already as part of the interface move we just did so this is the way we hide the nav bar now. Much love.

PR #3905, opened 2026-04-26
Audio improvements: safer shutdown, pacing, and per-thread config filtering

frangarcj:
## Summary

Audio improvements covering shutdown safety, streaming pacing, and per-thread configuration mismatch filtering.

### Changes

1. **cubeb: Safe shutdown via CubebAudioSharedState**
   - Extract callback-accessed state into a separate shared_ptr-owned struct
   - Destructor signals shutdown and waits for all in-flight callbacks
   - Prevents use-after-free crashes during rapid stream destruction

2. **SceAudio: Per-thread TidConfig mismatch filtering**
   - Track per-thread audio config (len/freq/mode) in AudioOutPort::tid_configs
   - Silently drop audio from threads using stale configurations
   - Preserves type and volume settings during port reconfiguration

3. **Fix movie audio pacing and queue timing**
   - Use get_stream_buffer_samples() to compute threshold for prebuffering
   - Platform-specific prebuffering (higher on Android, standard on other platforms)
   - Fix SDL3 wait timeout to not double the microseconds on some platforms

### DIFF
--- a/vita3k/modules/SceAudio/SceAudio.cpp
+++ b/vita3k/modules/SceAudio/SceAudio.cpp
@@ -146,8 +146,21 @@ EXPORT(int, sceAudioOutGetAdopt, SceAudioOutPortType type) {
 }
 
 EXPORT(int, sceAudioOutGetConfig, int port, SceAudioOutConfigType type) {
-    TRACY_FUNC(sceAudioOutGetConfig, type);
-    return UNIMPLEMENTED();
+    TRACY_FUNC(sceAudioOutGetConfig, port, type);
+    const AudioOutPortPtr prt = lock_and_find(port, emuenv.audio.out_ports, emuenv.audio.mutex);
+    if (!prt) {
+        return RET_ERROR(SCE_AUDIO_OUT_ERROR_INVALID_PORT);
+    }
+    switch (type) {
+    case SCE_AUDIO_OUT_CONFIG_TYPE_LEN:
+        return prt->len;
+    case SCE_AUDIO_OUT_CONFIG_TYPE_FREQ:
+        return prt->freq;
+    case SCE_AUDIO_OUT_CONFIG_TYPE_MODE:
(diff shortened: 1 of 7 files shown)


vita3k/modules/SceAudio/SceAudio.cpp
```diff
@@ -272,36 +305,46 @@ EXPORT(int, sceAudioOutSetConfig, int port, int len, int freq, int mode) {
...
         return RET_ERROR(SCE_AUDIO_OUT_ERROR_NOT_OPENED);
 
     // Save the port configuration
+    prt->type = current_port->type;
     prt->freq = set_freq;
     prt->len = set_len;
     prt->mode = set_mode;
+    prt->left_channel_volume = current_port->left_channel_volume;
+    prt->right_channel_volume = current_port->right_channel_volume;
+    prt->tid_configs = current_port->tid_configs;
```
imWatchful, line 346: Do you mean to copy over the old ports config here? If it's legitimately reconfigured (is that possible? idk). Then you would get a mismatch later with the new `config_match` check on Line 230, because try_emplace wouldn't write new data since the key exists already.

vita3k/modules/SceAudio/SceAudio.cpp
```diff
@@ -195,12 +214,23 @@ EXPORT(int, sceAudioOutOutput, int port, const void *buf) {
         return RET_ERROR(SCE_AUDIO_OUT_ERROR_INVALID_PORT);
     }
 
+    auto [cfg_it, inserted] = prt->tid_configs.try_emplace(thread_id, AudioOutPort::TidConfig{ prt->len, prt->freq, prt->mode });
```
imWatchful, line 217: Need a lock/sync here no?

bookmist:
Are you sure that audio port configuration can be different for the same port, but different threads? I found no such information in manual and disasm.

PR #3909, opened 2026-04-29
vita3k/CMakeLists.txt: Use Vita3K.icon for macOS app icon via actool instead of sips pipeline

ghost:
### Add Liquid Glass app icon for macOS 26
                                                                                                           
Adds a new app icon, it's really basic since I didn't really want to take away from the Vita3K icon and its design. The new icon uses the new `.icon` format which allows support for the glass-style transparency, specular highlights, and automatic theme adaptation (Default, Dark, Clear, Tinted) that the system provides natively.

The build system now uses `actool` to compile **Vita3K.icon** directly instead of the old `sips` resize pipeline, so the icon is generated correctly for the new format.

Here are some previews of the icons after I had designed them:

<details>
  <summary>Default</summary>
  Vita3K-iOS-Default-1024x1024@1x
</details>
<details>
  <summary>Dark</summary>
  Vita3K-iOS-Dark-1024x1024@1x
</details>
<details>
  <summary>Clear Light</summary>
  Vita3K-iOS-ClearLight-1024x1024@1x
</details>
<details>
  <summary>Clear Dark</summary>
  Vita3K-iOS-ClearDark-1024x1024@1x
</details>
<details>
  <summary>Tinted Light</summary>
  Vita3K-iOS-TintedLight-1024x1024@1x
</details>
<details>
  <summary>Tinted Dark</summary>
  Vita3K-iOS-TintedDark-1024x1024@1x
</details>


ghost:
I have to apologize, I hadn't checked the Xcode version requirement before submitting this PR since it had built fine on my machine.

The actool-based Liquid Glass icon support requires Xcode 26 or later, but the CI currently runs on Xcode 16.2 which cannot compile the .icon format.

This can properly be shelved until the build system safely migrates to Xcode 26.

bookmist:
You have nothing to apologise. Xcode 26 exists on github actions build machine, it's not default xcode. https://github.com/actions/runner-images/blob/main/images/macos/macos-15-arm64-Readme.md#xcode
I dont know if switching to xcode 26 is good or bad (maybe it disable support of old versions of macos or something else, I dont know macos at all) But at least it's possible on default github actions macos runner.

nn9dev:
As it stands, every Xcode version has a "minimum deployment target", and the minimum version for Xcode 26 is [macOS 11](https://developer.apple.com/xcode/system-requirements/). Coincidentally, macOS 11 is the version we target, and we specify this version in CMake using [`CMAKE_OSX_DEPLOYMENT_TARGET` here](https://github.com/Vita3K/Vita3K/blob/faa4a632f0879fdd50490aed79561969f0388175/CMakeLists.txt#L22) and configure boost with [cxxflags="-mmacosx-version-min=11.0" here](https://github.com/Vita3K/Vita3K/blob/faa4a632f0879fdd50490aed79561969f0388175/CMakeLists.txt#L149). 

All this to say, bumping the CI to use Xcode 26 should not cause any problems. 

However, I don't know how new icons interact with older (before macOS 15) versions of macOS, so that should be verified before this PR were to be accepted.

ghost:
In that case, I can try switching the workflow to Xcode 26 then.

As far as I know, it shouldn't break the icons in previous Mac version, what I think happens is that it would use the Default Icon (Light Mode) on those previous versions since that's how the icon looks like naturally.

The Liquid Glass design isn't from the .icns file or the icon, but the Assets.car file generated by Xcode 26 compiling the .icon file at runtime.

PR #3913, opened 2026-05-03
modules: lle kernel module "sysmodule"

bookmist:
This PR introduces Low-Level Emulation support for kernel modules and includes several improvements to the module loading system. It also contains some unrelated changes in the same files that were not intentionally separated.

For LLE sysmodule.skprx:
* Loading modules from bootimage.skprx
* ELF loading with automatic format detection based on the magic value
* New NIDs and modules for kernel-level functions (added manually).
* Implementation of several kernel-level functions . Most of them delegate to user-level counterparts.
* Writable tpidruro

Other changes, which are difficult to split:

* If memory cannot be allocated at a strictly specified address — return 0. (Previously — memory was overwritten.)
* Whenever possible, do not relocate modules.
* Patches are now read for eboot.bin only
* ELF dumps are now organized into subfolders named by application/game ID
* Correct implementation of LLE libfiber, even without an LLE sysmodule
* Some other kernel-level functions (I tried to implement LLE for modulemgr and processmgr, but no success at all.)


vita3k/modules/SceSysmem/SceSysclibForDriver.cpp
```diff
@@ -35,12 +42,47 @@ EXPORT(int, __aeabi_lmul) {
...
-EXPORT(int, __aeabi_uidiv) {
-    return UNIMPLEMENTED();
-}
-
-EXPORT(int, __aeabi_uidivmod) {
-    return UNIMPLEMENTED();
+EXPORT(SceUInt, __aeabi_uidiv, SceUInt a, SceUInt b) {
+    TRACY_FUNC(__aeabi_uidiv, a, b);
+    return a / b;
+}
```
imWatchful, line 65: You handle a 0 denominator case in `__aeabi_uidivmod` and `__aeabi_uldivmod`, wouldn't it also be needed here as well?

vita3k/modules/module_parent.cpp
```diff
@@ -273,7 +313,10 @@ int unload_module(EmuEnvState &emuenv, SceUID module_id) {
         return RET_ERROR(SCE_KERNEL_ERROR_MODULEMGR_NO_MOD);
     }
     LOG_INFO("Unloading module {} ({})", module_id, module->info.module_name);
-
+    if (module->info.segments[0].memsz == 0) {
+        LOG_WARN("Module {} has no segments, skipping unload", module->info.module_name);
+        return 0;
+    }
```
imWatchful, line 323: Is there no need to remove the fake modules here? Since we allow same process app restarts now this might(?) be an issue.

vita3k/modules/SceProcessmgr/SceProcessmgrForDriver.cpp
```diff
@@ -36,8 +39,11 @@ EXPORT(int, ksceKernelGetProcessLocalStorageAddr) {
     return UNIMPLEMENTED();
 }
 
-EXPORT(int, ksceKernelGetProcessLocalStorageAddrForPid) {
-    return UNIMPLEMENTED();
+EXPORT(int, ksceKernelGetProcessLocalStorageAddrForPid, SceUID pid, int key, Ptr<void> *out_addr, int create_if_doesnt_exist) {
+    TRACY_FUNC(ksceKernelGetProcessLocalStorageAddrForPid, pid, key, out_addr, create_if_doesnt_exist);
+    // See ksceKernelCreateProcessLocalStorage
+    *out_addr = key << 1;
```
imWatchful, line 45: Could the pointer be null here? might crash the emu

vita3k/interface.cpp
```diff
@@ -362,6 +364,18 @@ uint32_t install_contents(EmuEnvState &emuenv, const fs::path &path) {
...
+        if (patch.seg < MODULE_INFO_NUM_SEGMENTS) {
+            auto seg_ptr = sceKernelModuleInfo.segments[patch.seg].vaddr.cast<uint8_t>();
+            if (seg_ptr) {
+                LOG_INFO("Patching segment {} at offset 0x{:X} with {} values", patch.seg, patch.offset, patch.values.size());
+                memcpy(seg_ptr.get(mem) + patch.offset, patch.values.data(), patch.values.size());
+            }
+        }
+    }
+}
+
```
imWatchful, line 411: we didn't have one before either but maybe we should have some sort of write safety check to make sure we're writing within the segment

vita3k/modules/SceSysmem/SceSysclibForDriver.cpp
bookmist: fixed

vita3k/modules/module_parent.cpp
bookmist: add erase module record

vita3k/modules/SceProcessmgr/SceProcessmgrForDriver.cpp
bookmist: Original code does not contains check

vita3k/interface.cpp
bookmist: add check

imWatchful (dismissed):
LGTM 👍

imWatchful (approved):

PR #3634, opened 2025-07-19
gui: Apply Dynamic Font Rendering introduced in ImGui 1.20

saturnsky:
This PR implements the Dynamic Font Rendering feature introduced in ImGui 1.20. We can now dynamically render various fonts at different resolutions without the long initial loading time. This improvement is particularly significant when using CJK fonts, which previously required substantial time to bake all glyphs at startup.

### Concerns about Stability

Due to my limited expertise in graphics APIs, I cannot guarantee the stability of this PR. While I've closely followed the imgui modifications for implementation, I have significant doubts about whether `imgui_impl_sdl_vulkan` has been correctly ported, as Vulkan is outside my area of expertise.

### Testing Status

The logic in other parts should be ok. I've tested on Windows and Linux environments, and within my testing scope, no issues were found. However, I haven't been able to test on multi-monitor systems, and so on. Further testing by multiple people in various scenarios is needed.

### Credits

The modifications to `imgui_impl_sdl_gl3` and `imgui_impl_sdl_vulkan` are heavily based on the upstream changes to `imgui_impl_opengl3` and `imgui_impl_vulkan` by the ImGui maintainers.

Reference
- https://github.com/ocornut/imgui/commit/abe294bfd0bfff325bbaa000f2d688085f1d4696 
- https://github.com/ocornut/imgui/commit/dbb91a574f35e24697754f6344455fd43b97d46d

PR #3916, opened 2026-05-09
config: Add option to enable/disable font scaling.

Zangetsu38:
# About
- config: Add option to enable/disable font scaling.
Introduced a font scaling checkbox in the initial setup, with automatic font reload on completion.
Added the checkbox to the settings dialog as well, applying the change requires a restart.

### DIFF
--- a/vita3k/config/include/config/config.h
+++ b/vita3k/config/include/config/config.h
@@ -185,7 +185,7 @@ enum ScreenshotFormat {
     code(int, "keyboard-take-screenshot-alt", 0, keyboard_take_screenshot_alt)                          \
     code(int, "keyboard-pinch-modifier-alt", 0, keyboard_pinch_modifier_alt)                            \
     code(int, "keyboard-alternate-pinch-in-alt", 0, keyboard_alternate_pinch_in_alt)                    \
-    code(int, "keyboard-alternate-pinch-out-alt", 0, keyboard_alternate_pinch_out_alt)                 \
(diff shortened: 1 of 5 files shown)


nishinji:
Why start something new when there’s already a better PR?
https://github.com/Vita3K/Vita3K/pull/3634

Zangetsu38:
> Why start something new when there’s already a better PR? #3634

hmm yes, but need rebase it lol, here is before moving on sdl3, so pr is very outdated and big conflict

Macdu:
Adding an option to enable/disable font scaling is not a proper solution. Without font scaling, text looks super blurry on high-definition displays.
The issue is that even with only a regular full HD display, the font texture is 32Kx4K (that's 128 MB of VRAM, which is the same amount the PS Vita has in total) and as a consequence takes 3-4 seconds to load each time Vita3K is started. 
The correct solution is the idea in #3634, not this.

Zangetsu38:
> Adding an option to enable/disable font scaling is not a proper solution. Without font scaling, text looks super blurry on high-definition displays. The issue is that even with only a regular full HD display, the font texture is 32Kx4K (that's 128 MB of VRAM, which is the same amount the PS Vita has in total) and as a consequence takes 3-4 seconds to load each time Vita3K is started. The correct solution is the idea in #3634, not this.

Yes i know 🧍 i have just proposed it after nuru talking to can having option for can disable this

And like i said orher pr is outdated qnd request re works 🥲
Anyway i close

PR #3922, opened 2026-05-12
[TESTING NEEDED] kernel/thread: rework run_loop lifecycle (drop `ThreadToDo` + `something_to_do`)

xerpi:
Replace `ThreadToDo` and the `something_to_do` condvar with four boolean flags: `exit_requested`, `delete_requested`, `suspend_requested`, `single_stepping`. `run_loop()` returns void.

`call_level` brackets each frame: incremented on entry, decremented just before return.

All call-level exits (normal exit, guest return, delete) converge on a single block at the top of the loop that fires the end callback and parks dormant. This wakes `wait_thread_end` waiters on every exit path.
If deleted, return completely from all `run_loop()` nested levels.

### DIFF
--- a/vita3k/kernel/include/kernel/thread/thread_state.h
+++ b/vita3k/kernel/include/kernel/thread/thread_state.h
@@ -58,15 +58,6 @@ private:
     bool signaled = false;
 };
 
-// Internal
-enum class ThreadToDo {
-    remove,
-    run,
-    step,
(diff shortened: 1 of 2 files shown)


nishinji:
I can't get past the opening video in the Miku Diva series

imWatchful (approved):

PR #3923, opened 2026-05-14
kernel: replace global atomic with cleaner two-phase thread shutdown

xerpi:
Replace `exit_delete_all_threads_and_wait()` and a global `std::atomic<bool> shutting_down` with a cleaner two-phase flow:
- `process_exit()` blocks until all guest threads have exited (host-only)
- `request_process_exit()` fires the registered callback and returns immediately, safe to call from guest threads.

Threads are now SDL_DetachThread'd at creation. `thread_deleted_cond` on `KernelState` is used for join semantics in `process_exit()`.

Remove `shutting_down` from sync primitives; `exit_delete()` wakes blocked threads via `update_status(run)`, which all `status_cond` predicates already check.

`process_exit_callback` carries an `optional<AppLaunchRequest>` so `LoadExec` relaunches flow through the same path as normal exits. `AppLaunchRequest`/`AppLaunchReason` extracted to `emuenv/app_launch_request.h` so kernel can include it without a circular dependency.

After this, further simplification to `ThreadStatus` will be done in https://github.com/Vita3K/Vita3K/pull/3922.

### DIFF
--- a/vita3k/app/src/app_init.cpp
+++ b/vita3k/app/src/app_init.cpp
@@ -443,7 +443,7 @@ void deinit(EmuEnvState &state) {
     state.net.abort_all();
     state.http.shutdown_connections();
 
-    state.kernel.exit_delete_all_threads_and_wait();
+    state.kernel.process_exit();
 
     state.motion.reset_runtime();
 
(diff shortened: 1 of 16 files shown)


xerpi:
@imWatchful It would be nice if you could help test this a bit more!

xerpi:
Deadlock when exiting Wipeout (demo) fixed.

imWatchful (approved):
Great stuff

PR #3926, opened 2026-05-15
gui: fix keyboard input on macOS

aifunc:
## Summary

Fix keyboard input on macOS by using `QKeyEvent::nativeVirtualKey()` for physical key mapping.

Previous versions could not use keyboard input correctly on macOS because `nativeScanCode()` does not provide the macOS virtual key code
expected by `NativeScanCodeSet::MacOS`. This changes only the macOS path; other platforms continue to use `nativeScanCode()` unchanged.

## Testing

- Built and tested on macOS.
- Verified keyboard input works on macOS after this change.


imWatchful (approved):
Thanks!

PR #3927, opened 2026-05-15
vulkan: fix macOS swapchain setup with Qt 6.11

aifunc:
## Summary

Fixes a macOS Vulkan startup crash seen with newer Qt versions when booting a game.

On macOS, the Vulkan surface is backed by MoltenVK and requires a valid `CAMetalLayer` on the native Qt view. Newer Qt versions can leave the native view
layer in a state that is not suitable for swapchain creation, so this patch explicitly ensures that the view uses a configured `CAMetalLayer`.

This also adds a defensive swapchain validation check so Vita3K does not continue with an empty or invalid swapchain.

## Scope

This is macOS-specific for the Metal layer setup path. Windows and Linux do not use `CAMetalLayer` or MoltenVK's macOS surface path.

## Testing

- Built on macOS with Qt 6.11.1
- Verified game boot on macOS with Qt 6.10.0, 6.11.0, and 6.11.1

### DIFF
--- a/vita3k/renderer/src/vulkan/metal_layer.mm
+++ b/vita3k/renderer/src/vulkan/metal_layer.mm
@@ -19,5 +19,30 @@
 #import <QuartzCore/CAMetalLayer.h>
 
 extern "C" void *get_metal_layer_from_view(void *nsview) {
-    return (__bridge void *)((__bridge NSView *)nsview).layer;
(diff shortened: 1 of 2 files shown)


imWatchful:
Thanks for the PRs, for this specific one, the issue is a little bit bigger than just at the metal layer, and requires a different fix. Should be resolved by #3959

PR #3935, opened 2026-05-16
gui-qt: fix string/path/QString conversion for non ASCII strings

bookmist:
It's important for old Windows before 2019 (win8 and earlier)

### DIFF
--- a/vita3k/compat/src/compat.cpp
+++ b/vita3k/compat/src/compat.cpp
@@ -18,6 +18,7 @@
 #include <compat/functions.h>
 #include <compat/state.h>
 #include <config/state.h>
+#include <fmt/std.h>
 #include <util/log.h>
 
 #include <miniz.h>
@@ -127,13 +128,13 @@ bool load_from_disk(CompatState &state, const std::filesystem::path &cache_path)
     const auto db_path = cache_path / "app_compat_db.xml";
 
     if (!std::filesystem::exists(db_path)) {
-        LOG_WARN("DB not found at {}", db_path.string());
+        LOG_WARN("DB not found at {}", db_path);
         return false;
     }
 
     std::ifstream file(db_path, std::ios::binary | std::ios::ate);
     if (!file) {
-        LOG_ERROR("Could not open {}", db_path.string());
+        LOG_ERROR("Could not open {}", db_path);
         return false;
     }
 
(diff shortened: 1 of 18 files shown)


imWatchful:
Our CI format check is acting extremely weird for qt files. I have to use a very old version of clang-format for our CI to accept it, like 15 or something.

vita3k/compat/src/compat.cpp
```diff
@@ -167,7 +168,7 @@ bool install_db(CompatState &state, const std::filesystem::path &cache_path,
         }
 
         const auto out_path = cache_path / stat.m_filename;
-        if (!mz_zip_reader_extract_to_file(&zip, i, out_path.string().c_str(), 0)) {
+        if (!mz_zip_reader_extract_to_file(&zip, i, reinterpret_cast<const char *>(out_path.u8string().c_str()), 0)) {
```
imWatchful, line 171: Are you sure this is correct? Might be worth looking at the internals here to make sure `const char *` gets interpreted correctly here and not as a system code page.

vita3k/compat/src/compat.cpp
```diff
@@ -18,6 +18,7 @@
 #include <compat/functions.h>
 #include <compat/state.h>
 #include <config/state.h>
+#include <spdlog/fmt/bundled/std.h>
```
imWatchful: instead of including this maybe something like 
```
template <>  
struct fmt::formatter<std::filesystem::path> : ostream_formatter {};
``` 
in `util/fs.h` is a better idea?

vita3k/compat/src/compat.cpp
bookmist: miniz library uses utf-8 for paths. so u8string here is correct.

vita3k/compat/src/compat.cpp
bookmist: I'm not sure that including `std::filesystem` in widely used `util/fs.h` is good. Less headers, faster build. Also I like libraries code more then my.
But I change from `{fmt}`, included in spdlog to project's `{fmt}` (to avoid version conflict, cause (as I think) spdlog use project's fmt)

imWatchful (approved):
Awesome

Dicot0721:
> Our CI format check is acting extremely weird for qt files. I have to use a very old version of clang-format for our CI to accept it, like 15 or something. 

Would it be better to use `jidicula/clang-format-action`?

imWatchful:
> > Our CI format check is acting extremely weird for qt files. I have to use a very old version of clang-format for our CI to accept it, like 15 or something.
> 
> Would it be better to use `jidicula/clang-format-action`?

That looks pretty good yeah

PR #3936, opened 2026-05-16
io: replace better-enums lib to boost::describe

bookmist:
Including an entire library for a single enum feels excessive.

### DIFF
--- a/vita3k/io/src/io.cpp
+++ b/vita3k/io/src/io.cpp
@@ -91,17 +91,23 @@ SceSize get_directory_used_size(const VitaIoDevice device, const std::string &vf
 // * End utility functions *
 // ****************************
 
+static bool is_valid_output_path(const VitaIoDevice device) {
+    return !(device == VitaIoDevice::savedata0 || device == VitaIoDevice::savedata1 || device == VitaIoDevice::app0
+        || device == VitaIoDevice::_INVALID || device == VitaIoDevice::addcont0 || device == VitaIoDevice::tty0
+        || device == VitaIoDevice::tty1 || device == VitaIoDevice::tty2 || device == VitaIoDevice::tty3
+        || device == VitaIoDevice::music0 || device == VitaIoDevice::photo0 || device == VitaIoDevice::video0);
+}
+
 bool init(IOState &io, const fs::path &cache_path, const fs::path &log_path, const fs::path &pref_path, bool redirect_stdio) {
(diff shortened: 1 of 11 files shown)


vita3k/io/src/io.cpp
```diff
@@ -93,15 +93,14 @@ SceSize get_directory_used_size(const VitaIoDevice device, const std::string &vf
 
 bool init(IOState &io, const fs::path &cache_path, const fs::path &log_path, const fs::path &pref_path, bool redirect_stdio) {
     // Iterate through the entire list of devices and create the subdirectories if they do not exist
-    for (auto i : VitaIoDevice::_names()) {
-        if (!device::is_valid_output_path(i))
-            continue;
-        fs::create_directories(pref_path / i);
-    }
+    boost::mp11::mp_for_each<boost::describe::describe_enumerators<VitaIoDevice>>([pref_path](auto i) {
```
EXtremeExploit:
```suggestion
    boost::mp11::mp_for_each<boost::describe::describe_enumerators<VitaIoDevice>>([&pref_path](auto i) {
```

EXtremeExploit (changes requested):
pref_path should be by reference like most strings/other references to it

Dicot0721:
It seems that [dependencies](https://github.com/Vita3K/Vita3K/blob/4e03f06a0684e27fc4495c25a4751824d8aba723/.github/workflows/c-cpp.yml#L104) need to be updated?
image

vita3k/io/src/io.cpp
bookmist: fixed. thanks.

bookmist:
> It seems that [dependencies](https://github.com/Vita3K/Vita3K/blob/4e03f06a0684e27fc4495c25a4751824d8aba723/.github/workflows/c-cpp.yml#L104) need to be updated? image

No need, compile error was my mistake and now it's fixed. (I hope)

EXtremeExploit (approved):
awesome sauce

Dicot0721:
> No need, compile error was my mistake and now it's fixed. (I hope)

In fact, `enum_from_string` has an overloaded version that accepts a string-like argument (only the Linux build using older versions of Boost lack this, which is why the compilation fails only on Linux 😂).

```cpp
template <class S, class E>
bool enum_from_string(const S& name, E& e);
```

PR #3941, opened 2026-05-18
Update labeler.yml

nishinji:
Update the bot's tagging to adapt the latest repository


building.md
```diff
@@ -1,6 +1,6 @@
 # Build Vita3K
```
imWatchful, line 1: Since you're editing this, could you bring over the android build instructions from Vita3K-Android? (and maybe mention Qt binaries are required now for desktop)

imWatchful (approved):
<3

PR #3947, opened 2026-05-20
gui-qt: Did someone say themes? (100 times in discord?) (desktop only)

imWatchful:
Themes are back, qt style (and a qss refactor but who cares about that)!
Videos are always better than words:

https://github.com/user-attachments/assets/de5403dc-ec66-4dea-b605-93ca96703ffc

NOTE: Not all generated themes will come out looking good, do not open issues/ask for help with making a theme "work/look good/become readable". Additionally, until we switch to building our own Qt binaries, appimages and Linux builds might not be able to play at9 tracks through Qt Multimedia without explicitly setting ffmpeg as it's backend through an environment variable `QT_MEDIA_BACKEND=ffmpeg`, please do not open issues on that matter as well. 

P.S. Lockscreen? Never heard of her.


EXtremeExploit (approved):
ok

PR #3950, opened 2026-05-20
gxm/app: fix shutdown deadlocks caused by scegxm/clean up session controller

imWatchful:
Track all gxm sync objects, invalidate on shutdown instead of using bounded timeout coping. Resolves deadlocks during shutdown for Gravity rush, unit13 and others.
Clean up session controller
Move gxm shutdown to a designated function


EXtremeExploit (approved):
whatever goat

PR #3956, opened 2026-05-21
gui-qt: don't crash when deleting an archive locked by another process

troglodyte9:
after you install a .zip or .vpk, clicking yes on "delete archive files after install?" triggers fs::remove on those files. the problem is that it's using the throwing version of fs::remove. because this runs inside a qt slot connected to QDialogButtonBox::accepted, any exception that gets thrown just escapes straight into the event loop and completely crashes the emulator. 

this happens whenever an archive is still held open by another process, for example an archive manager left open on the .vpk being installed. use the non-throwing boost::system::error_code overload of fs::remove, collect the archives that could not be deleted, and show a warning that lists them instead of crashing.

### DIFF
--- a/vita3k/gui-qt/include/gui-qt/firmware_install_dialog.h
+++ b/vita3k/gui-qt/include/gui-qt/firmware_install_dialog.h
@@ -52,6 +52,7 @@ class FirmwareInstallDialog : public QDialog {
     Q_OBJECT
 public:
     explicit FirmwareInstallDialog(EmuEnvState &emuenv, QWidget *parent = nullptr);
+    FirmwareInstallDialog(EmuEnvState &emuenv, const QString &path, QWidget *parent = nullptr);
 
 Q_SIGNALS:
     void install_complete();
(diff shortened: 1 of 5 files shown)


imWatchful:
Rebase/force push, our CI just changed, github is tweaking out

troglodyte9:
alright

troglodyte9:
done

imWatchful:
This PR is not a bad shout, but this same bug exists inside of pkg installs, while our firmware install uses QFile::remove which would fail silently (and our drop implementation completely ignores the regular firmware dialog completely 🤪). Now this would require a bit more work to align everything properly and I was planning to do this in another PR, if you don't mind doing all of that I'd accept the PR otherwise I can have a bigger patch for this soon. Tell me how you're feeling

troglodyte9:
ill try my best

imWatchful:
Awesome you can dm me on discord if you need help.

troglodyte9:
bumping this, this should fix that behavior

imWatchful (approved):
Nice!

PR #3958, opened 2026-05-21
modules: introduce virtual modules (kubridge, fd_fix)

nn9dev:
A virtual module is a module that is implemented wholly at an HLE level, that is, there is no file actually being loaded from disk. This is nice in certain scenarios where the behavior or fix from a module is something that isn't a problem on an emulator due to lack of memory boundaries (kubridge) and other fun stuff (like fd_fix, or maybe repatch in the future).

Nevertheless, some homebrew titles check if a given module has been loaded, and we should have a nice system that not only acknowledges the module as loaded, but also slots in nice with the rest of the module system. A "fake" module system works well for this.

The motivation behind this PR is to make it so that neither kubridge nor fd_fix (or other future plugins that can be implemented within the emulator) need to be downloaded.

A few concerns:
- Currently kubridge and fd_fix are ""loaded"" with preload modules. maybe an option for this would be better? I guess it's not hurting anything atm
- CRC32 may not be preferred, but it's simple. open to suggestions

### DIFF
--- a/vita3k/modules/kubridge/kubridge.cpp
+++ b/vita3k/modules/kubridge/kubridge.cpp
@@ -23,13 +23,17 @@
 
 #include <module/module.h>
 
+#include <../SceSysmem/SceSysmemForDriver.h>
 #include <cpu/functions.h>
 #include <kernel/state.h>
+#include <kernel/types.h>
 #include <mem/functions.h>
 #include <modules/sysmem_state.h>
 #include <util/align.h>
 #include <util/log.h>
 #include <util/tracy.h>
+
+#include <cstdlib>
 
 TRACY_MODULE_NAME(kubridge);
 
@@ -231,4 +235,34 @@ EXPORT(void, kuKernelReleaseExceptionHandler, uint32_t exceptionType) {
         const Address old = emuenv.kernel.exception_handlers[exceptionType].exchange(0);
         LOG_INFO("kuKernelReleaseExceptionHandler: type={} old=0x{:08X}", exceptionType, old);
     }
+}
+
+EXPORT(void, kuKernelFlushCaches, const void *ptr, SceSize len) {
+    TRACY_FUNC(kuKernelFlushCaches, ptr, len);
+    STUBBED("Emulator has no real CPU caches");
+}
+
+EXPORT(SceUID, kuKernelAllocMemBlock, const char *name, SceKernelMemBlockType type, SceSize size, SceKernelAllocMemBlockKernelOpt *opt) {
+    TRACY_FUNC(kuKernelAllocMemBlock, name, type, size, opt);
+    SceUID res = CALL_EXPORT(ksceKernelAllocMemBlock, name, type, size, opt);
+    if (res < 0)
+        return res;
+
+    return CALL_EXPORT(kscePUIDOpenByGUID, 1, res);
+}
+
+EXPORT(int, kuKernelCpuUnrestrictedMemcpy, void *dst, const void *src, SceSize len) {
+    STUBBED("Regular memcpy used");
+    memcpy(dst, src, len);
+    return 0;
+}
+
+EXPORT(int, kuPowerGetSysClockFrequency) {
+    STUBBED("clock freq: 333");
+    return 333;
+}
+
+EXPORT(int, kuPowerSetSysClockFrequency, int freq) {
+    STUBBED("set clock freq success (0)");
+    return 0;
 }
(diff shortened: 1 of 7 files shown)


bookmist:
I have almost the same idea for my [LLE sysmodule implementation](https://github.com/Vita3K/Vita3K/pull/3913/changes#diff-259f5b87473c4e35e33bc7295a8fe5f359873ec181a918c7bbd51de19ce9e2c2R213-R227)
Some ideas:
* move check is module lle or hle inside load_module.
* module_name and file name are usually different (at least for Sce modules)

nn9dev:
> I have almost the same idea for my [LLE sysmodule implementation](https://github.com/Vita3K/Vita3K/pull/3913/changes#diff-259f5b87473c4e35e33bc7295a8fe5f359873ec181a918c7bbd51de19ce9e2c2R213-R227) Some ideas:
> 
> * move check is module lle or hle inside load_module.
> * module_name and file name are usually different (at least for Sce modules)

Ah, I see. I like the idea. I think the difference between our implementations is that for mine you can just provide a module name but yours is for an entry on disk (vs0:sys/external/filename.skprx). In that sense, mine is a bit more aimed at tai modules or tai kernel modules which don't exist in firmware, so the module could have any arbitrary name or path, and loads before the title with tai and not at runtime (like SceShaccCg, which is loaded manually by an app), though I guess bootimage also loads before everything else...

Right now, the code basically serves to put an entry in the modules list that can also be referred to by NID.

I appreciate the second point, I could certainly change my function to require a module_name as well as a file name like `load_virtual_module(emuenv, "SceMemDisk", "memdisk.skprx")`, or even a full fake path. If you want me to change anything or if you'd like to use my code as an extension to your PR, I would be open to that, or open to any other collaboration :D

bookmist:
I more like using standard ux0:tai/config.txt for configuration of loaded plugins and moving load_virtual_module as a part of load_module.
Also it avoid duplication of check if module is already loaded.

I'm thinking about putting almost all files in firmware to Core tab in settings and allowing user to choose which they want to HLE/LLE (of course automatic mode will be the best and most stable variant), but this will be in some (far) future.

config.txt sample
```
# Kernel plugins are started with taiHEN and are in this section
*KERNEL
ux0:tai/kubridge.skprx
ux0:tai/fd_fix.skprx
```
code sample
```
    LOG_INFO("Loading module \"{}\"", module_path);
    if (module_path.starts_with("vs0:sys/external/") || module_path.contains("kubridge") || module_path.contains("fd_fix")) {
        // check if module is LLEd or not
        // check only for this path because app modules are always LLEd and os0 modules can be loaded only by developer.
        const auto module_name = fs::path(module_path).stem().string();
        if (!is_lle_module(module_name, emuenv) || module_name == "kubridge" || module_name == "fd_fix") {
            LOG_INFO("Module {} is HLE. Skipping load.", module_name);```

**Comment by nn9dev** [comment]
I understand the point about taiHEN, but the main motivation was so that there would be no configuration needed for kubridge and fd_fix and that it would Just Work in the emulator. I don't see any downsides to having them "on" always, and I think it just makes it more of a hassle for someone to have to create a file in order to enable them. I suppose alternative solutions could be to make it a checkbox in settings or to have the emulator create a file with the content in it upon first start. What are your thoughts on that?

The second part of your response hinges upon the first part, and I would be open to refactoring it. In that case, it might be best to have the emulator create that config file, since then it would cause taiHEN to call load_module() for kubridge and fd_fix.

**Comment by imWatchful** [comment]
I like the hands-off approach more than a specific setting or creating a file personally.

File: vita3k/modules/kubridge/kubridge.cpp [hunk 1]
```diff
@@ -232,3 +236,33 @@ EXPORT(void, kuKernelReleaseExceptionHandler, uint32_t exceptionType) {
         LOG_INFO("kuKernelReleaseExceptionHandler: type={} old=0x{:08X}", exceptionType, old);
     }
 }
+
+EXPORT(void, kuKernelFlushCaches, const void *ptr, SceSize len) {
+    TRACY_FUNC(kuKernelFlushCaches, ptr, len);
+    STUBBED("Emulator has no real CPU caches");
```
bookmist, line 242: I'm not sure, but dynarmic jit cache exists and may be important, or not.

vita3k/modules/module_parent.cpp
```diff
@@ -306,6 +314,106 @@ SceUID load_module(EmuEnvState &emuenv, const std::string &module_path) {
     return load_module_data(module_buffer.data());
 }
 
+SceUID load_virtual_module(EmuEnvState &emuenv, const std::string &module_name) {
```
bookmist, line 317: It looks like it can be moved to kernel library. This function use only kernel and mem components, like load_self? so no external dependencies would be needed here.

PR #3962, opened 2026-05-24
gui-qt: add copy test environment summary to compatibility menu

### LINKED ISSUE #3586: [Feature Request]: UI Add a way to copy "Test environment summary"

"Test environment summary" is required for bug reporting but the only way to quickly get one is if the game has no open report. Which is often not the case. It's supper annoying to create manually or going to an old entry to copy it.
I propose you add a way to copy "Test environment summary" to clipboard OR that the new issue template is always available but as a reply to the existing issue if it already exists < That would be excellent for users.
If not at least copying the env. summary would help too.

Very useful for people helping make the compatibility list.

_No response_

### Platform (if relevant)

None

troglodyte9:
when a game already has a compatibility report, the context menu only lets you copy the vita3k build summary. this adds a second action to also copy the test environment summary (os, cpu, gpu, ram) so users can paste it into existing issue comments without having to manually type it out.

closes #3586


imWatchful (approved):

RetroGameTalk:
Awesome thanks 🙏

PR #3963, opened 2026-05-24
gui-qt: New theme

jojo11119:
This theme is based on the look of the previous ImGui.
For the animated backgrounds full firmware package installation is required.

iscrazyLimited_nocandoingAnythings


silver178:
Now this would be a good of 2 worlds here.

imWatchful (approved):
Revert all 🗿

PR #3965, opened 2026-05-25
ci: Add AppImage zsync update

EXtremeExploit:
Adds the zsync update information to appimages, so they can be updated with any appimage manager automatically or manually without manually going to the downloads or manual download.
No change should be needed on the release-collect side of things

Both arm and x86 are supported


imWatchful (approved):
Penguins rejoice!

PR #3967, opened 2026-05-25
vita3k: Rework linux default paths

EXtremeExploit:
Reworks how linux default paths are handled
* Now static_assets cant be in an XDG directory, since this didnt ever really happen and theres 0 reason for this to happen, and if it happens its not good
* Adds a fallback to $HOME using the passwd file
* Adds support for reading static files from `/usr/share/Vita3K` and the local variant
* Adds a few comments saying what each path is

* Removes a bunch of paths from app_list resolving, since ux0/vs0 cant really be in them, theoretically making app_list a little bit faster because it has to check less paths

### DIFF
--- a/vita3k/app/src/apps_list.cpp
+++ b/vita3k/app/src/apps_list.cpp
@@ -193,11 +193,6 @@ static std::string resolve_existing_path(const EmuEnvState &emuenv, const fs::pa
     const std::array<fs::path, 7> roots = {
         emuenv.pref_path,
         emuenv.default_path,
-        emuenv.shared_path,
-        emuenv.static_assets_path,
-        emuenv.base_path,
-        emuenv.config_path,
-        emuenv.cache_path,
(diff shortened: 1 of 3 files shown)


imWatchful (approved):

vita3k/app/src/apps_list.cpp
```diff
@@ -193,11 +193,6 @@ static std::string resolve_existing_path(const EmuEnvState &emuenv, const fs::pa
     const std::array<fs::path, 7> roots = {
```
Dicot0721, line 193:
```suggestion
    const std::array<fs::path, 2> roots = {
```

imWatchful: O i missed this...

EXtremeExploit: oups

PR #3968, opened 2026-05-25
i18n: Make use of static assets instead of hardcoded base

EXtremeExploit:
Basically title, shouldnt change anything except for the appimage, which right now there isnt any difference because there arent any translations available yet

### DIFF
--- a/vita3k/gui-qt/src/gui_language.cpp
+++ b/vita3k/gui-qt/src/gui_language.cpp
@@ -61,19 +61,18 @@ const std::array<UiLanguageOption, 23> k_ui_languages = { {
 
 std::unique_ptr<QTranslator> s_translator;
 
-QStringList translation_search_paths() {
-    const QString app_dir = QCoreApplication::applicationDirPath();
+QStringList translation_search_paths(fs::path &static_assets_path) {
     return {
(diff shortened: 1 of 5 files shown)


imWatchful (approved):
🐧

vita3k/gui-qt/src/gui_language.cpp
```diff
@@ -61,19 +61,18 @@ const std::array<UiLanguageOption, 23> k_ui_languages = { {
 
 std::unique_ptr<QTranslator> s_translator;
 
-QStringList translation_search_paths() {
-    const QString app_dir = QCoreApplication::applicationDirPath();
+QStringList translation_search_paths(fs::path &static_assets_path) {
```
Dicot0721, line 64: If it hasn't been merged, we can make `static_assets_path` a const ref 😣

imWatchful: It's been merged but I'm pretty sure there's gonna be another linux pr soon so it can be fixed then.

EXtremeExploit: Should the other references to it in the file/gui also be const?

PR #3969, opened 2026-05-26
ngs: refactor ngs/audio state ownership & some cleanup

imWatchful:
The point of this refactor is to split up ngs/audio state into guest-visible, internal logical processing and host runtime states. This allows us to serialize this state (and rebuild runtime state) in a much easier manner later on (😉). 
`ModuleData`, `Patch`, `Player`, and `Atrac9` were reworked around that split. Guest-facing reads now pull from the guest-visible side of `ModuleData`. Progress counters and guest-facing status remain in the public state structs, while queued decoded PCM, partial frame carryover, decoder history, and resampler replay history now live in logical state. FFmpeg/libatrac9 decoder instances, `SwrContext`s, and temporary decode buffers were moved into runtime state. A shared stereo rate-resampler helper was added so both modules can preserve enough logical history to rebuild resampler behavior cleanly without storing raw runtime objects.
Module setup now goes through `initialize_voice_data()`, so each voice/module pair gets its expected guest/logical/runtime state up front.
Some cleanup:
- buffered PCM is now managed through a shared `PCMFrameQueue` 
- decode paths reuse persistent runtime buffers instead of rebuilding temporary vectors as often

### DIFF
--- a/vita3k/modules/SceNgsUser/SceNgs.cpp
+++ b/vita3k/modules/SceNgsUser/SceNgs.cpp
@@ -218,18 +218,25 @@ EXPORT(SceInt32, sceNgsPatchGetInfo, ngs::Patch *patch, SceNgsPatchAudioPropInfo
         return RET_ERROR(SCE_NGS_ERROR_INVALID_ARG);
     }
 
+    patch->refresh_endpoints(emuenv.mem);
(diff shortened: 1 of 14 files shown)


EXtremeExploit:
miku diva x sound seems to work fine, tested menuing and music and all sounds play normally

EXtremeExploit (approved):

PR #3970, opened 2026-05-26
i18n: Make static_assets a const reference

EXtremeExploit:
addresses some of the suggestions from https://github.com/Vita3K/Vita3K/pull/3968 from @Dicot0721 

Also fixes the roots array size for apps_lists

### DIFF
--- a/vita3k/app/src/apps_list.cpp
+++ b/vita3k/app/src/apps_list.cpp
@@ -190,7 +190,7 @@ static std::string resolve_existing_path(const EmuEnvState &emuenv, const fs::pa
         return {};
     }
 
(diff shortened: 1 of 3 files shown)


imWatchful (approved):

vita3k/app/src/apps_list.cpp
```diff
@@ -190,7 +190,7 @@ static std::string resolve_existing_path(const EmuEnvState &emuenv, const fs::pa
         return {};
     }
 
-    const std::array<fs::path, 7> roots = {
```
Dicot0721, line 193: Oh, I forgot to mention that we can use const ref here… Nvm

Issue #3960, opened 2026-05-23
[Feature Request]: (Desktop) Allow for remembering size of rendering window, or setting a custom default size.


FurbyTime:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What feature are you suggesting?

Allow for Vita3k to either remember the size of the rendering window, or to set a new default size of the rendering window, in non-fullscreen scenarios.

### Why would this feature be useful?

For situations where full screen may not be the most useful of views, but users would still like the game to be displayed at larger than 1x in Windowed mode by default.

### Other emulators or software with a similar feature

Dolphin, MGBA, Ares, PPSSPP, etc.

### Platform (if relevant)

None

Daisouji:
We need this for a minute.

AvanMadisen:
As a 4K monitor user, I would very much like for this feature to be added.

PR #3971, opened 2026-05-26
gui-qt: add --exit-on-game-close cli flag

### LINKED ISSUE #2884: feature request: quit emulator after closing game
To improve the usefulness of the `-r,--installed-path TEXT:{}` command line option to automatically boot a game upon emulator start, it would be nice if the emulator could also automatically quit after the running game is closed (*home menu → pressing the cancel/back button - Cross in my configuration*). I couldn't find such an option.

A use case for this are game launchers (eg: Steam Deck) to abstract away most of the emulator interface. RPCS3 (PS3 emulator) can be used in such a way using the `--no-gui --fullscreen` options, as a real-world example.

troglodyte9:
adds a command line option to quit the emulator when the running game closes. useful for game launcher integrations like steam deck where the emulator should behave transparently without exposing its main window after the game exits.

usage: Vita3K --installed-path PCSG00001 --fullscreen --exit-on-game-close

closes #2884


Tartifless:
+1, but i guess it would be better as a GUI option

imWatchful:
I think for this to be a real feature/flag, I would much prefer if we were to take the rpcs3 route of the `--no-gui` flag. This would mean, skip the creation of `MainWindow` if this flag is invoked, and the exit path should also handle the "no MainWindow" case. This would implicitly exit the emulator when the emulated app is shut down as well. Now this is slightly more work than `--exit-on-game-close` but I think it would be a cleaner solution overall.

imWatchful:
Couple of small issues: 1) the --no-gui path shuts down the emulator even if the shutdown request came from `sceAppMgr` (games like God of war, uncharted etc, "relaunch" themselves). 2) The controls dialog is no longer receiving controller SDL events, so the user cannot rebind their mappings. 3) You hardcoded fullscreen on launch, instead of deriving that from the user's config, and always set the gamewindow to use the primary screen. Since we already have a value in gui_settings for `gw_geometry` that I didn't get around to using, it might be a good idea to knock #3960 out as well (just persist the geometry from however the user modified the screen - like mainwindow - no need for a new setting to set a new default size or anything).

EXtremeExploit:
what, noo this was awesome it just needed review D:

troglodyte9:
> what, noo this was awesome it just needed review D:

im redoing it since it got really messy

PR #3972, opened 2026-05-26
ngs: fix patch handle storage regression

imWatchful:
Fixes an NGS regression where I inflated patch layout and caused work buffer overruns in some games. This switches patch routing back to guest handles.


EXtremeExploit (approved):
k

PR #3974, opened 2026-05-26
fonts: Remove bundled fonts

EXtremeExploit:
Bundled fonts arent really needed, now that system fonts are used in the new overlay the odds of them being used is pretty much 0%, Arial and the others are pretty much guaranteed to be present, and firmware font is necessary for some games anyways, so even less people might use system fonts
Bundled fonts are only used in the overlay


imWatchful (approved):
search the string `imgui` in the codebase for an easter egg

PR #3978, opened 2026-05-28
io: remove unused files and dependencies

bookmist:


imWatchful:
`dialog` lost its transitive `emuenv` include I think

imWatchful (approved):

PR #3980, opened 2026-05-29
kernel: Fix elf dumping on 0 size dumps

EXtremeExploit:
Now that modules from bootimage can be loaded, some of them are treated as 0 sized modules, make it ignore them at least for the time being


bookmist:
#3984

EXtremeExploit:
thank you my goat

PR #3985, opened 2026-05-30
cpu: Fix log_code spinning forever on pc

EXtremeExploit:
Ever since the dawn of time watch code never worked, well it did, but after a few hundred/thousand instructions it would spin forever on the same instruction, making it useless
This was because IR doesnt handle calling it very well, so instead of using IR, just read the instruction directly from memory and parse it ourselves instead of making the IR call the function to translate

### DIFF
--- a/vita3k/cpu/src/dynarmic_cpu.cpp
+++ b/vita3k/cpu/src/dynarmic_cpu.cpp
@@ -22,7 +22,6 @@
 
 #include <mem/ptr.h>
 
-#include <dynarmic/frontend/A32/a32_ir_emitter.h>
 #include <dynarmic/interface/A32/coprocessor.h>
 #include <dynarmic/interface/exclusive_monitor.h>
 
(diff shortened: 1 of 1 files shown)


vita3k/cpu/src/dynarmic_cpu.cpp
```diff
@@ -141,21 +140,15 @@ class ArmDynarmicCallback : public Dynarmic::A32::UserCallbacks {
...
-
     void PreCodeTranslationHook(bool is_thumb, Dynarmic::A32::VAddr pc, Dynarmic::A32::IREmitter &ir) override {
         if (cpu->log_code) {
-            ir.CallHostFunction(&TraceInstruction, ir.Imm64((uint64_t)this), ir.Imm64(pc), ir.Imm64(is_thumb));
+            std::string disassembly = "invalid address";
+            if (!pc || !Ptr<uint32_t>{ pc }.valid(*parent->mem)) {
+                disassembly = "invalid address";
+            } else {
+                disassembly = disassemble(*parent, pc);
+            }
```
Dicot0721, line 150:
```suggestion
            if (pc && Ptr<uint32_t>{ pc }.valid(*parent->mem)) {
                disassembly = disassemble(*parent, pc);
            }
```

Macdu:
Your change does not have the same effect, it will only log the disassembly the first time an instruction is reached and never again afterwards (assuming we do not invalidate the instruction range).
I've used watch code quite a few times and never had any issue with it by the way (beside the fact that it's really slow).

EXtremeExploit:
Ah, i didnt check, i've seen watch code always hang in the same instruction on most games by the IR running TraceInstruction forever and some other people telling it kinda does that and asumed it was broken for everyone
And i didnt actually notice that some instructions werent being logged, ill probably take this again in the future but right now i have 0 idea why this happens if we have to keep using IR

PR #3986, opened 2026-05-30
util: Optimize string utils

Dicot0721:


bookmist:
split_string is never used, so it can be removed completely

Dicot0721:
> split_string is never used, so it can be removed completely

OK

bookmist (dismissed):

bookmist (approved):
I'm not sure if it has sense, but why not.

PR #3990, opened 2026-06-02
ci: Optimize format check workflow

Dicot0721:
- Use `sparse-checkout` to fetch only the root files and `vita3k` and `tools` folder.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -2,28 +2,36 @@ name: Format Check
 
 on:
   push:
-    paths-ignore:
-      - '**/*.md'
+    paths:
+      - .clang-format
+      - vita3k/**
+      - tools/**
+
   pull_request:
-    paths-ignore:
-      - '**/*.md'
+    paths:
+      - .clang-format
+      - vita3k/**
+      - tools/**
 
 jobs:
   format-check:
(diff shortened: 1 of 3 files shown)


.github/workflows/format.yml
```diff
@@ -2,28 +2,36 @@ name: Format Check
...
     steps:
       - name: Checkout
         uses: actions/checkout@v6
+        with:
+          sparse-checkout: ${{ matrix.path }}
 
       - name: Format check
         uses: jidicula/clang-format-action@v4.18.0
         with:
-          clang-format-version: '22'
```
imWatchful, line 28: Why remove this?

Dicot0721: Because it defaults to 22, and removing it might help keep it latest.

bookmist: So when latest will change we will have a sudden surprise

Dicot0721: Yes, the latest clang-format may change, but that will only happen after we manually update the action version, so it won’t be "sudden". Compared to before we started using this action, I think it’s good enough.

bookmist: ok. good if so.

imWatchful (approved):

PR #3997, opened 2026-06-04
config: Only allow native buffer mapping on android

EXtremeExploit:
* Better entry for mapping methods in UI code, instead of magic numbers, it now correctly represents the enum values
* logging about native buffer being only available in android is now critical
* Dont parse native buffer in config if its not available, make it fallback to disabled
* Small comment about native buffer being only in android


EXtremeExploit:
Oops closed by accident

imWatchful (approved):

PR #4001, opened 2026-06-06
io: stop reporting regular files as executable

imWatchful:
Stop reporting execute bits on regular files, only applying them towards directories. Fixes save/load bug and death softlock bug on uncharted. 
Still refresh save metadata if `slotParam` is missing.
Fix swapped save/load cancel confirmation strings in common dialog.


EXtremeExploit (approved):
ok

PR #4002, opened 2026-06-06
vita3k/renderer/gl: Implement Nearest and Bicubic

nishinji:
90% AI code.
I don't know anything about graphics, but the results seem to be an improvement over what we have now, at least.

| NEAREST | Bilinear |
|----------|----------|
| image | image |

| Bicubic | FXAA |
|----------|----------|
| image | image |

### DIFF
--- a/vita3k/renderer/src/gl/screen_render.cpp
+++ b/vita3k/renderer/src/gl/screen_render.cpp
@@ -21,6 +21,18 @@
 
 namespace renderer::gl {
 
+const SharedGLObject &ScreenRenderer::current_shader() const {
+    switch (filter) {
+    case Filter::FXAA: return m_render_shader_fxaa;
+    case Filter::Bicubic: return m_render_shader_bicubic;
+    default: return m_render_shader_nofilter; // Bilinear / Nearest
+    }
+}
+
+GLuint ScreenRenderer::current_sampler() const {
+    return (filter == Filter::Nearest) ? m_sampler_nearest : m_sampler_linear;
+}
+
 bool ScreenRenderer::init(const fs::path &static_assets) {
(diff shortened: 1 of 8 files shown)


silver178:
I'm still surprised ppl are contributing to OpenGL since vulkan is just a better, more compatible replacement.

Your heart's in the right place.

vita3k/renderer/src/gl/screen_render.cpp
```diff
@@ -21,6 +21,18 @@
 
 namespace renderer::gl {
 
+const SharedGLObject &ScreenRenderer::current_shader() const {
```
EXtremeExploit: this can be static, only used in this file

vita3k/shaders-builtin/opengl/render_main_bicubic.frag
```diff
@@ -5,8 +5,6 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
-#version 410 core
```
EXtremeExploit, line 8: why?

EXtremeExploit (approved):
i think its ok, just the 1 small why, iirc android's opengl and some drivers need the #version 410 core, else it doesnt work, that might not be necessary anymore tho

vita3k/shaders-builtin/opengl/render_main_bicubic.frag
nishinji: Because
https://github.com/Vita3K/Vita3K/blob/257464af2fea1e3497d6df75ca3d3bb276e909a8/vita3k/glutil/src/shader.cpp#L40-L43

PR #4003, opened 2026-06-06
Popcorn!

imWatchful:
This PR is mainly a rewrite of how kernel waits are represented and driven internally.

Up until now, a lot of thread waiting behavior lived in random places spread across `ThreadState`, sync primitives, display vblank handling, and parts of kernel HLE. That made suspension, callback capable waits, timeouts, thread end waits, debugger stepping, and pause/resume behavior fairly inconsistent. It also made the current implementation really hostile to any future work where waits need to be restored or re-armed from serialized state.

The core of this PR is moving all of that toward a shared wait model.

`ThreadState` has been reworked around a new, more granular model. `KernelThreadManager` now owns the state transitions that were previously spread out across unrelated code, and `CoreTiming` introduces guest-time based scheduling that kernel waits and timers can share consistently.

This also changes how waiters are tracked across sync primitives. Instead of loosely coupling waits to object-local state, waiters are now keyed by `thread_id + wait_sequence`, which lets mutexes, rwlocks, semaphores, condvars, event flags, simple events, timers, and message pipes all complete waits through the same signaled/timeout/canceled/deleted flow. Delay waits, signal waits, thread-end waits, and vblank waits now go through the same path as well.

Vblank waits and timer accounting now use guest time instead of host clock time, and display vblank ticking wakes threads through the kernel wait layer rather than the old thread status "system". 

ThreadMgr and LibKernel exports were cleaned up around the new model as well. Apart from routing more paths through the shared wait helpers, this also implements vector msgpipe send/receive entry points and their callback capable variants, fixes some of the info/reporting paths, and removes several places that were still relying on older thread-status model.

As per Macdu's advice, I also removed the old host-to-guest pointer constructor from `mem::Ptr`, since it had UB once page table mode is involved. That change forced a small refactor in a few places that were still depending on the incorrect behavior: `sceClibMspace*`, `sceClibStrchr`, `sceClibStrrchr`, `sceClibStrstr`, Tracy pointer logging, LLE `sceFiber` guest pointer handling, and NGS callback handles. 

`Block` ownership in `mem` was also reworked into proper RAII + move semantics, with a bit of extra functionality added for upcoming work where blocks will need to be rebound to existing allocations.

On the side, there are a couple of smaller cleanups as part of this PR: a shared async flip wake helper in the renderer/session path, forced flips on emulation pause/resume, and removal of some stale audio thread-status updates.

This PR fixes the Ratchet & Clank Collection regression where none of the games in the collection would boot, and also fixes cases where games requested an emulation pause but parts of the system kept running anyway. It might also fix some other random issues as well, only time will tell.


bookmist:
SceFiber.cpp is dead code. LLE libfiber.suprx is used now.

imWatchful:
> SceFiber.cpp is dead code. LLE libfiber.suprx is used now.

Should it all be UNIMPLEMENTED() then?

EXtremeExploit:
I think its fine to still work on fibers and keep them implemented in HLE even if they arent run, just in case, plus who knows, maybe one day we can get rid of LLeing fibers and doing HLE on them for a bit more performance and control or some edge case that just happens with lle

vita3k/kernel/include/kernel/thread/wait_state.h
```diff
@@ -0,0 +1,261 @@
...
+    void clear() {
+        *this = {};
+    }
+
+    static WaitState delay(const uint32_t delay_us, const bool callbacks_allowed = false) {
+        WaitState state;
+        state.type = WaitType::Delay;
+        state.target_value = delay_us;
+        state.callbacks_allowed = callbacks_allowed;
+        return state;
```
bookmist, line 120: Actual code in header (at least in header user more then only once) is not a good idea (At least for more then one-liner). Move this in wait_state.cpp

Zangetsu38:
> > SceFiber.cpp is dead code. LLE libfiber.suprx is used now.
> 
> Should it all be UNIMPLEMENTED() then? 

Nop, keep code like that is better, can using it without lle

Macdu:
This PR makes a single thread handle all synchronization primitives.
This is definitely going to have a performance cost, especially for weaker hardware.

imWatchful:
> This PR makes a single thread handle all synchronization primitives. This is definitely going to have a performance cost, especially for weaker hardware.

Only the guest time scheduling and timed callbacks are run on 1 thread, not the sync primitives themselves. The actual mutex/semaphore/condvar/eventflag/msgpipe wait and wakes still execute through the calling guest thread state. I ran some games on my Pixel 6a (fairly old phone) and didn't see any performance regressions anywhere, I'm open to suggestions as always though.

vita3k/kernel/include/kernel/thread/wait_state.h
imWatchful: I'd like to get some more opinions on this, these are basically "constructors". It would be a good idea soon to move sync primitives to a proper C++ api instead of what we have now, which would eliminate the need for these entirely.

frangarcj:
I like the direction of this PR. `WaitState` + `wait_sequence`, `KernelThreadManager`, and guest-time based scheduling are a much cleaner foundation than the old scattered `ThreadStatus` model.

I do see several correctness issues that I think should be fixed before merging.

### 1. Stale waiters can break queue semantics

Wait termination changes the thread-side `WaitState`, but queue cleanup is still mostly primitive-specific. This means stale `WaiterRef`s can remain in object queues.

The mutex unlock path is one example:

```cpp
const auto waiter = mutex->waiting_threads.front();
mutex->waiting_threads.pop_front();

if (!get_waiting_thread_snapshot(kernel, waiter, WaitType::Mutex, mutex->uid, snapshot)) {
    ...
    return SCE_KERNEL_OK;
}
```

If the first waiter is stale but the second waiter is valid, the mutex becomes free and the valid waiter is left sleeping.

This should continue pruning until either a valid waiter is found or the queue is empty, e.g. conceptually:

```cpp
while (!mutex->waiting_threads.empty()) {
    const auto waiter = mutex->waiting_threads.front();
    mutex->waiting_threads.pop_front();

    WaitingThreadSnapshot snapshot;
    if (!get_waiting_thread_snapshot(
            kernel, waiter, WaitType::Mutex, mutex->uid, snapshot)) {
        continue;
    }

    // transfer ownership + wake
    ...
    break;
}
```

The same invariant should apply to `CondVar::Any`, semaphore fairness checks, RW locks, etc. A stale queue entry should never consume a signal or affect whether another thread is allowed to acquire a primitive.

### 2. Terminating a thread during a CB wait can lose the original wait

During callback processing the active wait is moved out of `thread.wait`:

```cpp
thread.interrupted_wait.wait = thread.wait;
thread.interrupted_wait.sequence = thread.wait_sequence;
thread.interrupted_wait.valid = true;

thread.wait.clear();
thread.wait_sequence = 0;
```

But termination only checks:

```cpp
if (thread.has_wait_state())
    cancel_wait_locked(thread, WaitEndReason::Deleted);
```

So if termination happens while the callback is running, the real wait lives in `interrupted_wait` and is never completed.

Then `run_callback()` does:

```cpp
if (suspended_wait && !exit_requested && !delete_requested)
    restore_wait_after_callback_locked(*this);
else if (suspended_wait)
    interrupted_wait.clear();
```

which can discard the wait completely.

Meanwhile the outer wait loop is still doing:

```cpp
while (thread->wait.end_reason == WaitEndReason::None) {
    ...
    thread->state_cond.wait(...);
}
```

That can leave it waiting on an empty `WaitState`.

Termination/cancel/delete needs to operate on the active wait sequence regardless of whether it currently lives in `wait` or `interrupted_wait`.

Something along these lines would be safer:

```cpp
if (thread.has_wait_state()) {
    cancel_wait_locked(thread, thread.wait_sequence,
        WaitEndReason::Deleted);
} else if (thread.has_interrupted_wait_state()) {
    cancel_wait_locked(thread, thread.interrupted_wait.sequence,
        WaitEndReason::Deleted);
}
```

and the outer callback-capable wait loop should also explicitly observe thread termination.

### 3. VBlank CB waits don't actually use the callback-capable wait machinery

The VBlank wait is created correctly:

```cpp
kernel.thread_manager.begin_wait_locked(
    *wait_thread,
    WaitState::vblank(target_vcount, is_cb));
```

and callback notification sets:

```cpp
thread->callback_wakeup_pending = true;
thread->state_cond.notify_all();
```

But `wait_vblank()` only waits for:

```cpp
wait_thread->state_cond.wait(thread_lock, [&]() {
    return wait_thread->wait.end_reason != WaitEndReason::None;
});
```

So a normal callback notification wakes the condition variable, but the predicate stays false and no callback is processed.

This should use the same callback-aware wait helper as the other CB waits, e.g.:

```cpp
wait_for_wait_end_or_callbacks(
    kernel,
    thread_lock,
    wait_thread);
```

Otherwise `WaitVblank*CB` does not behave like the rest of the new CB-capable wait model.

### 4. Condvar waits must reacquire the mutex before returning on every path

Right now:

```cpp
if (end_reason != WaitEndReason::Signaled) {
    return map_wait_end_reason(...);
}

return mutex_lock_impl(
    kernel, mem, export_name, thread_id, 1,
    condvar->associated_mutex, weight,
    timeout, false, callbacks_allowed);
```

This means timeout/cancel/delete return without reacquiring the associated mutex.

The mutex should be reacquired regardless of how the condition wait ended, and that reacquisition should not reuse the original condvar timeout.

The flow should be closer to:

```cpp
const WaitEndReason end_reason = thread->wait.end_reason;

// finish/remove condvar wait first
...

const int relock_result = mutex_lock_impl(
    kernel,
    mem,
    export_name,
    thread_id,
    1,
    condvar->associated_mutex,
    weight,
    nullptr, // untimed reacquire
    false,
    callbacks_allowed);

if (relock_result < 0)
    return relock_result;

return map_wait_end_reason(
    end_reason,
    canceled_error,
    deleted_error);
```

Vita3K-Plus already had to fix this exact behavior for timed lwcond waits because Fios expects the calling thread to own the mutex again after the wait returns.

### 5. `CoreTiming::now_us()` can observe half of a pause transition

Pause currently does:

```cpp
guest_time_base_us = now_us_locked(current_host_now_us);
paused = true;
```

but readers do not take the mutex:

```cpp
uint64_t CoreTiming::now_us() const {
    return now_us_locked(host_now_us());
}
```

and `now_us_locked()` independently reads:

```cpp
if (paused.load(...))
    return guest_time_base_us.load(...);

const uint64_t base_us = guest_time_base_us.load(...);
const uint64_t resume_us = host_resume_time_us.load(...);

return base_us + (host_now_us - resume_us);
```

A reader can observe:

```text
paused = false
guest_time_base_us = newly updated paused value
host_resume_time_us = old resume timestamp
```

and add elapsed host time twice.

Since this clock drives timeouts and timers, I would make this state coherent instead of relying on three independent atomics.

The simplest safe version is probably just:

```cpp
uint64_t CoreTiming::now_us() const {
    const std::lock_guard<std::mutex> lock(mutex);
    return now_us_locked(host_now_us());
}
```

and then make `now_us_locked()` use the protected non-atomic state.

A sequence/versioned snapshot would also work if avoiding the lock is important.

### 6. MsgPipe vector size handling needs checked arithmetic

Both vector implementations currently do roughly:

```cpp
SceSize total_size = 0;

for (SceUInt32 i = 0; i < vectorCount; ++i) {
    vectors[i] = send_vector_data[i];
    total_size += vectors[i].size;
}
```

`SceSize` is 32-bit, so this can wrap.

For example:

```text
vector[0].size = 0xF0000000
vector[1].size = 0x20000000
```

produces a wrapped `total_size`, so the scratch allocation can be much smaller than the amount later copied.

I would accumulate in a wider type and validate before allocation:

```cpp
uint64_t total_size = 0;

for (...) {
    total_size += vectors[i].size;

    if (total_size > std::numeric_limits<SceSize>::max())
        return RET_ERROR(SCE_KERNEL_ERROR_INVALID_ARGUMENT);
}
```

Then also check the allocation:

```cpp
const Address scratch_addr =
    total_size ? alloc(mem, static_cast<SceSize>(total_size),
                     "msgpipe_send_vector")
               : 0;

if (total_size != 0 && scratch_addr == 0)
    return RET_ERROR(SCE_KERNEL_ERROR_NO_MEMORY);
```

`vectorCount` should also be validated before using it directly to size a host `std::vector`.

### Two additional things I would like before merging

The VBlank catch-up loop can execute every missed tick after a long host stall:

```cpp
for (uint64_t i = 0; i < ticks_to_run; ++i)
    run_vblank_tick(emuenv);
```

Each tick updates input/motion, notifies callbacks, and scans the complete thread list. I would strongly consider capping catch-up work or dropping excessive old ticks rather than producing a large burst after a suspend/stall.

And because this PR replaces the core wait state machine, I think targeted regression tests would be very valuable for:

```text
signal vs timeout
cancel/delete while waiting
stale waiter cleanup
CB interruption
thread termination during a CB wait
condvar timeout + mutex reacquire
pause/resume with armed timeouts
thread-end waits
```

The architecture is promising and I think this PR is worth continuing with. My request for changes is about getting the lifecycle invariants solid before this becomes the base for all kernel waits.

After these are addressed, it should also be rebased onto current `master` and CI rerun.

frangarcj (changes requested):
-

PR #4005, opened 2026-06-07
gui-qt: size key binding buttons to fit the capturing countdown text

troglodyte9:
the bind buttons used a fixed 100px width which clipped the "capturing...(5)" label shown while waiting for input. compute the width from the font metrics of that label plus the qss padding so the text is never truncated

### DIFF
--- a/vita3k/gui-qt/src/controls_dialog.cpp
+++ b/vita3k/gui-qt/src/controls_dialog.cpp
@@ -31,6 +31,7 @@
 #include <QCheckBox>
 #include <QCloseEvent>
 #include <QDialogButtonBox>
+#include <QFontMetrics>
 #include <QFrame>
 #include <QGridLayout>
 #include <QGroupBox>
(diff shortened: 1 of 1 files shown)


vita3k/gui-qt/src/controls_dialog.cpp
```diff
@@ -113,12 +114,19 @@ void style_binding_button(QPushButton *button, bool capturing = false) {
     gui::utils::refresh_theme_state(button);
 }
 
+int bind_button_width() {
```
EXtremeExploit: No need to make this a whole function, it can be inlined

EXtremeExploit (changes requested):

vita3k/gui-qt/src/controls_dialog.cpp
troglodyte9: done

EXtremeExploit (approved):
lgtm

Issue #2175, opened 2022-10-30
Crash on linux


Blad3forc3:
I get this crash and Vita3k closes automatically when scrolling through installed apps on the main screen

terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
Aborted (core dumped)

EXtremeExploit:
make sure you have enough space and/or ram in both the binaries location and pref path

Blad3forc3:
I have 32GB of RAM of which 28GB is free when run and 14TB of space on a
16TB hard drive where the preferences path is located

EXtremeExploit:
can you send logs? I don't think it will be useful but just in case it shows something

Blad3forc3:
The config file in the same folder as Vita3K is completely empty

EXtremeExploit:
of you delete cokfog file it will be regenerated, can you try that?

Blad3forc3:
An update on this, i can scroll up and down in the game list using the up and down arrows but as soon as i use the mouse to scroll it crashes

CoffeeBrewer64:
Are you definitely using xdg-desktop-portal?

scribam:
@Blad3forc3 is it still an issue with the latest build with xdg-desktop-portal installed?

Blad3forc3:
Still a random crash with just this error when run from a terminal

48821 Segmentation fault      (core dumped) ./Vita3K

scribam:
@Blad3forc3 thanks for the feedback.

To get more info, can you use gdb from a terminal?
```
$ gdb ./Vita3K
(gdb) run
<segfault happens here>
(gdb) backtrace
<offending code is shown here>
```

Blad3forc3:
No problem here is the backtrace

Thread 1 "Vita3K" received signal SIGSEGV, Segmentation fault.
0x000055555600b0d2 in gui::draw_app_context_menu(GuiState&, EmuEnvState&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) ()
(gdb) backtrace
#0  0x000055555600b0d2 in gui::draw_app_context_menu(GuiState&, EmuEnvState&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) ()
[...]

PR #4008, opened 2026-06-08
cmake: Dynamic link libc and stdc++

EXtremeExploit:
Brings back the normal way to link with stdc++ and libc
Originally it was static linking to avoid the missing libc version errors
But this only happened on standalone and we didnt have appimage back then

AppImage handles the libc stuff for us now, by having it normally linked again we might avoid some dependency bugs like #2175

So now the standalone will have a new requirement, to use a relatively modern libc version (either latest one or the same CI uses), which its fine as standalone is very rarely used, and if someone wants to use Vita3K the AppImage comes first

Need testing on super outdated systems (Ubuntu 22.04 LTS would be a good test since its the oldest ubuntu version thats still desktop supported)

### DIFF
--- a/vita3k/CMakeLists.txt
+++ b/vita3k/CMakeLists.txt
@@ -217,10 +217,6 @@ if(USE_DISCORD_RICH_PRESENCE)
 	target_link_libraries(vita3k PRIVATE discord-rpc)
 endif()
 
-if(LINUX)
-	target_link_libraries(vita3k PRIVATE -static-libgcc -static-libstdc++)
-endif()
-
 if(NOT ANDROID)
 	file(GLOB VITA3K_QT_TS_FILES CONFIGURE_DEPENDS
 		"${CMAKE_SOURCE_DIR}/i18n/qt/*.ts")


imWatchful (approved):

Macdu:
Is there really a reason to not statically link the standalone version?

EXtremeExploit:
> Is there really a reason to not statically link the standalone version?

Some libraries (possibly qt) might do weird stuff with libc, which can prompt to the issue i mentioned above, 
On AppImage libc will be handled by appimage, which is nice so it can change versions and do all the backwards compat, and any library that does weird stuff to libc can do so without worrying about the statics and all
On standalone the same/newer version will be needed, statically linking here yes solves the issue of the missing version, but at the possible cost of more bugs, plus standalone is more like an after-thought, only used in weird/specific scenarios which already requires Qt 6.11 and other libs, so adding the newer libc version to it is not a big deal and shouldnt be a problem since libc is the first thing to update

TLDR: I dont wanna deal with different libc version incompatibility and even less with bugs that arise with statically linking it now that appimage does it for us, and if we static link we cant let appimage do it for us :C

HylianSoul:
I tested the AppImage from the build 4049 (created from this commit, a few minutes ago) an the result is worse than before. Now, the window with the game does not work when I click play an the application is frozen. The binary from the ubuntu_latest.zip works correctly, as before.

PR #4015, opened 2026-06-11
external: Update SDL from 3.2.28 to 3.4.10

EXtremeExploit:


imWatchful (approved):
oh my peak

Zangetsu38 (changes requested):
android build is broken... Like i have see in my side before, when you update sdl, you need update some java code oàr c++ code code relate sdl func Func can disappear or renamed etc...

PR #4017, opened 2026-06-11
external/tracy: don't stop tracing on first AV or signal

bookmist:
Disable tracy crash handler. AV can be part of normal work of Vita3k


EXtremeExploit (approved):
Looks fine

Macdu (approved):

PR #4018, opened 2026-06-11
emuenv/app: Remove base_path

EXtremeExploit:
base_path is only used by other paths to create them, it is never used by emuenv or roots_paths by the emulator at all
So remove it to avoid confusion, like it already happened before

Plus it SHOULDNT be hardcoded used in the first place :D

### DIFF
--- a/vita3k/app/src/app_init.cpp
+++ b/vita3k/app/src/app_init.cpp
@@ -246,8 +246,6 @@ void init_paths(Root &root_paths) {
     fs::path internal_storage_path = fs::path(SDL_GetAndroidExternalStoragePath()) / "";
     fs::path vita_storage_path = internal_storage_path / "vita/";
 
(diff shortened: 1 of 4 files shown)


imWatchful (dismissed):

vita3k/app/src/app_init.cpp
```diff
@@ -426,14 +422,11 @@ bool init(EmuEnvState &state, Config &cfg, const Root &root_paths) {
 
     set_current_config(state, state.cfg.run_app_path.has_value() ? *state.cfg.run_app_path : "");
 
-    LOG_INFO("Base path: {}", state.base_path);
-#if defined(__linux__) && !defined(__ANDROID__)
```
Zangetsu38, line 430: no remove this
I put this in because on Windows they were all the same, it was mainly for Linux which is weird lol

imWatchful (approved):

PR #4024, opened 2026-06-15
Harden HTTP URL and response parsing & More

1YBJ1:
- Reworked HTTP URL parsing to avoid out-of-bounds reads on malformed or short URLs.
- Fixed parsing for URLs with credentials, query strings, fragments, and missing paths.
- Replaced throwing numeric conversions in HTTP status/content-length parsing with checked parsing.
- Fixed a Content-Length lookup bug that compared against the wrong header map.

### DIFF
--- a/vita3k/modules/SceHttp/SceHttp.cpp
+++ b/vita3k/modules/SceHttp/SceHttp.cpp
@@ -17,6 +17,7 @@
 
 #include <module/module.h>
 
+#include <algorithm>
 #include <cstring>
 #include <filesystem>
 #include <http/state.h>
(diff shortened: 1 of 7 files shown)


vita3k/modules/SceHttp/SceHttp.cpp
```diff
@@ -1464,7 +1474,7 @@ EXPORT(SceInt, sceHttpUriParse, SceHttpUriElement *out, const char *srcUrl, Ptr<
     if (require) {
         *require = internal_require;
     }
-    if (prepare < internal_require)
+    if (pool && out && prepare < internal_require)
```
EXtremeExploit, line 1477: Why?

EXtremeExploit (commented):
I think everything regarding http/net is fine, just one small comment
I dont know much about the rest so i will let someone else handle that part

EXtremeExploit:
Stale

PR #4025, opened 2026-06-17
external: Update fmt from 12.1.0 to 12.2.0

EXtremeExploit:
I think it fixes the issue of windows arm not compiling because of fmt

But the issue with macos remains because the ci itself is wrong now apparently


Dicot0721 (approved):

imWatchful (approved):
only had to rerun ci 15 times

EXtremeExploit:
WHT DO YOU MEAN???

PR #4027, opened 2026-06-20
Fix typeless copies for higher than 1.75x resolutions

nckstwrt:
In games like Uncharted: Golden Abyss, that use typeless copies in retrieve_color_surface_as_texture, if you used a resolution higher than 1.75x (specifically 1.975x was the turning point) then you would get severe colour corruption. This fixes that by ensuring the right data is sampled at any resolution.


Macdu:
Hello, thank you for the PR, I have quite a few issues with it so far unfortunately:
- This code has way too many assumptions that only apply to uncharted. For example it will break if we do typeless copies with something different than a 64-bit to 32-bit format copy (not that I'm aware of any game doing this, to be honest).
- A similar (bigger issue) is that all PS Vita games using gbuffering do it differently. I know multiple games which use the exact same texture for sampling the color/normals but instead just use a x-offset in the shader to decide whether to sample the color or the normals, these games would be broken by this PR.
- Your implementation uses a texture -> buffer -> compute shader -> buffer -> texture pass. If we are using a compute shader, we might as well just operate directly on storage images in the compute shader and just do texture -> compute shader -> texture, it would remove the two intermediate buffers.
- This shouldn't be applied if the resolution multiplier is 1.

Also the texture do not get broken at 1.975x, it just depends if parameters align properly or not (for example uncharted looks amazing at 5x)

nckstwrt:
Thanks for the review Macdu. Yes it really was just to solve the Uncharted problem, I had hoped it would be the same with other games. Fair point about the workflow, could just be reduced to a compute shader without the buffers as you say, might take a crack at that as it could be interesting for me to attempt.
I mentioned the 1.975x point just as it was useful for my own testing, but yes you are right about the sampling of course.

nckstwrt:
New version - tried to address all the points in one form or other:

1) Tried to dropped the Uncharted specific assumptions as best I could. The store is read through an R32G32_UINT view, so it works for any 64-bit to 32-bit (ratio 2) typeless cast regardless of the surface's actual format, instead of assuming RGBA16F.
2) Other gbuffer layouts (including the same-texture + x-offset games) are left alone. The path is guarded to that exact pattern, and everything else falls back to the existing byte reinterpret code unchanged. If you have a good example of a game with a different layout that might be worth looking into I could try and take a look to see what else I could include.
3) Removed both the intermediate buffers like you suggested so it's texture to compute to texture
4) And it will now be skipped entirely when the resolution multiplier is 1

Hopefully that's a bit more workable.

frangarcj (changes requested):
Thanks, I went through the current PR again and compared it with the newer implementation in your `all-enhancements` branch.

From Macdu's original review, I think points **1, 3 and 4 are already addressed** in the current PR:

* The compute path is now narrowly guarded to the supported 64-bit → 32-bit case instead of being applied to arbitrary typeless copies.
* The unnecessary texture → buffer → compute → buffer → texture round-trip is gone for the special path; the reinterpret is done image-to-image.
* The workaround is disabled at native resolution through:

```cpp
const bool use_compute_deinterleave =
    state.res_multiplier != 1.0f
    && bytes_per_pixel_in_store == 8
    && bytes_per_pixel_requested == 4
    && guard_ratio == 2
    && guard_native_store_col == 0
    && start_sourced_line == 0
    && (guard_sub_texel_byte % bytes_per_pixel_requested) == 0;
```

The remaining issue in the PR is Macdu's point about games that use the **same reinterpreted texture and select the two 32-bit words with a shader x/UV offset**.

The current PR still chooses one fixed half for the whole output:

```glsl
const uint word =
    (pc.half_index == 0u) ? words.x : words.y;
```

so it can't preserve a layout like:

```text
[A0 B0] [A1 B1] [A2 B2]
    ↓ 32-bit reinterpret
A0 B0 A1 B1 A2 B2
```

when the game expects both halves to remain in a single view.

However, I checked `all-enhancements`, and you already have a solution for this there:

```glsl
const uint word_idx =
    (pc.interleave != 0u)
        ? (x % pc.ratio)
        : pc.half_index;
```

together with the `has_phase_view` logic on the caller side.

That gives you both layouts:

```text
interleave = 1
A0 B0 A1 B1 A2 B2 ...
```

for games using one view + shader/UV offset, and:

```text
interleave = 0
A0 A0 A1 A1 ...
```

or the high-word equivalent for games using separate byte-offset views.

That looks like the missing generalization from the original review.

Could you bring that newer typeless-cast / `interleave` logic from `all-enhancements` into this PR and then rebase it onto current `master`?

After that I'd consider the original review points addressed and this should be ready for another full test/CI pass.

PR #4031, opened 2026-06-22
Add `USE_SYSTEM_FFMPEG` CMake option

j8r:
Add a new CMake option to use the system FFMPEG instead of the bundled one. This is necessary for Flatpaks (https://github.com/Vita3K/Vita3K/issues/1418).
Downloading pre-built artifacts can only be done through [module sources ](https://docs.flatpak.org/en/latest/module-sources.html), but it is preferable to build from source.

Inspired by RPCS3 https://github.com/RPCS3/rpcs3/blob/master/buildfiles/cmake/FindFFMPEG.cmake

I am by no mean an proficient in CMake, opened any improvements.

Vita3K compiled before the switch to Qt, which I'm no longer able to compile fully, likelly due to Qt version mismatch.


external/CMakeLists.txt
```diff
@@ -267,7 +267,15 @@ target_include_directories(vma INTERFACE
 if(ANDROID)
 	set(USE_STATIC_ANDROID_FFMPEG ON)
 endif()
-add_subdirectory(ffmpeg)
+if(USE_SYSTEM_FFMPEG)
+	add_library(ffmpeg INTERFACE)
+	message(STATUS "Using shared FFMPEG")
```
EXtremeExploit: Not a complete review, i dont think im the right one to review this as i dont know much about how ffmpeg is used in the project and all the modules and that jazz
```suggestion
	message(STATUS "Using system FFMPEG")
```

Macdu:
Not a good idea, Vita3K depends on some internal ffmpeg's headers, meaning even minor ffmpeg updates can break it

j8r:
@Macdu For Flatpak, the Sdk usually includes a stable FFMPEG, so could be fine? Updating the Sdk can of course break the build but then we can have the time tp report the error and fix it.

If not possible, it has to be compiled when building the Flatpak, because no internet connection is available during the build.

EXtremeExploit:
Is it not possible the use the prebuilt ffmpeg?

j8r:
Not possible. That's by design, it allows Flatpaks to be reproducible.
What I see is either we compile FFMPEG Vita3K fork beforehand, or use the already available FFMPEG in the Sdk.
Both ways the CMake built system has to be modified.

Dicot0721:
> If not possible, it has to be compiled when building the Flatpak, because no internet connection is available during the build.

What does "build" mean here? A CI workflow running on Flatpak's server? With so many SDKs that need to be downloaded online, they actually provide an offline environment?

j8r:
When building a Flatpak with the [Flatpak builder](https://docs.flatpak.org/en/latest/flatpak-builder.html), the commands are run in an offline sandbox.
Building can be done anywhere, locally or CI.

Dicot0721:
> When building a Flatpak with the [Flatpak builder](https://docs.flatpak.org/en/latest/flatpak-builder.html), the commands are run in an offline sandbox. Building can be done anywhere, locally or CI.

The module sources in the manifest file can be used to download prebuilt FFmpeg, right?
And you're right, the CMake script needs to be modified.

j8r:
Yes you can through the [module sources](https://docs.flatpak.org/en/latest/module-sources.html).
Cannot be through scripts like CMake or shell.

Dicot0721:
So you might need to change the title of this issue

j8r:
Good to me, feel free to edit this PR title with whatever fits the best.
I'll add to the description why it cannot be done another way.
Generally for Flatpak it is preferable to compiple from the source instead of downloading a pre-built artifact.

PR #4032, opened 2026-06-23
Add Flatpak equivalent of `/usr/share` to asset paths for potential future Flatpak builds

fpiesche:
This will support #1418. Using the regular cmake build+install process in a Flatpak bundle will get all the static assets deployed to the Flatpak sandbox's `/app/share/Vita3K` directory so a Flatpak-built vita3k currently cannot find e.g. the builtin shader files.

### DIFF
--- a/vita3k/app/src/app_init.cpp
+++ b/vita3k/app/src/app_init.cpp
@@ -358,6 +358,7 @@ void init_paths(Root &root_paths) {
         const constexpr char *static_asset_paths[] = {
             "/usr/local/share/Vita3K",
             "/usr/share/Vita3K",
+            "/app/share/Vita3K",
         };
 
         // Check both normal case and all lowercase paths


EXtremeExploit:
Does flatpak patch libc so that even when the program reads root it reads the sandbox instead?
If it doesnt do that you can do something similar as the appimage which knows if its in one, and takes the appimage path instead

fpiesche:
> Does flatpak patch libc so that even when the program reads root it reads the sandbox instead?

It does not. Flatpak assumes that applications use paths relative to their install prefix rather than hardcoded absolute paths (so eg. if the `vita3k` executable is in `~/.local/bin/vita3k` it should check in `~/.local/share/` rather than `/share` or `/usr/share`).

This is in fact how the CMake install behaviour via the `GNUInstallDirs` module works as well (and is supposed to) - `cmake --install` after a build will deploy the `vita3k` executable to `$CMAKE_INSTALL_PREFIX/bin`, any app data to `$CMAKE_INSTALL_PREFIX/share/Vita3K`, any libraries to `$CMAKE_INSTALL_PREFIX/lib/`, etc, to match whatever the user requested (with the default being usually `/usr/`).

It would be feasible to check for the `FLATPAK_ID` environment variable and reroute to `/app/share` instead but the "proper" behaviour would be to use a path relative to the vita3k executable's location - I'm not really enough of a C++ person to bludgeon that into shape, though...

EXtremeExploit:
Then i dont think the current change really changes anything since it will read /app/share/Vita3K outside the sandbox which doesnt really exist unless im not understanding it right
You can also just make it read from CMAKE_INSTALL_PREFIX, all you need to do is in the vita3k cmake to add a C define to whatever the value is and use that in the cpp file

fpiesche:
> Then i dont think the current change really changes anything since it will read /app/share/Vita3K outside the sandbox which doesnt really exist unless im not understanding it right

Flatpak is extremely heavily sandboxed by default. With this change, a regular Vita3k build would indeed attempt to check `/app/share/Vita3K` when run outside the sandbox (as well as the existing options), but with a Flatpak package the executable is run within the sandbox. The filesystem seen by applications in the sandbox is read-only space defined by the runtime; only `/app/` can be written to at build time (and is read-only at runtime).

At runtime, only `$XDG_[DATA|CONFIG|CACHE]_HOME` are writeable and persistent by default. Writes to anywhere in `$HOME` outside of the `$XDG_*_HOME` paths work, but do not persist between sessions.

Even read access to anything outside of the sandbox requires explicit permissions set on each Flatpak app, and write access to `$HOME` will be discarded when the application exits unless explicit persistence is requested in the permissions (at which point they will be persisted on the host in `~/.var/app/[appid]/` rather than the actual home directory).

> You can also just make it read from CMAKE_INSTALL_PREFIX, all you need to do is in the vita3k cmake to add a C define to whatever the value is and use that in the cpp file

That, however, is an excellent idea! Will give that a go later.

EXtremeExploit (approved):
lgtm

PR #4035, opened 2026-06-26
build: Increase macOS deployment target to 13.3

Dicot0721:
- Increase [macOS deployment target](https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_DEPLOYMENT_TARGET.html) (minimum required version) from 11.0 to 13.3.
- Due to [weak linking](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/WeakLinking.html), `std::to_chars` for floating-point types, used in `std::format("{:x}", /* floating-point */)` in Vulkan-Headers, requires a minimum version of 13.3 (just like [this comment](https://github.com/Vita3K/Vita3K/pull/3655#issuecomment-3234315053)); #3661 bypassed this requirement by setting the Xcode app version to 16.2, but as [GitHub updated the runner](https://github.com/actions/runner-images/issues/14167), Xcode 16.2 is no longer available on `macos-latest`.

### DIFF
--- a/.ci/setup-macos.sh
+++ b/.ci/setup-macos.sh
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-sudo xcode-select -s /Applications/Xcode_16.2.app/Contents/Developer
 brew install ccache create-dmg
 echo "$(brew --prefix ccache)/libexec" >> "$GITHUB_PATH"
 ccache --set-config=compiler_check=content
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,7 @@ include("external/dynarmic/CMakeModules/DetectArchitecture.cmake")
 set(CMAKE_CXX_STANDARD 23)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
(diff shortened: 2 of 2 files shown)


Dicot0721:
Another solution to this issue is to set the [macOS runner label](https://docs.github.com/en/actions/reference/runners/github-hosted-runners) to `macos-15` or `macos-15-intel` and keep Xcode at version 16.2, but this is probably not a long-term solution; sooner or later, the deployment target will need to be upgraded.

Refs:
- [C++ Language Support](https://developer.apple.com/xcode/cpp/#iso-c++standard-supported)

EXtremeExploit (approved):

Dicot0721:
In [this workflow run](https://github.com/Dicot0721/Vita3K/actions/runs/28287271452), when Xcode 16.2 was no longer pinned, the build succeeded on `macos-26` but failed on `macos-15`.

Perhaps this PR requires further consideration.

Dicot0721:
Well, macOS 13.3 required for `std::to_chars` is already among the highest requirements for almost all C++ features (the highest is macOS 14.0 required for `<memory_resource>`).

Updating from macOS 11.0 to 13.3 may be a good deal.

PR #4041, opened 2026-06-28
vita3k: add vci support

### LINKED ISSUE #3518: [Feature Request]: Add support for VCI files.

Implement VCI (Vita Cartridge Image) files. There's a Python code available [here](https://silica.codes/Li/VCI-TOOLS/src/branch/main/gc2nonpdrm.py), which extracts NoNpDrm from VCI.

Mote info is available [here](https://silica.codes/Li/GcToolKit).

VCI are 1:1 cartridge backups, instead of keeping games with fake licenses, these could be kept instead.

None.

### Platform (if relevant)

None

nishinji:
Support has been added for the VCI format, which is suitable for storing Vita game cards and decrypting them.
nonpdrm is not a backup of a game card (a nonpdrm dump cannot be used to accurately restore a game card).
For now, limited to gro0.
I wrote the code from scratch up to the creation of `klicense` and am reusing `decrypt_install_nonpdrm`.
Closes: #3518

### DIFF
--- a/vita3k/gui-qt/src/archive_install_dialog.cpp
+++ b/vita3k/gui-qt/src/archive_install_dialog.cpp
@@ -47,7 +47,7 @@ std::vector<fs::path> archives_in_dir(const fs::path &dir) {
         if (!entry.is_regular_file())
             continue;
         const auto ext = string_utils::tolower(entry.path().extension().string());
-        if (ext == ".zip" || ext == ".vpk")
+        if (ext == ".zip" || ext == ".vpk" || ext == ".vci")
             result.push_back(entry.path());
     }
     return result;
@@ -201,7 +201,7 @@ std::vector<fs::path> ArchiveInstallDialog::pick_archives() {
     auto *layout = new QVBoxLayout(&type_picker);
     layout->addWidget(new QLabel(tr("What would you like to install?"), &type_picker));
 
-    auto *btn_file = new QPushButton(tr("Select a file (.zip / .vpk)..."), &type_picker);
+    auto *btn_file = new QPushButton(tr("Select a file (.zip / .vpk / .vci)..."), &type_picker);
     auto *btn_dir = new QPushButton(tr("Select a directory (installs all archives inside)..."), &type_picker);
     auto *btn_cancel = new QPushButton(tr("Cancel"), &type_picker);
     layout->addWidget(btn_file);
@@ -231,7 +231,7 @@ std::vector<fs::path> ArchiveInstallDialog::pick_archives() {
             this,
             tr("Select Archive"),
             QString(),
-            tr("PlayStation Vita commercial software package (NoNpDrm/FAGDec) / PlayStation Vita homebrew software package (*.zip *.vpk);;PlayStation Vita commercial software package (NoNpDrm/FAGDec) (*.zip);;PlayStation Vita homebrew software package (*.vpk)"));
+            tr("PlayStation Vita commercial software package (NoNpDrm/FAGDec) / PlayStation Vita homebrew software package (*.zip *.vpk *.vci);;PlayStation Vita commercial software package (NoNpDrm/FAGDec) (*.zip);;PlayStation Vita homebrew software package (*.vpk);;Vita Cartridge Image (VCI) File (*.vci)"));
         if (path.isEmpty())
             return {};
         return { gui::utils::to_fs_path(path) };
@@ -246,7 +246,7 @@ std::vector<fs::path> ArchiveInstallDialog::pick_archives() {
     auto archives = archives_in_dir(gui::utils::to_fs_path(dir));
     if (archives.empty()) {
         QMessageBox::information(this, tr("No Archives Found"),
-            tr("The selected directory contains no .zip or .vpk files."));
+            tr("The selected directory contains no .zip, .vpk, or .vci files."));
         return {};
     }
     return archives;
--- a/vita3k/packages/src/exfat.cpp
+++ b/vita3k/packages/src/exfat.cpp
@@ -54,8 +54,9 @@ static uint64_t get_cluster_offset(const ExFATSuperBlock &super_block, uint32_t 
     const uint64_t sectors_per_cluster = 1ULL << super_block.spc_bits;
     const uint64_t cluster_size = sector_size * sectors_per_cluster;
 
-    // The cluster number is 0-based, so we need to add 1
-    return (static_cast<uint64_t>(cluster) + 1) * cluster_size;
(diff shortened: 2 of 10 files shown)


vita3k/gui-qt/src/archive_install_dialog.cpp
```diff
@@ -231,7 +231,7 @@ std::vector<fs::path> ArchiveInstallDialog::pick_archives() {
             this,
             tr("Select Archive"),
             QString(),
-            tr("PlayStation Vita commercial software package (NoNpDrm/FAGDec) / PlayStation Vita homebrew software package (*.zip *.vpk);;PlayStation Vita commercial software package (NoNpDrm/FAGDec) (*.zip);;PlayStation Vita homebrew software package (*.vpk)"));
+            tr("PlayStation Vita commercial software package (NoNpDrm/FAGDec) / PlayStation Vita homebrew software package (*.zip *.vpk *.vci);;PlayStation Vita commercial software package (NoNpDrm/FAGDec) (*.zip);;PlayStation Vita homebrew software package (*.vpk);;Vita Cartridge Image (VCI) File (*.vci)"));
```
EXtremeExploit, line 234: If the translations change then so they should also in  `i18n/qt/vita3k_en.ts`

vita3k/packages/src/exfat.cpp
```diff
@@ -54,8 +54,9 @@ static uint64_t get_cluster_offset(const ExFATSuperBlock &super_block, uint32_t
     const uint64_t sectors_per_cluster = 1ULL << super_block.spc_bits;
     const uint64_t cluster_size = sector_size * sectors_per_cluster;
 
-    // The cluster number is 0-based, so we need to add 1
-    return (static_cast<uint64_t>(cluster) + 1) * cluster_size;
+    // exFAT data clusters are numbered from 2, and the cluster heap begins
+    // cluster_sector_start sectors into the volume.
+    return static_cast<uint64_t>(super_block.cluster_sector_start) * sector_size + (static_cast<uint64_t>(cluster) - 2) * cluster_size;
```
EXtremeExploit, line 59: I assume this change only affects VCI images and not anything else right? was this tested to work with other dumps?

EXtremeExploit (changes requested):

vita3k/packages/src/exfat.cpp
nishinji: Exfat.cpp is written for the preinst firmware. The firmware installution works properly.

Zangetsu38: Don't touching it, is already fine

nishinji: The `get_cluster_offset` function in the extractor was the decisive problem.

```cpp
return (cluster + 1) * cluster_size;
```

This formula completely ignores exFAT's cluster heap start position (ClusterHeapOffset) and assumes the data region begins at offset 0. It only holds for Sony's PUP/firmware exFAT images, which have a fixed geometry where the heap starts exactly at the 3rd cluster. A real game card's gro0 (on the order of 1.8 GiB) has a different geometry, so the root directory's cluster position is off — it reads garbage, the entry types don't match, and it terminates immediately, producing no output.

Let me check whether `ExFATSuperBlock` has a field equivalent to ClusterHeapOffset. If it does, we can fix the formula correctly.

`ExFATSuperBlock` has everything needed for the correct calculation: `cluster_sector_start` (0x58, the cluster heap start sector) as well as `sector_bits` / `spc_bits`. So we can fix `get_cluster_offset` to use the correct exFAT formula.

That said, "zero output" can also come from a wrong carve position, so let me first confirm whether the offset my `find_gro0` returns is in bytes or in blocks (if it's blocks, the carve itself would be off).

`find_gro0` returns bytes via `* SECTOR_SIZE`, and the carve position (0x200 + 0x8000 = 0x8200) is correct too. So the cause is pinned down to a single point — the `get_cluster_offset` formula in `exfat.cpp`.

## Fix: `vita3k/packages/src/exfat.cpp`

Correct the cluster-number-to-byte-offset calculation to the proper exFAT formula.

Before:

```cpp
static uint64_t get_cluster_offset(const ExFATSuperBlock &super_block, uint32_t cluster) {
    const uint64_t sector_size = 1ULL << super_block.sector_bits;
    const uint64_t sectors_per_cluster = 1ULL << super_block.spc_bits;
    const uint64_t cluster_size = sector_size * sectors_per_cluster;

    // The cluster number is 0-based, so we need to add 1
    return (static_cast<uint64_t>(cluster) + 1) * cluster_size;
}
```

After:

```cpp
static uint64_t get_cluster_offset(const ExFATSuperBlock &super_block, uint32_t cluster) {
    const uint64_t sector_size = 1ULL << super_block.sector_bits;
    const uint64_t sectors_per_cluster = 1ULL << super_block.spc_bits;
    const uint64_t cluster_size = sector_size * sectors_per_cluster;

    // exFAT data clusters are numbered from 2, and the cluster heap begins
    // cluster_sector_start sectors into the volume. The old (cluster + 1)
    // form only held when the heap started exactly 3 clusters in (true for
    // Sony's PUP exFAT images, but not for real gamecard exFAT volumes).
    return static_cast<uint64_t>(super_block.cluster_sector_start) * sector_size
        + (static_cast<uint64_t>(cluster) - 2) * cluster_size;
}
```

## Why this fixes it — and why it doesn't break anything existing

The old formula `(c+1)·cs` and the correct formula `heap + (c−2)·cs` are equal only when `heap == 3·cs` (i.e., the cluster heap starts exactly at the 3rd cluster). PUP/firmware exFAT images are created by Sony's tools with this geometry, which is why the old formula worked. A real game card's gro0 has a different `cluster_sector_start`, so the root directory's cluster position was off, garbage was read, entry types didn't match, and it terminated immediately — hence zero output.

The new formula handles both cases correctly. Moreover, since the existing PUP extraction currently works, those images must have `heap == 3·cs`, so the new formula returns exactly the same offsets for PUP (mathematically, two linear functions with slope `cs` can only agree across multiple clusters if their intercepts match → `heap == 3·cs`). Existing paths such as firmware installation are therefore byte-for-byte unchanged.

Zangetsu38: I see 🤔

EXtremeExploit (approved):
I dont really have a VCI file to test, but i assume you do and it works fine

nishinji:
Of course, I've already tested VCI on several games and confirmed that it installs successfully.
To be honest, I'd like someone else to test it too.

EXtremeExploit:
it seems to work fine on my end

PR #4044, opened 2026-07-01
overlay: Add a menu to pause overlay

### LINKED ISSUE #1916: [Request] Add shut down hotkey
Hotkeys are a nice way to enhance the user control over emulators, Dolphin, PCSX2-QT, Duckstation and others offer the ability to shut down the virtual console/emulator from a user defined key/gamepad button. This is useful for couch gaming and users which launch games from a frontend like EmulationStation-DE and others.

image

### LINKED ISSUE #3813: [Feature Request]: Exiting from emulator using game controller

Option to quit emulator clicking BACK or B button on controller, or a controller clickable button to exit 

For e.g linux htpc, other linux gamebox with only controller connected

Moonlight can exit clicking B on main menu, PPSSPP have exit button

### Platform (if relevant)

None

EXtremeExploit:
Replaces the "Emulation Paused" text display with an actual (somewhat barebones looking) menu that can actually exit the game using the controller/keyboard/mouse

image

I wanted to display more stuff like the game icon and other stuff but not a lot is available from the overlay at the moment and i dont want to change everything just for tiny looks

probables TODOs: 
- [x] Fix Vita3K crashing when exiting any game with the overlay
- [x] Make it work while another overlay is already present (common dialog)
- [x] ~~Not hardcode the font~~

Fixes: #3813
Fixes: #1916


nishinji:
I'm posting this review of Claude because I'd like it to be merged as soon as possible. It is not 100% accurate.

**1. The pause menu can trigger Exit the moment it opens** — `pause_overlay.cpp:168`

`run_input_loop` seeds the button state as already-pressed (`last_button_state.fill(true)`) but does not do the same for touch (`last_touch_pressed = false`). `overlay_mouse.pressed` stays `true` for the whole Press→Release interval, so if you pause while holding the left mouse button, the very first poll fires `on_touch_pressed` at the last known coordinates. Exit's hit rect is dead centre of the screen (x 340–620 / y 278–324), so the session is killed with no confirmation.

**2. Exit does nothing on Android** — `session_controller.cpp:175`

`stop_requested()` is read from exactly one place, `main_window.cpp:1319`. Android's SDL loop never checks it, so Exit just resumes the game instead of closing it. The existing `post_app_launch_request({.reason = ProcessExit})` channel is consumed by both Qt and Android — using it would remove the need for this mechanism entirely.

**3. The menu is unusable when the pause came from Menu/Background** — `pause_overlay.cpp:152`

The overlay is shown whenever any pause bit is set (the reasons are OR'd), but Resume/Exit only clear `User`. With Android's `PAUSE_REASON_MENU` / `PAUSE_REASON_BACKGROUND`, `set_pause_reason` early-returns without doing anything, leaving the card on screen with no way out.

**4. The stop flag persists and closes the next game right after it launches** — `session_controller.cpp:197`

The flag is cleared in exactly one line inside `stop()`, and that line sits after the `phase == Idle || Stopping` early return. Neither `begin_launch()` nor `reset_session_tracking()` touches it — every other piece of session state is reset there, this one alone is missed. Once it gets stuck set, the next launch closes the game on the first 4 ms timer tick.

**5. The app title `string_view` is left dangling** — `session_controller.cpp:231`

The lock is released as the function returns and only the view escapes. `reset_app_state` / `abort_game_launch` call `clear()` from outside `stop()`'s lock. Close the game while paused and the main thread destroys the buffer while the render thread is walking it in `set_text`. Returning `std::string` by value fixes it.

EXtremeExploit:
> Exit does nothing on Android

Any idea on where the check should be in android? i actually dont know :sob: 

Other than that i addressed all the other suggestions/bugs

EXtremeExploit:
I tried to guess where the android check should be and what it should do, i cant really test if it works

PR #4065, opened 2026-07-26
ci: Fix caching

nishinji:
I'm not entirely sure, but this change seems to have made the cache work.

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -99,9 +99,9 @@ jobs:
         shell: pwsh
         env:
           VCPKG_ROOT: ${{ matrix.vcpkg_root }}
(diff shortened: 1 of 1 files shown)


EXtremeExploit:
are you sure?

EXtremeExploit (approved):

Dicot0721:
We used to cache packages instead of the default binary cache, but after migrating from classic mode to manifest mode (`vcpkg.json`), [the package installed directory](https://learn.microsoft.com/en-us/vcpkg/commands/common-options#install-root) changed, which may be the reason the cache is no longer valid.

If I'm not misunderstanding [the documentation](https://learn.microsoft.com/en-us/vcpkg/consume/binary-caching-default), this PR should work; otherwise, we may need to cache the directory `${{ github.workspace }}/vcpkg_installed/` to cache the packages.

PR #4063, opened 2026-07-23
external: Update dependencies

scribam:

PR #4077, opened 2026-08-03
vita3k/CMakeLists: Fix openssl ver

nishinji:
The build is terminated due to an error caused by a version mismatch.
Mistake from https://github.com/Vita3K/Vita3K/pull/4063.


scribam (approved):
Thanks!

PR #4089, opened 2026-08-11
renderer/gl: add workaroud for amd igpu

nishinji:
I went on vacation and tried to do some OpenGL development at my parents' house, but the screen was black for every game...
After debugging for a few hours with the help of Claude and RenderDoc, I discovered—as usual—that the problem was caused by some strange behavior in the AMD OpenGL driver. While this isn't a fundamental fix, disabling mask_bit emulation prevents the black screen from appearing.

I'll leave this as an open PR. I'd like to see if anyone else is experiencing the same problem so I can confirm that it's been fixed.

| Master | PR |
| -- | -- |
| image |  image |

### DIFF
--- a/vita3k/renderer/src/gl/renderer.cpp
+++ b/vita3k/renderer/src/gl/renderer.cpp
@@ -238,6 +238,24 @@ bool create(std::unique_ptr<State> &state, const Config &config) {
     gl_state.features.use_mask_bit = false;
 #else
     gl_state.features.use_mask_bit = true;
+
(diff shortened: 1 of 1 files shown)


Dicot0721:
I have an AMD integrated GPU, but I haven't encountered this issue. And I tested the build from your PR, it works. Maybe your GPU driver needs an update?

nishinji:
why I didn't think of it?
image

Dicot0721:
idk 🤷‍♂️
image
This is mine

Issue #4069, opened 2026-07-27
[BUG]: Crash (unhandled memory access) on game exit during audio backend reset — occurs across multiple games, not game-specific


lyymo:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Affected Build(s)

v0.2.1 4067

### Description of Issue

Device info

Device: Retroid Pocket 5 (RP5)
SoC: Snapdragon 865 (Adreno 650)
Android version: Android 13
Platform: Android
Vita3K version: v0.2.1, continuous build 4067 (downloaded from https://github.com/Vita3K/Vita3K/releases/tag/continuous)

Description
When exiting a game (returning to the main menu / closing the running title), the emulator crashes during the shutdown/cleanup sequence. This is not tied to a specific game — I've observed this crash on exit across several different titles, which suggests the issue lies in the emulator's shared shutdown/cleanup path rather than any single game's code.

Log excerpt (from the crash)

[16:07:57.381] |I| [SDL_main]: Shutting down game
[16:07:57.381] |D| [set_phase]: App session phase: Running -> Stopping
[16:07:57.398] |I| [save_pipeline_cache]: Saving pipeline cache...
[16:07:57.399] |I| [save_pipeline_cache]: Pipeline cache saved
[16:07:57.399] |I| [adhoc_thread]: Adhoc thread stopped
[16:07:57.400] |T| [close_file]: np_deinit: Closing file fd: 0x8
[16:07:57.519] |I| [set_async_compilation]: Asynchronous pipeline compilation is now disabled
[16:07:57.563] |I| [export_library_init_taihen]: taiHEN: HLE module initialized
[16:07:57.563] |D| [set_phase]: App session phase: Stopping -> Idle
[16:07:57.563] |D| [reset_android_session_audio]: Resetting Android audio backend 'SDL' after SDL session shutdown
[16:07:57.630] |C| [signal_handler]: Unhandled access to 0x201400

Analysis
The crash consistently occurs immediately after reset_android_session_audio logs "Resetting Android audio backend 'SDL' after SDL session shutdown". The very next line is an unhandled memory access at address 0x201400, which looks like an invalid/dangling pointer being dereferenced during (or right after) the SDL audio backend teardown on Android, rather than a game-specific asset or logic issue.

Steps to reproduce

Launch any game on this build (v0.2.1, continuous build 4067) on Android.
Play for a while.
Exit the game (return to main menu / close the running title).
[...]

PR #4090, opened 2026-08-11
Fixed crash on android when exiting a game 

Ayushkr3:
#4069 fix to crash that occurs due to double free and thread race. Fixed by reordering destructors, might be fixed with #4003 but this works fine for now and is relatively cheap and quick fix.

### DIFF
--- a/vita3k/kernel/include/kernel/thread/thread_state.h
+++ b/vita3k/kernel/include/kernel/thread/thread_state.h
@@ -64,6 +64,8 @@ struct ThreadState {
     SceUID id;
     Address entry_point;
 
+
+    CPUStatePtr cpu;
     Block stack;
     int stack_size;
     Block tls;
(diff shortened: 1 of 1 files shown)


frangarcj (changes requested):
The member reordering looks useful as a mitigation, but I don't think it proves the lifetime race is actually fixed.

II think the remaining race may be around `ThreadState` lifetime rather than just member destruction order.

Could you check that? The interesting order is:

```text
threads.erase()
process_exit() returns
ThreadState::~ThreadState()
Block::~Block()
deinit_mem()
```

The reorder may reduce the window, but if destruction still happens after `process_exit()` returns, the race is still there.

Ayushkr3:
yeah the race will still exist, i purposed this because i thought that #4003 will fix it and this was easy to review and quick fix. the issue is `CPUStatePtr cpu ` takes too much time to be destroyed inside thread_state.h. for more concrete solution how about we add something like `thread->cpu.reset()` or even better `thread=nullptr` inside `thread_function` in kernel.cpp in thread deleting stack. This will ensure that `~ThreadState` is called synchronously. what do you think

Issue #4093, opened 2026-08-13
[BUG]: sceJpegEncoderCsc. Pictures are not encoded properly


HylianSoul:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Affected Build(s)

v0.2.1 4074-496939b6

### Description of Issue

I indicated in the [compatibility repo](https://github.com/Vita3K/compatibility/issues/208) a bug in *Uncharted: Golden Abyss*. The game generates green jpg files every time you take a photo in the game. I write here because I compiled the source code I have tested some things, and I have found interesting information. I do not know where I must write about technical stuff. 

I modified this function (in ```vita3k/modules/SceJpegEnc/SceJpegEncUser.cpp```) to have visual information while I am testing.

```c++
EXPORT(int, sceJpegEncoderCsc, SceJpegEncoderContext *context, Ptr<uint8_t> outBuffer, Ptr<uint8_t> inBuffer, int32_t inPitch, int32_t inPixelFormat) {
    std::cout << "################################################" << std::endl;
    std::cout << "-- sceJpegEncoderCsc executed" << std::endl;
    
    TRACY_FUNC(sceJpegEncoderCsc, context, outBuffer, inBuffer, inPitch, inPixelFormat);
    auto inBufferData = inBuffer.get(emuenv.mem);
    std::cout << "inBufferData = " << static_cast<const void *>(inBufferData) << std::endl;
    auto outBufferData = outBuffer.get(emuenv.mem);
    std::cout << "outBufferData = " << static_cast<const void *>(outBufferData) << std::endl;

    std::cout << std::hex;
    for (int i = 0; i < 64; i++) {
        std::cout << static_cast<unsigned int>(inBufferData[i]) << " ";
    }
    std::cout << std::dec << std::endl;

    std::cout << "width = " << context->inWidth << std::endl;
    std::cout << "height = " << context->inHeight << std::endl;
    std::cout << "inPitch = " << inPitch << std::endl;
    std::cout << "inPixelFormat = " << inPixelFormat << std::endl;

    auto print_pixel = [&](int x, int y) {
    const uint8_t *p = inBufferData + y * inPitch + x * 4;

    std::cout
        << "Pixel (" << x << ", " << y << ") = ("
        << static_cast<int>(p[0]) << ", "
        << static_cast<int>(p[1]) << ", "
        << static_cast<int>(p[2]) << ", "
        << static_cast<int>(p[3]) << ")"
        << std::endl;
    };

[...]

PR #4096, opened 2026-08-21
modules/sceJpegEncUser: Add BGRA8888 CSC input support

saturnsky:
Add support for the SCE_JPEGENC_PIXEL_BGRA8888 pixel format to sceJpegEncoderCsc. Previously, the function only handled the SCE_JPEGENC_PIXEL_RGBA8888 pixel format.

This PR partially addresses #4093.

### DIFF
--- a/vita3k/codec/include/codec/state.h
+++ b/vita3k/codec/include/codec/state.h
@@ -282,7 +282,7 @@ struct PlayerState {
     ~PlayerState();
 };
 
-void convert_rgb_to_yuv(const uint8_t *rgba, uint8_t *yuv, uint32_t width, uint32_t height, const DecoderColorSpace color_space, int32_t in_pitch);
+void convert_rgb_to_yuv(const uint8_t *rgba, uint8_t *yuv, uint32_t width, uint32_t height, const DecoderColorSpace color_space, bool is_bgra, int32_t in_pitch);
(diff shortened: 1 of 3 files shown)


HylianSoul:
I have tested your pull request and now I have this bug when I take a photo the second time.

ProblemCapture

More information for other people: #4093

HylianSoul:
I inverted this condition of your pull request (```vita3k/codec/src/mjpeg.cpp```, line 81):

```
const AVPixelFormat input_format = is_bgra ? AV_PIX_FMT_BGRA : AV_PIX_FMT_RGBA;
```

to...

```
const AVPixelFormat input_format = is_bgra ? AV_PIX_FMT_RGBA : AV_PIX_FMT_BGRA;
```

And the second photo taked is...

photoUncharted


Is also necessary to change the variable name to avoid confusion.

saturnsky:
This was an expected issue. As explained in #4093, the game appears to pass an RGBA buffer to `sceJpegEncoderCsc` instead of BGRA. I suspect this is a separate issue related to surface cache handling and needs further investigation.

Currently, `sceJpegEncoderCsc` receives an RGBA8888 image but converts it as BGRA8888 to YUV, as requested by the argument, so the image cannot be displayed correctly. However, I believe the bug is in the path that provides the RGBA8888 buffer to `sceJpegEncoderCsc`, not in `sceJpegEncoderCsc` itself.

frangarcj (approved):
The scope of this PR is ok. The issue comes from another place.

PR #4100, opened 2026-08-26
motion: fix Steam Deck gyro — bundled SDL reports sensor timings in µs

elpendor:
Motion controls do nothing on a Steam Deck: sensors are enabled and data arrives, but the integrated orientation never advances, so games read a gyro that never turns.

The bundled SDL's Steam Deck driver passes **microseconds** into `SDL_SendJoystickSensor`'s `sensor_timestamp`, which SDL documents as nanoseconds:

```c
/* SDL release-3.2.x, src/joystick/hidapi/SDL_hidapi_steamdeck.c */
ctx->sensor_timestamp_us += ctx->update_rate_us;
SDL_SendJoystickSensor(timestamp, joystick, SDL_SENSOR_GYRO, ctx->sensor_timestamp_us, values, 3);
```

`motion.cpp` then converts ns -> us on top, so the deltas reaching `UpdateOrientation()` / `UpdateRotation()` are 1000x too small. Breaking on `SDL_SendJoystickSensor` on the device shows steps of 4000 -- 4 ms at the Deck's 250 Hz report rate, i.e. microseconds:

```
sensor_timestamp arg: 604000   (gyro)
sensor_timestamp arg: 608000
```

SDL fixed its side in [`e1af6236`](https://github.com/libsdl-org/SDL/commit/e1af6236) (shipped in `release-3.4.x`), but `external/sdl` pins a 3.2.x commit that predates it. The conversion came in `86526dea`, which is why builds since then regressed: **3829 has working motion, 3996/3998/4074 do not, and a build of this branch does** -- same Deck, same game.

This takes the elapsed time from a monotonic clock instead of the device. Only the delta between updates is ever used, so units cannot go wrong.

The trade-off: controllers that do report honest nanoseconds (DualSense, Switch Pro, Android sensors) now use the emulator's clock too. In practice that is immaterial -- `refresh_motion()` reads one timestamp per frame, ~16.7 ms apart, not the 4 ms sensor cadence -- but it does touch those paths.

If you would rather fix it at the root, bumping `external/sdl` to a `release-3.4.x` does the same job and leaves `motion.cpp` alone. Happy to send that instead.

Tested: CI green on this branch; gyro confirmed working in Gravity Rush on a Steam Deck with the resulting AppImage.


EXtremeExploit:
It might be easier to just update SDL rather than implementing a workaround

elpendor:
Oh wow, a fellow argentinian.

> It might be easier to just update SDL rather than implementing a workaround

Want me to do that instead? I can update the PR.

EXtremeExploit:
> Oh wow, a fellow argentinian.

hi

> Want me to do that instead? I can update the PR.

yes please, although iirc sdl next update changed the visibility of some methods that vita3k uses in android so look out for those with CI

elpendor:
> > Oh wow, a fellow argentinian.
> 
> hi
> 
> > Want me to do that instead? I can update the PR.
> 
> yes please, although iirc sdl next update changed the visibility of some methods that vita3k uses in android so look out for those with CI

Sure, I'll look into that. 

Gracias por el aviso, wachin.

elpendor:
Got the SDL update working, making the workaround unnecessary.

Needed two things on top. The visibility thing you mentioned is real, so VitaInputConnection and VitaTextEdit move into org.libsdl.app (which also kills the reflection VitaTextEdit was doing). And it aborted at startup on a device: prepare_frontend_runtime() inits SDL from MainActivity before SDLActivity sets up the JNI, and 3.4 gets as far as SDL_GetAndroidPackageName from there. Setting app metadata before init sorts it.

CI's green on all seven jobs. Deck runs a game with motion working, Windows and macOS open fine, Android starts with keyboard and controller detected (the subclasses this PR touches).

android/app/src/main/java/org/libsdl/app/VitaInputConnection.java
```diff
@@ -1,10 +1,16 @@
-package org.vita3k.emulator;
+// In SDL's own package, not Vita3K's. SDL 3.4 made SDLInputConnection and its
+// constructor package-private (libsdl-org/SDL 9820f655d, "android: reduce
+// visiblity as much as possible"), so a subclass has to live alongside it. The
+// alternative is patching the submodule, which is worse: it would have to be
+// re-applied at every SDL bump.
+package org.libsdl.app;
```
Dicot0721, line 6: Won't this pollute the SDL package?

elpendor: I don't see any other way around it, SDL 3.4 locked those classes, there's no trick that I can see to get around that.

Alternatives are, in my opinion, worse: patching SDL on every bump or rewrite the keyboard handling. Unless they unlock these upstream in the future.

Dicot0721: You're right

EXtremeExploit (dismissed):

nn9dev:
I apologize for the nitpick, but I notice four blocks of comments that seem AI generated. As they don't seem like very pertinent information, would you remove them? Otherwise the content looks good.

elpendor:
> I apologize for the nitpick, but I notice four blocks of comments that seem AI generated. As they don't seem like very pertinent information, would you remove them? Otherwise the content looks good.

Nah, it's okay. 

These were added as context, but I completely agree they're redundant.

Removed.

EXtremeExploit (approved):

PR #4106, opened 2026-08-29
kernel: Resolve imports by library, not just NID

nishinji:
A NID hashes the function name alone, so two modules exporting the same name collided in export_nids and the second was silently dropped. Omega Labyrinth hung on "Now Loading" because UnityNpToolkit's PrxInitialize lost to an unrelated one. Key function exports by (library NID, NID) as well and resolve imports against that first, falling back to the plain NID lookup.

| Master | PR |
| -- | -- |
| image |  image  |

### DIFF
--- a/vita3k/kernel/include/kernel/state.h
+++ b/vita3k/kernel/include/kernel/state.h
@@ -62,6 +62,8 @@ typedef std::map<SceUID, CallbackPtr> CallbackPtrs;
 typedef unordered_map_fast<uint32_t, Address> ExportNids;
 // A NID hashes the function name alone, so same-named exports from different libraries collide.
 typedef unordered_map_fast<uint64_t, Address> LibExportNids;
+// The plain NID entry outlives taiHEN and HLE redirects, so its owner is tracked apart from its address.
+typedef unordered_map_fast<uint32_t, uint32_t> ExportNidOwners;
 constexpr uint64_t lib_export_key(uint32_t library_nid, uint32_t nid) {
     return (static_cast<uint64_t>(library_nid) << 32) | nid;
 }
@@ -142,6 +144,7 @@ struct KernelState {
     std::mutex export_nids_mutex;
     ExportNids export_nids;
     LibExportNids export_nids_by_lib;
+    ExportNidOwners export_nid_owners;
     FuncBindingInfos func_binding_infos;
     VarBindingInfos var_binding_infos;
     ModuleUidByNid module_uid_by_nid;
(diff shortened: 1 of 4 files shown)


frangarcj (commented):
The `(library_nid, nid)` lookup makes sense and fixes the collision issue, but I think this introduces a regression with taiHEN export hooks.

`create_export_hook()` / the fallback export-hook path currently updates only:

`kernel.export_nids[func_nid] = hook_func`

With this PR, newly loaded imports resolve `export_nids_by_lib` first. So if an export hook is installed and another module importing that function is loaded afterwards, it can resolve to the original address from `export_nids_by_lib` instead of the hook.

The release path has the same issue when restoring the original export.

I think the hook state needs to keep the `library_nid` and update/restore `export_nids_by_lib[lib_export_key(library_nid, func_nid)]` as well.

It would also be good to test this sequence:

1. Load/export function
2. Install taiHEN export hook
3. Load a new module importing that function
4. Verify the new import resolves to the hook, not the original export

frangarcj (changes requested):
-

nishinji:
> The `(library_nid, nid)` lookup makes sense and fixes the collision issue, but I think this introduces a regression with taiHEN export hooks.
> 
> `create_export_hook()` / the fallback export-hook path currently updates only:
> 
> `kernel.export_nids[func_nid] = hook_func`
> 
> With this PR, newly loaded imports resolve `export_nids_by_lib` first. So if an export hook is installed and another module importing that function is loaded afterwards, it can resolve to the original address from `export_nids_by_lib` instead of the hook.
> 
> The release path has the same issue when restoring the original export.
> 
> I think the hook state needs to keep the `library_nid` and update/restore `export_nids_by_lib[lib_export_key(library_nid, func_nid)]` as well.
> 
> It would also be good to test this sequence:
> 
> 1. Load/export function
> 2. Install taiHEN export hook
> 3. Load a new module importing that function
> 4. Verify the new import resolves to the hook, not the original export

Fixed.

frangarcj (changes requested):
The taiHEN issue looks fixed now, thanks.

One small remaining case: `unload_func_exports()` still does `kernel.export_nids.erase(nid)` unconditionally. With colliding exports, unloading library B can remove the plain-NID fallback that actually belongs to library A (or an HLE export).

Could you make the plain fallback removal ownership-aware as well?

frangarcj (changes requested):
This fixes the normal colliding-library case, but I think there is still one edge case with redirected exports.

After `register_hle_override()`, both the library-keyed export and the plain NID entry can point to the same HLE stub. Then `unload_func_exports()` sees the same `unloaded_address` and may erase the plain HLE fallback even though it no longer belongs to the unloaded LLE export.

So comparing the current address is not quite the same as tracking ownership after redirects.

I think the simplest robust fix would be to track ownership of the plain-NID fallback separately from its current address.

For example, keep a small `export_nid_owners[nid] = library_nid` map when `export_nids.emplace()` actually wins. Redirects from taiHEN would only change the address, not the owner.

Then `unload_func_exports()` can remove the plain entry only when:

```cpp
export_nid_owners[nid] == library_nid
```

instead of comparing addresses.

For HLE overrides, the owner could be cleared/set to `0`, since the plain entry is no longer owned by that LLE library.

That should also make unload correct even after taiHEN/HLE redirects.

frangarcj (approved):

PR #4108, opened 2026-08-29
building.md: added Arch Linux building dependencies.

HylianSoul:
README.md: added dependencies required to execute on Linux.

I used a Arch Linux container using Distrobox (because is easier to get Qt 6.11). This is the reason that have a lot of packages: ninja failed because many elemental packages are not in the system.

Thanks to the *vita3k-bin* and *vita4k-git* AUR packages (from  EXtremeExploit ), I found the names of the Qt dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,6 @@ You can download the latest builds from [here](https://github.com/Vita3K/Vita3K/
     * [vita3k-git](https://aur.archlinux.org/packages/vita3k-git)<sup><small>AUR</small></sup>
   * Requirements:
     * xdg-desktop-portal
-    * pipewire-pulse
     * OpenGL or Vulkan runtime libraries
 * Android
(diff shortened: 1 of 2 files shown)


frangarcj (changes requested):
The Arch Linux build dependencies look good to me.
I would just avoid listing pipewire-pulse as a general Linux runtime requirement, since PipeWire/Pulse compatibility is system-dependent and shouldn't be required for every Linux setup.
Could you remove pipewire-pulse from the general requirements? The Arch build instructions themselves look fine.

HylianSoul:
OK. I have added the pipewire-pulse dependency because the emulator crashes when the sound of every game begun and I had not this dependency on my Arch Linux container.

I will delete this line and I will push again,

PS. The review requested is a missclick, sorry.

frangarcj (approved):

PR #4134, opened 2026-09-17
i18n: Move the interface translations to translation repository

nishinji:
Only 2 of the 21 languages Crowdin holds were ever committed here, and both were out of date. The interface translations move to [nishinji/vita3k_translations](https://github.com/nishinji/vita3k_translations), which syncs them with Crowdin and publishes the finished files as a release.

## Builds take the translations from that repository

`i18n/qt/*.ts` are deleted. `.ci/common.sh` downloads the latest release of the translations repository and unpacks it before the build:

- the compiled Qt `.qm` files into the build's `qt_translations` directory
- the Android `values-*/strings.xml` into `android/app/src/main/res`

## The Crowdin source is generated here

Crowdin needs the English strings, and only this repository has the Qt sources to extract them from. A new `qt-ts.yml` runs `lupdate` on every push to `master` that touches the Qt interface and uploads `vita3k_template.ts` as an artifact; the translations repository picks it up, merges it into every language and sends it to Crowdin. New strings reach translators without anyone editing a `.ts` by hand.


Zangetsu38:
I am not fan by adding pyton dependence 💀 
il you have other way please

nishinji:
> I am not fan by adding pyton dependence 💀 il you have other way please

Why?

Zangetsu38:
> Why?

I just don't want Python to become mandatory for the project...
And for a language-related thing like this, it's not necessary, there has to be another solution.

EXtremeExploit:
Python is already a dependency
https://github.com/Vita3K/Vita3K/blob/master/vita3k/lang/CMakeLists.txt#L16

Zangetsu38:
> Python is already a dependency https://github.com/Vita3K/Vita3K/blob/master/vita3k/lang/CMakeLists.txt#L16

AH, i have no view it, sorry lol

.github/workflows/qt-ts.yml
```diff
@@ -0,0 +1,42 @@
...
+      - master
+    paths:
+      - 'vita3k/gui-qt/**'
+      - 'vita3k/main.cpp'
+      - '.github/workflows/qt-ts.yml'
+
+jobs:
+  generate-template:
+    name: Generate Qt translation template
+    runs-on: ubuntu-24.04
```
Dicot0721:
```suggestion
    runs-on: ubuntu-latest
```

.ci/build-windows-msys2.sh
```diff
@@ -30,6 +30,7 @@ export CCACHE_DIR
 CCACHE_DIR="$(cygpath -u "C:/msys2-ccache")"
 
 cmake --preset "$PRESET"
+download_qt_translations "build/$PRESET/qt_translations"
```
EXtremeExploit, line 33: I belive this one isnt necessary as windows arm also uses build-desktop

EXtremeExploit: Nvm i think its good

.github/workflows/qt-ts.yml
```diff
@@ -0,0 +1,42 @@
...
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+    paths:
+      - 'vita3k/gui-qt/**'
+      - 'vita3k/main.cpp'
+      - '.github/workflows/qt-ts.yml'
```
EXtremeExploit, line 14: Should also include the overlay.json file in `i18n/lang`

nishinji: Is the `overlay.json` file ever updated?

EXtremeExploit: true...

EXtremeExploit:
i guess we can merge this for now, if it doesnt work for some reason then a quickfix can be done

EXtremeExploit (dismissed):

PR #4137, opened 2026-09-18
gxm/renderer: Support multiple immediate contexts

frangarcj:
## Summary

- Allow multiple immediate GXM contexts to coexist.
- Track immediate contexts independently for `sceGxmFinish`, destruction, cleanup and display submission.
- Restore the command list's renderer context before processing it.
- Avoid clearing the active renderer context when destroying a different context.

## Motivation

vitaGL's splash screen creates a second immediate GXM context while the main context is still alive.

Vita3K previously treated the immediate context as a process-wide singleton, causing the second `sceGxmCreateContext` call to fail with `SCE_GXM_ERROR_ALREADY_INITIALIZED`. As a result, vitaGL builds required the splash screen to be disabled.

## Testing

- Built the `vita3k` target successfully on macOS.
- `ctest --test-dir build/macos --output-on-failure`: 2/2 tests passed.
- `clang-format --dry-run --Werror` passes for all modified files.
- `git diff --check` passes.
- Tested vitaGL at commit `cadd460000ce50724705def29419535fd95de21d` with the splash screen enabled on Vulkan:
  - two immediate contexts are created simultaneously;
  - the splash context renders and is destroyed while the main context remains alive;
  - the main context resumes rendering and `sceGxmFinish` continues to work.
- Tested the lifecycle separately with two immediate contexts:
  - create A and B;
  - finish A and B;
  - destroy A;
  - finish B;
  - terminate GXM and clean up the remaining context.
- Runtime-tested the OpenGL path locally by temporarily enabling it on macOS:
  - the GXM lifecycle test creates two simultaneous immediate contexts and completes without GXM errors;
  - vitaGL reaches the second `sceGxmCreateContext` successfully;
  - existing macOS OpenGL 4.1 limitations later prevent full rendering, and the local compatibility workarounds used to reach that point are not part of this PR.


vita3k/modules/SceGxm/SceGxm.cpp
```diff
@@ -2263,8 +2270,13 @@ EXPORT(int, sceGxmDisplayQueueAddEntry, Ptr<SceGxmSyncObject> oldBuffer, Ptr<Sce
 
     // TODO: I do this because the sync function does not have access to the display state, but this is not great
     renderer::Context *active_renderer_context = nullptr;
-    if (emuenv.gxm.immediate_context != 0)
-        active_renderer_context = Ptr<SceGxmContext>(emuenv.gxm.immediate_context).get(emuenv.mem)->renderer.get();
+    if (emuenv.gxm.last_immediate_context != 0) {
+        const auto immediate_context = std::find_if(emuenv.gxm.immediate_contexts.begin(), emuenv.gxm.immediate_contexts.end(), [&](const auto &entry) {
```
Dicot0721:
```suggestion
        const auto immediate_context = std::ranges::find_if(emuenv.gxm.immediate_contexts, [&](const auto &entry) {
```

vita3k/modules/SceGxm/SceGxm.cpp
```diff
@@ -1300,8 +1308,9 @@ static int destroy_gxm_context(EmuEnvState &emuenv, Ptr<SceGxmContext> context_p
 namespace gxm {
 
 void destroy_all_contexts(EmuEnvState &emuenv, const bool force_backend_destroy) {
-    if (emuenv.gxm.immediate_context != 0) {
-        const int result = destroy_gxm_context(emuenv, Ptr<SceGxmContext>(emuenv.gxm.immediate_context), force_backend_destroy);
+    const std::vector immediate_contexts(emuenv.gxm.immediate_contexts.begin(), emuenv.gxm.immediate_contexts.end());
+    for (const auto &[context, context_addr] : immediate_contexts) {
```
Dicot0721: Why copy it into a vector instead of using it directly?

vita3k/modules/SceGxm/SceGxm.cpp
frangarcj: Updated to std::ranges::find_if, thanks!

vita3k/modules/SceGxm/SceGxm.cpp
frangarcj: The copy was there because destroy_gxm_context erases the current entry on success, so iterating the map with a normal range-for would invalidate the current iterator. I changed it to iterate the map directly, advance the iterator first, and then destroy the current context. This avoids the copy while still keeping failed entries registered for the later forced shutdown cleanup. Does this look better to you?

Dicot0721: Can we imitate the destruction of `emuenv.gxm.deferred_contexts` as below?

frangarcj: I looked into this and tested the deferred-context case explicitly. The same iterator pattern can be used there too, but the error handling cannot be identical. If a deferred context is still inside a command list, the non-forced cleanup returns SCE_GXM_ERROR_WITHIN_COMMAND_LIST. The current deferred code then erases it from tracking without destroying it; keeping it until the later forced shutdown is also unsafe, because by then kernel.process_exit() has already invalidated the guest context memory. I tested retrying the deferred destroy immediately with force=true, before process_exit(), and that succeeds and removes it cleanly. So I think it would make sense to make the immediate/deferred iteration style consistent in a follow-up PR, while handling the deferred WITHIN_COMMAND_LIST case specially. For this PR I would prefer to keep the immediate cleanup as it is rather than copy the current deferred erase-on-failure behavior.

Dicot0721: Oh, what I meant was to use only `empty()` and `begin()` instead of comparing the iterator to `end()`, the former doesn't require declaring multiple iterator variables and looks more concise.

EXtremeExploit (approved):
every game i tried seems to work as intended

PR #4139, opened 2026-09-18
Fix GXM downscale color surface misplacement and a related Vulkan crash

amurgshere:
## Summary

Four related fixes for GXM downscale color surfaces, a Vulkan crash that shows up when using them, and two OpenGL rendering issues, found while investigating a graphics bug in Super Stardust Delta (PCSF00019/PCSA00006).

- Fixes a Vulkan crash (unhandled vk::SystemError from a rejected vkQueueSubmit) caused by two render target size tracking bugs: an MSAA width/height doubling that wasn't reliably reverted, and a render pass renderArea computed from the surface's nominal size instead of the actual bound framebuffer size.
- Fixes misplaced/cropped content on GXM downscale color surfaces (SCE_GXM_COLOR_SURFACE_SCALE_MSAA_DOWNSCALE). These surfaces are allocated at their final logical size, but games issue region clip and viewport coordinates sized for the physical 2x canvas the hardware renders at internally before resolving down. The existing scaling logic only handled the opposite case (MSAA without downscale), so standalone downscale surfaces were left unscaled.
- Fixes an OpenGL only issue where render target color surfaces used as textures (for example in a multi tap blur reading a previous pass's output) never had their wrap mode set, so they defaulted to GL_REPEAT and could sample content from the wrong edge near the border. Vulkan was unaffected since its samplers are built per draw from the GXM texture's own address mode.
- Fixes another OpenGL only issue where `GL_PROGRAM_POINT_SIZE` was never enabled, so point sizes written by shaders were ignored and point-sprite particles (for example the green particles in Super Stardust Delta) did not show. Vulkan was unaffected since point size is taken from the shader there. A follow-up commit skips the `glEnable` on Android, where OpenGL ES has no such capability and the call would otherwise raise GL_INVALID_ENUM every frame.

## Issues

Addresses https://github.com/Vita3K/compatibility/issues/441#issuecomment-5730694974

## Testing scope

I have only tested this against Super Stardust Delta (PCSF00019/PCSA00006) on Windows, on both the Vulkan and OpenGL backends, where these bugs were originally found and reproduced (a stray red glow blob rendering onto the bottom right HUD icon, a fainter version of the same issue appearing at the top of the screen on OpenGL only, and green particles missing on OpenGL only). I have not tested against other titles that may use downscale color surfaces, so please treat this as a fix validated on one game rather than a broad regression sweep. Happy to test other affected titles if anyone can point me at one. The Android follow-up was checked on a Retroid Pocket 5 (Adreno 650, OpenGL ES 3.2) with Sparkle 2: the driver reported 548 invalid-enum errors in one run before the change and none after; point-sprite particles on Android OpenGL are not verified.

## Test plan

- [x] Super Stardust Delta (PCSF00019) on Vulkan: HUD icon glow bug gone, no crash
- [x] Super Stardust Delta (PCSF00019) on OpenGL: HUD icon glow bug gone, top of screen sliver gone, green particles now show, no crash
- [ ] Other titles using downscale color surfaces (untested, see testing scope above)


amurgshere:
Pushed a follow-up commit for the GL_PROGRAM_POINT_SIZE change. OpenGL ES has no such capability, so `glEnable(GL_PROGRAM_POINT_SIZE)` raised GL_INVALID_ENUM every frame on Android. The surface cache's `GLObjectArray::init` returns `glGetError() == GL_NO_ERROR`, so the leftover error made every second frame's new color surface texture look failed and nothing was presented (black flicker). Found on a Retroid Pocket 5 (Adreno 650, OpenGL ES 3.2) with Sparkle 2: 548 invalid-enum errors in one run before the change, none after. The new commit skips the call on Android; desktop behaviour is unchanged. I have not verified point-sprite particles on Android OpenGL as OpenGL ES has other issues as blockers on my device right now.

PR #4141, opened 2026-09-19
vulkan: Work around an Adreno shader compiler crash on the output register array

amurgshere:
## Summary

On Qualcomm Adreno 650 devices, Android Vulkan crashes the whole app with SIGSEGV inside `vkCreateGraphicsPipelines`. I hit it in Super Stardust Delta (PCSF00019) on a Retroid Pocket 5. It happens on the first level, around when the first enemy ships appear.

The crash happens inside the driver's shader compiler (`libllvm-qgl`, fault address `0x48`), not in Vita3K code.

## Cause

The driver crashes when a vertex shader writes only some lanes of a constant-indexed element of a `Private` `vec4` array of 19 or more elements. "Some lanes" means anything except a write that includes `.w`. Arrays of 18 or fewer elements are fine.

Vita3K's SPIR-V recompiler declares its output register bank (`outs`) as `REG_O_COUNT = 20 * 4`, which is 20 `vec4`. Any vertex shader that partially writes one of the bank's elements triggers the crash.

I found the trigger with a standalone Vulkan test app that compiles pipelines in a crash-isolated worker process. It automatically reduced the crashing game shader to a roughly 1.5 KB SPIR-V module. The measured boundary is 18 elements safe and 19 crashing. The reduced module contains no game content.

## Changes

Three commits, each of which builds on its own:

1. **Add an Adreno workaround setting (auto/on/off).**
   - New string config option `adreno-workaround` (default `auto`), stored globally and per-title.
   - `auto` enables the workaround when the Vulkan device's vendor ID is Qualcomm or its name contains "Adreno". `on` and `off` force it.
   - The setting is exposed in the Android GPU settings as an Auto/On/Off choice.
   - The renderer logs the resolved state at startup.
2. **Shrink the output register array on Adreno.**
   - When the workaround is active, the recompiler declares `outs` as 16 `vec4` instead of 20. Shaders write at most 15 output `vec4`, so nothing is lost.
   - The flag is also a new bit in the feature mask that keys the shader cache. Other GPUs keep their existing caches, and Adreno recompiles its shaders once.
3. **Report driver crashes during Vulkan pipeline creation on Android.**
   - `vkCreateGraphicsPipelines` is wrapped in a `sigsetjmp`/SIGSEGV guard, on Android only. On other platforms it just calls the function.
   - If the driver crashes, Vita3K logs the vertex and fragment shader hashes and then aborts. Before this, a driver crash was an unattributed native crash.
   - The guard is a diagnostic aid, not recovery. It exists so future driver bugs of this kind can be identified from a log.

## Testing

- **Retroid Pocket 5 (Adreno 650):**
  - Super Stardust Delta previously crashed on the first level, around when the first enemy ships appear. It now plays through level 7 with no crashes.
  - The log shows `Adreno workaround fixes: setting=auto device="Adreno (TM) 650" active=true`.
  - The driver-crash guard never fired.
- **Windows (OpenGL and Vulkan):** the game runs normally. The workaround stays off on non-Qualcomm GPUs, and the shader cache key is unchanged for them.
- **Builds:** this branch was built on its own, with the other branch's changes absent, for Windows (RelWithDebInfo) and Android (release APK).

## Notes

- I plan to report the driver bug to Qualcomm separately. The workaround stays useful for devices already in the field either way.
- The setting is deliberately not tied to a specific driver version, because the affected driver builds can't be enumerated. `off` lets you check whether a given device still needs the workaround.


Dicot0721:
Have you tried turning off "Enable Shader Cache" before pushing the AI?

amurgshere:
> Have you tried turning off "Enable Shader Cache" before pushing the AI?

Hey @Dicot0721 , 

I cleared the game's shader cache before each test run, so stale cached shaders weren't a factor. But I also just retried with 'Enable Shader Cache' off and the Adreno workaround off, and got the same crash.

This makes sense as the cache only stores Vita3K's generated SPIR-V, not driver output. The crash is in the Adreno driver's own shader compiler, inside `vkCreateGraphicsPipelines`, so it happens whether or not the SPIR-V came from the cache. I also ruled out the Vulkan pipeline cache by testing with it disabled, and the crash still occurred. I reproduced it outside Vita3K too, I got Claude to write a small standalone Vulkan test app, and then used `spirv-reduce` to cut a vertex shader from 33 KB to 1.5 KB. The trigger is a vertex shader that partially writes a constant-indexed element of a private `vec4[N]` array with N ≥ 19 (Adreno 650, tested on a Retroid Pocket 5). Vita3K's `outs` bank is 20 `vec4`, which hits this.

The workaround shrinks that bank to 16 `vec4`, which is enough because the generator emits at most 15. The flag is part of the shader cache key, so cached shaders are regenerated correctly. It's behind the `adreno-workaround` setting (auto/on/off, default auto, which activates only on Adreno), so other GPUs are unaffected.

Thank you for taking the time to review this, I am happy to take feedback and restructure it if you'd prefer it more contained. This is one of my favourite games for the PS Vita so really wanted to get it working on my Retroid Pocket 5 and this change allows it to run nicely.

Dicot0721:
So, if I understand correctly, this is a bug in the driver, not the emulator, right? (PCSF00019 works fine with 4096 on my phone.) You should use the Turnip driver instead of adding complex, redundant code to make the emulator work around the bug in the driver.

amurgshere:
> So, if I understand correctly, this is a bug in the driver, not the emulator, right? (PCSF00019 works fine with 4096 on my phone.) You should use the Turnip driver instead of adding complex, redundant code to make the emulator work around the bug in the driver.

Exactly, No problem that’s a fair call,  it’s your project, I want to get it working with the stock Adreno driver though as so many emulators require different drivers for each game, so I will just keep these changes in my local fork then, I will close the pr, feel free to delete it completely.i won’t raise any more PR for driver specific workarounds.

thank you.

PR #4142, opened 2026-09-19
kernel: Destroy a finished guest thread's state before waking process_exit

amurgshere:
## Summary

Stopping emulation could crash in the `ThreadState` destructor. I saw it on Android while stopping a game, and the race is not platform-specific.

## Cause

Guest threads are detached. `process_exit()` waits on `thread_deleted_cond` until the kernel's thread map is empty, then tears down the memory state.

`thread_function` erased its map entry and notified the waiter while still holding a `ThreadStatePtr` in a local. That reference was destroyed only when the function returned, after the notification. So `process_exit()` could wake and destroy the `MemState` while the last `ThreadState` was still alive. The `ThreadState` destructor then freed guest memory on a destroyed `MemState`.

## Change

`thread_function` now copies the thread id and processor id it still needs, then resets its `ThreadStatePtr` before taking the lock. The erase under the lock therefore destroys the state before `notify_all()` runs. The waiter can only wake once the state is gone.

The change is confined to `vita3k/kernel/src/kernel.cpp` (+7/−3). It doesn't change locking order or behaviour for running threads.

## Testing

- **Retroid Pocket 5 (Android):** repeated clean exits from Super Stardust Delta with no shutdown crash. Before the fix, exit crashed intermittently.
- **Windows:** builds and runs normally.
- **Builds:** this branch was built on its own for Windows (RelWithDebInfo) and Android (release APK).


EXtremeExploit (approved):
everything seems to work fine

